PR #284, opened 2015-03-16
StaveTie updated so that direction can be set, else falls back to normal...

jmahmud:
... assignement from first/last notes


jmahmud:
Just a few minor suggestions: main change is in the stavehairpin.js where I've added a beginPath & closePath (mimicking how crescendo.js works) as without it, was cause a duplication of surrounding glyphs.

tests/flow.html
```diff
@@ -5,7 +5,7 @@
   <title>VexFlow 2 - JavaScript Music Notation and Guitar Tab</title>
   <link rel="stylesheet" href="flow.css" type="text/css" media="screen" />
   <!-- VexFlow Sources -->
-  <script src="../build/vexflow-debug.js"></script>
+  <script src="../releases/vexflow-debug.js"></script>
```
0xfe: Is this line intentional?

0xfe:
Change looks good. Can you add tests please?

jmahmud:
HI Mohit - will definitely add tests and also see if I can do something with those hardcoded values in the stavehairpin class.  

Cheers.
Josh

tests/flow.html
jmahmud: Yes - it seems that the built vexflox-debug.js has move so I have updated the src...flow.html stopped working without this change - did I miss something perhaps?

0xfe: Running `grunt` will put files in `build/`. Also, `grunt watch` will keep `build/` updated as you change sources. The `releases/` directory is what I populate when I cut a new release and publish and NPM.

SalahAdDin:
:+1:

0xfe:
Thanks for these changes, and for adding the tests. Merged.

PR #289, opened 2015-04-22
adds a setX method to stave

ssoRL:
Allows the x position of a stave to be changed.

### DIFF
--- a/src/stave.js
+++ b/src/stave.js
@@ -94,10 +94,10 @@ Vex.Flow.Stave = (function() {
     setX: function(x){
       var shift = x - this.x;
       this.x = x;
-      this.glyph_start_x = x + 5;
-      this.glyph_end_x = x + this.width;
-      this.start_x = this.glyph_start_x;
-      this.end_x = this.glyph_end_x;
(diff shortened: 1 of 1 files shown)


gaborveszpremi:
Great, I am just thinking to wrap the score into lines, and I need setX() too.
But I think this is not 100% correct.
start_x should not always be equal to glyph_start_x
If I am not mistaken, this works only if you have no clef, and no key signature.
I use your "shift" variable instead.

Instead of this:

this.glyph_end_x = x + this.width;
this.start_x = this.glyph_start_x;
this.end_x = this.glyph_end_x;

I use this:

this.glyph_end_x += shift;
this.start_x += shift;
this.end_x += shift;

With this modification, the function works correctly on staves with modifiers.

Sorry for writing the code into a comment, I am working hard on understanding Git and finding the good practices to collaborate. :)

ssoRL:
OK, I made that change and everything seems to be working.
For small changes like this it's probably easiest to just post suggestions in the comments. If it was anything more you could branch my branch and then add a pull request there.

PR #291, opened 2015-05-08
add a vexflow typescript declaration file

ssoRL:
This only exposes the parts of vexflow that I needed for my project, so it could certainly be expanded on, but here's the start of a declarations file to use in a typescript project that uses vexflow.

### DIFF
--- a/contrib/typescript/vexflow.d.ts
+++ b/contrib/typescript/vexflow.d.ts
@@ -1,0 +1,140 @@
+
+declare module Vex {
+    export module Flow {
+        class BoundingBox {
+            x: number;
+            y: number;
+            w: number;
+            h: number;
+        }
+        export interface Note {
+            clef: string;
+            getBoundingBox(): BoundingBox;
+            setStave(stave: Stave): Note;
+            getAbsoluteX(): number;
+            getCategory(): string;
+            intrinsicTicks: number;
+            addAccidental(index: number, acci: Vex.Flow.Accidental): void;
+            addDotToAll(): void;
+            ticks: number;
+        }
+        class StaveNote implements Note {
+            clef: string;
+            constructor(options: {
+                keys?: string[], duration?: string,
+                clef?: string, auto_stem?: boolean
+            });
+            getBoundingBox(): BoundingBox;
+            setStave(stave: Stave): StaveNote;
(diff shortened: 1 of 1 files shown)


0xfe:
Hmmm... Is there a way to generate this programmatically? E.g., by adding annotations to the JS?

ssoRL:
There's <a href="https://github.com/SitePen/dts-generator">dts-generator</a> that does something like that for typescript files, but nothing for js that I could find.

fmilitao:
Hi,

[DefinitelyTyped](https://github.com/borisyankov/DefinitelyTyped) is probably (also?) a good place to pull the d.ts file since they have lots of useful typescript type definitions.

I also tried to use your definition file and I am getting errors when using "noImplicitAny: true" since not all functions include explicit types.

ssoRL:
Yes, the d.ts file here is intended more as a starting point than a definitive edition. Since vexflow doesn't have much in the way of a standard API, I just made something that exposed the parts that I needed. Your project might require some extension of that.
And yes, I think that perhaps this would be better suited on definitely typed.

rquiring:
I created a complete definition file and just [added it] (https://github.com/borisyankov/DefinitelyTyped/pull/5064) to [DefinitelyTyped] (http://definitelytyped.org/). Feel free add it to VexFlow.
Having gone through all the code manually, I don't think there is a way to generate the definitions. It was quite painful to get all the different init and return structs right, and I cannot imagine there is a realiable method of automating this.
Also, I noticed a few API inconsistencies, but I guess that's a separate issue. Search the file for the term "inconsistent" to see my suggestions.

0xfe:
Closing this. The DefintelyTyped definitions look more complete, and are also in the right place.

mscuthbert:
I've looked into this also and it appears that dts-gen and other automated tools still don't play well with es6 modules, so the answer from July 2015 remains the same.

mscuthbert:
I also notice that the DefinitelyTyped definitions are off from Vexflow's current ones (e.g., no openGroup() on SVGContext) -- should we be hosting them here so that we're more able to keep them up to date?

bneumann:
We (opensheetmusicdisplay) did an update once to deliver it to our users. It should be fairly up to date at definitely typed

PR #301, opened 2015-08-26
Enable Flat beams in Beam.generateBeams()

aaronmars:
This change allows beam slope calculations to be bypassed in the beam generation logic, and allows for a flat beam.  The current, default behavior of generateBeams() is unchanged.

The offset for the flat beam can be universally applied for all of the notes sent to generateBeams(), or it can be automatically calculated per beam group.

To enable flat beams with different heights per beam group, send in `flat_beams: true` to the generateBeams() options.

If the same height is desired for all of the beams in the set of notes sent to generateBeams(), a 2nd parameter, `flat_beam_offset`, can be passed along with `flat_beams: true`.  This parameter is ignored if `flat_beams` is `false` (the default).

I added 5 tests for the 2 modes described above with both up-stems and down-stems (and mixed, up and down).  I tried to put in as many edge cases as I could think of.

### DIFF
--- a/src/beam.js
+++ b/src/beam.js
@@ -176,6 +176,24 @@ Vex.Flow.Beam = (function() {
       this.y_shift = y_shift;
     },
 
+    // Calculate a slope and y-shift for flat beams
+    calculateFlatSlope: function() {
+      var highest_note = 0;
+      var lowest_note = 0;
+      for (var n = 0; n < this.notes.length; ++n) {
+        var stem = this.notes[n].getStemExtents();
+        var top = stem.topY;
+        if (top > highest_note) {
+          highest_note = top;
+        }
+        if (lowest_note === 0 || top < lowest_note) {
+          lowest_note = top;
+        }
+      }
+      this.slope = 0;
+      this.y_shift = (highest_note - lowest_note) * -this.stem_direction;
+    },
+
     // Create new stems for the notes in the beam, so that each stem
     // extends into the beams.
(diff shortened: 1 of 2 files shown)


src/beam.js
```diff
@@ -71,6 +71,7 @@ Vex.Flow.Beam = (function() {
       this.notes = notes;
       this.beam_count = this.getBeamCount();
       this.break_on_indices = [];
+      this.min_flat_beam_offset = 15;
```
0xfe: Can you put this in `render_options`?

0xfe:
Thanks for the commits. Can you paste in some screenshots of the test output here please?

In particular, I'm curious about the `flat_beam_offset` variable and how it affects the beams.

aaronmars:
Will do.  I'll also address the inline comment above.
Basically, `flat_beam_offset` can mess up the rendering if it is forced beyond the note heads (depending on the stem direction).

The 2 options to fix this (that I see) are:

* Override the supplied offset if it is going to cause this out-of-bounds issue.
* Throw an error if this occurs.

aaronmars:
The new functionality allows for flat_beam_offset to be ignored if it is going to cause a collision with the note heads.
Here are the tests (I added 2 since the original PR was sent).
flat_beams_tests_1
flat_beams_tests_2
flat_beams_tests_3

0xfe:
Thanks for the commit. Looks good.

Issue #302, opened 2015-09-16
Bug in calculation of partial beams with secondary breaks


aaronmars:
This can be reproduced in Vex.Flow.Test.Beam.breakSecondaryBeams() if the first voice is changed to:

```javascript
var notes = [
  newNote({ keys: ["f/5"], stem_direction: 1, duration: "16", dots: 1 }),
  newNote({ keys: ["f/5"], stem_direction: 1, duration: "32"}),
  newNote({ keys: ["c/5"], stem_direction: 1, duration: "16", dots: 1 }),
  newNote({ keys: ["d/5"], stem_direction: 1, duration: "32"}),
  newNote({ keys: ["c/5"], stem_direction: 1, duration: "16", dots: 1 }),
  newNote({ keys: ["d/5"], stem_direction: 1, duration: "32"}),

  newNote({ keys: ["f/5"], stem_direction: 1, duration: "16"}),
  newNote({ keys: ["e/5"], stem_direction: 1, duration: "16"}),
  newNote({ keys: ["e/5"], stem_direction: 1, duration: "16"}),
  newNote({ keys: ["e/5"], stem_direction: 1, duration: "16"}),
  newNote({ keys: ["e/5"], stem_direction: 1, duration: "16"}),
  newNote({ keys: ["e/5"], stem_direction: 1, duration: "16"})
];
notes.forEach(function(note) {
  if ('dots' in note && note.dots >= 1) {
    note.addDotToAll();
  }
});
```
You then get:
![partial_beam_bug](https://cloud.githubusercontent.com/assets/847926/9919829/2617f62c-5c84-11e5-90e3-b60788eba9a5.png)
...those first two partial beams at the 32nd level should be pointing left.

I have a fix for this, but it requires a re-write of `Beam.getBeamLines()`.  I'll send a PR for review once it's tested a bit more.

0xfe:
Thanks Aaron. Good catch, I'll wait for your fix.

On Wed, Sep 16, 2015 at 6:10 PM, Aaron <notifications@github.com> wrote:

> This can be reproduced in Vex.Flow.Test.Beam.breakSecondaryBeams() if the
> first voice is changed to:
>
> var notes = [
>   newNote({ keys: ["f/5"], stem_direction: 1, duration: "16", dots: 1 }),
>   newNote({ keys: ["f/5"], stem_direction: 1, duration: "32"}),
[...]

PR #303, opened 2015-09-30
Secondary breaks/partial beam fixes

aaronmars:
This fixes an edge case where partial beams were not pointing the correct direction when the secondary beam was being broken manually with `Beam.breakSecondaryAt()` (as reported in https://github.com/0xfe/vexflow/issues/302).

The fix required a re-write of `Beam.getBeamLines()`.

I also took the liberty of adding a new option to `Beam.generateBeams()` called `secondary_breaks `.  This parameter will force secondary breaks to be auto-applied per beam group, rather than relying on breakSecondaryAt() for each case.


0xfe:
Thanks for the change. This conflicts with your "flat beams" changes. Can you resolve?

aaronmars:
Resolved.

aaronmars:
OK.  The conflicts have been merged, and I had to do an extra commit to remove some temp files that were accidentally added by my merge tool.

0xfe:
Great! Thanks.

Issue #308, opened 2015-10-14
Bug in beam calculation with breaks


aaronmars:
The logic for calculating beams with secondary breaks has a bug in a specific case.  If the notes in the `AutoBeamFormatting.secondaryBreaks` test include the following for a quarter note-duration:
```javascript
newNote({ keys: ["f/5"], duration: "32"}),
newNote({ keys: ["f/5"], duration: "32"}),
newNote({ keys: ["f/5"], duration: "32"}),
newNote({ keys: ["f/5"], duration: "32"}),
newNote({ keys: ["f/5"], duration: "16", dots: 1}),
newNote({ keys: ["f/5"], duration: "32"}),
```
... the output is:
![screen shot 2015-10-14 at 10 08 44](https://cloud.githubusercontent.com/assets/847926/10491455/96506aea-725b-11e5-87c1-a4dbb1b99896.png)
(^ those first 4 notes are supposed to be 32nds)

I have a fix ready for this case, but I'm now a bit scared of other scenarios where the beaming/beam breaking logic fails.  Shall I submit this fix?

0xfe:
Yes, please submit -- I'll give it a look. Also, please add beam tests for these corner cases, and verify that the existing beam tests work correctly. I'm working on a framework for visual regression testing, so hopefully we won't have to rely on eyeballing the huge number of tests before submitting.

PR #309, opened 2015-10-14
Beam breaks bugfix

aaronmars:
This fixes issue https://github.com/0xfe/vexflow/issues/308.
When the first beam group has a secondary break preceded by notes of the same type, and the note following the break does not contain the beam at the target level, the beam is treated as a partial.

The test has been altered to include this case:
screen shot 2015-10-14 at 13 09 31
(Please compare to the screenshot in https://github.com/0xfe/vexflow/issues/308)


gristow:
Related to this -- and I'm happy to tackle this once the various active beam branches & issues are merged and closed  -- we are missing a method for a user to set the direction of a secondary beam manually. This is needed, for instance, in rhythmic patterns like 6/8 eighth-sixteenth-eighth-sixteenth, which can be written two ways:

screen shot 2015-10-14 at 7 31 07 pm

Where the first way indicates feeling the subdivision as three eighth-notes, and the second way indicates feeling it as two dotted-eighth long groups (a tuplet of sorts).  (Finale uses the first way, Sibelius the second by default.)

I have this worked out in a fork and will push separately once this is merged and I can fix code conflicts.

0xfe:
gristow@, Thanks that would be great. I'll try to get this reviewed and merged today.

BTW, it seems like the beaming code has gotten really gnarly, and probably needs to be "rethunk". I have some ideas on how to refactor and improve the code to better support these cases. (Also, cross-stave beaming has been clamored for for a while.)

aaronmars:
Agreed on all points.  My recent re-write fixed it only incrementally (and, sadly, introduced a regression).
It's way more complicated than one might think.  For now, I'll leave it as-is, and submit fixes to the current logic if more edge cases are found.
I feel a re-think is definitely in order, though.  Otherwise, we're just going to be patching the existing spaghetti over and over.

0xfe:
Where was the regression? I just checked in the visual regression test system, and it only found
a single test that changed (which I think is intentional): Duration Based Secondary Beam Breaks.

BTW, if you get a chance please try out the regression test tool. Instructions in `tools/visual_regression.sh`.

aaronmars:
The regression was not captured by the tests.  I found it during development while using the new auto-breaking functionality.
Having said that, it has been added now.  This visual regression tool is going to help a *lot*!  For some reason, I think I'll be able to trust it more than my eyeballs.

PR #320, opened 2016-01-18
Replacing StaveModifier handling system from using glyphs to modifiers only

panarch:
Happy new year!

At first, thanks so much for adding such kind of really nice visual regression test.
I made lots of change in this PR and it wasn't be possible without the visual regression test.

Changes in this PR are below,

## Replacing StaveModifier handling system from using glyphs to modifiers only
There were little bit special stave modifiers which added into stave begin or end: Clef, TimeSignature and KeySignature.

When stave.addClef(or addKeySignature, addTimeSignature) called, that stave modifier generates glyphs and add to stave right that time. But it makes stave hard to modify stave modifier values because stave.glyphs works independently with modifiers after added.
So, I removed current stave glyph system (in case of end_glyphs, I added it 2014...) and now stave only uses modifiers with self formatting.

#### Add StaveModifier.Position
Now StaveModifier has position which Modifier has, but there are new types BEGIN and END.
ABOVE, BELOW, LEFT, RIGHT positions does not require formatting but BEGIN and END requires formatting before drawing.

BEGIN and END are for Clef, TimeSignature, KeySignature and StaveBarline.

#### Make StaveBarline to be normal StaveModifier likes Clef, TimeSignature and KeySignature.
StaveBarline can have BEGIN or END positions.
Previously StaveBarline was a kind of exceptional case in case of stave's stavemodifier handling, but now it works similarly like other stave modifiers.

#### Auto alignment of BEGIN && END stave modifiers
Now, stave automatically fix order between clef, time, key and barline.
Like this example below, I didn't change test codes but result is different,

[Previous]
key_time_before
[Now]
key_time_auto_align

[Previous]
timesignature basic_time_signatures_a
[Now]
timesignature basic_time_signatures_b
[Diff]
timesignature basic_time_signatures_diff

#### Although test case which is over threshold 1.0 is only TimeSignature.basic_time_signatures but there were also another cases which cannot be ignored

Rhythm.Rhythm_Draw___32nd_note_rhythm_with_scratches 0.245074
Rhythm.Rhythm_Draw___16th_note_rhythm_with_scratches 0.0515266
Rhythm.Rhythm_Draw___beamed_slash_notes__some_rests 0.0175269
=> KeySignature and TimeSignature order changed. ( clef - time - key => clef - key - time )

Clef.Small_Clef_End_Test 0.0373351
Clef.Clef_End_Test 0.0272275
Clef.Clef_Test 0.0182327
Clef.Small_Clef_Test 0.00800095
=> Stave end modifier first padding value changed: 2 => 5


0xfe:
Hi Taehoon, This is a great change, thanks very much! And thanks for running the visual regression tests. I'll review this soon.

src/stave.js
```diff
@@ -319,14 +298,96 @@ Vex.Flow.Stave = (function() {
...
 
+    getModifiers: function(position) {
+      if (position === undefined) return this.modifiers;
+
+      return this.modifiers.filter(function(modifier) {
+        return modifier.getPosition() === position;
+      });
+    },
+
+    sortByCategory: function(items, order) {
```
0xfe: I think you can simplify this with `array.sort` that can take a comparison function.

src/stave.js
```diff
@@ -319,14 +298,96 @@ Vex.Flow.Stave = (function() {
...
+          if (order[items[j].getCategory()] > order[items[j + 1].getCategory()]) {
+            var temp = items[j];
+            items[j] = items[j + 1];
+            items[j + 1] = temp;
+          }
+        }
+      }
+    },
+
+    format: function() {
```
0xfe: Can you add a unit test for this function?

panarch: I had to write more details, thanks for good catching. I agree with you and it's better to be removed in the future, but for now I needed to use bubble sort not to break existing order between same category items.

I think Stave is better to allow only one clef, time, key and barline for each BEGIN and END positions. if someone wants to render exceptional special cases, they can use TimeSigNote or ClefNote, etc.
Currently, there are many test cases which add more than one BEGIN or END stavemodifiers into a single stave. I didn't want to break and modify existing test cases this time so that was the reason that I made that kind of custom bubble sort function.

How do you think of preventing adding more than one stavemodifiers which has BEGIN or END positions?

Current solution I thought is this,
1. Updating stave's handling modifiers not to allow multiple same position(BEGIN or END) && category stavemodifiers, during this update I can remove that sortByCategory function.
2. Adding setTimeSignature, setClef, etc(renaming from adding functions, but I will keep existing add functions for compatibility)
3. Updating all tests which adding more than one that kind of stavemodifiers into a single stave, splitting into multiple staves may be good way to change. targets will be test cases like these,
screen shot 2016-01-19 at 4 37 38 pm
screen shot 2016-01-19 at 4 37 58 pm

0xfe: On Tue, Jan 19, 2016 at 2:39 AM, Taehoon Moon <notifications@github.com>
wrote:

> In src/stave.js
> <https://github.com/0xfe/vexflow/pull/320#discussion_r50080487>:
>
> > +      });
> > +    },
> > +
> > +    sortByCategory: function(items, order) {
> > +      for (var i = items.length - 1; i >= 0; i--) {
> > +        for (var j = 0; j < i; j++) {
> > +          if (order[items[j].getCategory()] > order[items[j + 1].getCategory()]) {
> > +            var temp = items[j];
> > +            items[j] = items[j + 1];
> > +            items[j + 1] = temp;
> > +          }
> > +        }
> > +      }
> > +    },
> > +
> > +    format: function() {
>
> I had to write more details, thanks for good catching. I agree with you
> and it's better to be removed in the future, but for now I needed to use
> bubble sort not to break existing order between same category items.
>
> I think Stave is better to allow only one clef, time, key and barline for
> each BEGIN and END positions. if someone wants to render exceptional
> special cases, they can use TimeSigNote or ClefNote, etc.
> Currently, there are many test cases which add more than one BEGIN or END
> stavemodifiers into a single stave. I didn't want to break and modify
> existing test cases this time so that was the reason that I made that kind
> of custom bubble sort function.
>
> How do you think of preventing adding more than one stavemodifiers which
> has BEGIN or END positions?
>
It sounds like a good idea, but I think it might not be backwards
compatible. Will it break existing users?

I think the custom sort is fine if there's a good reason for it, and it
seems like this is a good reason.


> Current solution I thought is this,
> 1. Updating stave's handling modifiers not to allow multiple same
> position(BEGIN or END) && category stavemodifiers, during this update I can
> remove that sortByCategory function.
> 2. Adding setTimeSignature, setClef, etc(renaming from adding functions,
> but I will keep existing add functions for compatibility)
> 3. Updating all tests which adding more than one that kind of
> stavemodifiers into a single stave, splitting into multiple staves may be
> good way to change. targets will be test cases like these,
> [image: screen shot 2016-01-19 at 4 37 38 pm]
> <https://cloud.githubusercontent.com/assets/2025065/12412578/15b2ce48-becb-11e5-91d7-410f7fce22f9.png>
> [image: screen shot 2016-01-19 at 4 37 58 pm]
> <https://cloud.githubusercontent.com/assets/2025065/12412579/15b491d8-becb-11e5-8caa-7ecc69a9234a.png>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/0xfe/vexflow/pull/320/files#r50080487>.
>



-- 
Mohit Muthanna [mohit (at) muthanna (uhuh) com]

panarch: Allowing only one stavemodifier for each category + position combination will certainly have affect to current tests, but that test examples (adding multiple time signature, key or clef to a single stave) are not used in real world so I think most of existing users will not experience any change. 

And there's another good thing, users will be able to change existing stave modifiers(clef, time or key) although they already added clef, time or key signature.

Even if there exists users who want to draw multiple clefs(or etc) in a row in the single stave, they can do by using clefnote of timesignote.

Hmm... may be we can keep both addClef..+ and setClef...+
It might be better to only simply setting deprecated flag comments to that add functions.
How do you think about current add functions? (addClef, addEndClef, addTimeSignature, addEndTimeSignature and addKeySignature)

panarch: I just added sortByCategory unit test to stave_tests.

And today I implemented stave modifier set functions, I didn't push yet;
setClef, setEndClef, setKeySignature, setEndKeySignature, setTimeSignature and setEndTimeSignature.

We can keep both add and set. It will not also break any existing tests and no effect to existing users.
Another thing I'd like to get your opinion is about updating current test cases.
I think it's better to replace all current addClef and etc add functions test cases to use set functions except some test cases which add multiple stave modifiers into a single stave.
How do you think of that?
Both set and add are ok to use but in usual cases I think set functions fit more.

0xfe: Thanks for adding the `sortByCategory` tests. Also for the `set` functions.

Re: test cases, please add new tests instead of replacing, just so that we don't accidentally break `add` in the future. We can deprecate both `add` and the `add` tests, and update any documentation to make it clear. How does that sound?

SalahAdDin:
Finally, one update for VexFlow!

src/stave.js
panarch: Sounds nice! I agree with you. Some kind of removing work is better for the future.

I added commits of `set` function implementations and also added a new test case.

This test case also verify working of changing begin & end stavemodifier after drawing.
screen shot 2016-01-22 at 5 41 02 pm

There exists too many test cases which use `add` because some functions like `addClef` or `addTimeSignature` are very essential so used in many other kinds of test cases. How to make deprecated might be better to have a discussion more.

0xfe:
Hi Taehoon, Is this good to go? (I'm not sure if you're waiting on me.)

panarch:
@0xfe Yes it is!
I was preparing the next step, I will update those to issue page.

0xfe:
Looks great. Thanks for the contribution. I'll run some checks on my side then merge.

panarch:
Thanks! And there are two more updates I did. I will send PR after merged since those are independent of this PR, it's on my fork - https://github.com/panarch/vexflow/commits/slimerjs.

One is StaveConnector simple improvements,
and you may like the other which is about simplifying test steps; Replace from jsdom & xmldom to [SlimerJS](http://slimerjs.org/) So, no more manual text cache update is required.

0xfe:
Everything looks good. Merged!

I'm curious about your SlimerJS support -- it sounds great. Does it work on Linux and OSX?

panarch:
Thanks!!
Yes, both Linux and OSX are no problem to use SlimerJS.
In case of desktop, there's no more things to do. But someone wants to run SlimerJS in linux server, because SlimerJS is not headless so additional package [xvfb](http://tobyho.com/2015/01/09/headless-browser-testing-xvfb/) is required. I tested SlimerJS also works well in my ubuntu server.

It just runs Firefox so we don't need to worry about any kind of node.js lack of features.

I'll add PR about StaveConnector tiny improvements first. :) thanks!

src/stavebarline.js
```diff
@@ -23,21 +23,45 @@ Vex.Flow.Barline = (function() {
   };
 
   Vex.Inherit(Barline, Vex.Flow.StaveModifier, {
-    init: function(type, x) {
+    init: function(type) {
```
0xfe: Looks like this change broke VexTab :-(

`src/barnote.js` creates `BarLine` elements and supplies the `x` parameter to the constructor.

0xfe: It looks like a simple fix. I'm pushing it now and adding some tests.

PR #322, opened 2016-02-05
StaveConnector: NONE type, allowing multiple texts and support brace type x_shift

panarch:
Three updates to StaveConnector!

1. NONE type
Added new StaveConnector type NONE, users can use NONE type when they want to render only text without drawing connector.
2. Allowing multiple texts
Unlike Stave, it was not possible to add multiple texts to StaveConnector (this feature was what I did 2 years ago.. haha), now StaveConnector supports adding multiple texts.
3. Brace x_shift support
There are some cases which both brace and bracket staveconnectors added. Like the example below,

Modified test case (below) now uses all NONE type, multiple texts and brace x_shift.

* Before

screen shot 2016-02-05 at 2 44 16 pm

* After

screen shot 2016-02-05 at 2 44 20 pm


0xfe:
Thanks. Merged.

PR #330, opened 2016-03-20
Tuplet Improvements

### LINKED ISSUE #329: Tuplets / Triplets not fully supported
** QUICK FIX requested. Detailed instructions given. This is a half-hour job.

See graphics below for cases not handled: two notes of unequal duration that together form a triplet (or in the general case, n notes of unequal duration that together form an m-tuplet).

The tuplet module takes two integer prarmeters,
num_notes and 
beats_occupied

Currently, parameter num_notes sets the number of notes to be affected, in this case 2, but also sets the glyph above the bracket, which should be 3 in this case. 

Notice also that the scaling is incorrect so that notes sounding together are not printed vertically aligned. In this case, all note durations should be scaled to take up 2/3 of their usual horizontal space.

In the most general case, tuplets require three integers to be specified: the number of notes falling under the bracket, and the ratio that the notes are scaled by. In the most common case, the ratio is 2/x (i.e. quintuplets are scaled by 2/5). By convention, in the 2/x cases the bracket says only x (3 for triplets, 5 for quintuplets), otherwise the (inverted) scaling ratio is shown. 

It would be possible to cover the great majority of useful cases with only two integers of input, by assuming that the numerator of the scaling ratio is 2. The module currently takes this numerator (as beats_occupied) but NOT the important denominator. It would be a very quick fix to the code to change the spec so that the second input "beats_occupied" functioned as the denominator here; both for glyph and scaling purposes.

The most general implementation would need to take both a numerator and denominator for the scaling ratio; the third graphic below shows a score by Stockhausen that requires this notation. This is correct, meaningful notation which should ideally be representable, but these are "advanced" cases that are not used nearly as often as those in the first graphics, which are extremely common and should definitely be represented. 

It may be slightly more complicated to do this more general fix, since modules other than tuplet.js would need to be touched, as the input type to tuplet is changed, but the easier fix mentioned above would only affect this module, and current correct functionality would be maintained (e.g. num_notes = 3, beats_occupied = 2).

tuplets1
tuplets2
klavierstucke_ stockhausen _nr2-klavierstuck1-anfang

gristow:
This PR resolves #329.

- Added support for nested tuplets, and autodetection of tuplet collisions by implementing a tuplet stack on `tickable`:
image
- Refactored option `beats_occupied` to `notes_occupied` in `tuplet.js`, to clarify that the two parameters `num_notes` & `notes_occupied` together form the tuplet ratio, and that they may in fact have nothing to do with the beat.
- Added extensive commenting of `tuplet.js` to better document parameters & options.
- Added `location`, `ratioed`, `bracketed` properties to `options` in `tuplet.js`
- Deprecated use of `beats_occupied`, `setBeatsOccupied` and `getBeatsOccupied` in `tuplet.js` for `notes_occupied` variants.


SalahAdDin:
:+1:

panarch:
Thanks for starting from my branch :+1:

0xfe:
Merged. Thanks for doing this! If you could update the wiki with some of your findings, that would be great.

gristow:
Thanks -- I updated the wiki. (I hadn't really realized how hard writing documentation is.) Hopefully it'll be clear and helpful in the future.

SalahAdDin:
Yes, it's very hard write the documentation, like sphinx packages too.

0xfe:
Wow, good doc!

PR #337, opened 2016-04-28
Feature/delete notes

khisakuni:
- Defines method `Vex.Flow.Voice.deleteNote`
- Takes in a number for the index of the `StaveNote` to be "deleted"
- The `StaveNote` will be replaced with a rest of equal duration
- The updated voice is returned.

I'm working on an app that uses vexflow to allow users to interactively score music. Right now I have a wrapper around vexflow that has methods that modifies vexflow objects, but I think it would be useful to add some of these methods to the vexflow library. Other methods would be things like `Vex.Flow.raisePitch` and `Vex.Flow.lowerPitch` which would mutate vexflow objects. What do you think?

### DIFF
--- a/src/voice.js
+++ b/src/voice.js
@@ -50,7 +50,7 @@ Vex.Flow.Voice = (function() {
       this.voiceGroup = null;
     },
 
-    // Inserts rest of equal duration at index
+    // Inserts rest of equal duration at provided index
     deleteNote: function(index) {
       var duration = this.tickables[index].duration + "r";
       this.tickables[index] = new Vex.Flow.StaveNote({duration: duration, keys: ['b/4']});
--- a/src/voice.js
+++ b/src/voice.js
@@ -50,14 +50,11 @@ Vex.Flow.Voice = (function() {
       this.voiceGroup = null;
     },
 
-    sayHi: function() { return "hi"; },
-
+    // Inserts rest of equal duration at index
     deleteNote: function(index) {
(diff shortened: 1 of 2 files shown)


mscuthbert:
I wouldn't think that "deleteNote" would replace it w/ a rest.  Would "noteToRest()" be just as good a term?

Silverwolf90:
In general, mutating VexFlow objects after specific parts of their lifecycle can lead to unexpected behavior. It's probably fine in this case, but I don't these kind of methods should be built into VexFlow.

SalahAdDin:
@Silverwolf90 so, what methods recommends you?

Silverwolf90:
There are no specific methods to recommend. VexFlow was not designed with the intention of mutating objects at any point in time.

If you want to dynamically change things at any point in time, I would not use VexFlow as your data model. I would suggest rolling your own model, and whenever this model changes you rebuild all the VexFlow objects and re-render.

0xfe:
Hi @khisakuni, Thanks for the PR.

@Silverwolf90 is spot on here -- VexFlow is a rendering engine, and we don't want to couple it too tightly with a UI model. It's much better to have a separate model, one that you can mutate more easily, and generate the final score from it. This will give both VexFlow and the UI a lot more flexibility, and they can be improved upon, optimized, and maintained independently.

Also, putting this stuff inside VexFlow adds constraints to the design and the API of VexFlow, and vastly increases its complexity, which is something I'd strongly like to avoid.

PR #344, opened 2016-06-05
Stroke dasharray

gristow:
The `context.setLineDash` functions in both `SVGContext` and `CanvasContext` were not actually resulting in a dashed line. I've fixed the underlying API calls in both cases, added some more obvious visual test elements for the one object that uses these (`StaveLine`), and included blessed, current and diffed images.

**Before:**
image

**After:**
image

**Diffs:**
staveline simple_staveline
staveline staveline_arrow_options


0xfe:
Thanks @gristow. Will review and merge.

0xfe:
Merged, thanks!

PR #347, opened 2016-06-09
fix(stavenote.js): getBoundingBox should not add extra height to chords

### LINKED ISSUE #346: StaveNote.getBoundingBox() inconsistent height results
When StaveNote contains one whole note, `StaveNote.getBoundingBox()` returns the correct height:
image

But when two whole notes are present, extra height is added:
image

Although not when stemmed notes are present:
image

I haven't had a chance to look at the code for this one yet, but will do so in the next week or so.

gristow:
Fixes #346 -- two-lines of code erroneously placed in a for-loop caused an extra half-staff-line of space to be added to `getBoundingBox().height` for every note after the first in a `StaveNote` in stemless rhythms (whole notes & breves).

This PR addresses & corrects that:
image

### DIFF
--- a/src/stavenote.js
+++ b/src/stavenote.js
@@ -505,9 +505,9 @@ Vex.Flow.StaveNote = (function() {
             min_y = Vex.Min(yy, min_y);
             max_y = Vex.Max(yy, max_y);
           }
-          min_y -= half_line_spacing;
-          max_y += half_line_spacing;
         }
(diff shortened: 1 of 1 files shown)


0xfe:
Good catch!

0xfe:
FYI, the visual regression tests also show a fix! :-)

Before:
image

After:
image

Diff:
image

PR #355, opened 2016-06-16
Added support for 3,9,10 string tabstaves, needs visual regression test

A-Manning:
Passed all other tests, but I have yet to figure out visual regression tests on my system.

### DIFF
--- a/src/clef.js
+++ b/src/clef.js
@@ -195,7 +195,7 @@ Vex.Flow.Clef = (function() {
       var glyphScale;
       var glyphOffset;
       switch(this.stave.getOptions().num_lines) {
-        case 10:
+        case 10: //TODO: Test with Visual Regression
           glyphScale = 69; 
           glyphOffset = 26;
           break;
--- a/src/clef.js
+++ b/src/clef.js
@@ -195,6 +195,14 @@ Vex.Flow.Clef = (function() {
       var glyphScale;
       var glyphOffset;
       switch(this.stave.getOptions().num_lines) {
+        case 10:
+          glyphScale = 69; 
+          glyphOffset = 26;
+          break;
+        case 9:
+          glyphScale = 62; 
+          glyphOffset = 20;
+          break;
         case 8:
           glyphScale = 55;
           glyphOffset = 14;
@@ -214,6 +222,10 @@ Vex.Flow.Clef = (function() {
         case 4:
           glyphScale = 23;
           glyphOffset = -12;
+          break;
+        case 3:
+          glyphScale = 16;
+          glyphOffset = -18;
           break;
       }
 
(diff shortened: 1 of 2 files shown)


Silverwolf90:
Sorry that this has been sitting around for so long. I've recently had a bit of time to do some maintenance.

A few things are making me reluctant to merge this:

1. Now that we have real glyph metrics, we should be able to automatically determine the necessary size of the TAB glyph based on some proportion of the staff height. So I'd rather rip out this switch statement than extend it.
2. The `Vex.Tuning` object isn't scaling well. The naming convention is awkward (e.g. `standard5bass`) because we're forced to encode string counts and instrument in the name of the scale. Maybe changing the structure so we could access stuff like `Vex.Tuning.guitar[7].dropA`. But that could also get out of hand.

I'm gonna give this a bit more thought.

A-Manning:
Agreed on both points, feel free to close. 
Maintaining a dictionary of reasonable common, named tunings isn't a bad idea though, there are only so many.

0xfe:
Yeah, I don't see named tunings as a big issue either -- it makes it simpler for things like VexTab, where you can just add `tuning=standard7` (or some such) to quickly configure the tab. 

Agreed about automatically determining TAB glyph size.

Silverwolf90:
I tried to look for an independent dataset of guitar tunings but couldn't find anything :) So I guess we're kinda stuck defining this on our own. In that case, we should establish better conventions because this ad-hoc naming is leading to inconsistency/weirdness:

* The guitar tunings don't have `guitar` as part of their name but `standardbass` does have the word `bass`
* `standardBanjo` is camelCase
* `standardd` is not camelCase
* `dropd` is not camelCase
* `dagdad` is [far more widely known in the opposite direction ](https://en.wikipedia.org/wiki/DADGAD), in fact tunings are more typically declared from the lowest (in pitch) strings first. To have the tuning name reflect the opposite order most people would expect is about as unintuitive as it gets.
* `eb` generally I think of this as "half step down" -- suggests that a tuning could have multiple names

I think my main problem is the naming is inherently opinionated and, in our case, biased towards guitarists. These characteristics are more appropriate in something higher-level like VexTab than in VexFlow.

What about defining tunings like this:

```javascript
const tunings = [
  {
    instrument: 'guitar',
    strings: ['E/5', 'B/4', 'G/4', 'D/4', 'A/3', 'E/3'],
    names: ['standard', 'eadgba'],
  }
];
```

This data is far more flexible and useful for people building applications on top of VexFlow. Maintaining a simple string mapping is very high-level idea.

0xfe:
I like your proposal for the structured tunings. (Not super convinced about `instrument` though, because `guitar` itself has lots of sub-categories.)

Although I see the naming inconsistencies, I wouldn't consider this high on my priority list to fix -- this is the kind of thing I'd let stew for longer until we can be more confident of what we really want.

Silverwolf90:
Agreed!

Silverwolf90:
I'm going to close this now. I've opened issue #514 which points to this PR.

PR #357, opened 2016-06-16
ES6 modules

aaronmars:
Remove the concept of a `Vex` global in the sources, and re-introduce it at build time for compatibility with existing code.

Use Rollup.js to build a bundle that behaves exactly like the existing global-based one.


aaronmars:
Travis failed.  Will re-send with the fix.

PR #358, opened 2016-06-16
Es6 modules

aaronmars:
Remove the concept of a Vex global in the sources, and re-introduce it at build time for compatibility with existing code.

Use Rollup.js to build a bundle that behaves exactly like the existing global-based one.


Silverwolf90:
Thanks so much for doing this!

Any particular reason you didn't use `export default`? Wouldn't that be ideal when we have use the class syntax?

aaronmars:
I've found named exports to be more flexible and declarative.  When there's just one thing exported, I guess it's just a matter of preference.  I have no problem going that way.

Having said that, I don't think it makes a difference when we go to class syntax... 

```
export class Accidental extends Modifier {
  // ....
}
```
then:
```
import { Accidental } from './accidental';
```
just as we have now.

0xfe:
Going through all the changes, this is really wonderful! I can't wait to create the dependency graph of how modules use one another.

Few questions:
- Does rollup just give you ES6 module syntax or does it come with an ES6 transpiler? i.e., if I start using ES6 syntax in my code, what happens?
- What is the process of adding a new file? Just update `index.js`?
- Why do you have just `import './glyph'` in `tremolo.js` instead of `import { Glyph } from './glyph'`? Can we keep it all consistent?

aaronmars:
> Does rollup just give you ES6 module syntax or does it come with an ES6 transpiler? i.e., if I start using ES6 syntax in my code, what happens?

Rollup just transforms the module syntax. In this state, your ES6 syntax will remain un-transpiled.  I have a working branch that integrates babel into the toolchain!  Looks good so far.  Let me know if you want a PR with that now, or if you want it to chill for a bit as-is.

> What is the process of adding a new file? Just update `index.js`?

Exactly!  The dependencies are defined in code now, rather than some build config looking at the filesystem.  What's also awesome here is the "public" API is defined by index.js.  If it's not in there, it's not part of the Vex.Flow namespace.

> Why do you have just `import './glyph'` in tremolo.js instead of `import { Glyph } from './glyph'`? Can we keep it all consistent?

That's part of the weirdness I mentioned earlier regarding `Flow.renderGlyph()`.  That function is defined globally in glyph.js, so by including it without a name, renderGlyph() gets defined, but the Glyph class name is not imported (it's not used in tremolo.js).  A linter looking for unused variables would indicate an error if you did `import { Glyph } from './glyph';`, and then never used it.

0xfe:
Thanks for the answers.

- The babel stuff sounds great and I def want to get it soon, but lets put it in a followup PR.

- re: `renderGlyph`, how about we kill this now and move it to `Vex.Flow.Glyph.renderGlyph`, i.e., make it a static method of `Vex.Flow.Glyph`? (A quick search show about 15 references to it.) Does that work?

aaronmars:
Seems legit.  Let's go for it.
I'll update this guy.

0xfe:
Merged! Thanks very much for this totally excellent PR!

Silverwolf90:
Really exciting stuff! :)

mscuthbert:
Very cool.  Are the mapping files working with the new version?  It looks like the sources array in the vexfloow-debug.js.map now has absolute paths on the filesystem (`"/Users/mmuthanna/git/vexflow/src/vex.js"`) instead of relative paths (`"../src/vex.js"`), which is a potential security concern (OMG now I know the hidden folder where 0xfe keeps his git files! It's `~/git`!).

0xfe:
Ugh! Thanks for catching that before the evil hack3rz did. :-)

0xfe:
Seems like a rollup idiosyncrasy. Fixed!

aaronmars:
Thanks!

PR #362, opened 2016-06-18
Expose Vex.Flow.Font to fix glyphs.html

infojunkie:
The file glyphs.html uses `Vex.Flow.Font` to display the glyphs, but that object was no longer being exposed after the ES6 rewrite. Fixed this by adding `Vex.Flow.Font` to index.js.

### DIFF
--- a/src/glyphs.html
+++ b/src/glyphs.html
@@ -39,10 +39,7 @@
     }
   </style>
 
-  <script src="vex.js"></script>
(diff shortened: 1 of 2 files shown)


aaronmars:
Thanks for this.  Sorry I missed it!

0xfe:
Thanks. Merging.

PR #367, opened 2016-06-20
Add Getter and Setters for staff spacing

mscuthbert:
Adds getters and setters for staff spacing tweaks in this.options. — not 100% sure if options is supposed to need getter and setter access, but these should redraw lines anyhow.


mscuthbert:
Sorry -- I didn't realize this was going to get so many merges as well.  closing and will include just 34c91e7

PR #368, opened 2016-06-20
adding getter/setter methods for staff spacing

mscuthbert:
Adds getters and setters for staff spacing tweaks in this.options. — not 100% sure if options is supposed to need getter and setter access, but these should redraw lines anyhow.

### DIFF
--- a/src/stave.js
+++ b/src/stave.js
@@ -96,6 +96,23 @@ export class Stave {
     this.resetLines();
     return this;
   }
+
+  /**
+   *  The space above and below staff lines is measured in staff lines 
+   */
+  getSpaceAboveStaffLines() { return this.options.space_above_staff_ln; }
+  setSpaceAboveStaffLines(space) {
+    this.options.space_above_staff_ln = space;
+    this.resetLines();
+    return this;
+  }
+  getSpaceBelowStaffLines() { return this.options.space_below_staff_ln; }
+  setSpaceBelowStaffLines(space) {
+    this.options.space_below_staff_ln = space;
+    this.resetLines();
+    return this;
+  }
+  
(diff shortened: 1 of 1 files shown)


0xfe:
Thanks Michael... not sure that we should support changing staff spacing post-construction. I'd prefer to just put this in a doc (and actively discourage this pattern.)

(Also, this whole file is a mess, which is a different discussion.)

mscuthbert:
Got it -- I suppose an important thing I'm unclear on (and maybe has not yet been decided) is: what are private properties and what aren't in the Vexflow world.  For instance, there is no `Stave.getOptions()` method; does this mean `s = new Stave(); s.options.line_spacing = 10` is a no-no, or is this okay?  For a while it looked like Vexflow was going in the Java direction: no method = don't touch.  But it looks more like it's going in the Python direction.  I'd like that -- if so feel free to close this and I'll submit docs instead.

0xfe:
Yep. There shouldn't be anything that is directly accessed without a getter/setter. I went through a big cleanup a while ago where I tried to remove any direct references to instance variables (but there might be some leaks since then.) The goal here is to keep the API surface from getting out of control.

So your example of `s.options.line_spacing = 10` is not okay and that option should be provided at construction time. If you need to change the spacing on an already constructed object, then create a new object. (It's just simpler to maintain invariants this way.)

BTW, I'm open to adding a getter for `line_spacing` but I'd prefer not to unless there's a good reason for it. It's probably better if you maintain the model for your staves out-of-band.

mscuthbert:
Thanks -- a lot of these routines appeared to be things that were helpful back before .getYForLine() and .setLineConfig() were added I think it's something that can be closed w/o accepting.  Thanks!

Issue #361, opened 2016-06-18
Migrate to ESLint


Silverwolf90:
As we start to move over to ES6, I wanted to propose using ESLint over JSHint.

Reasons:

* More flexible, powerful
  * Larger set of more configurable rules
  * Supports custom rules, we could create our own `vexflow` specific set of eslint rules that suggest best practices, or warn of improper usage.
* Some errors can be automatically fixed with a CLI flag: `eslint --fix`
* It's currently winning the linter wars
  * JSCS team decided to make 3.0 the last major version and join the ESLint team

I personally use the [airbnb](https://github.com/airbnb/javascript) style guide's [`eslint-config`](https://www.npmjs.com/package/eslint-config-airbnb), with some tweaks. For the most part it's a great starting place, but I find it overly strict in some cases. But in general, VexFlow would benefit significantly from having stricter linting rules.

Silverwolf90:
My ideal `.eslintrc` would look something like this:
```json
{
  "extends": "airbnb",
  "rules": {
    "max-len": [2, 80],
    "no-param-reassign": [0],
    "no-shadow": [0],
    "no-use-before-define": [0],
    "strict": [2, "global"],
    "no-else-return": [0],
    "space-before-function-paren": [2, "never"],
    "no-case-declarations": [0],
    "func-names": [0],
    "prefer-template": [0],
    "no-confusing-arrow": [0],
    "no-multi-spaces": [0],
    "new-cap": [0],
  }
}
```

aaronmars:
Yes! I support this 100%.

mscuthbert:
I like this a lot too (and thanks for the great link to the airbnb code).  One thing that the new class system proposal would run afoul of is the `export class Thing ...` paradigm -- I think I like the airbnb idea of imports at the top of the file and exports at the bottom of the file.

Silverwolf90:
WIth the above config:

[Here's the result of running `eslint --fix`](https://github.com/Silverwolf90/vexflow/commit/67618d41b16107994e557b50e83fc4fa8f511279)

[Here's an example of a `annotation.js` after manually editing remaining lint errors](https://github.com/Silverwolf90/vexflow/commit/f24967dacf98f9baf9c3116ed18c4e22e2f75ea3)

0xfe:
Looks great. I like the final `annotation.js` too.

[...]

PR #369, opened 2016-06-21
Use eslint: accidental, formatter, stavenote, and modifiercontext

Silverwolf90:
This is a WIP. But I figured I'd open a PR if anyone wants to comment as I go.

- [x] accidental.js
- [x] modifiercontext.js
- [x] stavenote.js
- [x] formatter.js
- [x] Add eslint grunt task


0xfe:
A lot of this looks really good. Can't tell how much of this is machine translated, and how much you're doing by hand.

src/accidental.js
```diff
@@ -422,44 +448,53 @@ export class Accidental extends Modifier {
...
-    L("Rendering: ", this.type, acc_x, acc_y);
+    const start = this.note.getModifierStartXY(this.position, this.index);
+    let accX = ((start.x + this.x_shift) - this.width);
+    const accY = start.y + this.y_shift;
+    L('Rendering: ', this.type, accX, accY);
 
     if (!this.cautionary) {
       // Render the accidental alone.
-      Glyph.renderGlyph(this.context, acc_x, acc_y,
+      Glyph.renderGlyph(this.context, accX, accY,
```
Silverwolf90: I'm not a fan of this style with longer lists of arguments. I would prefer to put each argument on their own line.

src/accidental.js
```diff
@@ -422,44 +448,53 @@ export class Accidental extends Modifier {
...
     } else {
       // Render the accidental in parentheses.
-      acc_x += 3;
-      Glyph.renderGlyph(this.context, acc_x, acc_y,
-                           this.render_options.font_scale, this.paren_left.code);
-      acc_x += 2;
-      Glyph.renderGlyph(this.context, acc_x, acc_y,
+      accX += 3;
+      Glyph.renderGlyph(this.context, accX, accY,
+                           this.render_options.font_scale, this.parenLeft.code);
```
0xfe: This alignment looks weird (also the next few lines.) I think `this.context` and `this.render_options` should be column-aligned.

Silverwolf90:
Everything after and including `ab4a032` is done by hand.

src/accidental.js
0xfe: I'm okay with argument-per-line, but I don't get the additional level of indentation. It hurts my eyes :-)

src/accidental.js
Silverwolf90: What about

```
Glyph.renderGlyph(
  this.context,
  accX,
  accY,
  this.render_options.font_scale,
  this.paren_left.code
);
```

0xfe: Yes, this works.

Silverwolf90: Cool, that's what I was trying to say in that other comment, should have provided an example.

src/accidental.js
```diff
@@ -465,36 +465,42 @@ export class Accidental extends Modifier {
...
   // Render accidental onto canvas.
   draw() {
-    if (!this.context) {
+    const {
+      context,
+      type, position, note, index, cautionary,
+      x_shift, y_shift, width,
+      accidental, parenLeft, parenRight,
+      render_options: { font_scale },
+    } = this;
```
Silverwolf90: how do you feel about this pattern?

Silverwolf90: I find it helpful to logically group properties when destructuring, particularly with a lot of properties. But it's not necessary.

0xfe: Doesn't look very readable -- not clear how all that is destructured from `this`.

Silverwolf90: Wouldn't that rule out all multi-line destructuring calls?

0xfe: Multi-line destructuring is okay. I guess I didn't understand how those variables were destructured from `this`, and now I do. It's definitely succinct. I don't have a strong opinion here, so you can leave it as is.

Silverwolf90:
Ok well, getting these files into a decent state is taking longer than I thought. 

I would suggest, when these 4 files are complete and agreed upon, that we run `lebab` (for arrow functions, let/const, shorthand properties) and `eslint --fix` on the entire codebase. That should be pretty safe. But the code will require a lot of manual edits and minor refactorings to get into as good of shape as the one's I'll be doing in this PR.

Silverwolf90:
Any ideas on how we would migrate? One option would be to use both jshint and eslint, keeping a list of files that can be used with eslint and the remaining files use jshint. Or we could disable lint errors from failing the build until we've run through the entire codebase.

0xfe:
Maybe start a wiki on instructions on running `lebab` and `eslint` and maybe some common gotchas or other hints to help migrate, and we can split the work. Also, we don't have to go to the level of detail that you're going into right now (which is great, btw), we can improve it incrementally.

So, send a PR with what you have now (with passing build and tests), and we'll hand maintain a list of files for eslint. Between you, @aaronmars (if he's interested), and me, even if we do 1-2 files a day each, we could have it all done in a week probably and get rid of jshint.

aaronmars:
Sign me up for this effort. I'm down to take a slice of the files and help with the conversion

0xfe:
Great, thanks Aaron. Also paging @mscuthbert and @gristow incase they're interested :-)

mscuthbert:
Oh, I'm interested :-)  My experience from switching lint systems on a large project about two years ago would be to just make a big leap and switch the lint to eslint immediately, and get that in as a PR with an extremely minimal config setting that maybe just flags one thing, so we know it's working.  And get the documentation up to date to match the reality of how things are going.  Then make successive pull requests that each enable one or two new settings and fix all the errors that come with them.

The drawback is that there'll be a short-term period where our lint situation is worse than before, since we're not running tests that are even as stringent as what we had before.  If this repository were github.com/nasa/mars-rover or jarvis/artificial-heart, then that'd be a problem, but given that the stakes of the possibility of a global variable leaking or a line being too long are...not that high, it think the risk is worth saving the time.

gristow:
Also happy to help with these conversions, though I'm tied up with "real" work for the next ten days or so. 

Would it be worth deciding who'll tackle which files/classes? (My first picks would be: beam.js, fraction.js, svgcontext.js, canvascontext.js, raphaelcontext.js, and renderer.js. Though, I'm also happy to tackle others as well.)

Regarding Raphael, I wonder if we're ready to deprecate it completely at this point?

mscuthbert:
OH Yes;  happy to do anything to deprecate Raphael.  :-)

Silverwolf90:
@0xfe Yeah I'll try and set up wiki page up within the next couple days

@mscuthbert To be cynical -- since VexFlow has typically been a slow moving project -- it's possible that this conversion might not be a "short-term period." So I'm personally hesitant to convert the codebase rule-by-rule instead of file-by-file. Going file-by-file, it seems that the work is more focused, and files are either to-do or done, rather than the entire codebase being in a half-state.

Silverwolf90:
Oh and :+1: to deprecating Raphael

Silverwolf90:
Once the build passes this should be good to go. I might have gone a bit overboard... hopefully this isn't too much at once.

Still no dice on the visual regression tests (I may just bite the bullet and get a VM set up). I don't know much about Travis -- can it store build artifacts?

0xfe:
Wow, lots going on here. Thanks, let me go through everything today.

0xfe:
Code looks great, but it looks like some subtle new bug is introduced that messes with the formatting. (I think it's in `StaveNote`.)

```
You have       11 fail(s):
Three_Voice_Rests.Auto_Adjust_Rest_Positions___Three_Voices__1 42.7349
Auto_Beaming.Beam_Across_All_Rests_with_Stemlets 32.4864
Rests.Auto_Align_Rests___Beamed_Notes_Stems_Up 13.001
Auto_Beaming.Beam_Across_All_Rests 11.4103
Three_Voice_Rests.Auto_Adjust_Rest_Positions___Three_Voices__2 3.45775
Three_Voice_Rests.Auto_Adjust_Rest_Positions___Two_Voices 1.58174
Rests.Auto_Align_Rests___Beamed_Notes_Stems_Down 0.828317
Rests.Auto_Align_Rests___Single_Voice__Default_ 0.667154
Rests.Auto_Align_Rests___Single_Voice__Align_All_ 0.219455
Rests.Auto_Align_Rests___Multi_Voice 0.116768
Auto_Beaming.Break_Beams_on_Middle_Rests_Only 0.114961
```

three_voice_rests auto_adjust_rest_positions___three_voices__2_blessed
three_voice_rests auto_adjust_rest_positions___three_voices__2_current
three_voice_rests auto_adjust_rest_positions___three_voices__2
three_voice_rests auto_adjust_rest_positions___two_voices_blessed
three_voice_rests auto_adjust_rest_positions___two_voices_current
three_voice_rests auto_adjust_rest_positions___two_voices
three_voice_rests auto_adjust_rest_positions___three_voices__1_blessed
three_voice_rests auto_adjust_rest_positions___three_voices__1_current
three_voice_rests auto_adjust_rest_positions___three_voices__1
rests auto_align_rests___beamed_notes_stems_up_blessed
rests auto_align_rests___beamed_notes_stems_up_current
rests auto_align_rests___beamed_notes_stems_up
rests auto_align_rests___multi_voice_blessed
rests auto_align_rests___multi_voice_current
rests auto_align_rests___multi_voice
rests auto_align_rests___single_voice__align_all__blessed
rests auto_align_rests___single_voice__align_all__current
rests auto_align_rests___single_voice__align_all_
rests auto_align_rests___single_voice__default__blessed
rests auto_align_rests___single_voice__default__current
rests auto_align_rests___single_voice__default_
rests auto_align_rests___beamed_notes_stems_down_blessed
rests auto_align_rests___beamed_notes_stems_down_current
rests auto_align_rests___beamed_notes_stems_down
auto_beaming beam_across_all_rests_with_stemlets_blessed
auto_beaming beam_across_all_rests_with_stemlets_current
auto_beaming beam_across_all_rests_with_stemlets
auto_beaming break_beams_on_middle_rests_only_blessed
auto_beaming break_beams_on_middle_rests_only_current
auto_beaming break_beams_on_middle_rests_only
auto_beaming beam_across_all_rests_blessed
auto_beaming beam_across_all_rests_current
auto_beaming beam_across_all_rests

Silverwolf90:
Ok I'll take a look at that soon.

Silverwolf90:
Good to go!

Silverwolf90:
[I made a wiki page.](https://github.com/0xfe/vexflow/wiki/Migrating-to-ESLint) Let me know if anything else should be added.

[Also I would suggest we return conversation back to the main issue](https://github.com/0xfe/vexflow/issues/361) if we are content with the code looks in this PR.

0xfe:
Thanks Cyril. So I'm still seeing regressions in the rest positioning. Not sure if you're seeing the same:

```
Three_Voice_Rests.Auto_Adjust_Rest_Positions___Three_Voices__1 42.7349
Auto_Beaming.Beam_Across_All_Rests_with_Stemlets 32.4864
```

three_voice_rests auto_adjust_rest_positions___three_voices__1_blessed
three_voice_rests auto_adjust_rest_positions___three_voices__1_current
three_voice_rests auto_adjust_rest_positions___three_voices__1
auto_beaming beam_across_all_rests_with_stemlets_blessed
auto_beaming beam_across_all_rests_with_stemlets_current
auto_beaming beam_across_all_rests_with_stemlets

0xfe:
Wait, i think i may have screwed up here. BRB.

0xfe:
Yep, silly me just assumed that the regression tests would kick of a rebuild before running. Everything looks great. Merging!

PR #370, opened 2016-06-22
Fix call to Formatter.AlignRestsToNotes() to remove  keyword

aaronmars:
Apparently, this worked with the old call style, but it's not necessary to use `new` to call the static function.

### DIFF
--- a/src/formatter.js
+++ b/src/formatter.js
@@ -328,7 +328,7 @@ export class Formatter {
     if (!voices || !voices.length) throw new Vex.RERR("BadArgument",
         "No voices to format rests");
     for (var i = 0; i < voices.length; i++) {
-      new Formatter.AlignRestsToNotes(voices[i].tickables, align_all_notes);
+      Formatter.AlignRestsToNotes(voices[i].tickables, align_all_notes);
     }
   }
 


0xfe:
Yikes, not sure how that ever worked.

aaronmars:
Yeah! I was a bit surprised myself.

Silverwolf90:
[The ESLint rule `no-new`  will catch situations this.](http://eslint.org/docs/rules/no-new)

[MDN has a good explanation of how `new` works.](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/new)

> When the code` new Foo(...)` is executed, the following things happen:

>1. A new object is created, inheriting from `Foo.prototype`.
2. The constructor function `Foo` is called with the specified arguments, and with this bound to the newly created object. `new Foo` is equivalent to `new Foo()`, i.e. if no argument list is specified, `Foo` is called without arguments.
3. The object returned by the constructor function becomes the result of the whole new expression. If the constructor function doesn't explicitly return an object, the object created in step 1 is used instead. (Normally constructors don't return a value, but they can choose to do so if they want to override the normal object creation process.)

0xfe:
Dammit, did I forget to merge this?

Anyhow, it looks like it got fixed by the linter. :-)

PR #374, opened 2016-06-29
Lint: clef, stave, staveconnector,  stringnumber,  tuning, vibrato

Silverwolf90:
See title.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,7 +10,12 @@
     "no-else-return": [0],
     "no-multi-spaces": [0],
     "no-param-reassign": [0],
-    "no-restricted-syntax": [0],
+    "no-restricted-syntax": [
+      2,
+      "DebuggerStatement",
+      "LabeledStatement",
+      "WithStatement",
+    ],
     "no-shadow": [0],
     "no-use-before-define": [0],
     "prefer-template": [0],
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,7 +14,7 @@
       2,
       "DebuggerStatement",
       "LabeledStatement",
-      "WithStatement",
+      "WithStatement"
     ],
     "no-shadow": [0],
     "no-use-before-define": [0],
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,18 +2,19 @@
   "extends": "airbnb-base",
   "root": true,
   "rules": {
-    "max-len": [2, 100],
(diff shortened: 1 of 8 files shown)


.eslintrc.json
```diff
@@ -2,18 +2,19 @@
   "extends": "airbnb-base",
   "root": true,
   "rules": {
-    "max-len": [2, 100],
```
0xfe: are these rule changes intentional?

0xfe: oh, i see you just sorted them. np.

src/stringnumber.js
```diff
@@ -27,7 +27,11 @@ export class StringNumber extends Modifier {
     let shift_left = 0;
     let shift_right = 0;
 
-    let i, num, note, pos, props_tmp;
+    let i;
+    let num;
+    let note;
+    let pos;
+    let props_tmp;
```
0xfe: this is a bit excessive, i think... why not put them in a single line?

aaronmars: I'd say this block is in need of a little refactor anyway.  Whether these are one line or multiple lines, perhaps they shouldn't be there at all, and rather declared when needed (that's what `let` is for!)

Silverwolf90: I agree with @aaronmars. Putting each declaration on a new line is definitely uglier but highlights a code smell.

src/stringnumber.js
```diff
@@ -38,25 +42,35 @@ export class StringNumber extends Modifier {
         pos = num.getPosition();
         const props = note.getKeyProps()[num.getIndex()];
```
mscuthbert: For me the real code smell occurs over the next three lines where the same `i` is reused in two internal looks -- it would seem to me that the outer loop is unnecessary (I'm sort of surprised that it works at all, but the inner loop leaves i == nums.length so it exits the outer loop when it leaves the inner loop.  I don't know if there's a linter in the world that would catch this, however.

0xfe: Yeah, there's a lot of yuck here.

0xfe:
Looks good. Merging.

FYI, Cyril, I just realized that the visual regression tests are run as part of the github Travis CI integration, so you can always check the output of the tests for your PR right here. (Click "show all checks" -> "details" and scroll down.) If there are regressions, you'll see a the list of failed tests sorted by phash difference at the bottom of the log.

PR #378, opened 2016-07-01
Fix GhostNote and add tests

panarch:
Hi Mohit,

It's so glad to see recent great improvements!
Here is a tiny commit.
I found we don't have GhostNote test cases and so it was missed out during es6 converting work.

I fixed it to work and added tests not to be forgotten in the future :)

Thanks!

screen shot 2016-07-01 at 3 39 26 pm
screen shot 2016-07-01 at 3 39 31 pm


0xfe:
Thanks for adding these tests Taehoon! Looks good. Merging.

PR #380, opened 2016-07-04
AttrNoteGroup; which is for ClefNote, TimeSigNote and BarNote multi-voice support

panarch:
Hi Mohit,
Here's new class! AttrNoteGroup (I simply named as AttrNoteGroup because that kind of ClefNote etc are categorized by "Attributes" in MusicXML.. :) thanks for any other name recommendation if you have)

Current Attributes Notes system works in case of single-voice well,
but with current "ignore_ticks" option, it cannot handle multi-voice & multi-staff notes.

Here is an example below, ClefNotes are not formatted correctly.
screen shot 2016-07-04 at 6 22 17 pm

I firstly thought modifying Formatter to support this kind of formatting, but it will make Formatter more complicated and Formatter already has this kind of formatting feature, Modifier.

It is similar with GraceNoteGroup, but it covers ClefNote, TimeSigNote and BarNote which can be categorized by "Attributes".
AttrNoteGroup makes Modifier for StaveNote using that kind of attributes notes.
So it makes Formatter handle multi-voice & multi-staff examples properly.

screen shot 2016-07-04 at 6 22 44 pm
screen shot 2016-07-04 at 6 22 50 pm
screen shot 2016-07-04 at 6 22 54 pm

And, I didn't change other test cases which uses ClefNote, TimeSigNote or BarNote yet, but I think it's better to change those examples, too.
It will be better not to use that notes without AttrNoteGroup as we don't use GraceNote without GraceNoteGroup, how do you think of this?
And because current "ignore_ticks" works only for specific case (single-voice) and it can be also replaced by Modifier system, so it may be good to make it deprecated.
Only other class which has "ignore_ticks" statement is TextNote but it actually has no test cases for that.

Thanks!


src/attrnotegroup.js
```diff
@@ -0,0 +1,109 @@
...
+//
+// This file implements `AttrNoteGroup` which is used to format and
+// render attr notes; ClefNote, TimeSigNote and BarNote.
+
+import { Vex } from './vex';
+import { Flow } from './tables';
+import { Modifier } from './modifier';
+import { Formatter } from './formatter';
+import { Voice } from './voice';
+import { GraceNoteGroup } from './gracenotegroup';
```
0xfe: It doesn't look like you're using `GraceNoteGroup` in this file. (mistake?)

0xfe:
Thanks Taehoon. I like the examples, and it looks like a good abstraction. Let me think of an alternate name for this. :-)

0xfe:
Will this work with any modifiers? Maybe we can call it `ModifierGroup`?

mscuthbert:
On the Multi Staff test, when the r.h. turns to alto clef and 9/8, can the l.h. change clefs at the same time?  It looks like the only thing that hasn't been tested -- does the alignment work properly if one staff has more than one of ClefNote, TimeSignatureNote, etc. and the other only has one?

For naming are these PreTickNotes ?  Or something that implies that they appear before the tick to which they are attached?

src/attrnotegroup.js
panarch: Oh, it's obviously, my mistake :) I'll update.

Firstly, I tested using GraceNoteGroup which works similarly (it also convert Note to Modifier).
And after I finished implementing, I had to remove that but forgot.

panarch:
The key role of this class is providing clef change in multi-voice environment because clef-change is usual case in real world scores but others are not.
But what this class actually do is converting `Note[]` to `Modifier`, so any kind of `Note` can be a member of this, although there's no need of doing that.

@mscuthbert Yeah, it doesn't matter, this class's role is very simple, just gathering `Note[]` and run `Formatter` and make it as a `Modifier`. one or more `ClefNote`, `TimeSigNote` any thing ok.

I used postfix `Group` because `GraceNoteGroup` which works similarly (`GraceNote[]` to `Modifier`) already uses `Group` postfix, I think it might be better to use `Group` for consistency.
May be `Group` postfix for any kind of class which has role of converting `Note[]` to `Modifier`.

And it actually be able to work with any `Note` (even `StaveNote` is also possible), so rather than using `Attr` prefix for specific `Note`, I think `Modifier` prefix looks better :)

So, how about `ModifierNoteGroup`?
`ClefNote`, `TimeSigNote` and `BarNote` are `Note` but not recommended to use directly without `Group`, something,,, `Note` but it is actually not `Note`.

0xfe:
Looks like my the message I sent via e-mail didn't make it here. :-/

Okay, this could be used on either side, and could be helpful for other things to. Taehoon, let's just call this `NoteSubGroup` and take it from there.

Also can you add a doc to the wiki with some steps on how to use it? Thanks!

panarch:
Oh, yes I didn't get any messages.

`NoteSubGroup` looks good, thanks!
I'm now back to work so I will update both renaming and wiki soon!

panarch:
Ok, I fininshed remaining and added "How do I display mid-measure clef?" to The VexFlow FAQ!

0xfe:
Thanks Taehoon. Looks good. Merging.

panarch:
@0xfe By the way, could you add me to wiki contributors list? :)

0xfe:
Yes, of course.

panarch:
Thanks!

PR #397, opened 2016-07-12
Fix broken articulation positions within stave (and remove hard-coded articulation glyph metrics)

Silverwolf90:
I've been wanting to fix this for literally years! 

I wanted to open up a little work I've done around articulations. This is a pretty substantial refactor around the Articulation.draw() method which also includes work done on Glyph to override the default origin.

### The following pairs are [before-fix/after-fix] - UPDATED (Jul 15, 2016)

image
image

image
image

image
image

image
image

image
image

image
image

image
image

image
image


Silverwolf90:
Hmm, looks like some articulations below the staff are too far away.

.eslintrc.json
```diff
@@ -25,6 +25,7 @@
     "prefer-template": [0],
     "quote-props": [0],
     "space-before-function-paren": [2, "never"],
-    "strict": [2, "global"]
+    "strict": [2, "global"],
+    "yoda": [2, "never", { "exceptRange": true }]
```
Silverwolf90: Added this because I think it's better to have ranges in the form of 

```
lowerbound < value && value < higherbound
```

src/glyph.js
```diff
@@ -105,7 +105,7 @@ export class Glyph {
   }
 
   static getOutlineBoundingBox(outline, scale, x_pos, y_pos) {
-    const bboxComp = new BoundingBoxComputation(x_pos, y_pos);
```
Silverwolf90: removing these because the some glyphs don't have origins located on their path (ie: tenuto marks) and including the origin in these cases incorrectly expands the bounding box

src/articulation.js
```diff
@@ -12,12 +12,95 @@
...
+    }
+  } else if (note.getCategory() === 'tabnotes') {
+    if (note.hasStem()) {
+      if (stemDirection === Stem.UP) {
+        return stave.getYForBottomText(textLine - 2);
+      } else {
+        return stemTipY;
+      }
+    } else {
+      return stave.getYForBottomText(textLine - 2);
```
Silverwolf90: I have no idea why these magic  numbers are being used... but they were there before.

src/articulation.js
```diff
@@ -12,12 +12,95 @@
...
-import { StaveNote } from './stavenote';
 import { Glyph } from './glyph';
+import { Stem } from './stem';
 
 // To enable logging for this class. Set `Vex.Flow.Articulation.DEBUG` to `true`.
 function L(...args) { if (Articulation.DEBUG) Vex.L('Vex.Flow.Articulation', args); }
 
+const { ABOVE, BELOW } = Modifier.Position;
+
+const mathToIncrement =
```
0xfe: This function is just used once in the code (AFAICT) with a constant mathFn. It makes `snapLineToStaff` harder to read, so doesn't seem worth it.

src/tables.js
```diff
@@ -241,110 +241,19 @@ Flow.textWidth = text => 6 * text.toString().length;
...
+  'a-': { code: 'v25', between_lines: true }, // Tenuto
+  'a^': { code: 'va',  between_lines: false }, // Marcato
+  'a+': { code: 'v8b', between_lines: false }, // Left hand pizzicato
+  'ao': { code: 'v94', between_lines: false }, // Snap pizzicato
+  'ah': { code: 'vb9', between_lines: false }, // Natural harmonic or open note
+  'a@a': { code: 'v43', between_lines: false }, // Fermata above staff
+  'a@u': { code: 'v5b', between_lines: false }, // Fermata below staff
+  'a|': { code: 'v75', between_lines: false }, // Bow up - up stroke
+  'am': { code: 'v97', between_lines: false }, // Bow down - down stroke
+  'a,': { code: 'vb3', between_lines: false }, // Choked
```
0xfe: Love that the glyph bounding boxes eliminate the need for the shifts.

.eslintrc.json
0xfe: Thanks, I was wondering about this -- couldn't see where the yoda code was.

src/articulation.js
0xfe: Thoughts here?

0xfe:
Also, it looks like just (SVG?) tab is affected by the bottom annotations being too far away.

src/articulation.js
0xfe: Not sure either, what if you remove them? Does it affect the tab articulations?

Silverwolf90:
Sorry that was really unclear because I didn't mention this. The pairs are before/after *not* canvas/svg.

Silverwolf90:
The most obvious improvement is here:

### before
image

### after
image

Silverwolf90:
Ok I've done a lot of digging. I'm gonna go a bit overboard to try and explain my thoughts as clearly as possible:

`Stave.option` has two properties: `top_text_line` and `bottom_text_line`. As far as I can tell, they're supposed to represent the default placement for objects above/below the stave. Despite having `text` in their name they are used by many objects -- via `getYForTopText`/`getYForBottomText` --  for generic positioning around a stave.

(It's important to note here, that `getYForTopText` and `getYForBottomText` use `getYForLine(line)` under the hood and `getYForLine` expects the `line` argument to be a top-down 0-based index. ie: the top-most line is line `0`)

They two properties are initialized like this:

* [`top_text_line` is set to `1` by default](https://github.com/0xfe/vexflow/blob/master/src/stave.js#L47)
* [`bottom_text_line` is set to `num_lines + 1`, so a standard 5 line staff results in a `6`](https://github.com/0xfe/vexflow/blob/master/src/stave.js#L68)

So these values seem weird. Using top-down 0-based stafflines, `top_text_value` would be sitting within the staff.

Here is the start of the problem: these values are actually not *absolute* staff lines, but directional and relative. So relative to staff line 0, and for `top_text_line` a *positive* value moves **towards the top** of the canvas,  and for `bottom_text_line` a *positive* value moves **towards the bottom** of the canvas. Knowing that, we can convert the values into absolute stafflines:

* top_text_line: `-1`
* bottom_text_line: `6`

These values are still weird. If the staff lines have a zero-based (staff top: 0, staff bottom: 4) index it seems that the `top_text_line` is only 1 line away from the staff  while `bottom_text_line` is 2 lines away from the staff.

So basically... this is a bug right? `bottom_text_line` should be set to `numLines` not `numLines + 1`?

Silverwolf90:
Also seems like `staveNote.getStemExtents()` and `stem.getExtents()` are broken for beamed stems. In the following image I'm hiding the beams and printing out a green dot for each method. 

image

At the very least, since the dots are overlapping, they're consistently broken...

0xfe:
Thank Cyril. You are correct that they are relative positions, and `text` is misleading here -- initially I used it to track the position of text annotations on the top and bottom of the staff, but then overloaded it to all annotations. Ideally, those would be called `top_annotation_position` and `bottom_annotation_position`. These are actually used in conjunction with the `top_text_line` and `text_line` in the `format()` function of all the modifiers.

I don't think the defaults are a bug -- they represent the best placement for the first (closest) annotation.

The `getStemExtents()` thing though looks pretty bad... and you can see how they diverge slowly with every note (per beam.) It's almost like there's a leaky y-pixel increment somewhere in one of the variables that sum to the height.

src/articulation.js
Silverwolf90: Agreed

Silverwolf90:
Right, so these defaults attempt to to preserve 1 staff space between text/staff on either side. I think a better solution to achieve this, without making default positions asymmetric, is to put the burden of shifting based on orientation onto the annotation itself. [This is something I've already done with the articulations in this PR.](https://github.com/0xfe/vexflow/pull/397/files#diff-8b6993baa34f67c389f21f81cec0b25bR187).
 
In any case, given this asymmetry, any glyph based elements (which can now shift their origin) **should not** be using these functions to get their `y` values.

I should write some documentation about this too. Because it's something I *only now* understand completely.

Silverwolf90:
It also begs the question, how opinionated is VexFlow intended to be?

It's unavoidable with higher-level formatting decisions. But it seems to me that the `Stave` is too basic of a class to make any assumptions about what "looks best" farther than a set of consistent, sane defaults. I think the position of annotations relative to the staff is greatly over stepping boundaries here, and as I've exemplified, it is certainly not obvious.

src/articulation.js
```diff
@@ -56,126 +139,75 @@ export class Articulation extends Modifier {
...
-      } else { // Without a stem
-        top = stave.getYForTopText(this.text_line - 1);
-        bottom = stave.getYForBottomText(this.text_line - 2);
-      }
+    const { x } = note.getModifierStartXY(position, index);
+    const shouldSitOutsideStaff = !canSitBetweenLines || isTab;
+
+    const y = {
+      [ABOVE]: () => {
+        glyph.setOrigin(0.5, 1);
```
Silverwolf90: @0xfe tried linking to this in my previous comment but it doesn't seem to anchor to the right point

src/articulation.js
Silverwolf90: Yup.

image

Silverwolf90: Wait sorry no, wrong screenshot.

image

0xfe: Okay, yeah, so this pushes the modifiers further out. It explains the constants, but still not why `getYForBottomText` doesn't give you the right value automatically for the stave.

0xfe:
Re: opinionated, VexFlow shouldn't be opinionated at all (atleast in the lower level elements like Staves), and agree with you about consistent sane defaults... although in this case I'm not sure that I could get both consistent and sane for the bottom, but maybe I didn't try hard enough. What happens if you set `bottom_text_position` to `numLines`? (It does sound like a bug, thinking about this more.)

src/articulation.js
0xfe: I guess this is because of the `bottom_text_position` default being `numLines + 1`?

Silverwolf90:
I've set `bottom_text_position` to `num_lines` and now the articulations are symmetrical without any hardcoded numbers! But I'm sure lots of tests were affected as a result of this change.  We should see how everything comes out. But here are the articulation tests, top pictures are `master`, bottom pictures represent current state of this PR. I updated the original post with these images as well.

image
image

image
image


image
image


image
image


image
image


image
image

image
image

image
image

For some reason eslint is failing in the build but passing on my system... have to sort that out.

0xfe:
Thanks. Will run this when I'm back. (BTW, the tab articulations are further out than I'd like -- maybe override the constants for the tab stave?)

Silverwolf90:
Moved them closer. Extracted out a function to get the initial offset from the articulation's starting y value.

0xfe:
Thanks. Still a few issues:

### Text too close to stave
staveline staveline_arrow_options_blessed
staveline staveline_arrow_options_current
staveline staveline_arrow_options

### Articulations too far from stave
articulation articulation___up_stroke_down_stroke_blessed
articulation articulation___up_stroke_down_stroke_current
articulation articulation___up_stroke_down_stroke

## Failures (most of these are okay)
```
You have       19 fail(s):
Articulation.Articulation___Up_stroke_Down_Stroke 25.2195
StaveLine.StaveLine_Arrow_Options 9.94362
Articulation.TabNote_Articulation 7.3859
Articulation.Articulation___Staccato_Staccatissimo 5.59049
Articulation.Articulation___Fermata_Above_Below 5.52052
Annotation.Test_Justification_Annotation_Stem_Down 5.32743
Annotation.Test_Justification_Annotation_Stem_Up 3.42785
Beam.Complex_Beams_with_Articulations 2.65166
StaveNote.Center_Aligned_Note_with_Articulation 2.58623
Articulation.Articulation___Marcato_L_H__Pizzicato 1.36549
Articulation.Articulation___Accent_Tenuto 1.21678
Stave.Stave_Text_Test 1.08376
Annotation.TabNote_Annotations 0.952887
TextBracket.TextBracket_Styles 0.209995
Articulation.Articulation___Inline_Multiple 0.209491
Stave.Multiple_Line_Stave_Text_Test__Raphael_ 0.187604
Annotation.Bottom_Annotation 0.0642813
Articulation.Articulation___Snap_Pizzicato_Fermata 0.0148498
PedalMarking.Custom_Text_1 0.010428
```

Silverwolf90:
For the  "Articulations too far from stave," only the top articulations move because their blessed positions were asymmetric and my changes correct this. Do you want me to shift both the top and bottom articulations towards the staff by a half space?

0xfe:
Yes, please. Post an image of the same test with the half-space shift on both sides.

Silverwolf90:
I meerged in the latest `master` and had to make some more changes. Things are looking ok. I've altered `Articulation.format` so that it increments the `modifierCtx.state.text_line` according to the height of the glyph with a half space margin. But there are some glaring issues with how articulations are positioned which means that they'll never appear perfectly consistent. I've outlined the problems in a comment.

image

image

src/tables.js
```diff
@@ -9,6 +9,7 @@ const Flow = {
   STEM_HEIGHT: 32,
   STAVE_LINE_THICKNESS: 2,
   RESOLUTION: 16384,
+  INITIAL_ARTICULATION_OFFSET: -0.5,
```
0xfe: I'm trying to minimize (or get rid of) these global constants. Prefer `class.CONSTANT` instead, so you could have `Articulation.INITIAL_OFFSET` located in `articulation.js`.

0xfe:
Thanks Cyril. The `0.5` offset definitely looks better. Is this good to go (modulo my comment about the constant)?

src/tables.js
Silverwolf90: Are these values supposed to be configuration? If they are, why scatter the config across the modules instead of a centralized location? If it's not configuration, then I agree.

Silverwolf90:
I've spotted a couple issues.

Silverwolf90:
Should be good now if the build passes!

src/tables.js
0xfe: I'd like to only put things here that are used across multiple classes, otherwise keep them along with the class. It makes things more cohesive, simpler to maintain, and less likely to rot.

Also, I think it's cleaner to say `VF.Articulation.INITIAL_OFFSET = 1` than `VF.INITIAL_ARTICULATION_OFFSET = 1`. Same for the other values here, but I just haven't had a chance to clean them up.

Silverwolf90: Fixed.

0xfe:
Cool thanks. Changes look good (modulo a few oddities in the text), but overall much better. Merging.

Issue #372, opened 2016-06-29
Secondary beam breaks doesn't work with tuplets


aaronmars:
If auto-beaming is configured to work with secondary breaks, tuplets are not counted correctly.
Fo example, if I have a 16th note triplet, and auto breaks are to be on 8th note boundaries, the tuplet notes' fractional values are not considered.

Here's an example of the bug:
<img width="529" alt="screen shot 2016-06-29 at 09 11 25" src="https://cloud.githubusercontent.com/assets/847926/16459989/10eb0914-3dda-11e6-90b3-5ea2ced780e3.png">

This is generated with `Beam.generateBeams(notes, { secondary_breaks: '8' });`
The 16ths in the tuplet are being treated like full-value 16ths rather than all three of them adding up to the space of a single 8th note.

aaronmars:
OK.  I got the secondary beam issue worked out, but there's still an issue with tuplet bracketing and placement (see image).  I'll work that out and include it in the fix.

<img width="535" alt="screen shot 2016-07-03 at 06 39 11" src="https://cloud.githubusercontent.com/assets/847926/16545743/2daa3a40-40e9-11e6-8b9f-0b1bb784de88.png">

This is just called with
```javascript
const beams = Vex.Flow.Beam.generateBeams(notes, { secondary_breaks: 8 });
```

aaronmars:
Well, this snowballed a bit.  Here are the things I'd like to clean up here.

- That initial issue where secondary beams are not broken correctly when tuplets are involved.
- Issue where a tuplet will have a bracket if it does not include the first note of the beam group.
- Issue where the tuplet numbers and brackets are mis-aligned
- Issue I found while rendering a quarter note triplet that has 8th notes in it (new screenshot below).  I believe the ratio is wrong in this case.  Shouldn't it just be '3' rather than '4:2'? (I'm pretty sure it should be bracketed too)

<img width="532" alt="screen shot 2016-07-07 at 08 04 42" src="https://cloud.githubusercontent.com/assets/847926/16658007/8647fe66-4419-11e6-97bd-3216a3d5c3cc.png">

0xfe:
Thanks for investigating Aaron. Do you have tentative changes to fix some of the issues?

gristow:
[...]

PR #398, opened 2016-07-16
Refactor stems/beams

Silverwolf90:
Still  a WIP, tab is broken.

Important things to note:

* The logic `StaveNote.getStemExtents()` has been removed and now delegates to `Stem.getExtents()`
* Went through `Beam` and did a lot of renaming, most variable names were really vague and I think it's a lot easier to understand now.
* A lot of logic in `beam.applyStemExtensions` was just plain wrong mainly due to incorrect assumptions around stem coordinates because it's a [very confusing part of the codebase](https://github.com/0xfe/vexflow/wiki/Development-Gotchas) -- that method has been greatly simplified
* `Formatter.FormatAndDraw` now passes `options.stave` to `formatter.formatToStave` which is required to trigger the `postFormat`  lifecycle (this also explains why the stem extents were wrong in the articulation PR)

### DIFF
--- a/tests/stavenote_tests.js
+++ b/tests/stavenote_tests.js
@@ -373,7 +373,7 @@ VF.Test.StaveNote = (function() {
       tickContext.setX(10);
       tickContext.setPadding(0);
 
-      equal(tickContext.getWidth(), 15.5988);
+      VF.Test.almostEqual(tickContext.getWidth(), 15.5988, 0.0001);
     },
 
     showNote: function(note_struct, stave, ctx, x, drawBoundingBox) {
--- a/tests/stavenote_tests.js
+++ b/tests/stavenote_tests.js
@@ -373,7 +373,7 @@ VF.Test.StaveNote = (function() {
       tickContext.setX(10);
       tickContext.setPadding(0);
 
-      equal(tickContext.getWidth(), 16);
+      equal(tickContext.getWidth(), 15.5988);
     },
 
     showNote: function(note_struct, stave, ctx, x, drawBoundingBox) {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,9 @@
 {
   "extends": "airbnb-base",
   "root": true,
+  "env": {
+    "browser": true
+  },
   "rules": {
     "camelcase": [0],
     "yoda": [2, "never", { "exceptRange": true }],
--- a/src/beam.js
+++ b/src/beam.js
@@ -550,7 +550,9 @@ export class Beam {
         currentExtreme = stemTipY;
         extremeY = note.getNoteHeadBounds().y_bottom;
         extremeBeamCount = note.getBeamCount();
-      } else if (stem_direction === Stem.UP && (currentExtreme === 0 || currentExtreme > stemTipY)) {
+      } else if (
+        stem_direction === Stem.UP && (currentExtreme === 0 || currentExtreme > stemTipY)
+      ) {
         currentExtreme = stemTipY;
         extremeY = note.getNoteHeadBounds().y_top;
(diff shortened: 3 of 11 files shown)


gristow:
This is great, so much easier to understand what's happening now!

aaronmars:
@Silverwolf90 , do you think this could help with my tuplet headaches in https://github.com/0xfe/vexflow/issues/372 ?

Silverwolf90:
Branches merge together without conflicts. I took the group of notes with the misaligned bracket:

Here's the test you added:

image

Here's that group of notes with previously mis-aligned tuplet:

image

aaronmars:
That is wonderful news.  I'll stop trying to fix it with the current code and wait until your changes here are merged.

0xfe:
Thanks for the PR, Cyril. I'm traveling right now, and will review as soon as I get a chance.

0xfe:
Most of the tests have reasonable diffs in the regression test, but this one was really weird:

StaveNote.Center_Aligned_Note_with_Articulation (blessed, current, diff):

stavenote center_aligned_note_with_articulation_blessed
stavenote center_aligned_note_with_articulation_current
stavenote center_aligned_note_with_articulation

Silverwolf90:
Looks like it fixed a bug :)

The test's title is Center_Aligned_Note_with_Articulation, and the blessed image doesn't have an articulation, while the current one has a fermata!

Silverwolf90:
But Tab stems are stil broken. I started digger further and it seems that the `StaveNote` and `TabNote` stem lifecycles are very distinct. So I'm going to make them as consistent as possible.

0xfe:
Ha, looks like it did fix a bug. :-)

Silverwolf90:
Ok, things should be back to normal-ish. But almost everything will have minor changes.

The stem lifecycle between stavenotes and tabnotes is now more consistent.

Started removed a lot of the shifting that compensates for the `Stem.WIDTH`, but couldn't remove it from everywhere.

After fixing a lot of stem logic, stem connections to noteheads got screwed. The solution seemed to be replacing the hard-coded notehead widths the *real* bounding box width.  For now I just replaced the property with a getter, but this is a temporary hack. 

Ultimately, `tables.js` should not be responsible for anything related to glyph metrics. It should only contain data concerning music theory and music engraving semantics . Everywhere in the codebase that accesses this `head_width` should switch to getting the notehead width from a locally instantiated `Glyph`.

Silverwolf90:
Also something is definitely wrong with Travis. We should be seeing tons of diffs.

0xfe:
This sounds great. I'm still traveling so will try to get to this as soon
as I get a moment.

0xfe:
Okay, finally able to look at this. Lots of differences in tests, but all sane. The only visual nit is displaced note heads. It seems like this change has shifted them about a pixel away from the stem, making them look slightly detached when you zoom in. Want to take a quick look?

tests/stavenote_tests.js
```diff
@@ -373,7 +373,7 @@ VF.Test.StaveNote = (function() {
       tickContext.setX(10);
       tickContext.setPadding(0);
 
-      equal(tickContext.getWidth(), 16);
+      equal(tickContext.getWidth(), 15.5988);
```
0xfe: This _might_ cause platform-specific errors due to floating point variations. Does qunit have an `almostEqual`? (or perhaps define `amostEqual(value, expectedValue, errorMargin)`).

Silverwolf90:
Ok this is good to go. Added the `almostEqual` assertion and shifted the displaced noteheads so that they overlap the stem.

0xfe:
All good. Great change!

PR #399, opened 2016-07-21
Remove rollup in favor of webpack for super-fast incremental rebuilds. 

Silverwolf90:
Throwing this out there... I was unable to get fast incremental rebuilds working with rollup (@aaronmars do you know how?). If it's possible I'm happy to stick with it. But I couldn't figure it out, so I replaced rollup with webpack (using `grunt-webpack`).

See the rebuilds in action by running: `grunt webpack:watch`. On my machine, the initial build takes `~8s`, and a rebuild takes `~800ms`. Makes for a much nicer dev experience.

The `watch:scripts` task has been removed in favor of `webpack:watch` (in order to take advantage of webpack's built-in file watching). The task for building _without_ watching is `webpack:build`.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,9 @@
 {
   "extends": "airbnb-base",
   "root": true,
+  "env": {
+    "browser": true
+  },
   "rules": {
     "camelcase": [0],
     "yoda": [2, "never", { "exceptRange": true }],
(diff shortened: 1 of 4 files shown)


aaronmars:
Yeah!  I don't have a problem with this at all.  Rollup might be a bit bare-bones.

0xfe:
This is way faster -- thanks for the change. (Bonus, because I prefer webpack too.)

PR #400, opened 2016-07-22
Add StaveModifier to build, use StaveModifier.Position in tests

panarch:
Hi Mohit,

Here's a single tiny commit!
I found StaveModifier is missing in current build, so I added.
And I also modified StaveModifier test code to use StaveModifier class explicitly.


0xfe:
Thanks Taehoon. Sorry about the slow response -- I just got back from vacation, I will look at it shortly.

panarch:
No problem at all, you don't need to hurry :)

Thanks!

0xfe:
Good catch. Merging.

PR #407, opened 2016-07-30
Improve glyph+stem proportions.

Silverwolf90:
~~This is still a WIP -- looks like down-stemmed gracenote flags aren't working as expected. But I wanted to open this up in case anyone had any objections. I'll fix the gracenotes shortly.~~

----

Now that glyph widths are not hard-coded, we can more easily increase glyph sizes to be more standard (and more visually pleasing).

Before this change, the default stem height was 32px. But this is non-standard given a stave with a default 10px staff space. Typically in standard music notation, the default stem height is 3.5 staff spaces (ie: 35px given a 10px staff spcae). The reason it was set to 32px is because the noteheads were too small, so a 35px stem looked too long. I've increased the size of the noteheads, in order for their height to take up an entire staff space. And increased the Stem to the standard 35px.

The crux of this change was replacing the `head_width` property of notehead glyphs in `tables.js`. It was replaced with a `.getWidth(fontScale)`.

I also extracted some of the common font scales into the `Flow` object.

Here are some of the tests with the new glyph sizes:

image

image

image

image

image

image

@mscuthbert and @gristow, do think this is a visual improvement?


0xfe:
Can you set `Vex.Flow.STAVE_LINE_THICKNESS = 1` and post the same test images?

Silverwolf90:
image

image

image

image

image

image

Silverwolf90:
Note that viewing them inline in the github comments distorts the image, view the raw image to make sure you don't get any unnecessary anti-aliasing artifacts.

0xfe:
It actually looks pretty good. It's more symmetric and touches both stave lines evenly.

Silverwolf90:
I agree that it looks better. Although, with the 1px staff line the half/quarter notehead extends past the staffline by literally a single pixel (it might even just be overlapped anti-aliasing artifacts). But despite that, I think it looks significantly better. Much more natural, and much more reminiscent of traditional engraving.

Silverwolf90:
Ideally one would size the notehead based on the staff space, rather than estimating the font size to match the staff space.

0xfe:
Yes, although, I've never heard of anyone change the staff spacing, so I think this is good.

Silverwolf90:
> Yes, although, I've never heard of anyone change the staff spacing, so I think this is good.

Auto-correct may have butchered what you meant to say? I don't quite follow.

0xfe:
Sorry, the butchering was all me. I meant to say that people rarely (if ever) change the staff spacing, so it's probably not worth the effort of estimating the font size.

Silverwolf90:
This kind of touches on an issue I've been meaning to raise...

I'd contend that no-one fiddles with the staff spacing (specifically the `spacing_between_lines_px` property) because very little elements respect that value. It doesn't work as expected. For example, the stem's default height is hardcoded 35 pixels, not 3.5 staff spaces. So modifying the staff space doesn't impact the stem's height, even though it should. The staff space should be the canonical unit that **all** layout and positioning uses. But it's treated as this on-the-side-unit that is occasionally used.

Ossias exemplify this problem really well. VexFlow does not easily support assembling and formatting 2 differently sized staves.

gristow:
Big improvement! Cyril, I think you're right when you say:
> Ideally one would size the notehead directly based on the staff space, rather than estimating the font size to roughly match the staff space.

Otherwise, it seems we'll always have slightly uneven notehead-staveline overlap one way or the other. Even with these vast improvements, at a very large size noteheads are still not perfectly vertically aligned, regardless of `stave_line_thickness`. (Pictures at bottom.)

But maybe the exact notehead sizing doesn't matter as long as it's vertically centered. An idea: we could flag noteheads that are meant to take up exactly one staveline of space for special Y centering. (Are there ones that don't? Is a quarter rest a notehead, for instance?) Something along the lines of:

```javascript
// Code to vertically center a notehead on a stave line, or in a stave space.
// This thoroughly untested code could go somewhere in notehead.js

const line = this.getLine(); // the notehead's staveline
const stave = this.stave;
const glyph = this.getGlyph();
const glyphHeight = glyph.getMetrics().height; // the bbox height of glyph

// default y placement:
const y1 = stave.getYForNote(line); 
// y for one staveline lower (spacing_between_lines + staveLineThickness):
const y2 = stave.getYForNote(line + 1); 
// the actual distance between those ys, which takes into account
// staveline thickness:
const targetHeight = y2 - y1;
// And the visual midpoint between those two ys, which is where we want
// the very middle of the notehead to be:
const targetCenterY = y1 + (targetHeight / 2) - (staveLineThickness / 2);
// Plus a little math to put the notehead there:
const y = targetCenterY - (glyphHeight / 2);
this.y = y;
```

I have not actually tried that code out... wonder what you think of it as an idea?

And here's the unevenness I see at super zoom when I build this branch:
At `STAVE_LINE_THICKNESS = 2`:
screen shot 2016-07-30 at 10 27 18 pm
At `STAVE_LINE_THICKNESS = 1`:
screen shot 2016-07-30 at 10 26 53 pm

Silverwolf90:
@gristow that feels like a band-aid to me. The glyph's origin should be at the exact middle of the notehead's height.  And `stave.getYForLine()` should be returning the right values. Given those two things, the notehead will be perfectly centered without the need for further adjustment.

I think I've fixed it now though. Commits incoming.

image
image

src/svgcontext.js
```diff
@@ -334,7 +336,7 @@ export class SVGContext {
       height *= -1;
     }
 
-    this.rect(x, y, width - 0.5, height - 0.5, this.attributes);
```
Silverwolf90: I had to take out these adjustments to get Barlines to behave as expected. Were these to mitigate anti-aliasing artifacts?

0xfe: Yes, IIRC, the only way you could get 1px lines (without anti-aliasing) at the time was to set the width to 0.5. Not sure if this is still the case.

Silverwolf90: This is still the case, but that is by design. Also shifting by 0.5 does not solve the problem, it just pushes the anti-aliasing problem onto lines with even-numbered widths.

In both canvas and svg, a line's origin is in the middle, ie: the width is evenly distributed on both sides. The combination of integer coordinate values with an odd pixel width ends up distributing a half-pixel on each side, causing the anti-aliasing engine takes over. So for odd numbered pixel widths, you need to shift by 0.5 to avoid anti aliasing artifacts, but for even numbered pixel widths, you don't.

Silverwolf90:
Thoughts at this point?

Random Sampling:

image

image

image

image

image

image

image

image

image

image

image

0xfe:
I think the proportions look pretty good overall. Minor nit: the muted notes on the tab look slightly off center (vertically). I patched in this change, and it looks like a bunch of tests are broken?

Silverwolf90:
Ah right, no `Object.assign` support in phantomJS (used by `grunt-qunit`). Added a babel plugin to transform it.

Silverwolf90:
Fixed the muted notes. The culprit was more hardcoded numbers!

Silverwolf90:
image

0xfe:
Something weird going on in the fret/string numbering tests:

1) Second chord in second bar has `3` pushed out.
2) Third chord in second bar has a new `1` (which is probably a bug you unintentionally fixed.)


## Current
stringnumber fret_hand_finger_in_notation_current

## Blessed
stringnumber fret_hand_finger_in_notation_blessed

0xfe:
All the other tests look good, BTW.

Silverwolf90:
Hmm I think this messed up horizontal formatting for flagged stem-up notes. Stem down notes are getting formatted closer than identical notes with stem up.

Silverwolf90:
[Actually, since the formatter uses the width of a stavenote (which I wouldn't recommend on a pass), it looks like this like this has been an issue for 5 years.](https://github.com/0xfe/vexflow/blob/4b904561e920fc23b134b022e36b805da7374707/src/stavenote.js#L312).

So this PR hasn't introduced any problems and is good to go.

0xfe:
Looks great. Merging.

PR #408, opened 2016-08-01
Tuplet beam breaks

aaronmars:
Added code to inspect all tuplets in a beam group for stem direction and bracketing.

Fixed the duration inspection for tuplet notes that determines if a note is able to be beamed vs. how many ticks it actually occupies.

Added a test.

There is one diff in the visual regression that reflects the removal of a wrongly-placed triplet bracket.


0xfe:
Thanks for the fix Aaron. Looks good. Merging.

Issue #411, opened 2016-08-07
Vertical formatting results in lots of collisions when mixing elements and changing sizes


Silverwolf90:
This relates to a lot of things I've been bringing up lately (positioning, boudingboxes, units, hardcoded values, etc). So I figured I should showcase the current state of vertical formatting:

![image](https://cloud.githubusercontent.com/assets/1631625/17462011/57e1096a-5c96-11e6-839a-3983db1bee12.png)

https://jsfiddle.net/vrd5757s/1/

Fundamentally, the mixing of elements should not be a complex case to handle: advance the position of the next element by some value derived from the previous element. There are implementation details around this, but the basic mechanism is the same. The current formatting lifecycle, specifically the `ModifierContext` (see https://github.com/0xfe/vexflow/issues/409) and the `.format()` methods make this much more convoluted then it needs to be, while also enforcing an inflexible ordering of `Modifiers` which is problematically opinionated.

rvilarl:
@AaronDavidNewman the test is there in #1238 

![image](https://user-images.githubusercontent.com/22865285/142896887-be15dc4e-089c-4182-84d8-26108c9fd86f.png)

AaronDavidNewman:
I should have a chance to look at this in the next week or so

rvilarl:
@AaronDavidNewman is this one now closed?

AaronDavidNewman:
yes this can be closed, thought it was

rvilarl:
@Silverwolf90 @0xfe @ronyeh please close.

Issue #401, opened 2016-07-23
We need to explicitly define the object lifecycle


Silverwolf90:
Two major problems: 

- VexFlow has no clearly defined object lifecycle
- VexFlow provides no support in assembling objects the right way

This is something that that test's really get wrong. They all construct, setup,  and draw things differently. It's horrifyingly inconsistent. We're always telling people to look at the tests to learn how to use VexFlow's more obscure parts, but as I'm looking through them now. It's chaos. We can't expect people to look at the tests to get any clarity on how to use this library. 

I'm bringing this up because I was trying to make a very reasonable change to `StaveNote`:

Currently, a `StaveNote` sets the `x` value for it's `Noteheads` and `Stem` in it's `render()` method. But this is clearly the wrong place. We shouldn't be mutating anything in `render()`. So I moved that logic into `.postFormat()`. Since that's clearly the the place to set the `x` values. However, doing so broke *tons of tests* because lots of them are doing things in the wrong order. But that fact would have never been apparent when the tests were written.

Silverwolf90:
Another related question: when should `Glyph`'s be initialized? I think, ideally, all `Glyph` based `Elements` would construct their glyphs at the same point in the `Element` lifecycle.

Provided that `Elements` have methods used for configuration *after construction*, I think they should be initialized right before formatting (ie: after the object has been been fully configured).

0xfe:
Right now I think there's too much initialization in the constructor, causing a lot of issues with configuration after construction. I think as much as possible should be pushed out to the preformatting stage.

PR #412, opened 2016-08-07
WIP/Blocked: Extend staff lines to middle line if too far above/below the stave

Silverwolf90:
*This is a WIP but currently blocked by reasons explained below*

Concerning Stem lengths, see Behind Bars, page 14, *Stem Length, Notes on Ledger Lines:*

> Stems for notes on more than one ledger line extend to the middle stave-line
(marked below):

The PR introduces changes that implement this behvaior:

image
*sidenote: the beams should not be flat*

Unfortunately, trying to implement this behavior exemplifies a big problem:
**The stave isn't necessarily set on all elements when formatting occurs.** So what should be a simple implementation, results in ~~~57~~ ~270 failed tests [because of this new error](https://github.com/0xfe/vexflow/compare/master...Silverwolf90:extend-stems-to-staff-middle?expand=1#diff-14d05dc59b818c53fb8a5a1cd38c1a22R139).

### Proposal
It should be a new requirement that the stave is set on all formatted elements **before** formatting occurs, which would also enable some big improvements to vertical formatting (see https://github.com/0xfe/vexflow/issues/411).

This also relates to better defining the formatting lifecycle (see https://github.com/0xfe/vexflow/issues/401).


0xfe:
I think it's a totally reasonable requirement to set the stave on all formatted elements. There are cases where users render elements without staves, but I don't think they ever use the formatter.

We should also try to infer the stave as much as we can, to prevent the user from having to call `setStave` on everything (e.g., in modifiers.) In that stem example of yours, where is the stem created? Is it possible to set the stave when `setStave` is called for the relevant note? (or during the note's `preFormat`?)

Silverwolf90:
Agreed that we should infer as much as possible,

The stem is created on construction, as are the noteheads. `staveNote.setStave` already cascades down to the noteheads. And in this PR I implement `Stem.setStave` which is also called from `staveNote.setStave`. We could definitely add the `Modifiers`.

Since requiring the stave be set before formatting was not previously a requirement, and it broke a lot of our tests, this would be a breaking change.

Silverwolf90:
Ok this is a bigger change than I thought. When I earlier said 57 failing tests I must have accidentally been filtering some tests, because the number is in fact ~270. 

[Also we'll need to change the tutorial.](https://github.com/0xfe/vexflow/wiki/The-VexFlow-Tutorial#step-2-add-some-notes--run-) It currently has code with the stave being set **after** formatting.

And we'll need to deprecate the usage of `voice.draw(ctx, stave)` in favor of only using the `ctx` argument argument: `voice.draw(ctx)`.

0xfe:
Re: deprecating `voice.draw`. There are two ways this should work:

### 1) Add the stave to the formatter

```javascript
voice.addTickables(notes);
var formatter = new VF.Formatter().joinVoices([voice]).format([voice], 400, {stave: stave});
voice.draw(context);
```

### 2) Call `voice.setStave` after adding tickables and call the formatter normally

```javascript
voice.addTickables(notes).setStave(stave);
var formatter = new VF.Formatter().joinVoices([voice]).format([voice], 400);
voice.draw(context);
```

It looks like 1 works, but 2 doesn't, and it's because the formatter (line 396) only calls `voice.preFormat` if a stave is provided. Can you pull the `preFormat` out of the branch in this change too?

Silverwolf90:
Agreed on the second, not so sure about the first. What about multiple voices across multiple staves? What does passing the `stave` parameter to `format()` mean then?

0xfe:
1) is just a helper option for single staves and it already exists, 2) is certainly the recommended approach, but I just discovered that it doesn't work while updating the tutorial.

(also, I just noticed that `postFormat` is called only if the stave is supplied to `Formatter.format`.)

Silverwolf90:
Cool, that all sounds good then.

Silverwolf90:
Actually, I'm not sure the line that calls `voice.preFormat()` makes any sense at all At best, it makes sense as a "convenience"  if the the staves aren't set and the `stave` is passed in. Assuming that `voice.setStave(stave)`  calls `setStave` on each `tickable`, then that would break support for cross-staff voices.

0xfe:
How about if it calls `setStave` only if a stave isn't already set?

Silverwolf90:
Nevermind! On second glance, there's nothing technically wrong with the line (or the conditional, it's necessary since the `stave` might not be provided). But the block does feel very out of place to me. The `stave` is passed to `preFormat` **only** for this little bit of logic. I'm not really convinced that passing the `stave` option to `formatter.format` provides any convenience over using `voice.setStave(stave)` provided it works as expected -- given those two code snippets you posted, they seem equally verbose.

[Looking at this code...](https://github.com/0xfe/vexflow/blob/73c19abd57e63ae9dcf832574f3904f338a85e0f/src/formatter.js#L558) If I remember correctly, only postFormatting if `opts.stave` existed was working around the original issue of the `stave` not always being set. Now that stave will be available, we don't have to guard the call to `postFormat`.

I think we should remove the `stave` option in the formatter completely and **only** use `voice.setStave(stave)` to set the stave on a voice. And for individual notes, one can use `staveNote.setStave(stave)`.

Also I think I found out why why `voice.setStave(stave)` wasn't working...[ it looks like it doesn't call `setStave` on its `tickables`](https://github.com/0xfe/vexflow/blob/master/src/voice.js#L80). Seems to me that it definitely should be doing that.

Silverwolf90:
In fact, all `postFormat` related functionality become obsolete, since we can now move stems/beams to the beginning of the formatting lifecycle.

0xfe:
I think it is intentional that `voice.setStave` doesn't call `setStave` on all its tickables. Ideally, a setter should not have unpredictable side effects, and I think this is why `voice.preFormat` is required to be called after the voice is "ready."

If we make `setStave` here update all its tickables, then we probably should also make sure that `addTickable` sets the stave when new notes are added after `setStave` is called. But I think that sucks too.

I propose the following:

* Leave `setStave` as is -- a setter that updates `this.stave`.
* Add a function called `Voice.updateStave(stave)` (or some such) which sets the stave on all the tickables to the argument `stave`. If `stave` is not provided, then use `this.stave`. If `this.stave` isn't set, then do nothing.
* Replace the stave setting code in `voice.preFormat` with `this.updateStave`.

I think this will make the API a bit clearer and simpler to reason about.

Silverwolf90:
RE: `voice.setStave` not calling `setStave` on tickables

That's inconsistent with the "infer as much as possible" strategy that we're applying in `staveNote.setStave`, currently the call to `setStave` cascades down into the `Stem` and `Noteheads`. I don't think we should have two conflicting strategies for "parents" to manage their "child objects"

0xfe:
Yes -- I think that all those cases should really be called `updateStave`. It's too late to do anything now, but we are in a situation where we don't know exactly what a setter function will do to an object. To me that's a pretty big smell.

Re: purpose, it's to separate the API from the internal details. Ideally, users (and internal vexflow classes) will call `note.setStave()` instead of setting `note.stave` directly. It's much simpler to refactor code that is written this way (ignoring the new ES6 getter/setter thingies for now.)

So I recommend going with `updateStave` for this, and I'll work on cleaning up the rest of the API such that setters remain predictable.

Silverwolf90:
Yeah `updateStave` is a much better name. I'm definitely in favor of
renaming and making things consistent.

Re: purpose

I wasnt asking about the purpose of getter/setter methods, but specifically
the notion of having a stave on the voice at all. The `Voice` doesn't have
a corresponding `getStave` method, leaving direct property access the only
way to access the stave. Since it appears there are no instances of
accessing the `stave` on a voice, it suggests that's it's fundementally not
being used. If it's dead code then we should probably remove it.

0xfe:
Yes, removing it and replacing it with `updateStave` sounds good. (I don't think it had any purpose outside of as a helper to update the staves on the tickables.)

Silverwolf90:
I've got a lot of test changes incoming... still ~90ish tests left to fix.

Silverwolf90:
Hmm damn. `grunt-qunit` uses PhantomJS which doesn't support es6 -- we need to transpile the tests if we want to use ES6 in the tests with `grunt-qunit`.

Silverwolf90:
I've yet to rename `setStave` to `updateStave` and use the `Factory` -- some of the first tests I initially edited before some our discussion will need another pass I think.

0xfe:
Yeah, the tests are ES5 unfortunately. I just ran into that too.

Silverwolf90:
I'm gonna break up this PR into smaller ones, with the test refactoring first, then the behavior changes.

PR #414, opened 2016-08-11
Basic grace tab notation implementation

bforian:
The core classes of this feature are
GraceTabNote and GraceTabNoteGroup, based upon the implementation of existing classes GraceNote and GraceNoteGroup.


src/gracetabnotegroup.js
```diff
@@ -0,0 +1,147 @@
...
+import { Flow } from './tables';
+import { Modifier } from './modifier';
+import { Formatter } from './formatter';
+import { Voice } from './voice';
+import { TabTie } from './tabtie';
+
+// To enable logging for this class. Set `Vex.Flow.GraceNoteGroup.DEBUG` to `true`.
+function L(...args) { if (GraceTabNoteGroup.DEBUG) Vex.L('Vex.Flow.GraceTabNoteGroup', args); }
+
+export class GraceTabNoteGroup extends Modifier {
```
0xfe: Do we really need this class? Can't we just use `GraceNoteGroup` instead? (I'd like to minimize duplicated code and logic as much as possible.)

src/tables.js
```diff
@@ -228,18 +230,18 @@ Flow.integerToNote.table = {
 };
 
 
-Flow.tabToGlyph = fret => {
+Flow.tabToGlyph = (fret, scale) => {
```
0xfe: Since you're changing the `tabToGlyph` signature here, you'll need to set a default value for scale (so we don't break existing users.)

src/tables.js
```diff
@@ -9,6 +9,8 @@ const Flow = {
   STEM_HEIGHT: 35,
   STAVE_LINE_THICKNESS: 1,
   RESOLUTION: 16384,
+  TABLATURE_FONT: '10pt Arial',
+  TABLATURE_GRACE_FONT: '7.5pt Arial',
```
0xfe: Can you move these constants to `tabnote.js` and `gracetabnote.js`? I'd like to keep the configurables close to the classes.

src/gracetabnote.js
```diff
@@ -0,0 +1,32 @@
...
+// to be rendered on a tab stave.
+//
+// See `tests/gracetabnote_tests.js` for usage examples.
+
+import { TabNote } from './tabnote';
+import { Flow } from './tables';
+
+export class GraceTabNote extends TabNote {
+  static get CATEGORY() { return 'gracetabnotes'; }
+  static get SCALE() { return 0.66; }
```
0xfe: So I think we could simplify this by moving `SCALE` to `this.render_options.scale`, and setting it in the constructor.

src/gracetabnote.js
```diff
@@ -0,0 +1,32 @@
...
+  constructor(note_struct) {
+    super(note_struct, false);
+    this.render_options.y_shift = 1;
+  }
+
+  getCategory() { return GraceTabNote.CATEGORY; }
+
+  getScale() { return GraceTabNote.SCALE; }
+
+  getFont() { return Flow.TABLATURE_GRACE_FONT; }
```
0xfe: Same thing here. Rather than a static constant, this is probably better as part of `render_options`. So all these getters can remain in `tabnote.js` only, and the only thing you'll need here is the constructor and `CATEGORY`.

0xfe:
Thanks for the change bforian -- mostly looks good. Most of my comments are stylistic -- can you address them please? Also, can you post images of the test output here?

bforian:
Thanks Mohit, I’ll run through your comments quickly and send you updates & test output soon.

bforian:
All your comments have been applied, I send you the test output:
grace-tab-simple
grace-tab-slurred

0xfe:
Thanks bforian. Looks good overall. I made some micro adjustments to y-center the notes and push up the slurs. Merging!

bforian:
Awesome, thanks!

src/gracetabnote.js
Silverwolf90: I think this is problematic. Removing the static property means that the scale can't be configured globally. I think this is a more typical use-case than needing to set the scale on a per-instance basis. If you keep the static `GraceTabNote.SCALE` and have `graceTabNote.render_options.scale` default to it, then both global and local styling are available.

PR #417, opened 2016-08-15
Refactor tuplet tests with factory

Silverwolf90:
See title.

### DIFF
--- a/src/factory.js
+++ b/src/factory.js
@@ -69,8 +69,6 @@ export class Factory {
     }
     this.renderQ = [];
     this.stave = null; // current stave
-
-    this.StaveNote = this.StaveNote.bind(this);
   }
 
   getOptions() { return this.options; }
--- a/src/factory.js
+++ b/src/factory.js
@@ -20,6 +20,7 @@ import { System } from './system';
 import { TickContext } from './tickcontext';
 import { Tuplet } from './tuplet';
 import { Voice } from './voice';
+import { Beam } from './beam';
 
 // To enable logging for this class. Set `Vex.Flow.Factory.DEBUG` to `true`.
 function L(...args) { if (Factory.DEBUG) Vex.L('Vex.Flow.Factory', args); }
@@ -32,7 +33,7 @@ function X(message, data) {
   L(this.name + ':', message, data);
 }
 
-function setDefaults(params, defaults) {
+function setDefaults(params = {}, defaults) {
   const default_options = defaults.options;
   params = Object.assign(defaults, params);
   params.options = Object.assign(default_options, params.options);
@@ -68,6 +69,8 @@ export class Factory {
     }
     this.renderQ = [];
     this.stave = null; // current stave
+
+    this.StaveNote = this.StaveNote.bind(this);
   }
 
   getOptions() { return this.options; }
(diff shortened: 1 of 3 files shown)


src/factory.js
```diff
@@ -68,6 +69,8 @@ export class Factory {
     }
     this.renderQ = [];
     this.stave = null; // current stave
+
+    this.StaveNote = this.StaveNote.bind(this);
```
Silverwolf90: By doing this we get the really elegant:  `[...noteStructs].map(vf.StaveNote)` that can be seen in the tests

tests/tuplet_tests.js
```diff
@@ -2,482 +2,507 @@
...
+
+/*
+eslint-disable
+no-undef,
+no-var,
+object-shorthand,
+func-names,
+wrap-iife,
+vars-on-top,
+ */
```
0xfe: We don't have a linter covering `tests/`. Do you need this?

src/factory.js
0xfe: I think this should be a local function in the test file (maybe as a wrapper called `newStaveNote`). It certainly does not belong here.

tests/tuplet_tests.js
```diff
@@ -2,482 +2,507 @@
...
-    },
+  // Ideally this would be using arrow syntax...
+  var set = function(propertyName) {
+    return function(value) {
+      return function(object) {
+        object[propertyName] = value;
+        return object;
+      };
+    };
+  };
```
0xfe: This is clever, but probably too clever? I had to squint to understand what all this did. :-/

0xfe:
Thanks Cyril. Looks good overall. Minor comments.

tests/tuplet_tests.js
Silverwolf90: Nope.

tests/tuplet_tests.js
Silverwolf90: Unfortunately, ES5 syntax obfuscates how simple this is. I'd argue that there's nothing clever about this because it's just a curried function. 

Here's it refactored to use arrow functions: 

```javascript
var set = (propertyName) => (value) => (object) => {
  object[propertyName] = value;
  return object;
};
```

or with a [`curry` decorator:](https://lodash.com/docs#curry)

```javascript
var set = curry((propertyName, value, object) => {
  object[propertyName] = value;
  return object;
});

src/factory.js
Silverwolf90: Can you elaborate on why this does not belong here? This seems appropriate if the point of the Factory is to provide a common "construction context" for objects, unless there's a specific reason to allow usage of these methods as unbounded functions -- but I can't think of one, since that's what plain constructors are for. I suppose it's important to note that I didn't add this specifically for the tests, but to minimize boilerplate/duplication around usage of the Factory _in general_. (And I think this should be extended to every factory method)

Binding instance methods on construction is not an uncommon pattern in the JS ecosystem, for example, `class` based React components often employ this pattern.

tests/tuplet_tests.js
0xfe: Yeah, the arrow syntax is much simpler to read.

src/factory.js
0xfe: My main issue here is that it makes `StaveNote` a special method in the class, which this pattern is only used for the test. I'm open to having something like this here, but maybe in a separate PR where it applies across all factory  methods.

Silverwolf90: Agreed, that's a better idea.

Silverwolf90:
Issues fixed.

0xfe:
Looks great, merging. Thanks!

PR #420, opened 2016-08-16
Refactor grace note tests with factory

Silverwolf90:
Added `GraceNote` and `GraceNoteGroup` to the `Factory`.

### DIFF
--- a/src/factory.js
+++ b/src/factory.js
@@ -20,6 +20,9 @@ import { System } from './system';
 import { TickContext } from './tickcontext';
 import { Tuplet } from './tuplet';
 import { Voice } from './voice';
+import { Beam } from './beam';
+import { GraceNote } from './gracenote';
+import { GraceNoteGroup } from './gracenotegroup';
 
 // To enable logging for this class. Set `Vex.Flow.Factory.DEBUG` to `true`.
 function L(...args) { if (Factory.DEBUG) Vex.L('Vex.Flow.Factory', args); }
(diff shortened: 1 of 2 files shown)


src/factory.js
```diff
@@ -118,6 +121,19 @@ export class Factory {
...
 
+  GraceNote(noteStruct) {
+    const note = new GraceNote(noteStruct);
+    if (this.stave) note.setStave(this.stave);
+    note.setContext(this.context);
+    return note;
+  }
+
+  GraceNoteGroup(params) {
+    const group = new GraceNoteGroup(params.notes, params.slur);
```
Silverwolf90: should this method have a `params.options` (for consistency with the other factory methods) even though the `GraceNoteGroup` constructor's second argument is a boolean?

src/factory.js
```diff
@@ -173,6 +189,19 @@ export class Factory {
...
 
+  Beam(params) {
+    params = setDefaults(params, {
+      notes: [],
+      options: {
+        autoStem: false,
+      },
+    });
+
+    const beam = new Beam(params.notes, params.options.autoStem).setContext(this.context);
```
Silverwolf90: (Note that this code is also duplicated in  https://github.com/0xfe/vexflow/pull/417 )

The inverse of the other comment, should the `params.options` exist here even though beam doesn't take an `options` object in the constructor?

src/factory.js
0xfe: If it's unused, you can leave it for now. It's easy to add it if we need it without breaking users.

src/factory.js
0xfe: Yes, that is fine. This is actually a good case for it, and a way to break out of long unnamed parameter lists.

0xfe:
Looks good. Merging.

0xfe:
I merged this, but then noticed in the visual regression tests that the ratioed tuplets are broken. (They don't show ratios anymore.)

Silverwolf90:
Looking into it.

PR #423, opened 2016-08-19
Refactor accidental tests with factory

Silverwolf90:
Added a `paddingBetween` option to `Formatter.SimpleFormat`

Some diffs will be significantly different due to scale/dimensions changing, but the notation itself has not been changed (as far as I could tell).


0xfe:
Looks good. Merging.

0xfe:
I think this may have broken one of the accidental tests. The third chord (on both voices) below look like they're colliding.

screen shot 2016-08-21 at 3 15 25 pm

Silverwolf90:
Hmm, this actually looks to be a bug in `StaveNote.formatByY` I think -- PR incoming

PR #426, opened 2016-08-21
Extract out builder commit behavior into functions 

Silverwolf90:
WIP - But wanted to get your initial thoughts on this

---

This refactors the `Builder` with a new property: `onCommit`. It's an array of functions which will will be run in sequence and passed `(currentOptions, note, factory)` -- the expected usage is around matching key/values in the `currentOptions` object and then transforming the `note` in some way. The `currentOptions` object is formed by the assigning the piece options into the builder options

I extracted the `id` setting into its own function. And also implemented articulations in this way:
```
c4/4[articulations='staccato.above,tenuto.below']
```

Most importantly, this provides the user with a hook for arbitrary extension.


src/easyscore.js
```diff
@@ -227,39 +262,40 @@ class Builder {
 
   commitPiece() {
     L('commitPiece');
-    if (!this.factory) return;
-    const options = this.piece.options;
-    const stem = options.stem || this.options.stem;
-    const clef = this.options.clef;
+    const { factory } = this;
```
Silverwolf90: Sorry this file is bit noisy, just a bit of minor cleanup.

0xfe:
I like the general idea of customized hooks. How about if these hooks were added by the classes themselves? So `articulations.js` would look something like this:

```javascript
class Articulations {
  ...
  static easyScoreOption(currentOptions, note, factory) { ... }
  ...
}

EasyScore.addNoteOption('articulations', Articulations.easyScoreHook)`
```

So when new elements are created, we don't need to modify `easyscore.js`, and keep it free from the potential mess.

Silverwolf90:
Yeah that sounds fine to me!

Silverwolf90:
Thought about this a bit more:

While I do think these functions need to go *somewhere else*, I'm not sure they should go in as statics on existing classes. These functions might not relate to a specific class.  We already have the `setId` function, which totally generic. And I can imagine others (ie: style, color, config, metadata) might not be class specific either. Where do we put those?

0xfe:
I think the generic ones can stay here (for now). Only element-specific ones go in as static methods to the relevant classes.

Silverwolf90:
Ok that sounds good.

Silverwolf90:
When the `Builder` is `reset` would you expect the commit hooks to persist, or get removed.

0xfe:
I think they should persist, because they're not part of the state.

src/easyscore.js
```diff
@@ -271,6 +284,10 @@ export class EasyScore {
     this.options = Object.assign({
       factory: null,
       builder: null,
+      commitHooks: [
+        setId,
```
0xfe: Can you add a `setClass` too that calls `addClass` on the element?

0xfe: (and an element can have multiple classes, so maybe comma separated)

Silverwolf90: Should this support multiple tokens? And if so, separated by comma or space or...?

Silverwolf90: Whoops didn't see your last comment

Silverwolf90:
Do you want me to refactor the Bach test to use this? How do you feel about the syntax for the articulations hook?
```
c4/4[articulations='staccato']
c4/4[articulations='staccato,tenuto']
c4/4[articulations='staccato.above,tenuto.below']
```

Silverwolf90:
Should we shorten to `artics` or `artic` or leave as is?

0xfe:
I think `articulations` is good.

0xfe:
Don't refactor the Bach tests as yet. I'm still mutilating them.

src/easyscore.js
```diff
@@ -214,54 +220,69 @@ class Builder {
...
+  if (id === undefined) return;
+
+  note.setAttribute('id', id);
+}
+
+function setClass(options, note) {
+  if (!options.class) return;
+
+  options.class
+    .split(',')
```
0xfe: Better to use `.split(/\s*,\s*/)` so `class="red, blue, green" will work too.

0xfe: Never mind, I just noticed the `trim`.

0xfe:
Is this good to go?

Silverwolf90:
I don't feel that the tests are in the right place, but otherwise this is good to go.

Silverwolf90:
Do we want a new file for EasyScore hooks? Do we put the articulation hook in the articulation tests? (That feels a bit weird -- it would be unlike every other test in that file)

0xfe:
I think leaving everything in `easyscore_tests.js` is fine.

Silverwolf90:
Cool, then my last thought might be to split up the `drawOptionsTest` into a test for each hook (without rendered output -- just to test that the parsing and expected objects get created/attached). But that might not be necessary. If you're content, then I think this is good to go!

0xfe:
It doesn't hurt to have, but I don't think it's necessary, especially since we'll be using these features in many other tests.

src/easyscore.js
```diff
@@ -271,12 +292,18 @@ export class EasyScore {
     this.options = Object.assign({
       factory: null,
       builder: null,
+      commitHooks: [
+        setId,
+        setClass,
+        Articulation.easyScoreHook,
```
0xfe: I think the hook should be registered in `articulation.js`, not here. Ideally there will be no articulation code here, so new elements don't ever have to touch this file. You can implement this with an `EasyScore` static method. So the end of `articulation.js` will look like this:

```javascript
EasyScore.registerDefaultCommitHook(Articulation.easyScoreHook);
```

src/articulation.js
```diff
@@ -198,6 +198,25 @@ export class Articulation extends Modifier {
     return true;
   }
 
+  static easyScoreHook({ articulations }, note, factory) {
```
0xfe: Instead of `factory` how about passing in the `EasyScore` instance itself, and add accessor `getFactory`. This would future-proof it a bit.

src/easyscore.js
Silverwolf90: >new elements don't ever have to touch this file.

Is that so problematic? It seems natural that the default EasyScore config is tightly coupled to the EasyScore class. Won't that make it easier to follow the "evolution" of the EasyScore defaults, ie: via git blame? Moving the registration of default hooks into their respective files unnecessarily fragments something that I think belongs together. And if the defaults ever end up changing drastically, then those changes don't permeate the codebase, but stay in an isolated place.

src/articulation.js
Silverwolf90: :+1:

src/easyscore.js
Silverwolf90: What about a separate EasyScore config file?

0xfe: It's not problematic, but I think each element should have all it's logic, constants, hooks, etc. all in one place (yes, there's still lots of room for improvement in the existing code). I don't think the keyval options really need to be tightly coupled to EasyScore -- they're by design meant to allow for extensions. 

Also, I don't think that following the evolution of EasyScore is that useful (at least compared to having cohesive units that can be managed separately.)

Silverwolf90: While I do agree that a module should be cohesive, I don't agree that the registration of a default commit hook satisfies this.

How does someone reading the code identify what the default `EasyScore` behavior is? When discovering a JS library/class I typically look for some sort of centralized configuration object -- that feels much more intuitive and discoverable. 

But this also leads me to questioning the API, because `EasyScore.addCommitHook` is append only. Don't we need to support completely overriding the defaults? VexFlow provides a basic set, but I easily might want my own set of more specialized commit hooks. What if instead of an `addCommitHook` method we provide a `options.commitHooks` array on construction which defaults to a `static` `EasyScore.defaultCommitHooks`?

src/articulation.js
Silverwolf90: Just realized this isn't currently possible because the `Builder` calls the commit hooks and doesn't have access to the `EasyScore` instance.

src/easyscore.js
Silverwolf90: Actually yeah I think the regex is clearer!

src/easyscore.js
0xfe: It seems like you're optimizing for the reader, and I'm optimizing for the writer.

I prefer the latter, but in the interest of moving forward, let's leave the defaults in one place, and revisit if this becomes an issue.

src/articulation.js
0xfe: How about passing in the builder then (instead of the factory)?

Silverwolf90: Done

0xfe:
Thanks. Looks like there's a regression in the tests:

## Blessed



## Current

Silverwolf90:
Fixed.

0xfe:
Merged. Thanks for the change.

PR #427, opened 2016-08-21
Add shortcut to `Factory` constructor.

0xfe:
This makes the common case of just setting a DOM selector more elegant. @Silverwolf90 let me know what you think of this. (I'm in two minds about it.)


Silverwolf90:
I don't have a particularly strong opinion -- but it feels a bit premature. Is such minimal configuration the common case?

0xfe:
Drat, I read your previous 'sounds fine to me reply' (on e-mail) and assumed it was for this PR! woops!

0xfe:
To answer your question, I was writing up a tutorial for EasyScore, and most of the examples started like this:

```javascript
vf = new VF.Factory({renderer: {selector: 'boo'}})
```

...which got annoying after a while. (I don't know if this really will be the common case in general, but it was for me :-)

Silverwolf90:
What about relying on static methods instead of polymorphic constructors?

```javascript
vf = VF.Factory.fromSelector(selector)
```

Dunno. Just an idea, don't feel strongly either way!

0xfe:
Yeah, I think that's a better idea. I'll fix master.

PR #428, opened 2016-08-21
Fix two voice note formatting

Silverwolf90:
image

The bug is that `keyProps.line` and `stave.getYForLine()` are incompatible -- so we have to convert `keyProps.line`. I wrote about this in the ["Development Gotchas"](https://github.com/0xfe/vexflow/wiki/Development-Gotchas) wiki page.

Worth a look at the diffs.

### DIFF
--- a/src/stavenote.js
+++ b/src/stavenote.js
@@ -255,11 +255,11 @@ export class StaveNote extends StemmableNote {
       // See https://github.com/0xfe/vexflow/wiki/Development-Gotchas
       const topNotBottomY = topNote
         .getStave()
-        .getYForLine(Math.abs(5 - topKeys[0].line) + HALF_NOTEHEAD_HEIGHT);
+        .getYForLine(5 - topKeys[0].line + HALF_NOTEHEAD_HEIGHT);
 
       const bottomNoteTopY = bottomNote
         .getStave()
-        .getYForLine(Math.abs(5 - bottomKeys[bottomKeys.length - 1].line) - HALF_NOTEHEAD_HEIGHT);
+        .getYForLine(5 - bottomKeys[bottomKeys.length - 1].line - HALF_NOTEHEAD_HEIGHT);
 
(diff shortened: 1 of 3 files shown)


0xfe:
Thanks. That's pretty ugly. :-/

0xfe:
Looks like this causes some aligned stave notes (even across staves) to get shifted.

See last two notes below:

screen shot 2016-08-21 at 6 07 37 pm

And last two notes on first two staves below:

screen shot 2016-08-21 at 6 09 40 pm

Silverwolf90:
Fixed.

0xfe:
Still some off alignment in tests:

### Current (Formatter - StaveNote Justification)
screen shot 2016-08-21 at 8 48 56 pm

### Blessed (Formatter - StaveNote Justification)
screen shot 2016-08-21 at 8 48 50 pm

### Current (NoteSubGroup - MultiStaff)
screen shot 2016-08-21 at 8 48 30 pm

### Blessed (NoteSubGroup - MultiStaff)
screen shot 2016-08-21 at 8 48 20 pm

Silverwolf90:
(The names of the tests are flipped) 

The StaveNote justification tests was a silly mistake, used `<=` instead of `<`.

But the `NoteSubGroup` tests , and previous `Formatter - Multiple Staves No Justification` diffs are related to `.joinVoices` being used to join all formatted voices, instead of only voices on the same stave. 

`.joinVoices` has been a source of confusion for lots of people over many years -- we should probably reconsider this part of the API and make the construction of `ModifierContexts` easier to understand. The fact that this is leaking into the tests shows that even with review, improper usage is difficult to spot.

0xfe:
Thanks Cyril -- I'm running the visual tests now. Re: `joinVoices`, hopefully the `System` class will take care of most of the common cases, and users will never have to see `joinVoices` again.

0xfe:
All good. Merging. Thanks for tracking this down!

PR #429, opened 2016-08-22
Add new `Registry` class to manage VexFlow elements.

0xfe:
This is a first pass at the `Registry` class. It allows us to index and keep track of VF elements in a document and query by various attributes. This allows us to build simpler APIs:

For example, we could add a Tie or Slur simply based on element IDs:

```javascript
vf.Tie('foo', 'bar');
```

Or color all elements in a specific class red:

```javascript
registry.getElementsByClass('specialnotes').forEach(n => n.setColor('red'));
```

Or we could build stylesheets that post-format generated notation (strawman):

```javascript
registry.applyStyle('newstyle.css')
```

There are two ways to use the registry:

1) Explicitly register elements with a registry via `registry.register(element)`. This could be something a factory or builder does.
2) Use `Registry.enableDefaultRegitstry(registry)` to auto register any new elements with the specified registry.

It's not complete yet, but I'm sending this out for early review.


0xfe:
This is good to go. A cool free feature of this class is the ability to get stats on a rendering. E.g., here's a count for all the elements in the proportional formatter test:

```
"System: 1"
"StaveNote: 21"
"Stem: 21"
"Glyph: 149"
"NoteHead: 21"
"Tuplet: 3"
"Voice: 5"
"Stave: 5"
"Barline: 10"
"Clef: 5"
"TimeSignature: 5"
"StaveConnector: 1"
"TickContext: 16"
```

Silverwolf90:
This looks great! Nitpicky stuff inline.

Maybe we should transpile the tests for the `qunit` task only.

src/registry.js
```diff
@@ -0,0 +1,123 @@
...
+    });
+    elem.onRegister(this);
+    return this;
+  }
+
+  getElementById(id) {
+    return this.index.id[id] ? this.index.id[id][id] : null;
+  }
+
+  getElementsByAttr(attrName, value) {
```
Silverwolf90: Should we rename this to `getElementsByAttribute` for consistency with `Element`? It's a bit verbose but I think the consistency is more important.

src/registry.js
```diff
@@ -0,0 +1,123 @@
...
+      class: {},
+    };
+    return this;
+  }
+
+  // Updates the indexes for element 'id'. If an element's attribute changes
+  // from A -> B, make sure to remove the element from A.
+  updateIndex({ id, name, value, oldValue }) {
+    const elem = this.getElementById(id);
+    if (oldValue !== null && this.index[name][oldValue]) delete this.index[name][oldValue][id];
```
Silverwolf90: These two lines are pretty dense. Can we break these lines up into blocks?
```javascript
if (oldValue !== null && this.index[name][oldValue]) {
  delete this.index[name][oldValue][id];
}
```

src/registry.js
```diff
@@ -0,0 +1,123 @@
...
+//
+// 1) Explicitly call `element.register(registry)`, or,
+// 2) Call `Registry.enableDefaultRegistry(registry)` when ready, and all future
+//    elements will automatically register with it.
+//
+// Once an element is registered, selected attributes are tracked and indexed by
+// the registry. This allows fast look up of elements by attributes like id, type,
+// and class.
+
+export class X extends Error {
```
Silverwolf90: Can we create an Error class factory to avoid the duplication?

```javascript
function makeErrorType(name) {
  return class extends Error {
    constructor() { ... }
  };
}
```

src/registry.js
0xfe: Done.

src/registry.js
0xfe: Done.

src/registry.js
0xfe: Good idea. Added `Vex.MakeException` and refactored a few other custom exceptions to use it (Factory, EasyScore, Parser.)

0xfe:
Thanks for the review. Addressed all your comments.

0xfe:
If there are no more issues, I'll go ahead and merge.

Silverwolf90:
Yup, it's good!

PR #435, opened 2016-08-28
WIP: Render first 32 measures of Bach Minuet

0xfe:
This is a test to flesh out the `System`, `EasyScore` and `Factory` APIs. I hope to put the entire first part of the piece here. This will also be useful as a regression test when its done.

Current progress:

Open issues:
* Lots of duplication in building systems. `System` should support `measures` parameter so creating multi-measure systems is easy.
* Time signature is not vertically centered.
* Registry and IDs are super helpful. Adding classes to `EasyScore` keyvals will be enormously helpful configuring multiple similar notes.
* `EasyScore` keyvals for Articulations and Annotations may be helpful.
* ~~`StaveConnector` types need `Factory` shortcuts (like position and justification)~~
* Factory should register elements created by it instead of default registry.

### DIFF
--- a/src/factory.js
+++ b/src/factory.js
@@ -162,16 +162,16 @@ export class Factory {
       text: 'p',
       vJustify: 'below',
       hJustify: 'center',
-      font_family: 'Times',
-      font_size: 14,
-      font_weight: 'bold italic',
+      fontFamily: 'Times',
+      fontSize: 14,
+      fontWeight: 'bold italic',
       options: {},
     });
 
     const annotation = new Annotation(params.text);
     annotation.setJustification(params.hJustify);
     annotation.setVerticalJustification(params.vJustify);
-    annotation.setFont(params.font_family, params.font_size, params.font_weight);
+    annotation.setFont(params.fontFamily, params.fontSize, params.fontWeight);
     annotation.setContext(this.context);
     return annotation;
   }
--- a/src/factory.js
+++ b/src/factory.js
@@ -42,13 +42,6 @@ function setDefaults(params = {}, defaults) {
   return params;
 }
 
-const positionMap = {
-  above: Modifier.Position.ABOVE,
-  below: Modifier.Position.BELOW,
-  left: Modifier.Position.LEFT,
-  right: Modifier.Position.RIGHT,
-};
-
 export class Factory {
   constructor(options) {
     L('New factory: ', options);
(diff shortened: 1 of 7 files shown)


0xfe:
I think this is in a reasonable state to merge (mainly because of the improvements to the other classes.)

I can continue to fix the open issues in later PRs.

tests/bach_tests.js
```diff
@@ -0,0 +1,133 @@
...
+      system.addStave({
+        voices: [
+          voice(notes('(G3 B3 D4)/h, A3/q', {clef: 'bass'}), {time: '3/4'})
+        ]
+      }).addClef('bass').addKeySignature('G').addTimeSignature('3/4');
+      system.addConnector().setType('brace');
+      system.addConnector().setType('single_right');
+      system.addConnector().setType('single_left');
+
+      id('d1').addModifier(0, vf.Annotation({text: 'p', vJustify: 'below'}));
```
Silverwolf90: Real dynamics glyphs are available through `TextDynamics` or `TextNote`.

src/staveconnector.js
```diff
@@ -49,6 +49,21 @@ export class StaveConnector extends Element {
     };
   }
 
+  static get typeString() {
+    return {
+      single_right: StaveConnector.type.SINGLE_RIGHT,
```
Silverwolf90: Can we move to camelCased aliases?

src/factory.js
```diff
@@ -152,6 +156,51 @@ export class Factory {
     return acc;
   }
 
+  Annotation(params) {
+    params = setDefaults(params, {
+      text: 'p',
+      vJustify: 'below',
+      hJustify: 'center',
+      font_family: 'Times',
+      font_size: 14,
```
Silverwolf90: Can we camelCase these too

Silverwolf90:
What are you using for reference? Some stuff on IMSLP shows that the first bar has two voices in the left hand:

image

tests/bach_tests.js
0xfe: Done.

src/factory.js
0xfe: Done.

src/staveconnector.js
0xfe: Done.

0xfe:
Reference: I have a Shirmer's Library copy of Bach pieces on my desk. :-)

0xfe:
Addressed your comments, PTAL.

Silverwolf90:
Looks good to me!

mscuthbert:
Great!

Resident musicologist here: the piece is not by Bach -- it's from a collection of pieces by other composers that JS Bach copied out into a notebook for his wife.  :-)

0xfe:
JSB: The original music pirate.

PR #438, opened 2016-08-28
Reference the correct position of vexflow-debug.js

Huji:
The reference needs to be updated for glyphs.html to work out of the box.

### DIFF
--- a/src/glyphs.html
+++ b/src/glyphs.html
@@ -39,7 +39,7 @@
     }
   </style>
 
-  <script src="../build/vexflow-debug.js"></script>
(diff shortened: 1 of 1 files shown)


src/glyphs.html
```diff
@@ -39,7 +39,7 @@
     }
   </style>
 
-  <script src="../build/vexflow-debug.js"></script>
+  <script src="../releases/vexflow-debug.js"></script>
```
0xfe: Hi, This `build/` is actually the correct path. It requires you to build the library before you can use it. This is the same for the tests and other `.html` files. Thanks.

PR #439, opened 2016-08-28
Make font defintions readable

Huji:
The difference in file size is minimal, but the glyph codes will much easier to read and follow.

Also, a change to one glyph will be easier to identify in a diff.

If you approve this, I will also propose it with the other font files.


0xfe:
Thanks @Huji. This file is generated via `src/transform.html`. This format looks good to me, but please also make sure to update `transform.html` so this doesn't get clobbered the next time it's generated.

Huji:
Will make a new pull request shortly, using a fork.

Issue #440, opened 2016-08-28
Using ES6 code will not allow using vexflow in some modern browsers (as of yet)


Huji:
In several places, ES6 syntax has been introduced for `const` declaration, but not every browser supports it yet. It appears Firefox won't support it until [version 51](https://bugzilla.mozilla.org/show_bug.cgi?id=1263355).

I think this issue should be left open, and a new release should not be created, until all major browsers support declarations like `for (const a in b)`. Until then, releases should be based on revisions prior to the implementation of ES6-style code.

Silverwolf90:
VexFlow is transpiled to ES5 using Babel. This allows us to write the source in ES6 while supporting ES5 compliant browsers. If you need to support IE <=10 then there is an unsupported legacy build available in `/releases`.

Huji:
No, IE <= 10 is not my concern. I am talking about Firefox 48 (latest release) which still doesn't support `for (const a in b)` syntax.

aaronmars:
Perhaps this is just a bug? I'd think `(let a in b)` would be appropriate here.  According to MDN, `for ... in` is supported in Firefox: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

Huji:
Yes, `for .. in` is support in firefox, what is not supported is for **const** a in b

0xfe:
But shouldn't Babel transpile this down to what any (reasonable) modern browser will support?

aaronmars:
All I'm saying is that `const` makes no sense here since the variable being declared is, by definition, mutable. Maybe I'm missing something here, but I guess I'd just like to see if the same error pops up if `let` is used instead.

Huji:
I agree with that. I think the use of `const` here is uncalled for.

0xfe:
Oh, right `const` doesn't make sense here, and our `eslint` rules won't let that code through anyway. @Huji Do you know where this is coming from? I did a quick search in `src/` and didn't see that pattern anywhere.

Huji:
@0xfe I just marked all those places where a `const` was inside a for loop. In all those cases use of `const` is uncalled for, since the object is mutable by definition.

Huji:
Here is another issue that also has to do with ES6:

At the very end of `src/vex.js` there is a line which reads:

```
export { Vex };
```

That throws an error message in Firefox:

```
SyntaxError: export declarations may only appear at top level of a module
```

[...]

PR #442, opened 2016-08-29
Removing unnecessary const declarations

Huji:
This will partially address #440


0xfe:
Closing this: working as intended.

PR #450, opened 2016-09-02
Ignore both "build" and "build/"

Huji:
This is helpful because I often bypass the building process by creating a symlink from build/ to releases/ and it is best if git ignores a symlink (which is a file) just like it ignores a directory.


0xfe:
Thanks. Merging.

PR #451, opened 2016-09-03
EasyScore accidentals

infojunkie:
Added support for all accidentals in the EasyScore syntax.

Actually, the original EasyScore regex for accidentals is broken https://github.com/0xfe/vexflow/blob/master/src/easyscore.js#L133: the regex `[b#n]+` means that accidentals such as `####` and `b##bnnn` would be allowed. This is also fixed in my PR.


Silverwolf90:
Cool, looks good to me!

0xfe:
Looks good, will merge. Thanks for the change.

PR #452, opened 2016-09-04
Reset notehead displacements after changing stem direction (Fixes #436)

### LINKED ISSUE #436: Incorrect notehead displacement when `.setStemDirection` is called after construction
~~Take a look at the `EasyScore - Draw Basic` test. The chord's noteheads are incorrectly displaced.~~

Take a look at what happens if you call `setStemDirection(Stem.DOWN)` after the construction of a stem-up chord with displaced noteheads.

image

My guess is that the noteheads only get built on construction, and so changing the stem after construction does not change placement. That being said, the displacement of noteheads is certainly a kind of "formatting" and not something to be done on construction. However, that's a larger change, and in the meantime we should rebuild when `setStemDirection()` is called.

Silverwolf90:
Related test included. Feel free to close if not necessary.

### DIFF
--- a/src/stavenote.js
+++ b/src/stavenote.js
@@ -426,6 +426,8 @@ export class StaveNote extends StemmableNote {
 
   // Calculates and stores the properties for each key in the note
   calculateKeyProps() {
+    this.keyProps = [];
+
     let lastLine = null;
     for (let i = 0; i < this.keys.length; ++i) {
       const key = this.keys[i];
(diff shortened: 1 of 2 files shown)


0xfe:
I fixed this with the commit-hooks merge, so this is not necessary.

FYI, this change would have broken the auto beam tests. :-)

Silverwolf90:
Hmm yeah you're right, do you know why? Did you run into that same issue?

0xfe:
Yes, ran into it. When `setStemDirection` is called after `setStave`, the new note heads lose their stave (and hence their y-positions.) I fixed it by calling `setStave` on the new note heads after reconstruction.

Issue #349, opened 2016-06-11
Proposed steps to prepare for SMuFL support


Silverwolf90:
I wanted to get this discussion going again, as I've been giving a bit of thought to VexFlow lately. I don't have much time these days, but I figured I could lay down my thoughts on how we could get to SMuFL support and maybe we could agree on a plan.

I previously worked on a branch to support SMuFL ([branch](https://github.com/Silverwolf90/vexflow/tree/remove-hardcoded-head-width) related to https://github.com/0xfe/vexflow/issues/181), and while I got it into a working state it was quite a mess. Such a drastic change permeates through the entire codebase, and I'm not sure the current codebase is good enough to support such a large change in a clean manner. Although, the new testing infrastructure should also help significantly.

The first things that need to be done is to improve font infrastructure without doing anything SMuFL specific:
- **Rely _only_ on glyph metrics provided by automatic font processing**
  - Get accurate bounding box information for each glyph
    - In the font files, the height (eg: `Vex.Flow.Font.glyphs['v0'].ha`) values seem to be completely off. Take a look at the following image, but ignore horizontal/vertical alignment of the box on the glyph, it should be obvious that the height of some of the boxes clearly don't fit the glyphs.
      ![image](https://cloud.githubusercontent.com/assets/1631625/15986828/fdce9166-300a-11e6-916a-a4a8a0fecc47.png)
  - Bounding box data is generally not available on a per-glyph basis in most font encodings. But there are tools to extract the bounding box from an SVG path (I've not used them though):
    - https://github.com/icons8/svg-path-bounding-box
    - https://www.npmjs.com/package/svg-bounding-box
- **Implement horizontal/vertical origin alignment on the `Glyph` class**
  - Some kind of api like: `glyph.setAlignH('center')` `glyph.setAlignV('baseline')`
  - This removes need for low-level fine-tuning in high-level classes, ie: the `Ornament` class shouldn't need to worry about pixel amounts to center the `Glyph` at a point
  - I implemented this functionality in that old branch, probably a lot of that code can be re-used
- **Use appropriate glyphs for articulations that are `above`/`below`** 
  - VexFlow articulations only use the glyph meant for placement above a note, even if it placed below. But Gonville has specific glyphs with origin points meant for articulations placed below a note
  - Articulations should not need to rely on these awkward `shift_up` and `shift_down` properties in `Vex.Flow.articulationCodes.articulations`
- **!!! Remove all hardcoded numbers around glyph positioning !!!**
[...]

PR #454, opened 2016-09-04
Remove hardcoded accidental metrics

Silverwolf90:
Removes hardcoded glyph metrics in `tables.js`. Lots of small diffs are expected.

Working towards https://github.com/0xfe/vexflow/issues/349


0xfe:
Looks good. Lots of visual diffs.

One noticeable regression is with cautionary accidentals, where the glyph isn't centered between the parenthesis.

screen shot 2016-09-04 at 12 00 43 pm

Silverwolf90:
It's better now, since the double sharp looks good. I suspect the bounding boxes for the flat aren't perfect because they still look like there's too much padding on the right. This might not be fixable until SMuFL support.

Silverwolf90:
image

0xfe:
How hard would it be to add a shift to the flats? They do kinda look off.

Silverwolf90:
Not hard, just gross. Basically have to reinstate hardcoded constants per accidental.

0xfe:
Yes, gross I agree, but with these changes, less gross than before. We only need shift values for the flats in `accidentalCodes`. (It sucks, but that's our burden, not the user's.)

mscuthbert:
If it's a consolation, eventually flats will need to be a special case in bounding box specification anyhow since the stem of a flat or double flat is the only musical character that a slur is allowed to intersect with (under normal circumstances).  So you can be content that special flat bounding box adjustments were going to be inevitable eventually anyhow.  :-)

Silverwolf90:
Fixed. Added a shift to pretty much all accidentals so symmetrical left/right padding (ie: identical values) produces symmetrical results.

image

@mscuthbert Good point! But that information is ideally extracted from `SMuFL` metadata, in your example, [the `accidentalFlat` glyph has a cutout that provides this y coordinate (in staff spaces) without us having to store it.](http://w3c.github.io/smufl/gitbook/specification/bbox-cut-outs.html)

0xfe:
Looks good. Merging.

PR #456, opened 2016-09-05
Refactor beam tests

Silverwolf90:
This is nearly done.

But for some reason **Outlier Beam** is looking weird -- notice the beam slopes:

image

Haven't figured out why yet.

### DIFF
--- a/src/factory.js
+++ b/src/factory.js
@@ -32,6 +32,8 @@ import { GraceNoteGroup } from './gracenotegroup';
 import { EasyScore } from './easyscore';
 import { ClefNote } from './clefnote';
 import { PedalMarking } from './pedalmarking';
+import { TabNote } from './tabnote';
+import { TabStave } from './tabstave';
 
 // To enable logging for this class. Set `Vex.Flow.Factory.DEBUG` to `true`.
 function L(...args) { if (Factory.DEBUG) Vex.L('Vex.Flow.Factory', args); }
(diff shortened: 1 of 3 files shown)


0xfe:
Is this ready for review?

Silverwolf90:
I haven't really looked into the wrong beam angles I highlighted above yet. However, I don't think there's an issue with the test. Could you take a look at the `outlier` test's code? If you don't see anything wrong, and the other diffs are acceptable, we could merge and open a separate issue for this bug.

Silverwolf90:
On another note, I think the beam tests aren't thorough enough given the sheer variety that is encountered in printed music. And there are a lot of tests which don't provide any clear value over others. This is actually the case with a lot of VexFlow's tests, but the beam tests are a particularly good example.

For beams, combinations of the following should be tested:

- Durations
- Ascending slopes
- Descending slopes
- Flat slopes
- Beams sitting inside the staff
- Beams sitting outside the staff
- Variety of Widths
- Concave Note Groups
- Convex Note Groups
- Tight note groups (ie: close together, small intervals)
- Loose voice groups (ie: far apart, big intervals)
- Note groups starting on a line
- Note groups starting in a space

The most thorough test suite would take the Cartesian product of all of these,  although that is rather unmanageable :) But with a decent sampling of precise combinations will go a long way.

VexFlow's current `Beam` class don't respond to all these cases (like inside/outside staff) but handling all of this is certainly something to strive for.

0xfe:
Sorry, I'm traveling right now, so a bit slow... will review as soon as I get a chance.

Silverwolf90:
[Looked into the weird beams and found this bug](https://github.com/0xfe/vexflow/issues/485). Since there's nothing wrong with my changes I'm going to merge this in shortly.

Issue #457, opened 2016-09-06
`Factory` method conventions


Silverwolf90:
So far, the `Factory` methods seem bring a bit more consistency to object construction. Now that we've implemented a variety of objects, it might be worth defining some conventions. I think there's a bit of inconsistency that could be ironed out.

Two forms of Factory method arguments exist so far:
1. An object with all config properties at the top level
   - The object is passed straight into the respective constructor
   - eg: `StaveNote(noteStruct)`, `GraceNote(noteStruct)`
2. An object with properties and a nested `options` object
   - The object is picked apart and the values passed appropriately -- this could be to a constructor or to subsequent method calls.
   - eg: `Beam(params)`, `Stave(params)`, `Annotation(params)`

For the latter, it's not really clear to me how the `options` object is intended to be used. Take the defaults for `Annotation` :

``` javascript
{
  text: 'p',
  vJustify: 'below',
  hJustify: 'center',
  fontFamily: 'Times',
  fontSize: 14,
  fontWeight: 'bold italic',
  options: {},
}
```

Since the font/justification properties are at the "top level", what goes in the `options` object? Not that they shouldn't be "top level" properties, but the distinction between the two tiers is not clear to me.

0xfe:
I agree this is a good time to raise and rethink this. The inconsistency in `StaveNote` has bothered me too, but I'm not entirely sure how to rationalize, because I think it's pretty reasonable.

I think it's important to note that the `Factory` API is not set in stone and we can iterate on it. My intended pattern for factory methods was something like:

```javascript
Factory.Element(params, factoryOptions)
```

Above `params` should look like this:

```javascript
{
  param1: ...,
  param2: ...,
  options: { ... }
}
```

The `paramX` fields are **required** and `options` include everything optional. So for `Annotation`, ideally we'd do this:

```javascript
{
  text: 'p',
  options: {
    vJustify: 'below',
    hJustify: 'center',
    fontFamily: 'Times',
    fontSize: 14,
    fontWeight: 'bold italic',
  }
}
```

It's not like that now, but perhaps we should settle on it?

0xfe:
Looking at `factory.js` now, it looks like most of the methods are sane. Only `Annotation` is inconsistent. (Ignoring `StaveNote` and `GraceNote` for now.)

0xfe:
And thinking about `StaveNote` and `GraceNote`, perhaps we should:

```javascript
{
  keys: [...],
  duration: ...,
  options: {
[...]

PR #458, opened 2016-09-06
Refactor clef tests

Silverwolf90:
Not sure about the `Factory.prototype.ClefNote` options. But that's related to a more general question around Factory methods here in https://github.com/0xfe/vexflow/issues/457


0xfe:
Good to go?

Silverwolf90:
Yup! Following our discussion in https://github.com/0xfe/vexflow/issues/457 the `ClefNote` factory method `options` has `size` and `annotation` since they are optional for construction.

PR #460, opened 2016-09-11
Fix BoundingBox mergeWith function

panarch:
Hi Mohit,

I found a bug which is very simple but obvious, so I fixed and here is the PR.
Current BoundingBox mergeWith function was not working properly for some cases;
Box1 contains Box2 -> error calculating new_w or new_h

Thanks!
Taehoon


panarch:
hmm... build failed while installing slimerJS.

Same ci test result of my fork is this, done without any issue, https://travis-ci.org/panarch/vexflow/builds/159087097

Silverwolf90:
Lint looks fine on my end! Merging. Thanks for the fix!

PR #461, opened 2016-09-12
Add sori and koron to gonville_all font

Huji:
WIP, please don't merge yet


infojunkie:
You need to clone `gonville_all.js` into `vexflow_font.js`: the latter is where VF gets its glyphs. Actually, maybe you don't need to modify `gonville_all.js` at all, since your glyphs are not coming from Gonville. 

Also, would be great to add tests to the following places:
- `Vex.Flow.Test.Accidental.microtonal`
- `VFT.EasyScore.accidentals`
- `VFT.EasyScore.drawAccidentalsTest`

Huji:
That can't be it. Even after making the change you suggested and rebuilding the project, I still don't see the said glyphs on transform.html (I do see them on glyphs.html though). Besides, the content of vexflow_font.js should be updated using transform.html and if that page doesn't show the glyphs, I shouldn't be updating the vexflow_font.js by hand.

@0xfe can you please take a look here and tell me what I am doing wrong?

infojunkie:
There are a number of issues here. 

- transform.html hasn't been updated to reflect the latest syntax used by VexFlow. Compare the output of the "Glyph Code" section (i.e. `Vex.Flow.Font = {...`) to the current contents of vexflow_font.js:  `export var Font = {...` here: https://github.com/0xfe/vexflow/blob/master/src/fonts/vexflow_font.js#L1

- also, the reason that your additions to gonville_all.js are not showing up in transform.html is that the variable `Vex.Flow.Font` is already defined by vexflow-debug.js. Maybe in the past it wasn't included in the same manner, but that's what is happening now. If you modify gonville_all.js to read `Vex.Flow.Font.Original = {`, you will see your additions inside the `Original` key.

In summary, transform.html needs to be updated to reflect the latest changes in VexFlow.

0xfe:
Hey folks, Sorry about the delay -- I'm currently traveling and behind on these reviews. My really quick thoughts:

- Please create a new file under fonts for these new glyphs. Maybe `fonts/quarter_tone` or `fonts/international` or something like that.
- Add a `.md` doc to `fonts/` and document the new glyphs, particularly the sources (bravura).

`transform.html` is really old, and its goal was to pick out and visualize glyphs from `goneville_all.js`, which generates a blob of text that is pasted into `vexflow_font.js`. It's really ugly and hacky, but it did the job and wasn't used frequently enough to bother improving it.

So... ideally, we need something like this:

- A new NodeJS script (perhaps `tools/generate_fonts.js`) which takes glyphs from sources like `goneville_all.js`, `quarter_tone.js` etc. and generates `vexflow.js`. We could simply copy the maps from `transform.html` and add the new quarter tones there.
- Remove `transform.html`  or replace with some simple `.html` to only display all the glyphs from a font file.

The above would be ideal, but if you want to do it quick, then a decent stopgap would be to modify `transform.html` so it picks fonts from both `goneville_all.js` and your new `quarter_tone.js` file. (So in the least, I'd like the new glyphs in a separate file with the sources documented.)

How does that sound?

infojunkie:
Sounds good to me. The `glyphs.html` file already exists to display all glyphs.

Huji:
Alright, I am going to focus on adding the new glyphs for now, and will get to fixing `transform.html` at a later time. And @0xfe your wish is granted: the new glyphs are now in a new font file called `src/fonts/microtonal.js`

I also added the code to `vexflow_font.js` as suggested by @infojunkie and now the new glyphs correctly show in `src/glyphs.html`

And as you already know, I have also updated `src/tables.js` to define new accidental code for the two new accidentals: `s` for sori and `k` for koron.

Now I am going to add the tests.

Huji:
@infojunkie I am not familiar with EasyScore, and can't tell what to put for must pass or must fail. Can you please help me?

My ideas are that these must pass: c#s, cs, ck, cbk

And these must fail: cs#, ckb, csk, cks

What are your thoughts?

infojunkie:
EasyScore is a parser to add VexFlow elements. To support EasyScore, you should add all possible allowed combinations to the regex on this line: https://github.com/0xfe/vexflow/blob/master/src/easyscore.js#L134

And then add the tests that you mentioned.

Huji:
@infojunkie I reviewed some credible books containing notation using Iranian accidentals; it seems like combined accidentals (like bemol + koron) are generally avoided. The standard key signatures only include single accidentals, and modulations within the same piece are generally written in a way that doesn't require combined accidentals. Therefore, I marked all combos as mustFail in easyscore test. Pushing now.

infojunkie:
Great! Will pull here and review.

infojunkie:
Pulled and looks good so far. It would be good to add new accidentals to `VFT.EasyScore.drawAccidentalsTest`.

Thanks!

infojunkie:
Also, I found that page that describes the use of flat koron and sharp sori: https://sites.google.com/site/240edo/about24-edo. From my point of view, it would be beneficial to include them in the allowed set of accidentals - VexFlow acting as an inclusive system, instead of being prescriptive.

Huji:
Yes, Vaziir had foreseen flat koron and sharp sori, but I didn't find examples of it in more modern notation. But your point is well taken and we should be inclusive.

The questions is should I create new glyphs for these combined accidentals? Or is there a way to tell VexFlow to show two accidentals back to back?

infojunkie:
As far as I understand the code, we need to list the accidental combinations explicitly, there is no provision for chaining arbitrary sequences.

Huji:
In that case, it'll take me some time to create the new combined glyphs for sharp sori and flat koron. For now, I added the tests by modifying the drawAccidentalsTest function.

Silverwolf90:
What's the status on this PR at the moment?

Huji:
I think it can actually be merged as is. I can add combined glyphs later (they are almost never used anyway).

I just need someone else to confirm that this does the job correctly (in my tests, it does).

infojunkie:
Tests pass for me too.

Silverwolf90:
According to @0xfe's first response, we still need a markdown file that declares sources for the new glyphs. And we need a mechanism for generating this file again when we add new glyphs. At least a modification to `transform.html` that merges this new `microtonal` font.

src/fonts/microtonal.js
```diff
@@ -0,0 +1,9 @@
+Vex.Flow.Gonville = {
```
Silverwolf90: This should probably be `Vex.Flow.Microtonal` or something, right?

src/fonts/microtonal.js
```diff
@@ -0,0 +1,9 @@
+Vex.Flow.Gonville = { 
+'glyphs': {
+ 'vd0': { 'x_min': -10.890625, 'x_max': 299.4375, 'ha': 294, 'o': 'm 44 174 b 51 174 47 174 49 174 b 68 173 55 174 61 174 l 287 112 l 551 40 b 615 20 617 22 609 23 b 626 0 622 16 626 8 b 615 -22 626 -9 622 -18 b 613 -23 613 -23 613 -23 b 613 -23 613 -23 613 -23 b 287 -113 613 -24 597 -29 l 68 -174 b 53 -176 61 -176 57 -176 b 39 -172 47 -176 43 -174 b 27 -151 31 -167 27 -159 b 39 -129 27 -141 31 -133 b 230 -74 43 -124 20 -131 l 370 -36 l 468 -9 b 498 0 484 -4 498 -1 b 468 8 498 0 484 2 l 370 34 l 230 73 b 40 126 28 129 43 124 b 27 149 31 131 27 140 b 44 174 27 161 34 170 m 205 110 l 205 300 b 205 330 245 330 245 300 l 245 300 l 245 -300 b 245 -330 205 -330 205 -300 l 205 -300 l 205 110 l 345 90 m 345 90 l 345 330 b 345 360 385 360 385 330 l 385 330 l 385 -270 b 385 -300 345 -300 345 -270 l 345 -270 l 345 90 '},
```
Silverwolf90: Must we stick with the `v<hex>` convention for the glyph keys? @0xfe is there a particularly reason for it? Could these keys just be `sori` and `koron`?

Ultimately this will all be deprecated when we move to SMuFL but that could still be for some time...

src/fonts/microtonal.js
```diff
@@ -0,0 +1,9 @@
...
+ 'vd1': { 'x_min': -20, 'x_max': 320, 'ha': 257, 'o': 'm -8 200 b -8 210 8 200 16 200 l 20 148 -199 l 23 -615 b 0 -629 9 -629 6 -629 l -21 -612 l -21 -201 l -21 216 l -20 200 m 16 200 l 310 0 l 240 0 l 16 140 l 16 -120 l 240 0 l 310 0 l 16 -200 '}
+},
+ 'cssFontWeight': 'normal', 'ascender': 1903, 'underlinePosition': -125, 'cssFontStyle': 'normal', 'boundingBox': { 'yMin': -2065.375, 'xMin': -695.53125, 'yMax': 1901.578125, 'xMax': 1159.671875 },
+ 'resolution': 1000, 'original_font_information': { 'postscript_name': 'Gonville-18', 'version_string': 'Version 0.1.8904 ', 'vendor_url': '', 'full_font_name': 'Gonville-18', 'font_family_name': 'Gonville-18', 'copyright': 'No copyright is claimed on this font file.', 'description': '', 'trademark': '', 'designer': '', 'designer_url': '', 'unique_font_identifier': 'FontForge 2.0 : Gonville-18 : 17-3-2010', 'license_url': '', 'license_description': '', 'manufacturer_name': '', 'font_sub_family_name': '18' },
```
Silverwolf90: Also this metadata for Gonville surely doesn't apply, right?

src/fonts/microtonal.js
Huji: Will be fixed in the upcoming push

src/fonts/microtonal.js
Huji: You are correct, they'll all eventually go away. Until then I would rather keep things consistent.

src/fonts/microtonal.js
Huji: I am happy to change it; suggestions?

src/fonts/microtonal.js
Silverwolf90: Fair enough

src/fonts/microtonal.js
Silverwolf90: If the glyphs look acceptable when rendered, lets just remove all this metadata. Most of it isn't used by VexFlow in the first place.

Huji: Done.

PR #465, opened 2016-10-03
Removed unnecessary Array.prototype.slice() call

pete-otaqui:
ES6 rest parameter (`...args` in this case) returns a true array, not an "arguments" object, so no need for this indirection


Silverwolf90:
Good catch! Merging.

PR #486, opened 2016-12-18
Refactor Curve tests

Silverwolf90:
Diffs will be due to stave position/size changes


0xfe:
Way more readable. Looks great. Thanks.

PR #487, opened 2016-12-18
Refactor TextNote tests

Silverwolf90:
Diffs due to staves changing position/size.


0xfe:
Thanks. Looks good to me.

PR #488, opened 2016-12-18
Add TextBracket to Factory; refactor TextBracket tests

Silverwolf90:
Note the destructured `TextBracket` constructor parameters. I'm personally very fond of this pattern (I use it extensively with my React projects since all React components take a single `props` object). Makes default values really convenient without cluttering up the assignments in the body.


0xfe:
Thanks, yes, I like the destructured constructor parameters too. It's convenient but also explicit, so a reader looking at the API knows at first glance what the fields should look like.

Issue #489, opened 2016-12-18
StaveModifiers placed at the end of the Stave collide with StaveTies without end notes.


Silverwolf90:
See the collision at the end of the stave in this example:

![image](https://cloud.githubusercontent.com/assets/1631625/21296197/2c0abc56-c55e-11e6-80f7-bd9eba3aa891.png)

StaveModifiers placed at the beginning do *not* exhibit this issue (although the spacing leaves something to be desired):

![image](https://cloud.githubusercontent.com/assets/1631625/21296204/4f1d9b96-c55e-11e6-935a-00750fd3aaa4.png)

0xfe:
This is because `Stave.setTieEndX()` does not accommodate for modifiers.

From `src/stave.js`:

```javascript
  getTieStartX() { return this.start_x; }
  getTieEndX() { return this.x + this.width; }
```

So while `getTieStartX()` is based on `Stave.start_x` which changes as modifiers are added, `getTieEndX()` is just the last horizontal pixel on the stave.

Silverwolf90:
Cool, simple fix then.

PR #490, opened 2016-12-18
Refactor StaveTie tests

Silverwolf90:
~So I really got a little more experimental... but I think this managed to reduce boilerplate really clearly and isolates the parts being tested.~

~I've always found the string+test mapping in each `Start` function to be pretty annoying. So I got rid of it and rely on iteration through the keys using a `runModule` test helper.~

~Definitely looking for feedback on this!~

Also bug https://github.com/0xfe/vexflow/issues/489 was found during the course of this refactor.

### DIFF
--- a/src/factory.js
+++ b/src/factory.js
@@ -358,7 +358,9 @@ export class Factory {
       first_indices: [0],
       last_indices: [0],
       text: null,
-      options: {},
+      options: {
+        direction: undefined,
+      },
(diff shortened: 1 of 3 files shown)


Silverwolf90:
Actually, I ended up changing my mind on the structure! The original way I setup made didn't guarantee an order of tests since they were set up as object keys.

PR #491, opened 2016-12-18
Refactor GhostNote tests

Silverwolf90:
* Added `GhostNote` to `Factory`
* Added a `.getVoices` method to `Factory` so that the created voices can be accessed.
* Continuing with the `createTest` pattern I've been refining over the past few PRs

### DIFF
--- a/src/factory.js
+++ b/src/factory.js
@@ -33,6 +33,7 @@ import { EasyScore } from './easyscore';
 import { ClefNote } from './clefnote';
 import { PedalMarking } from './pedalmarking';
 import { TextBracket } from './textbracket';
+import { GhostNote } from './ghostnote';
 import { TabNote } from './tabnote';
 import { TabStave } from './tabstave';
 import { TextNote } from './textnote';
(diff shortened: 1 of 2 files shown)


src/factory.js
```diff
@@ -157,6 +159,14 @@ export class Factory {
     return note;
   }
 
+  GhostNote(noteStruct) {
+    const ghostNote = new GhostNote(noteStruct);
+    if (this.stave) ghostNote.setStave(this.stave);
+    ghostNote.setContext(this.context);
+    this.renderQ.push(ghostNote);
+    return ghostNote;
```
0xfe: These four lines are duplicated all over this class. I wonder if we could replace them with something like `return this.setupElement(ghostNote)`.

No action needed here -- just a thought.

0xfe:
Looks good. (Added a comment for discussion.)

src/factory.js
Silverwolf90: Totally agree. Might make the most sense to do after most elements are in the Factory. That should more-or-less coincide with the completion of the test refactoring.

PR #492, opened 2016-12-19
Make Curve tests more consistent with recent test pattern

Silverwolf90:
* Moved `createTest` out of object
* Alias `VF.Test.runTests` to `run`


Silverwolf90:
Diffs pass. Merging.

PR #495, opened 2016-12-21
Reset noteheads when `StaveNote#setKeyLine` is called; Refactor threevoice_tests.js;

Silverwolf90:
Take a look at the diffs. They are significant due to the new test setup affecting the automatic rest alignment.

Some of these tests relied on formatting code that was executed *only if* missing a stave (which really doesn't make sense to me why any kind of multi-voice formatting would be attempted in this case). Anyway, we are soon making this an invalid state. So now, the rests are formatted by code that formats StaveNotes with a stave. But the output is not good. The vertical adjustments will have to be completely rewritten to take real bounding box metrics into account (but this can only happen once the new stave requirement is in place), and changes will need to be made to support >2 voices.

See comment in code for more details.

### DIFF
--- a/src/stavenote.js
+++ b/src/stavenote.js
@@ -57,7 +57,21 @@ export class StaveNote extends StemmableNote {
   static format(notes, state) {
     if (!notes || notes.length < 2) return false;
 
-    if (notes[0].getStave() != null) return StaveNote.formatByY(notes, state);
+    // FIXME: VexFlow will soon require that a stave be set before formatting.
+    // Which, according to the below condition, means that following branch will
+    // always be taken and the rest of this function is dead code.
+    //
+    // Problematically, `Formatter#formatByY` was not designed to work for more
+    // than 2 voices (although, doesn't throw on this condition, just tries
+    // to power through).
+    //
+    // Based on the above:
+    //   * 2 voices can be formatted *with or without* a stave being set but
+    //     the output will be different
+    //   * 3 voices can only be formatted *without* a stave
+    if (notes[0].getStave()) {
+      return StaveNote.formatByY(notes, state);
+    }
 
     const notesList = [];
 
@@ -725,6 +739,7 @@ export class StaveNote extends StemmableNote {
   setKeyLine(index, line) {
     this.keyProps[index].line = line;
     this.note_heads[index].setLine(line);
+    this.reset();
     return this;
   }
 
--- a/src/stavenote.js
+++ b/src/stavenote.js
@@ -281,7 +281,7 @@ export class StaveNote extends StemmableNote {
       const areNotesColliding = bottomNoteTopY - topNotBottomY < 0;
 
       if (areNotesColliding) {
-        xShift = topNote.getVoiceShiftWidth();
+        xShift = topNote.getVoiceShiftWidth() + 2;
         bottomNote.setXShift(xShift);
       }
     }
--- a/src/stavenote.js
+++ b/src/stavenote.js
@@ -738,7 +738,6 @@ export class StaveNote extends StemmableNote {
 
   setKeyLine(index, line) {
     this.keyProps[index].line = line;
-    this.note_heads[index].setLine(line);
     this.reset();
     return this;
   }
--- a/tests/threevoice_tests.js
+++ b/tests/threevoice_tests.js
@@ -3,594 +3,241 @@
  * Copyright Mohit Muthanna 2010 <mohit@muthanna.com>
  */
 
+ /*
+ eslint-disable
+ no-var,
+ no-undef,
+ wrap-iife,
+ func-names,
+ vars-on-top,
+ max-len,
+ object-shorthand,
+ prefer-arrow-callback,
+  */
+
 VF.Test.ThreeVoices = (function() {
+  function concat(a, b) { return a.concat(b); }
+
+  function createThreeVoicesTest(noteGroup1, noteGroup2, noteGroup3, setup) {
+    return function(options) {
+      var vf = VF.Test.makeFactory(options, 600, 200);
+      var stave = vf.Stave().addTrebleGlyph().addTimeSignature('4/4');
+      var score = vf.EasyScore();
+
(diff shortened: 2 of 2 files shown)


0xfe:
Can you paste in some visual diffs please?

Silverwolf90:
Ah yeah no problem, I'll make a post with those and walk through the changes too. Looking over the diffs again, I think some will need explanation.

src/stavenote.js
```diff
@@ -725,6 +739,7 @@ export class StaveNote extends StemmableNote {
   setKeyLine(index, line) {
     this.keyProps[index].line = line;
     this.note_heads[index].setLine(line);
+    this.reset();
```
Silverwolf90: This change is important. `StaveNote#reset()` rebuilds the `NoteHeads`, which clearly must be done after changing the line of a key. While this is a bug fix, it affects some of the test output in a negative way.

0xfe:
Also, at a high-level the three voice code is very fragile, and I'm not sure if we want to keep it this way. This is one of those cases where the benefit is not worth the cost, and we should probably look for a more general way to solve this problem.

src/stavenote.js
Silverwolf90: And actually the line above it probably needs to be removed (and `setLine` does nothing other than set a `line` property)

Silverwolf90: Also worth mentioning that while we could update the notehead individually, I think it's better to use the `.reset()` as a catch-all "something has changed, please rebuild" (which also indicates that formatting is happening before `.format` is called, but that's another story).

Silverwolf90:
First of all there are some (initially unexpected) diffs around rests in beams. These changes are related to [the line I added and described here.](https://github.com/0xfe/vexflow/pull/495/files#diff-127a29276b44875a1cde4a2ae983b9c9R742)

I recently found out that rests within beams are always aligned. But `Formatter#AlignRestsToNotes` uses `StaveNote#setKeyLine` which was broken until the above change in this PR. The rest wasn't shift position because the `NoteHead` (which contains the rest glyph) wasn't being updated. Now that it is being updated, we are seeing the following position changes (this is not all of them, but a sampling):

#### Auto_Beaming
image
image
image
auto_beaming beam_across_all_rests_with_stemlets_blessed
auto_beaming beam_across_all_rests_with_stemlets_current
auto_beaming beam_across_all_rests
auto_beaming break_beams_on_middle_rests_only_blessed
auto_beaming break_beams_on_middle_rests_only_current
auto_beaming break_beams_on_middle_rests_only

#### Tuplet.Complex_Tuplet
image
image
image

And here are the diffs for the tests I changed glyphs (remember that the tests with 3 voices s are now being formatted  with algorithm that wasn't designed for 3 voices):

three_voice_rests auto_adjust_rest_positions___three_voices__1_blessed
three_voice_rests auto_adjust_rest_positions___three_voices__1_current
three_voice_rests auto_adjust_rest_positions___three_voices__1
three_voice_rests auto_adjust_rest_positions___three_voices__2_blessed
three_voice_rests auto_adjust_rest_positions___three_voices__2_current
three_voice_rests auto_adjust_rest_positions___three_voices__2
three_voice_rests auto_adjust_rest_positions___two_voices_blessed
three_voice_rests auto_adjust_rest_positions___two_voices_current
three_voice_rests auto_adjust_rest_positions___two_voices
three_voice_rests three_voices____1_blessed
three_voice_rests three_voices____1_current
three_voice_rests three_voices____1
three_voice_rests three_voices____2_complex_blessed
three_voice_rests three_voices____2_complex_current
three_voice_rests three_voices____2_complex
three_voice_rests three_voices____3_blessed
three_voice_rests three_voices____3_current
three_voice_rests three_voices____3

Silverwolf90:
I added some padding to the displaced notes. Which improves some of the output. But concerning rests, those basically stay the same as the previous set of diffs. We'll need to rewrite `StaveNote.formatByY`.

Going to merge.

### New diffs
accidental multi_voice_blessed
accidental multi_voice_current
accidental multi_voice
auto_beaming beam_across_all_rests_blessed
auto_beaming beam_across_all_rests_current
auto_beaming beam_across_all_rests_with_stemlets_blessed
auto_beaming beam_across_all_rests_with_stemlets_current
auto_beaming beam_across_all_rests_with_stemlets
auto_beaming beam_across_all_rests
auto_beaming break_beams_on_middle_rests_only_blessed
auto_beaming break_beams_on_middle_rests_only_current
auto_beaming break_beams_on_middle_rests_only
dot multi_voice_blessed
dot multi_voice_current
dot multi_voice
formatter stavenote_justification_blessed
formatter stavenote_justification_current
formatter stavenote_justification
rests auto_align_rests___multi_voice_blessed
rests auto_align_rests___multi_voice_current
rests auto_align_rests___multi_voice
stavenote center_aligned_note___multi_voice_blessed
stavenote center_aligned_note___multi_voice_current
stavenote center_aligned_note___multi_voice
three_voice_rests auto_adjust_rest_positions___three_voices__1_blessed
three_voice_rests auto_adjust_rest_positions___three_voices__1_current
three_voice_rests auto_adjust_rest_positions___three_voices__1
three_voice_rests auto_adjust_rest_positions___three_voices__2_blessed
three_voice_rests auto_adjust_rest_positions___three_voices__2_current
three_voice_rests auto_adjust_rest_positions___three_voices__2
three_voice_rests auto_adjust_rest_positions___two_voices_blessed
three_voice_rests auto_adjust_rest_positions___two_voices_current
three_voice_rests auto_adjust_rest_positions___two_voices
three_voice_rests three_voices____1_blessed
three_voice_rests three_voices____1_current
three_voice_rests three_voices____1
three_voice_rests three_voices____2_complex_blessed
three_voice_rests three_voices____2_complex_current
three_voice_rests three_voices____2_complex
three_voice_rests three_voices____3_blessed
three_voice_rests three_voices____3_current
three_voice_rests three_voices____3
tuplet complex_tuplet_blessed
tuplet complex_tuplet_current
tuplet complex_tuplet

PR #498, opened 2016-12-22
Refactor Dot tests

Silverwolf90:
dot multi_voice_blessed
dot multi_voice_current
dot multi_voice


Silverwolf90:
Updated with diffs.

Silverwolf90:
This change is also dependent on https://github.com/0xfe/vexflow/pull/499

(The above diffs I believe were run with the above fix, which is why there is more width in the current version)

PR #500, opened 2016-12-23
Lint voices_test.js; slight refactor of "Full Voice Mode" test

Silverwolf90:
Basically just linting, and making sure the stave is set early.


Silverwolf90:
voice full_voice_mode_test_blessed
voice full_voice_mode_test_current
voice full_voice_mode_test

PR #511, opened 2016-12-26
Fix blurry canvas rendering on retina screens

Silverwolf90:
Use `window.devicePixelRatio` when setting up canvas


Silverwolf90:
## Comparison

#### Before:


#### After:
image

ppanyukov:
Good day, not sure if this is fixed? I'm referring to the blurred canvas rendering on MacBook Pro with retina on the http://www.vexflow.com and http://www.vexflow.com/vextab/tutorial.html

Some rendering switch (Canvas/SVG) on the sites would be nice feature?

PR #512, opened 2016-12-26
Rename `selector` to `elementId`

Silverwolf90:
Pretty much every occurrence of the name `selector`/`sel` is incorrect. [The `Renderer` class uses `document.getElementById` to get a DOM Element](https://github.com/0xfe/vexflow/pull/512/files#diff-5326222f837d36fd2bff476584c07621R129) so in fact passing in a valid selector wouldn't even work. An id on its own, eg: `'container'`, is not the same as a selector which uses an id, eg:  `'#container'`.

### DIFF
--- a/src/canvascontext.js
+++ b/src/canvascontext.js
@@ -98,8 +98,7 @@ export class CanvasContext {
   }
 
   resize(width, height) {
-    return this.vexFlowCanvasContext.resize(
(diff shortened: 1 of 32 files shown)


Silverwolf90:
The other PRs should be merged in before this, there will likely be a lot of minor conflicts.

PR #516, opened 2016-12-27
Lint tests

Silverwolf90:


0xfe:
Wow, you're on a roll here. I just logged back in to a ton of notifications! :-)

Good stuff!

PR #517, opened 2016-12-28
Refactor transform.html; include new Microtonal glyphs in generated font

Silverwolf90:
Self explanatory.


Silverwolf90:
The generated font file causes no diffs 🎉

PR #527, opened 2017-01-12
 stroke outline of NoteHeads

shimondoodkin:
just implemented stroke outline of NoteHeads

not sure how to check if it goes well with other things

not sure how to add it to documentation

I had a need to for bi-colored notes to color colored notes of sharp notes

colorednotes

### DIFF
--- a/src/glyph.js
+++ b/src/glyph.js
@@ -86,14 +86,15 @@ export class Glyph extends Element {
    * @param {number} point The point size to use.
    * @param {string} val The glyph code in Vex.Flow.Font.
    * @param {boolean} nocache If set, disables caching of font outline.
+   * @param {boolean} stroke If set, also strokes the outline.
    */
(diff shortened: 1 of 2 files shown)


shimondoodkin:
I believe it is safe to merge, and will not be interfering with other things.

usage:
staveNote.setKeyStyle(noteIndex, {fillStyle: color, headStrokeStyle:color2, headStrokeWidth:3 });

SalahAdDin:
@shimondoodkin Awesome!

impactcolor:
@shimondoodkin is there an update to the new notehead.js?
I tried to integrate your update but the notehead.js file has changed

ronyeh:
Just looking through old PRs. I agree there should be more flexible ways to style individual glyphs.

However, I would suggest that we "archive" (i.e., close) this since the last activity is over four years ago. If there is new momentum to make this happen, we can open a new PR and reference this one.

0xfe:
Agreed.

Issue #552, opened 2017-05-31
Running "webpack:build" (webpack) task 95% emit


match08:
grunt  ->   Running "webpack:build" (webpack) task 95% emit

match08:
![image](https://cloud.githubusercontent.com/assets/616730/26624659/22d2e6c6-4624-11e7-8cc2-b0e43c2b2074.png)

Issue #541, opened 2017-03-20
Webpack dependency is incompatible with current config object


eliot-akira:
In `package.json`, the Webpack dependency is specified as `"*"`. It installs version 2.x, with breaking changes in the config object - so the VexFlow build is interrupted with an error.

**Changing the Webpack dependency to `"1.*"`** got the build to complete.

Perhaps this can be included in PR #525? Updating the config object to Webpack 2.x might take some effort.

mmathys:
@0xfe did you abandon this project? :(

Issue #528, opened 2017-01-21
Missing dependency on babel-core


tinne:
Building from source as documented leads to an error:
```
$ grunt
Running "eslint:target" (eslint) task

Running "webpack:build" (webpack) task                                                                                Version: webpack 1.14.0

ERROR in Cannot find module 'babel-core'
Warning: Task "webpack:build" failed. Use --force to continue.

Aborted due to warnings.
```

Manually installing babel-core works:

```
$ npm install babel-core
vexflow@1.2.83 D:\Projekte\Musiknotation\vexflow
`-- babel-core@6.22.1
  +-- babel-generator@6.22.0
  | +-- detect-indent@4.0.0
  | `-- jsesc@1.3.0
  +-- babel-helpers@6.22.0
  +-- babel-register@6.22.0
  | `-- home-or-tmp@2.0.0
  |   `-- os-tmpdir@1.0.2
  `-- slash@1.0.0

npm WARN optional Skipping failed optional dependency /chokidar/fsevents:
npm WARN notsup Not compatible with your operating system or architecture: fsevents@1.0.17
```

(Platform is node 7.4.0 on Windows 64bit.)

infojunkie:
Addressed in https://github.com/0xfe/vexflow/pull/525

PR #559, opened 2017-07-18
build(package.json): add webpack version, add babel-core dependency

gristow:
Fresh npm installs were pulling Webpack 3.0 — but 1.x is required for
the rest of the dependencies. This commit fixes that.

babel-loader requires babel-core to be installed locally or globally.
The missing dependency is added here.


gristow:
Fixes #552, Fixes #541, fixes #528

PR #577, opened 2017-08-26
EasyScore factory was still using selector instead of elementId

mgladdish:


gristow:
Good catch, thanks; merging.

PR #585, opened 2017-10-23
Fix colorization of slashed notes

balazskiss:


gristow:
Good catch -- style should apply to slashed noteheads as well.

To my taste, instead of nesting the `if(this.style)` twice in each branch of the `if(this.note_type ==='s')`, this would be better as:

``` javascript
if (this.style) {
  this.applyStyle(ctx);
}

if (this.note_type === 's') {
  const staveSpace = this.stave.getSpacingBetweenLines();
  drawSlashNoteHead(ctx, this.duration, head_x, y, stem_direction, staveSpace);
} else {
  Glyph.renderGlyph(ctx, head_x, y, glyph_font_scale, this.glyph_code);
}

if (this.style) {
  this.restoreStyle(ctx);
}
```

If you wouldn't mind pushing a commit along those lines I'll merge.

balazskiss:
Thanks for the review @gristow ! I have updated my code.

gristow:
Looks good, merging.

PR #595, opened 2017-12-13
Fixed ESLint error

matthras:
ESLint Error: Unexpected template string expression on Line 231. Looks like template literals were intended instead of quotation marks.

Picked this one up when using Vexflow within create-react-app


mscuthbert:
Thanks!  Merging.

PR #603, opened 2018-02-26
Text styling patch

### LINKED ISSUE #602: TextNote does not call applyStyle() before draw()
Should this be happening?  I can't seem to apply fillStyle or strokeStyle to text, which leaves my text in the same color as the last note drawn.  I can make a fix, but just making sure this isn't omitted for a reason.

mscuthbert:
replace ctx.save() and ctx.restore() with this.applyStyle(ctx) and this.restoreStyle(ctx)

Fixes #602


gristow:
Michael, the changes to `TextNote` look good to me.

I have mixed feelings about adding package-lock.json to .gitignore (see for instance [this discussion of the issue](https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5)). But am also not sure we really want to be tracking this in commits. @0xfe, do you feelings on this one way or the other?

0xfe:
Yeah, I'm not sure why I did that -- I think it should be checked into the repo, so builds are reproducible.

(Thanks for the PR, @mscuthbert!)

mscuthbert:
got it -- just removed from my .gitignore -- I'll let whoever makes a build after this check in their version, since I'm not sure that my .package-lock.json is absolutely correct, and I'd rather have a definitive version checked in than mine.

gristow:
Looks good to me -- merging!

Issue #376, opened 2016-06-29
Editorconfig


aaronmars:
I really like the idea of .editorconfig (http://editorconfig.org).  It allows a project to define some basics (like indentation, newlines, etc.).  This way, editor applications can conform to these settings per-project.

I am a perfect use case as I prefer 4-space indentation, and I'd like my editor to default to that.  VexFlow uses 2 spaces, so if it has an .editorconfig, I don't have to constantly change the settings in my editors/IDEs.

The best thing is this just helps catch eventual linter errors before they happen, and it helps maintain code style.  It's totally opt-in since it requires a compatible editor.

Is this an OK thing to add?

mscuthbert:
Please, please! I haven't done too many pushes to the Vexflow codebase, but when I have it's always needed an extra roundtrip (before Lint was added) because I used 4 spaces instead of 2.  :-)

0xfe:
Yes, lets try it out. (I've never used it before.)

peterankelein:
I realize this is an older discussion but since it was never done, hope you don't mind I took it on as a first contribution.

rvilarl:
@0xfe this one can be closed.

PR #623, opened 2018-06-18
Issue #376 - add .editorconfig

peterankelein:
Addressed #376 and added the .editorconfig. As you're extending airbnb as a base for eslint, I used their base editorconfig with an additional line so trailing whitespace isn't trimmed in .md files. The rest is standard.

```
root = true

[*]
indent_style = space
indent_size = 2
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true
end_of_line = lf
# editorconfig-tools is unable to ignore longs strings or urls
max_line_length = null

[*.md]
trim_trailing_whitespace = false

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+end_of_line = lf
+# editorconfig-tools is unable to ignore longs strings or urls
+max_line_length = null
+
+[*.md]
+trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,6 @@
 root = true
 
-[*.js]
+[*]
 indent_style = space
 indent_size = 2
 charset = utf-8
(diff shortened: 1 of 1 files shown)


0xfe:
Thanks for the change -- overall it looks good, however, I'm not a fan of "catch-all" wildcards because it could cause confusion in the future. Can you replace `*` with `*.js` (or something more specific?)

peterankelein:
I've made the change. Are contributors mostly working on the JS files or do they branch out into the environment setup or other file types (HTML, RUBY, JSON etc)? The reason I ask is that the wildcard is pretty standard in .editorconfig setups. You define your global settings such as the 2 space indenting, then drill down to the specifics for individual file types. Omitting the '*' and targeting *.js leaves any other file types unaffected unless I add an entry for each one. If no one's touching them, it won't matter and can always be added later.

Pete

0xfe:
Got it -- I guess I'm not familiar enough with `editorconfig` to have a strong opinion on this.

I think right now since this entire repo is JS, it's not a big deal, but I'll let you decide.

peterankelein:
Cool. I reverted it back to how I originally had it.

0xfe:
Merged. Thanks!

Issue #627, opened 2018-07-05
Delayed ornaments get shifted to the left instead of right


sschmidTU:
Doing setDelayed(true) on a VF.Ornament, the ornament gets shifted to the left, instead of halfway between the corresponding note and the next one.
Here, the red arrows point to where the delayed ornaments should be placed:
![delayed_ornaments_bug](https://user-images.githubusercontent.com/33069673/42347618-b585f70c-80a6-11e8-8b3d-d14a2fcc53a9.png)
(the first turn is not delayed, so it's placed correctly)

I set up an interactive code demo on jsfiddle that you can modify and see that setDelayed(true) will always cause this issue:
https://jsfiddle.net/sschmidTU/6es9n407/

~~The issue is probably in `if (this.delayed)` in ornament.js~~
Edit: this part seems fine,

There are two other bug sources with signatures and multiple bars, see comment below.

I'm trying to fix it and make a pull request, though i'm having runtime issues with my VexFlow build (TickContext).

sschmidTU:
I did some more testing and found that clef, key and time signature also do not shift the ornaments to the right, though they should.

Here's a jsfiddle demo with 4 example tests:
https://jsfiddle.net/sschmidTU/L2dcbkqy/

Test 1: VexFlow's working internal test, Works fine. (code slightly adapted)
Test 2: Test 1 with added signatures. Causes missing right-shift of ornaments on its own
Tests 3&4: Test 1 with multiple bars, a different source of missing right-shift.

Tests 3 and 4 are nearly identical, 4 is for trying to fix the issue, feel free to experiment with it. 

It looks about like this in jsfiddle's display:
![image](https://user-images.githubusercontent.com/33069673/42576242-39981e10-8522-11e8-828e-ac95b0d60add.png)

I'll continue to try and fix the code in VexFlow in a PR, but any help would be welcome.

bneumann:
Hi, so while using vexflow @sschmidTU and I found some bugs in the formatting for modifiers, ornaments and grace notes. The fixes are in our pending PR #632

bneumann:
If @sschmidTU has no objection this can be closed

sschmidTU:
yes, fixed by #632. only need a new release ;)

PR #632, opened 2018-07-19
Fixed gracenote offset after redraw

bneumann:
Hi,

we had some problems in opensheetmusicdisplay/opensheetmusicdisplay#293 that grace notes shifted to the right everytime we redraw stuff. We do this a lot because we can zoom in and out and also add debugging layers on top. Anyways I found that the GraceNoteGroup does this ```alignGraceNotesWithNote``` everytime it draws, thus adding the offset again and again which causes the whole group to shift.

My fix checks if the group is already aligned and skips that. It's rather quick and dirty (adding a prop after creation) but so is the mentioned function. Why is there formatting done during draw anyways?

### DIFF
--- a/src/gracenotegroup.js
+++ b/src/gracenotegroup.js
@@ -164,21 +164,21 @@ export class GraceNoteGroup extends Modifier {
         - extraPx.extraLeft
         + that.getSpacingFromNextModifier();
 
-        grace_notes.forEach(graceNote => {
-          const tick_context = graceNote.getTickContext();
-          const x_offset = tick_context.getX();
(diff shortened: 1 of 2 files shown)


sschmidTU:
So, this fixes grace note offset and ornaments offset (#627), with multiple bars/redraws.
Merge would be highly welcome, as [OSMD](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay) has some huge rendering bugs without this fix (see #627).
No breaking changes, all visual regression tests pass.

sschmidTU:
Thanks for the quick merge as well! :)

Issue #607, opened 2018-03-29
bad dots position (Y) for rest note


sug1no:
pls refer to the following test:

https://jsfiddle.net/47mxtjsf/

vexflow dot line for rest note is differenct from {finale, museScore}:
![image](https://user-images.githubusercontent.com/3293067/38068275-4572bf5c-334b-11e8-9021-847bfb8c7bc5.png)

b346a917125ab4b57bd5eaa28e3dd8de07818fbe fix #288 but ignore 
`Vex.Flow.durationToGlyph.duration_codes[*].type.r.dot_shiftY`   (https://github.com/0xfe/vexflow/blob/e107adf1b7c4512b4ec73d46b0e607332b5535b8/src/tables.js#L744) values(applied to Dot at https://github.com/0xfe/vexflow/blob/58aafbf34c6a63908e82692302ad2c2ea2a00b57/src/stavenote.js#L810)

0xfe:
Thanks -- I think this is a straightforward fix, but will look into it when I get a chance.

mscuthbert:
The Finale behavior may be a font issue; in my version of F25 the dot is perfectly centered, which is I believe the right behavior.

gristow:
I've tracked the problem down to the format method in `Dot.js`, where regardless of what `dot_shiftY` had been set to, it is overwritten: https://github.com/0xfe/vexflow/blob/d8763b9cb030c94208cd45a38b878d4efc6e40e1/src/dot.js#L75

@0xfe do you still want to take this one, or should I go after it? (It'll take me a bit to dig into what the code around this is doing...)

0xfe:
@gristow Thanks for tracking it down. I'm unfortunately too swamped to look at it right now, however will eventually get to it if no one else does first.

gristow:
No worries, @0xfe. It turned out to be a one character fix.

PR #633, opened 2018-07-26
Fix dotted note dot y-shifting each draw

sschmidTU:
Before this fix, dotted note dots get y-shifted a half line each render/draw:
image
We redraw a lot in OSMD compared to Vexflow tests, so we noticed [this issue](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/300) pretty quickly.

The bug was introduced by [fixing quarter dotted rest dot positioning](https://github.com/0xfe/vexflow/commit/298cd8813e0a545c31fb312bd9f7dfa31e634e17) ([#607](https://github.com/0xfe/vexflow/issues/607)).

Rest note and stave note shiftY seem to be handled differently, stave notes remember previous y-shift.
The dot_shiftY could also be refactored to operate the same for stave notes and rests (= vs +=), but this fix handles the difference.
All visual regression tests pass and i haven't noticed other issues, using this for a while in OSMD.
Dotted notes are perfect now.

By the way, most physical sheet music editions use a quarter plus eigth rest instead of a dotted quarter, easier to read and can align with other notes in the staff.
Beethoven Sonata Op. 106 "Hammerklavier" beginning:

image

### DIFF
--- a/src/dot.js
+++ b/src/dot.js
@@ -57,7 +57,7 @@ export class Dot extends Modifier {
         if (Math.abs(line % 1) === 0.5) {
           // note is on a space, so no dot shift
           half_shiftY = 0;
-        } else if (!note.isRest()) {
+        } else {
           // note is on a line, so shift dot to space above the line
           half_shiftY = 0.5;
           if (last_note != null &&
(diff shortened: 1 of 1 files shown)


0xfe:
Good catch. Thanks for the fix. Merging.

PR #634, opened 2018-07-27
fix NoteSubGroup x-shift each render

sschmidTU:
Before, NoteSubGroup x-offset was reapplied every render, similar to what was mentioned in #632:
```tick_context.setX(x + x_offset);```

This resulted in NoteSubGroups moving x-wise every draw():
image

I improved and reapplied the fix from #632. Now, GraceNoteGroups and NoteSubGroups use the same align function in modifier.js, whereas before they both had an align function that was duplicate code. I tried moving the alignment to a format method, because it should rather be there than in draw(), but when i tried it seemed like it was too early and didn't work.

In TickContext I added the possibility to represent x-positioning as base x plus offset, so we can simply overwrite the x-offset and not apply the x-offset of the previous draw()-call to the next one.

This could be a good method to use in general, though it'll still be possible to simply get and set x in TickContext.

This PR fixes both in-measure clef (as NoteSubGroup) and grace note rerender shift in [OSMD](github.com/opensheetmusicdisplay/opensheetmusicdisplay/).

All Qunit tests pass btw.

### DIFF
--- a/src/gracenotegroup.js
+++ b/src/gracenotegroup.js
@@ -153,31 +153,7 @@ export class GraceNoteGroup extends Modifier {
     }
 
     this.setRendered();
-    const that = this;
-    function alignGraceNotesWithNote(grace_notes, note) {
-      // Shift over the tick contexts of each note
-      // So that th aligned with the note
-      const tickContext = note.getTickContext();
-      const extraPx = tickContext.getExtraPx();
-      const x = tickContext.getX()
-        - extraPx.left
-        - extraPx.extraLeft
-        + that.getSpacingFromNextModifier();
-
-      grace_notes.forEach(graceNote => {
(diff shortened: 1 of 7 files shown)


0xfe:
Thanks for the change. It looks like you're getting more familiar with the codebase. I think it's a good idea to have a few additional tests to prevent a regression here (of the general reapplying of offsets.) Can you please add?

0xfe:
Also, please run the linter so the CI doesn't complain.

sschmidTU:
@0xfe Yes, rerender tests for offsets will be nice, i'll add them, and lint 'em all.

sschmidTU:
@0xfe I added 3 tests for multiple draws of grace notes, ornaments and notesubgroups.

0xfe:
Looks great! Merging.

Issue #350, opened 2016-06-11
SMuFL support + status of Gonville support


Silverwolf90:
(Might be worth looking at this proposal first: https://github.com/0xfe/vexflow/issues/349 -- Once those tasks are completed this issue becomes more relevant)

In the [old issue on SMuFL support](https://github.com/0xfe/vexflow/issues/181), we agreed on not using the metadata but a different set of hardcoded metrics, and I think that was a mistake. Being able to rely on SMuFL metadata entirely for layout will simplify the codebase. Now that I've used SMuFL pretty extensively, I'm in favor of going **all-in** on SMuFL, rather than ending up in some half-way state because we're trying to support Gonville too. It would still be a big net positive if we could use the SMuFL metadata, even at the cost of supporting Gonville. 

I do think we need to seriously consider what is gained from supporting Gonville. It's non-standard, has a minimal glyph set, completely lacks metadata, and adds complexity to the codebase to support it alongside SMuFL. Basically, now that SMuFL exists, I do not think VexFLow should aim to support non-standard music fonts.

If we did want to continue supporting Gonville alongside SMuFL we have to store data about Gonville that allows us to approprately transform and consume Gonville glyphs as if they were SMuFL compliant
- We need the following data on each Gonville glyph:
  - Horizontal origin alignment
  - Vertical origin alignment
  - Orientation for multi-line segments
    - All [multi-line segments](http://www.smufl.org/version/latest/range/multiSegmentLines/) in SMuFL are oriented horizontally but in Gonville _some_ are oriented vertically
  - SMuFL glyph name (eg: `accidentalSharp`) for the respective VexFlow Gonville glyph identifiers (eg: `vb9`)
- Some Gonville glyphs coincidentally match up with SMuFL (like noteheads), so it might only be necessary to maintain a mapping of deviations from SMuFL
- Provide SMuFL compliant metadata for Gonville glyphs

mscuthbert:
I agree completely.  It would be far easier to do this change once, support Bravura, and make a SMuFL compliant version of Gonville later (if someone loved the look of that font) than to let the code support both SMuFL and non-SMuFL fonts.

My sense though is that SMuFL support will have enough people who want to have it in Javascript that it's worth creating a SMuFL API independent of VexFlow (obviously one that is incorporated into VexFlow also and under the same license) and then have VexFlow use that API for all its font calls.  It'd make testing easier I'd think.

Silverwolf90:
[...]

Issue #181, opened 2014-03-19
VexFlow should be able to "plug-in" alternate fonts.


0xfe:
A good start would be the lilypond font or [SMUFL](http://smufl.org).

This means:

1) Factoring out direct references to Gonville font indexes (e.g., "v51", "v7b", etc.) and putting them in a separate table (e.g., "closed_head": "v51", "16th_rest": "v8a", etc.).

2) The table must include all parameters related to the glyphs, e.g., scale, grace_note_scale, x_shift, y_shift, etc, bounding_box, etc.

3) The Glyph renderer should be able to render different JSON font formats.

jmahmud:
Hi all

I realise this thread is quite old now (almost a year to the date!) but I was wondering whether Bravura or general SMuFL has been incorporated within Vexflow main branch?  I have a fork of the source at the moment but would like to incorporate SMuFL as an option for rendering glyphs.  Could you let me know@

* if available in VF, how can it be selected OR
* how I might incorporate the changes made by Silverwolf90 into my branch.

Any information would be greatly appreciated!

Cheers
Josh

SalahAdDin:
What's about this?

i want to do a music sheet editor and i want to do a tool bar with the common music symbols, i don't know how implements the VexFlow glyph fonts for do it. :/

SalahAdDin:
The SMuFL font used here is Bravura, right?

Silverwolf90:
@SalahAdDin yes this is Bravura

SMuFL is not in the main branch of VexFlow. And the work done so far is definitely not ready to move into the main branch. It goes without saying that trying to support SMuFL is particularly intrusive, as changes are required throughout the entire codebase. And it's perhaps one that it's not ready for (went through a lot of pain to get it to the point that it's at).

As far as how fonts are stored, you can see the data here: https://github.com/0xfe/vexflow/blob/master/src/fonts/vexflow_font.js

It just stores a serialized version of the glyph outlines (with some other data as well). If you want to add new glyphs, you could probably use something like [opentype.js](https://nodebox.github.io/opentype.js/) to parse a font file and serialize the outlines of the glyphs you want into a similar structure.

SalahAdDin:
I didn't know that this is so hard.

jmahmud:
Hi Cyril

Is this something which you're willing to allow others to contribute to or
are you wanting to continue work on it before making it more public?

Thanks
Josh



[...]

PR #635, opened 2018-07-29
SMuFL mappings in tables

mscuthbert:
Hoping that this will be a useful contribution for an eventual push to allow for the complete SMuFL font plug in, letting a smooth transition from Gonville to SMuFL font plugin take place.  Gives a mapping for almost all glyphs from SMuFL names to Vexflow codepoints so that these can start being replaced in the hardcoded code and then eventually using the existing .json files, other SMuFL fonts could be used for an extended character set.

This pull request only contains the mappings for SMuFL to Gonville.  Next possible steps include allowing any glyph to be accessed in places such as TextNote and lyric annotations using the SMuFL name.  Then replacing references to codepoints to SMuFL names, and then creating plugable fonts.

A step towards #349 and esp. #350 ("SMuFL glyph name (eg: `accidentalSharp`) for the respective VexFlow Gonville glyph identifiers (eg: `vb9`)")

7 glyphs in Gonville could not be mapped -- three seem to be duplicate notehead glyphs (e.g., noteheads that are the same for quarter as for eighth and smaller notes), two are accidentals that cannot be found in the extensive SMuFL list (slashed double flat, v90, and an alternative 3/4 sharp, v7a), an unused marcato alternative? v5a, and an alternate positioning for a down arrow (v11).

### DIFF
--- a/src/tables.js
+++ b/src/tables.js
@@ -994,6 +994,192 @@ Flow.durationToGlyph.duration_codes = {
   },
 };
 
+// For future collaboration with the SMuFL Standard Music Font Layout
+
+Flow.smufl = {};
+
+// add references between smufl glyph names and code points.
+Flow.smufl.to_code_points = {
+  // staff brackets and deviders (e000-e00f)
+  bracketTop: 'v1b',
+  bracketBottom: 'v10',
+
+  // barlines (e030-e03f)
+  barlineTick: 'v6f',
+
+  // repeats (e040-e04f)
+  segno: 'v8c',
+  coda: 'v4d',
+
+  // clefs (e050-e07f)
+  gClef: 'v83',
+  cClef: 'vad',
+  fClef: 'v79',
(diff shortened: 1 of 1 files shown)


mscuthbert:
Pinging @Silverwolf90 who has done so much so far on this.

mscuthbert:
AH! This was already done by @Silverwolf90 in https://github.com/Silverwolf90/vexflow/blob/refactor/extract-hard-coded-glyphs/src/fonts/gonville_metrics.js  -- not sure if it's better to extract from there or to take this in lieu of the difficulty in merging with #181 from 2014.

Silverwolf90:
I'd keep this one! Looks like my old branch included the extraction of hard-coded values, which is at least partially obsolete with some of the work around glyphs from a couple years ago.

0xfe:
This is really great -- I've been wanting to switch everything to SMuFL for a long time.

Also, related, `tables.js` is getting huge, maybe it should be a directory `tables/` with files broken up more neatly.

PR #645, opened 2018-08-28
Add square rectangle note heads

PieterHartzer:
Added two new note heads, square and rectangle.

image

In order to get the stem aligned with the right I had to make the heads that specific width until #643 is fixed.  Certain note lengths might still align incorrectly then.


0xfe:
Thanks for the change. `vexflow_font.js` is typically auto-generated from other fonts in the same directory. What you have there is good, but can you also add something like `square_heads.js` to the directory with your new glyphs, so they don't get clobbered?

PieterHartzer:
When these are generated, how does it determine the next code (`vd2`)?  Shall I use a completely different code (starting at `c01`) when moving them into a new file?
Also, I will make it more generic so that it's not only square but all custom glyphs instead.

PieterHartzer:
Custom glyphs are now in a [separate file](https://github.com/0xfe/vexflow/pull/645/files#diff-e0749974c4676140faf8f62fd37a63e6) and merged into the font in [index.js](https://github.com/0xfe/vexflow/pull/645/files#diff-1fdf421c05c1140f6d71444ea2b27638).
I kept the codes following on from the existing font instead of making them custom.  Please let me know if I need to change this.

0xfe:
Thanks Pieter -- sorry I wasn't clear enough. It's okay to have the glyphs in `vexflow_font.js`, just *also* have them in a separate file so the sources can be tracked.

To answer your question, the font is not auto generated -- I usually do it by hand (I need to figure out a process for it.)

So, no need for `import CustomGlyph`. :-)

0xfe:
Thanks Pieter, but still not there, sorry :-)

Here are (hopefully) clearer instructions:

* Add the new glyphs to `vexflow_font.js`
* Keep them in you custom glyphs file
* Remove the unnecessary imports (now that the glyphs are already imported via `vexflow_fonts.js`

PieterHartzer:
@0xfe I forgot to push.  Sorry.  It should now be there.  I also merged master in and resolved conflicts.

0xfe:
Great!

PR #646, opened 2018-08-29
Stem not linking to note head at correct place

PieterHartzer:
Added code to allow the stem to connect to the custom note head at the expected place.

`stem_up_y_offset` and `stem_down_y_offset` are both needed because the stem meets at the bottom of the note head for some, and bottom/top for others depending whether it's up or down.

Example from the tests:
image

### DIFF
--- a/src/stem.js
+++ b/src/stem.js
@@ -54,6 +54,10 @@ export class Stem extends Element {
     this.isStemlet = options.isStemlet || false;
     this.stemletHeight = options.stemletHeight || 0;
 
+    // Changing where the stem meets the head
+    this.stem_up_y_offset = options.stem_up_y_offset || 0;
+    this.stem_down_y_offset = options.stem_down_y_offset || 0;
+
     // Use to adjust the rendered height without affecting
     // the results of `.getExtents()`
     this.renderHeightAdjustment = 0;
@@ -84,8 +88,9 @@ export class Stem extends Element {
 
   // Gets the entire height for the stem
   getHeight() {
+    const y_offset = (this.stem_direction === Stem.UP) ? this.stem_up_y_offset : this.stem_down_y_offset; // eslint-disable-line max-len
     return ((this.y_bottom - this.y_top) * this.stem_direction) +
-           ((Stem.HEIGHT + this.stem_extension) * this.stem_direction);
+           ((Stem.HEIGHT - y_offset + this.stem_extension) * this.stem_direction);
   }
   getBoundingBox() {
     throw new Vex.RERR('NotImplemented', 'getBoundingBox() not implemented.');
@@ -128,11 +133,11 @@ export class Stem extends Element {
     if (stem_direction === Stem.DOWN) {
       // Down stems are rendered to the left of the head.
       stem_x = this.x_begin;
-      stem_y = this.y_top;
+      stem_y = this.y_top + this.stem_down_y_offset;
     } else {
       // Up stems are rendered to the right of the head.
       stem_x = this.x_end;
-      stem_y = this.y_bottom;
+      stem_y = this.y_bottom - this.stem_up_y_offset;
     }
 
     const stemHeight = this.getHeight();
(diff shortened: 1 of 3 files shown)


src/tables.js
```diff
@@ -472,6 +472,17 @@ Flow.parseNoteData = noteData => {
...
+
+    // If we have keys, try and check if we've got a custom glyph
+    if (noteData.keys !== undefined) {
+      const regexp = /^([a-gx]?)\/(\d?)\/?([a-z][0-3])$/;
+      const result = regexp.exec(noteData.keys[0]);
+
+      // We have a custom glyph specified after the note eg. /X2
+      if (result && result.length === 4) {
+        type = result[3]; // Set the type to the custom note head
+      }
```
0xfe: Not sure I understand what's being attempted here. (Also, the regex stuff up there is probably going to be fragile, is there a better way to do this?)

src/tables.js
```diff
@@ -185,16 +185,16 @@ Flow.keyProperties.note_glyph = {
...
 
   /* Triangle */
-  'T0': { code: 'v49', shift_right: -2 },
-  'T1': { code: 'v93', shift_right: 0.5 },
-  'T2': { code: 'v40', shift_right: 0.5 },
-  'T3': { code: 'v7d', shift_right: 0.5 },
+  'T0': { code: 'v49', shift_right: -2, stem_up_y_offset: -4, stem_down_y_offset: 4 },
+  'T1': { code: 'v93', shift_right: 0.5, stem_up_y_offset: -4, stem_down_y_offset: 4 },
+  'T2': { code: 'v40', shift_right: 0.5, stem_up_y_offset: -4, stem_down_y_offset: 4 },
+  'T3': { code: 'v7d', shift_right: 0.5, stem_up_y_offset: -4, stem_down_y_offset: 4 },
```
0xfe: Is there any case when the up and down offsets will be different? Why not just have one field here `y_offset` and apply it the same way to the up and down cases?

src/stem.js
```diff
@@ -128,11 +133,11 @@ export class Stem extends Element {
     if (stem_direction === Stem.DOWN) {
       // Down stems are rendered to the left of the head.
       stem_x = this.x_begin;
-      stem_y = this.y_top;
+      stem_y = this.y_top + this.stem_down_y_offset;
```
0xfe: with a single offset, this line would be the same

src/stem.js
```diff
@@ -84,8 +88,9 @@ export class Stem extends Element {
 
   // Gets the entire height for the stem
   getHeight() {
+    const y_offset = (this.stem_direction === Stem.UP) ? this.stem_up_y_offset : this.stem_down_y_offset; // eslint-disable-line max-len
```
0xfe: this could be `const y_offset = this.y_offset * (this.stem_direction === Stem.UP ? 1 : -1)`

src/tables.js
PieterHartzer: The type for the note defaults to `n`.  When, for example, `keys: ['a/5/x3']` is specified (circle-x note head) then I set the type to `x3`, the custom note head identifier.  The reason for this is because the width of the head is calculated wrong for custom note heads.
An alternative would be to implement a new `Flow.durationToGlyph.duration_codes` for each custom head, but that seems a lot of duplication.

The regex can be replaced by `str.split('/');` but I am not sure whether about performance impact and believed the regex to be faster.

src/tables.js
PieterHartzer: Yes there is a case where we need two, currently only for triangle heads.
For triangles the up stem needs to be longer than the down stem.

image
image

An option could be to move the actual glyphs so that the starting point is at the bottom of the glyph. We will then have
image
instead of the current
image

src/stem.js
PieterHartzer: Agreed, when we have a single offset it could be

src/tables.js
0xfe: Thanks for the explanation and the example, that makes sense.

src/tables.js
0xfe: Yes please, `str.split('/')` is much clearer about intent and far less fragile.

0xfe:
Looks good. Merging. Thanks.

PR #648, opened 2018-09-04
specify -highlight-color option to 'compare'.

sug1no:
this PR specify the  -highlight-color option to 'compare'.

diff image is changed as follows:

image

anyway, I think last release binaries does not contains some recent changes. (8e86d583d65322d1b461efdd0b9805de46925b98)


0xfe:
Looks good. Thanks.

PR #653, opened 2018-09-14
chore: fix 20 npm vulnerabilities by updating npm (6.4.1) etc.

sschmidTU:
This mainly updates npm to 6.4.1. Without this, npm audit shows 20 moderate vulnerabilities when having vexflow as an npm dependency in another project, and 35 vulnerabilities, one high, when doing npm install in vexflow. There shouldn't really be a security concern, but one shouldn't have these warnings then, either.

After this, it's 0 vulnerabilities for Vexflow as a npm dependency in another project, and 4 low vulnerabilities, one moderate for Vexflow devDependencies. All tests still pass, seems like no breaking changes.

`npm audit` now runs automatically at the end of npm install with npm 6.x, and it doesn't look nice to have these warnings when using Vexflow as a dependency.
image

The npm 6.4.1 update is enough to eliminate these 20 vulnerabilities, so that's the most important change.

The other 5 vulnerabilities from Jscs and grunt-release are only devDependencies of Vexflow, and should hardly be a concern at all.

Npm Audits before/after:
[npm-audits-before-after-dependency-updates.zip](https://github.com/0xfe/vexflow/files/2383705/npm-audits-before-after-dependency-updates.zip)
Run `npm audit` (npm@6) to check these yourself.
https://docs.npmjs.com/cli/audit
https://docs.npmjs.com/getting-started/running-a-security-audit

Jscs is unmaintained and merged with Eslint, so to eliminate the vulnerability one would have to fork the npm package or switch to Eslint.

For grunt-release, I could ask them to update their dependencies.

If this makes sense and is merged, it would be nice to have a new Vexflow release at some point, but no hurries!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -987,37 +987,6 @@
       "integrity": "sha1-RqoXUftqL5PuXmibsQh9SxTGwgU=",
       "dev": true
     },
-    "bl": {
-      "version": "1.0.3",
-      "resolved": "https://registry.npmjs.org/bl/-/bl-1.0.3.tgz",
-      "integrity": "sha1-/FQhoo/UImA2w7OJGmaiW8ZNIm4=",
-      "dev": true,
-      "requires": {
-        "readable-stream": "~2.0.5"
-      },
-      "dependencies": {
-        "process-nextick-args": {
-          "version": "1.0.7",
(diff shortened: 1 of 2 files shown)


0xfe:
Thank you for doing this! Merging.

PR #657, opened 2018-10-02
changed to save and restore svgcontext.lineWidth.

### LINKED ISSUE #656: svgcontext: lineWidth is not restored
test case: http://jsfiddle.net/bLhogjv0/

image

red line width:

* canvas context: 1: OK.
* svg context: 10: NG.
  - [ ] save/restore lineWidth.

sug1no:
* this PR {save/restore} svgcontext.lineWidth to fix #656.

* Previously, stroke results of {svg and canvas} context were different, but now they are similar.
* many tests are affected by this PR as follows (is an excerpt): 

image

### DIFF
--- a/src/svgcontext.js
+++ b/src/svgcontext.js
@@ -646,6 +646,7 @@ export class SVGContext {
         width: this.shadow_attributes.width,
         color: this.shadow_attributes.color,
       },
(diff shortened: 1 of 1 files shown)


0xfe:
Nice catch! Merging. Thanks!

PR #659, opened 2018-10-04
fix(articulation): Grace notes can now have articulation

sschmidTU:
GraceNotes weren't able to have articulation before, because articulation.js only checked whether the note's category is 'stavenotes', not 'gracenotes'. GraceNote extends StaveNote.

Of course it would be ideal to check whether something is a superclass of StaveNote, but Javascript isn't good enough with type detection and inheritance, right?

Now we can have some funky grace note articulations:
image

Tests pass.

### DIFF
--- a/src/articulation.js
+++ b/src/articulation.js
@@ -55,8 +55,9 @@ const getTopY = (note, textLine) => {
   const stave = note.getStave();
   const stemDirection = note.getStemDirection();
   const { topY: stemTipY, baseY: stemBaseY } = note.getStemExtents();
+  const noteCategory = note.getCategory();
 
-  if (note.getCategory() === 'stavenotes') {
+  if (noteCategory === 'stavenotes' || noteCategory === 'gracenotes') {
     if (note.hasStem()) {
       if (stemDirection === Stem.UP) {
         return stemTipY;
@@ -66,7 +67,7 @@ const getTopY = (note, textLine) => {
     } else {
       return Math.min(...note.getYs());
     }
-  } else if (note.getCategory() === 'tabnotes') {
+  } else if (noteCategory === 'tabnotes') {
     if (note.hasStem()) {
       if (stemDirection === Stem.UP) {
         return stemTipY;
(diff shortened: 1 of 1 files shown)


0xfe:
Great change! (Will review this and other PRs soon, sorry for the delay.)

mscuthbert:
Here's how music21j checks for classes:

```javascript
export class ProtoM21Object {
    constructor() {
        this._storedClasses = undefined;
    }

    get classes() {
        if (this._storedClasses !== undefined) {
            return this._storedClasses;
        }
        const classList = [];
        let thisConstructor = this.constructor;
        let maxLinks = 20;
        while (
            thisConstructor !== null
            && thisConstructor !== undefined
            && maxLinks
        ) {
            maxLinks -= 1;
            if (thisConstructor.name === '') {
                break;
            }
            classList.push(thisConstructor.name);
            thisConstructor = Object.getPrototypeOf(thisConstructor);
        }
        classList.push('object');
        this._storedClasses = classList;
        return classList;
    }
}
```

and then you can do things like:
```
if (n.classes.includes('StaveNote')) { ... }
```

sschmidTU:
@0xfe no problem, thanks!
@mscuthbert thanks for the tipp! I've read about Javascript type detection and prototypes and so on, but from what i remember it's messy and not completely reliable.
I'm also not sure if classes.includes('StaveNote') would be true for GraceNote, because it only extends StaveNote and has its own class name. Javascript definitely has issues with inheritance.

mscuthbert:
@sschmidTU -- I think it'd work in this case; when you extend a class, then the class's prototype becomes the previous class's constructor.  It won't work if you have no idea how a particular object was created, but as long as you're working within an ES6 class framework without anonymous classes, it has always worked for us, even with pseudo-multiple-inheritance.

sschmidTU:
@mscuthbert I tried to implement it after your example, but it didn't recognize the 'class'.

```javascript
var constructor = note.constructor;
if (constructor.name === 'StaveNote' || Object.getPrototypeOf(constructor) === 'StaveNote') {
```
the if-clause resolves to false for a GraceNote note.

on the other hand, what's in the PR works:
```javascript
const noteCategory = note.getCategory();
  if (noteCategory === 'stavenotes' || noteCategory === 'gracenotes') {
```
this resolves to true.

I'm calling this on a grace note with an articulation.

In any case, i don't think we'll add another StaveNote extending class so soon, so i wouldn't worry too much about making the code more generalized for this PR.

Still thanks for the interesting class check algorithm! It does seem JS prototypes are pretty stable as long as you have a class search function like that.

sschmidTU:
I refactored the stave note class check into a method, so at least we can easily change it.

0xfe:
Hey folks -- PR looks good.

Michael, thanks for the class check snippet. It is however a bit to hacky to use here. I generally try to avoid prototype gymnastics because they can be fragile and fail in unexpected ways (making it tricky to debug.)

Also, `getCategory` was kinda intended for this kind of explicit type checking, and we can make it support hierarchies (e.g., with `inCategory(..) bool`) but no need to do that unless we really need it.

PR #676, opened 2018-11-13
#670: Fraction.copy: support 'number' type argument to fix Fraction.equls(number).

### LINKED ISSUE #670: rhythm whole note & double whole note display incorrect
Bellow condition is always true for whole note and false for double whole note.
[`if (Flow.durationToFraction(duration).equals(0.5)) {`](https://github.com/0xfe/vexflow/blob/master/src/notehead.js#L55)

change this line to `if (Flow.durationToFraction(duration).value() == 0.5) {` or make Fraction.equals support numeric parameter fix it.

```javascript
  equals(compare) {
  if(typeof compare === 'number') return this.value() === compare;
    const a = Fraction.__compareA.copy(compare).simplify();
    const b = Fraction.__compareB.copy(this).simplify();

    return (a.numerator === b.numerator) && (a.denominator === b.denominator);
  }
```

breve

sug1no:
This PR fixes #670.

@dolphinxx san: Thank you for your detailed comment!

- Since `Fraction.subtract` supports numeric argument, changed to support numeric argument with `Fraction.copy` (not `Fraction.equals`).

- npm test (with master):
```shell
 (670-fix_durationToFraction=)
$ git checkout master && npm start && git checkout @{-1} && npm test
...
Results stored in ./build/images/diff/results.txt
All images with a difference over threshold, 0.01, are
available in ./build/images/diff, sorted by perceptual hash.

You have 4 fail(s):
StaveNote.StaveNote_Draw___Slash 0.50734
Rhythm.Rhythm_Draw___slash_notes 0.426776
StaveNote.StaveNote_BoundingBoxes___Treble 0.201863
Rhythm.Rhythm_Draw___beamed_slash_notes__some_rests 0.191847
```

- {1/2, 1} slash note head now correctly rendered.
- StaveNote.StaveNote_BoundingBoxes: half rest boundingBox is better than before.
  https://github.com/0xfe/vexflow/blob/58d6e1a420ab0cdb545e461432f15b15eff107c0/src/stavenote.js#L529

image

<http://jsfiddle.net/eo47gvmc/4/> :
image

### DIFF
--- a/src/fraction.js
+++ b/src/fraction.js
@@ -220,6 +220,9 @@ export class Fraction {
 
   // Copies value of another Fraction into itself.
   copy(copy) {
+    if (typeof copy === 'number') {
+      return this.set(copy || 0, 1);
+    }
     return this.set(copy.numerator, copy.denominator);
   }
 


mscuthbert:
Good catch of subtle bug @sug1no -san and @dolphinxx !

mscuthbert:
Can we add a test for Fraction.subtract(2) etc. to make sure bug never reoccurs?  Thanks!

PR #683, opened 2018-11-27
export StaveTempo

sschmidTU:
Without this we can't use `stave.addModifier(new StaveTempo(...))`, which is more powerful than `stave.setTempo()`.

PR #684, opened 2018-11-27
fix double quote lint warning

sschmidTU:
fix warning introduced by #683


sug1no:
Thank you for the quick response!

mscuthbert:
Thanks commenter, submitter, and merger!

PR #687, opened 2019-01-02
Add .gitattributes file / update dependencies to latest stable versions

mvolganli:
I added a `.gitattributes` file to prevent git from converting LF linebreaks to CRLF when cloning the project on Windows (because of the option `core.autocrlf=true`) to avoid the following eslint error when running `grunt`:

> Expected linebreaks to be 'LF' but found 'CRLF'

> https://eslint.org/docs/rules/linebreak-style#using-this-rule-with-version-control-systems

The second commit is for updating dependencies to latest stable versions. I had to modify two `.eslintrc` files to disable new rules that were applied by default and the `Gruntfile` because the configuration for `grunt-contrib-uglify` changed. I also added the webpack option `libraryExport` to expose the Vex global object in the generated vexflow-debug.js file. I removed the babel plugin `add-module-exports` which was previously used for this purpose.


0xfe:
Thanks, merging.

PR #688, opened 2019-01-05
Fix repeat barline alignment

mvolganli:
This change ensure that REPEAT_BEGIN barlines are correctly aligned in a system with multiple staves.

Before:

before

After:

after

### DIFF
--- a/src/stave.js
+++ b/src/stave.js
@@ -76,9 +76,11 @@ export class Stave extends Element {
     if (!this.formatted) this.format();
 
     this.start_x = x;
-    const begBarline = this.modifiers[0];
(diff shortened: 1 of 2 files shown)


src/stave.js
```diff
@@ -76,6 +76,10 @@ export class Stave extends Element {
     if (!this.formatted) this.format();
 
     this.start_x = x;
+    const begBarline = this.modifiers[0];
```
0xfe: I don't think `Stave` is guaranteed to have modifiers, and this will crash in that instance.

Maybe wrap this block in `if (this.modifiers.length > 0) { ... }`.

PR #697, opened 2019-03-18
Simplified visual regression tests

panarch:
Hello Mohit,
Hope you are doing well, long time no see!

Recently while I introduced VexFlow to my friend, I found the current test can be broken in some cases.

SlimerJS does not provide support for the latest Firefox browser, so if someone already has installed Firefox on the machine, then `generate_svg_images.js` may fail due to the version conflict.

And nowadays there are many better alternative ways to generate images rather than using SlimerJS, so I applied.

I replaced SlimerJS & rsvg-convert to jsdom & node-canvas.
Now jsdom + node-canvas generates png images directly using `tools/generate_png_images.js`

I changed test script dependencies, so current travis-ci test will fail until you release new one.

Thanks,
Taehoon


0xfe:
Hi Taehoon! Good to see you back!

Thanks for this -- I was itching to fix this for a while. Looks great. Merging.

(I'll cut a new release shortly.)

PR #701, opened 2019-04-25
Fix beam.js partial beam rendering bug,

panarch:
Fix to render properly on the case, [8th - 32nd - 16th] beam(32nd partial beam is rendered incorrectly).
Add a partial beam test case.
——
I added a new test case for verifying this, and all existing test cases are passed.

Previous:
IMG_0151

Current:
IMG_0152

Thanks!

### DIFF
--- a/src/beam.js
+++ b/src/beam.js
@@ -687,7 +687,11 @@ export class Beam extends Element {
 
       // Check to see if the next note in the group will get a beam at this
       //  level. This will help to inform the partial beam logic below.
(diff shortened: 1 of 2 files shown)


0xfe:
Nice job. You dug into one of the hairiest parts of the code. :-)

panarch:
Though it was little bit hard to understand at first, all the codes were inside a single `beam.js` file so it was good to read. :)

mscuthbert:
Code looked good to me.

PR #703, opened 2019-05-03
Fix dotted-beamed-displaced chord note rendering bug,

panarch:
Dots are rendered incorrectly on the specific condition below:
=> Displaced (ex. c/4 + d/4 in treble clef) chord note has dot and "beam".

Update StaveNote setBeam function to re-calculate extra px to ignore flag and consider beam properly

I simply updated Dot.Basic test case to also consider beamed chord note case.

Before
IMG_0160

Current
IMG_0161

Thanks!

### DIFF
--- a/src/stavenote.js
+++ b/src/stavenote.js
@@ -362,6 +362,12 @@ export class StaveNote extends StemmableNote {
     this.calcExtraPx();
   }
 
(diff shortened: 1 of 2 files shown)


0xfe:
Looks good. Thanks.

PR #707, opened 2019-06-18
Fix beam direction in partial beam cases

24seconds:
Hi, I am GeunYeong Cheon and I found that beam direction is slightly different as I thought in specific partial beam case. I added one test case in `partial_beam` test.

To get a correct beam direction, looking up the upper level beam lines'
direction is needed. Therefore, I made lookupBeamDirection to check
direction.

I tested using `npm test` with `release version` and `build version` - build version includes my modification -.  Also after running `npm test` and it shows that all test cases are passed except the one that I added.

Before
Beam Partial_Beam_Blessed

Current
Beam Partial_Beam_Current

Difference
Beam Partial_Beam

Thanks!

### DIFF
--- a/src/beam.js
+++ b/src/beam.js
@@ -35,6 +35,10 @@ const getStemSlope = (firstNote, lastNote) => {
   return (lastStemTipY - firstStemTipY) / (lastStemX - firstStemX);
 };
 
+const BEAM_LEFT = 'L';
+const BEAM_RIGHT = 'R';
+const BEAM_BOTH = 'B';
+
 export class Beam extends Element {
   // Gets the default beam groups for a provided time signature.
   // Attempts to guess if the time signature is not found in table.
@@ -652,10 +656,10 @@ export class Beam extends Element {
     }
   }
 
-  // return upper level beam direction. L for Left, R for Right, B for Both
+  // return upper level beam direction.
   lookupBeamDirection(duration, prev_tick, tick, next_tick) {
     if (duration === '4') {
-      return 'L';
+      return BEAM_LEFT;
     }
 
     const lookup_duration =  `${Flow.durationToNumber(duration) / 2}`;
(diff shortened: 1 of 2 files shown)


0xfe:
Thanks for the PR, @24seconds!

@mscuthbert @gristow thoughts on those screenshots?

gristow:
Yes -- a definite improvement!

Unrelated to this PR, and I know I say this every time a beaming question comes up --  but there are some larger underlying issues in the beaming algorithm. I am unfortunately buried with other work these days, but if anyone else wants to think about them, they are:

  - Notes that overflow the current beam-group's fraction are pushed so that they begin the next beam group, even if that causes the next beam group to straddle a required beam break-point in the measure.
  - Partial beams (or "fractional beams" as Elaine Gould calls them in Behind Bars) should:
     a. Point right if they are at the beginning of the next-higher subdivision in the measure or the first note in a beam, regardless of where it falls metrically
     b. Point left if they are anywhere else in relation to the next-higher subdivision.
    (our current routine has all partial beams that are not at the beginning of a beam group, or after a manually defined secondary beam break, pointing left. Note also that while Elaine Gould's images in Behind Bars for this are correct, p. 158, her description of the algorithm is wrong and ignores issues of smaller subdivision fractional beams)
  - Beam groups that begin other than a metric beam break point (for instance, a beam group beginning on the second sixteenth note in 6/8 following a rest) would have incorrect partial directions even if we fix the prior issue because the beaming algorithm is ignorant of where the first note is in relation to the beat/metric beam break point.

0xfe:
Thanks @gristow! Very helpful. Reviewing.

src/beam.js
```diff
@@ -652,8 +652,31 @@ export class Beam extends Element {
...
+      return 'L';
+    }
+
+    const lookup_duration =  `${Flow.durationToNumber(duration) / 2}`;
+    const prev_note_gets_beam = prev_tick < Flow.durationToTicks(lookup_duration);
+    const next_note_gets_beam = next_tick < Flow.durationToTicks(lookup_duration);
+    const note_gets_beam = tick < Flow.durationToTicks(lookup_duration);
+
+    if (prev_note_gets_beam && next_note_gets_beam && note_gets_beam) {
+      return 'B';
```
0xfe: Can we use constants here instead of strings?

24seconds: Thanks for reviewing! That means is it okay to just declare constants in `lookupBeamDirection` function?
image

0xfe: Make them `const` scoped to this file (at the top). So `const BEAM_LEFT = 'L'`. And use them in the caller.

24seconds: Yap

24seconds: At first I declared consts in `lookupBeamDirection` scope but I make `const` scoped to beam file now

0xfe:
Woops. This fell off my radar. Looks good, merging. Thanks! :-)

PR #719, opened 2019-10-28
Suggested addition for a dynamic code base

spianoDev:
I have made an additional ReadMe for anyone who wants to use the code on a note by note basis.

### DIFF
--- a/DynamicVexFlow.md
+++ b/DynamicVexFlow.md
@@ -1,0 +1,68 @@
+# Quick Start Guide for setting up Flexible Uses of VexFlow
+
+## Install dependencies:
+```
+$ npm install vexflow
+```
+
+## HTML script requirement:
+Be sure to include this script in the html or layout.hbs file or the VexFlow render will not work.
+```html
+<script src="https://unpkg.com/vexflow/releases/vexflow-debug.js"></script>
+```
+## Begin by using EasyScore
+In HTML File: This places where the staff will go
+```html
+<div class="music-render" id="new-song" >
+</div>
+```
+In CSS File: This sets the staff size that will render on the page
+```
+.music-render {
+       margin: 10px auto;
+       padding: 10px;
+       background-color: rgba(255, 255, 255, 0.85);
+       width: 600px;
+       height: 150px;
+   }
+```
+In JavaScript File: This will establish the requisite variables needed to create a new staff
+```javascript
+let vf = new Vex.Flow.Factory({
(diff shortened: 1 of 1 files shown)


spianoDev:
> Looks like a nice quickstart guide!
> 
> However, a .md file in the main directory is probably not the best way to go place it There aren't many files in the main directory and for .mds, the Readme is the only one.
> And if we had a .md file for every Wiki page, the main directory would be flooded.
> https://github.com/0xfe/vexflow
> 
> Why not write a Wiki entry instead?
> https://github.com/0xfe/vexflow/wiki
> Not sure if you have permissions though.

Thank you for suggestion... I will look into that this weekend!

0xfe:
Thanks for reviewing @sschmidTU.

@spianoDev, the Wiki is publicly editable, feel free to add a new page. :-)

mix3d:
I don't understand why you `npm install vexflow` but then just use the unpkg import in the html. Seems redundant.

spianoDev:
> I don't understand why you `npm install vexflow` but then just use the unpkg import in the html. Seems redundant.

When I tried to run the program without the html script, I was getting an error message referring to the VexFlow Factory. Adding the script cleared that error and allowed the project to render the score.

mix3d:
>When I tried to run the program without the html script, I was getting an error message referring to the VexFlow Factory. Adding the script cleared that error and allowed the project to render the score.

Right, but if you `npm uninstall vexflow` you'll have the same result. The `npm install` isn't doing anything in your example. 

Really, vexflow needs an ESM compatible build, to use instead of one that pollutes the global namespace; many linters and bundlers do not like this.

I've found a workaround by importing the code directly (see below), but this is far from ideal. The NPM module is not using the `.npmignore` file to simplify what's in the module: https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package
 
Example:
```
import Vex from 'vexflow/src/index.js'
const VF = Vex.Flow

... etc
```

spianoDev:
> Right, but if you `npm uninstall vexflow` you'll have the same result. The `npm install` isn't doing anything in your example, and is misleading.

I will have to play with it again to see if that is true. I cannot remember if I installed globally (I suspect not) or locally in the specific repo I was building. My current schedule is a bit nutty, so it will likely be another couple of weeks before I can dig back into this project. Thanks for commenting!

spianoDev:
I submitted to the wiki... please feel free to message me with any edits or improvements!

PR #725, opened 2019-11-10
bugfix(svgcontext.js): restore scale on resize & state save/restore

### LINKED ISSUE #723: SVGContext's Scale does not stick after Renderer.resize()
A simple update to the resize function to call this.scale() with its own state should be sufficient. I may even submit a PR for it.

mix3d:
Fixes #723


mix3d:
Fixes #725

mix3d:
Not sure why the CI failed though...

PR #732, opened 2019-11-24
Add `!` to Banner comment

mscuthbert:
Many code minimizers will retain comments that begin with `/**!` or `/*!` so if the code goes through a minimizer (even a minimizer of a minimizer) this will help others keep their license obligations.


0xfe:
Thanks, good PR.

PR #746, opened 2020-03-05
Limit CanvasContext to max browser width/height, warn

### LINKED ISSUE #742: Canvas Backend breaks at height > 32767 (browser limit)
Using the Canvas backend, you can resize a canvas to height=32767 and it renders fine, but resize to height=32768 and it breaks, showing this in Chrome:
image

Try it here:
https://jsfiddle.net/2Lsme8yd/
(setting the height back to 32767 fixes the problem)

This is because browsers limit the size of a canvas element:
https://stackoverflow.com/a/11585939/10295942

(It's funny that they just break without a real error message though.
And that a 16 bit signed integer seems to be the somewhat arbitrary limit. I guess you can fit width+height in one 32-bit register then.)

Btw, there's a npm module canvas-size that tells you the current browser's maximum canvas size, which could be used:
https://stackoverflow.com/a/53677532/10295942

So, @0xfe what do you think would be the best way to deal with this? OSMD logs a warning and simply reduces the height (and width) to the maximum of 32767.
I suppose Vexflow could do the same, outputting a warning with Vex.W(). Or exit with 1? ;)

I can make a PR for this, just wanted to discuss how to proceed first.

sschmidTU:
fix #742

old behavior:
browser canvas element crashes when width or height > 32767 used on renderer.resize()
https://jsfiddle.net/2Lsme8yd/

new behavior:
renderer.resize() automatically crops dimensions to 32767 and gives warning
https://jsfiddle.net/tm9qaLzb/

npm test ran without fails.

(note that this only applies to CanvasContext, not SVGContext)

### DIFF
--- a/src/canvascontext.js
+++ b/src/canvascontext.js
@@ -19,10 +19,10 @@ export class CanvasContext {
     return 32767; // Chrome/Firefox. Could be determined more precisely by npm module canvas-size
   }
 
-  static CheckCanvasDimensionsForBrowserLimit(width, height) {
+  static SanitizeCanvasDims(width, height) {
     if (Math.max(width, height) > this.CANVAS_BROWSER_SIZE_LIMIT) {
       Vex.W(
-        'canvas dimensions exceed browser limit. reducing/cropping to ' +
+        'Canvas dimensions exceed browser limit. Cropping to ' +
         this.CANVAS_BROWSER_SIZE_LIMIT
       );
       if (width > this.CANVAS_BROWSER_SIZE_LIMIT) {
@@ -120,9 +120,7 @@ export class CanvasContext {
   }
 
   resize(width, height) {
-    width = parseInt(width, 10);
-    height = parseInt(height, 10);
-    [width, height] = this.CheckCanvasDimensionsForBrowserLimit(width, height);
+    [width, height] = this.SanitizeCanvasDims(parseInt(width, 10), parseInt(height, 10));
     return this.vexFlowCanvasContext.resize(width, height);
   }
 
--- a/src/canvascontext.js
+++ b/src/canvascontext.js
@@ -5,6 +5,8 @@
 //
 // Copyright Mohit Cheppudira 2010
 
+import { Vex } from './vex';
+
 /** @constructor */
 export class CanvasContext {
   static get WIDTH() {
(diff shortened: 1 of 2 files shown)


src/canvascontext.js
```diff
@@ -13,6 +15,26 @@ export class CanvasContext {
...
     return 400;
   }
+  static get CANVAS_BROWSER_SIZE_LIMIT() {
+    return 32767; // Chrome/Firefox. Could be determined more precisely by npm module canvas-size
+  }
+
+  static CheckCanvasDimensionsForBrowserLimit(width, height) {
+    if (Math.max(width, height) > this.CANVAS_BROWSER_SIZE_LIMIT) {
+      Vex.W(
+        'canvas dimensions exceed browser limit. reducing/cropping to ' +
```
0xfe: Fix capitalization please 'Canvas dimensions exceed browser limit. Cropping to ...'

src/canvascontext.js
```diff
@@ -13,6 +15,26 @@ export class CanvasContext {
   static get HEIGHT() {
     return 400;
   }
+  static get CANVAS_BROWSER_SIZE_LIMIT() {
+    return 32767; // Chrome/Firefox. Could be determined more precisely by npm module canvas-size
+  }
+
+  static CheckCanvasDimensionsForBrowserLimit(width, height) {
```
0xfe: Maybe a shorter function name: `SanitizeCanvasDims`

src/canvascontext.js
```diff
@@ -98,7 +120,10 @@ export class CanvasContext {
   }
 
   resize(width, height) {
-    return this.vexFlowCanvasContext.resize(parseInt(width, 10), parseInt(height, 10));
+    width = parseInt(width, 10);
+    height = parseInt(height, 10);
+    [width, height] = this.CheckCanvasDimensionsForBrowserLimit(width, height);
```
0xfe: You can move the `parseInt` into the above `SanitizeCanvasDims` (which is a better place for it.)

src/renderer.js
```diff
@@ -45,7 +45,12 @@ export class Renderer {
 
   static buildContext(elementId, backend, width, height, background) {
     const renderer = new Renderer(elementId, backend);
-    if (width && height) { renderer.resize(width, height); }
+    if (width && height) {
+      if (backend === Renderer.Backends.CANVAS) {
+        [width, height] = CanvasContext.CheckCanvasDimensionsForBrowserLimit(width, height);
```
0xfe: Looks like you're already calling this inside `resize`, so we don't need this here, no?

sschmidTU: ah, right, i thought this was separate. removing this call.

src/canvascontext.js
sschmidTU: I assume like this:
`[width, height] = this.SanitizeCanvasDims(parseInt(width, 10), parseInt(height, 10));`
becuse moving it into the `SanitizeCanvasDims` function would mean i 'd need to call parseInt again for width and height for `this.vexFlowCanvasContext.resize(width, height);`, which would be duplicated.

src/canvascontext.js
sschmidTU: yup, sounds good

src/canvascontext.js
sschmidTU: done

sschmidTU:
changes are all implemented, and tested again.
thanks for reviewing by the way!

src/canvascontext.js
0xfe: Okay, that's fine, thanks.

0xfe:
Looks good. Thanks for the PR! Merging :-)

PR #764, opened 2020-04-06
Export StaveText in Vex.Flow object

jeroenlammerts:


0xfe:
Thanks, didn't realize that this was unexported.

PR #765, opened 2020-04-06
Fix link to the "build" instructions

rowild:
The link to the build instructions lead to an empty wiki page. I assume it should point to the "build and Release" page, right? If not, please delete this PR.


0xfe:
Thank you.

PR #767, opened 2020-04-06
Fix a note, which was an octave to high

rowild:
In measure 6 of the Bach Minuet Demo test, the B natural was written as "B5/q", whereas it should be "B4/q".


0xfe:
Thanks.

Issue #467, opened 2016-10-12
Beam.generateBeams algorithm gives wrong results on dotted rhythms


andiejs:
The code below should automatically beam the notes in groups of 1/4 notes (i.e. as in 4/4 time); instead it beams like: (dotted-quarter) (eighth, eighth) (eighth, eighth) (eighth).

Looks like the algorithm is wrong in that it groups the dotted-quarter as one group (since it can't break it) and then starts a new group at the next eighth. Probably would be better to keep track of the fact that 1/8 of the next group was already used up.

Thanks!
Andie

```javascript
VF = Vex.Flow;

// Create an SVG renderer and attach it to the DIV element named "boo".
var div = document.getElementById("boo")
var renderer = new VF.Renderer(div, VF.Renderer.Backends.SVG);

// Configure the rendering context.
renderer.resize(500, 500);
var context = renderer.getContext();
context.setFont("Arial", 10, "").setBackgroundFillStyle("#eed");

// Create a stave of width 400 at position 10, 40 on the canvas.
var stave = new VF.Stave(10, 40, 400);

// Add a clef and time signature.
stave.addClef("treble").addTimeSignature("4/4");

var notes = [
  new VF.StaveNote({ keys: ["a/4"], duration: "4d" }).addDotToAll(),
  new VF.StaveNote({ keys: ["b/4"], duration: "8" }).
    addAccidental(0, new VF.Accidental("b")),
  new VF.StaveNote({ keys: ["c/4"], duration: "8" }),
  new VF.StaveNote({ keys: ["d/4"], duration: "8" }),
  new VF.StaveNote({ keys: ["e/4"], duration: "8" }).
      addAccidental(0, new VF.Accidental("b")),
  new VF.StaveNote({ keys: ["d/4"], duration: "8" })
  ];

var beams = VF.Beam.generateBeams(notes, {groups : [ new Vex.Flow.Fraction(1,4) ]});
Vex.Flow.Formatter.FormatAndDraw(context, stave, notes);
beams.forEach(function(b) {b.setContext(context).draw()})
// Connect it to the rendering context and draw!
stave.setContext(context).draw();
```

Silverwolf90:
Yup, looks like a bug.

Screenshot of case:

<img width="431" alt="screen shot 2016-11-18 at 4 44 37 pm" src="https://cloud.githubusercontent.com/assets/1631625/20451481/b4428c60-adae-11e6-94f1-7fabaff2d661.png">

[...]

PR #784, opened 2020-04-26
Bugfix/auto beam

cmigliorini:
This, originally written by @andiejs, appears to fix https://github.com/0xfe/vexflow/issues/467

### DIFF
--- a/src/beam.js
+++ b/src/beam.js
@@ -168,7 +168,8 @@ export class Beam extends Element {
 
     function createGroups() {
       let nextGroup = [];
-
+      // number of ticks in current group
+      let currentGroupTotalTicks = new Fraction(0, 1);
       unprocessedNotes.forEach(unprocessedNote => {
         nextGroup = [];
         if (unprocessedNote.shouldIgnoreTicks()) {
@@ -176,10 +177,9 @@ export class Beam extends Element {
           currentGroup = nextGroup;
           return; // Ignore untickables (like bar notes)
         }
-
         currentGroup.push(unprocessedNote);
         const ticksPerGroup = tickGroups[currentTickGroup].clone();
-        const totalTicks = getTotalTicks(currentGroup);
+        const totalTicks = getTotalTicks(currentGroup).add(currentGroupTotalTicks);
 
         // Double the amount of ticks in a group, if it's an unbeamable tuplet
         const unbeamable = Flow.durationToNumber(unprocessedNote.duration) < 8;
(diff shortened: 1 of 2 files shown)


cmigliorini:
Finally managed to get the visual regression test working. I do have a failure here...

Blessed
Auto_Beaming More_Automatic_Beaming_Blessed

Current
Auto_Beaming More_Automatic_Beaming_Current

Diff
Auto_Beaming More_Automatic_Beaming

But we add this, which is correct
Auto_Beaming Auto_Beaming_With_Overflow_Group

...where previous version would produce
Auto_Beaming Auto_Beaming_With_Overflow_Group

which is better to me.

I don't know what to think: failed test corresponds to a somewhat strange music, but still is a failed test. Let me know what you think.

mscuthbert:
If I found the test properly, the meter is 9/8: 
```
      var voice = score.voice(score.notes(
        'c4/8, g4/4, c5/8., g5/16, a5/4, a5/16, (c5 e5)/16, a5/8'
      ), { time: '9/8' });
```

The failed test could go either way, though I would prefer the new/current version.  The high A quarter note subverts the norms of the meter (I would rewrite it as two eighth notes, tied); thus beaming the notes on either side wouldn't necessarily be the correct thing to do.  So, is it better to beam what you can, or only beam if the whole group is right?  I prefer the latter as it says "watch out--your intuition is going to be wrong"

cmigliorini:
Agreed. I'd be happy to rewrite the test adding tied notes -- although I'm not familiar with EasyScore and could not find how to tie notes :/

I could also give a go to adding time signature but I guess that would break _all_ tests so it might be another story...

cmigliorini:
Matter of fact, now you mention the fact that it's 9/8: `Auto_Beaming.More_Automatic_Beaming_Current` does not actually beam eights by three when you split the quarters. With
```
 var voice = score.voice(score.notes(
     'c4/8, g4/8, g4/8, c5/8., g5/16, a5/8, a5/8, a5/16, (c5 e5)/16, a5/8'
), { time: '9/8' });
```

commit f4753d368df84627464bfa32c5af83c1bce0b175 gives:

Auto_Beaming More_Automatic_Beaming_Current

cmigliorini:
Oh, got it. You need to specify groups

```js
var beams = VF.Beam.applyAndGetBeams(voice, undefined,
  VF.Beam.getDefaultBeamGroups('9/8'));
```

so now we don't break the test. I also verified in [this branch](https://github.com/cmigliorini/vexflow/tree/improve-more-automatic-beaming) that when splitting high quarter A it still work.

I would say that blessed test works "by chance", as it would probably give unexpected beaming should the measure continue with an eight and a quarter, so perhaps I should also bring 5785e1a36ebb01b67030304fa42b6b3533488584 into this branch, breaking visual regression test for the better?

0xfe:
This is certainly better. A few comments:

- Add a few more tests to show the corner cases covered by this change (and please add screenshots to this PR.)
- Clean up the code (remove spurious `// ANDY` comments, and rename variables to match VexFlow style: `ticksRunningTotal` instead of `tickrunningtotal`, or better still `totalTicks`.)
- The while loop there with repeated subtraction is weird, why not division?

cmigliorini:
Trying to remove the `while` loop made me realize that we need to call `nextTickGroup()` at each step, especially when there are irregular groups.

So I kept the loop and moved `nextTickGroup()` call inside it with appropriate tick count.

So I added 3 more tests cases.

Current images for all 4 new test cases are

## Simple overflow on a simple measure
This would probably be the most frequent use case : a dotted quarter followed by two eighths on a 4/4 measure.
Auto_Beaming Auto_Beaming_With_Overflow_Group

## Autobeaming on an irregular pattern
I believe this test was not already present: a standard 4/4 measure with an irregular grouping pattern (3/8, 3/8, 2/8).

Auto_Beaming Automatic_Beaming_4_4_with__3__3__2_Pattern_and_Overflow

## Same with an overflow on first group
Here, dotted quarter overflows first group, algorithm properly groups following two eights and final two.

Auto_Beaming Automatic_Beaming_4_4_with__3__3__2_Pattern

## Same with an 8/4 measure
This time we have an 8/4 measure and only 4 quarters worth of groups (3/8, 2/8, 3/8), this test makes sure the code properly groups 3, 2, 3, 3, 2, 3, even with an overflow in first group.

Auto_Beaming Automatic_Beaming_8_4_with__3__2__3_Pattern_and_Overflow

0xfe:
Looks great. Thanks for addressing this. Merging.

(Also, looks like a formatter bug there -- when there's a tiny note next to a large duration note, not enough space on the left of the large note. E.g., the whole next to the 16th in your last test.)

PR #786, opened 2020-04-27
Add chord symbol glyphs to vexflow fonts

AaronDavidNewman:
In support of chord symbols, issue 776.


tools/smufl/config/valid_codes.js
```diff
@@ -196,4 +196,19 @@ module.exports = {
   vexNoteHeadMutedBreve: 'vf',
   vexNoteHeadRectBlack: 'vd4',
   vexNoteHeadRectWhite: 'vd5',
+
+  csymDiminished:null,
```
0xfe, line 200: Few stylistic notes:

- Move this block before the `vex...` blocks.
- Add a comment above the block specifying that these are chord symbols.
- Add a space between the colon and the null

src/fonts/bravura_glyphs.js
```diff
@@ -1,1055 +1,1151 @@
 export const BravuraFont = {
-  'glyphs': {
```
0xfe, line 2: You mentioned in the bug that you used a different Bravura OTF file to generate these, can you please add that OTF file to this repo please?

AaronDavidNewman:
>> You mentioned in the bug that you used a different Bravura OTF
When I created the new fork repository, the otf file was correct.  So I must have had a stale version previously.  Style tweaks complete.

0xfe:
Looks great, thanks. Merging.

PR #790, opened 2020-05-03
chord symbol contribution

AaronDavidNewman:
Please have a look and let me know if anything is missing.  I don't know how to do the visual regression tests but I did create test cases.  This is for issue 776


AaronDavidNewman:
cleaned up some of the problematic symbols, very simple kerning adjustments for e.g. A diminished.

0xfe:
Hi Aaron, This looks good at a first glance, and I'd like to review it more thoroughly. Would you mind attaching screenshots of the tests to this PR please?

src/chordSymbol.js
```diff
@@ -0,0 +1,593 @@
...
+      },
+      ')': {
+        code: 'csymParensRightTall',
+        scale: 0.8,
+        y_offset: 1
+      },
+      'rightParen': {
+        code: 'csymParensRightTall',
+        scale: 0.8,
+        y_offset: 1
```
0xfe, line 119: Can you move these `scale` and `y_offset` to the font metrics files please? We're trying to remove constants from the code.

src/chordSymbol.js
```diff
@@ -0,0 +1,593 @@
...
+  }
+
+  addGlyph(glyph, parameters) {
+    parameters = parameters == null ? {} : parameters;
+    parameters.glyph = glyph;
+    parameters.symbolType = ChordSymbol.SymbolTypes.GLYPH;
+    return this.addSymbolBlock(parameters);
+  }
+
+  addGlyphOrText(text, parameters) {
```
0xfe, line 416: Not thrilled with this function name. What exactly is the purpose of this?

src/chordSymbol.js
```diff
@@ -0,0 +1,593 @@
...
+
+  getCategory() { return ChordSymbol.CATEGORY; }
+
+  // Set font family, size, and weight. E.g., `Arial`, `10pt`, `Bold`.
+  setFont(family, size, weight) {
+    this.font = { family, size, weight };
+    return this;
+  }
+
+  setUseKerning(val) {
```
0xfe, line 451: prefer `enableKerning`

src/chordSymbol.js
```diff
@@ -0,0 +1,593 @@
...
+    } else /* CENTER_STEM */ {
+      x = this.note.getStemX() -  this.getWidth() / 2;
+    }
+    this.symbolBlocks.forEach((symbol) => {
+      const sp = this.isSuperscript(symbol);
+      const sub = this.isSubscript(symbol);
+      let curY = y;
+
+      if (sp) {
+        curY -= 8;
```
0xfe, line 556: These constants should be in metrics files. (Also, how does this remain stable as the font sizes change?)

src/chordSymbol.js
```diff
@@ -0,0 +1,593 @@
+// [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
```
0xfe, line 1: Please lowercase the filename (to stay consistent with other files.)

src/chordSymbol.js
```diff
@@ -0,0 +1,593 @@
...
+    };
+  }
+
+  // Glyph data
+  static get GLYPHS() {
+    return {
+      'diminished': {
+        code: 'csymDiminished',
+        scale: 1,
+        y_offset: 0
```
0xfe, line 64: Since this is a new file, please use `camelCase` for all variable names as per JS conventions.

src/chordSymbol.js
```diff
@@ -0,0 +1,593 @@
...
+
+        if (symbol.symbolType === ChordSymbol.SymbolTypes.GLYPH &&
+          symbol.glyph.code === ChordSymbol.GLYPHS.over.code) {
+          lineSpaces = 2;
+          symbol.x_offset += -2;
+        }
+
+        if (j > 0 && instance.symbolBlocks[j - 1].symbolType === ChordSymbol.SymbolTypes.GLYPH &&
+          instance.symbolBlocks[j - 1].glyph.code === ChordSymbol.GLYPHS.over.code) {
+          symbol.x_offset += -2;
```
0xfe, line 221: Style: `xOffset`

AaronDavidNewman:
Sure, take your time.  Screen shots added.

src/chordSymbol.js
AaronDavidNewman: done

src/chordSymbol.js
AaronDavidNewman: done

src/chordSymbol.js
AaronDavidNewman: fonts come in sizes, i forgot that.  I added some test cases for it and adjust for font size, to some effect (but it still drifts as fonts get super large/small).  I may have too many test cases now.

src/chordSymbol.js
AaronDavidNewman: done

src/chordSymbol.js
AaronDavidNewman: image

src/chordSymbol.js
AaronDavidNewman: Here is the description, feel free to suggest a better name:
```
  // ### addGlyphOrText
  // Add a glyph for each character in 'text'.  If the glyph is not
  // available, use text from the font.  E.g.:
  // `addGlyphOrText("(+5#11)")`
  // will use text for the '5' and '11', and glyphs for everything else.
```

src/chordSymbol.js
AaronDavidNewman: done

src/chordSymbol.js
0xfe: Did you push the update? (I still see `chordSymbol.js`)

src/chordSymbol.js
0xfe: Okay, this is fine -- thanks.

src/chordSymbol.js
AaronDavidNewman: I did.  I'm guessing this is just the github review widget.  When I pull the fork, it is lower case and it won't even compile otherwise.

AaronDavidNewman: well, hang on I take that back.  It does not show up but I did make the change.  Let me look into it.

AaronDavidNewman: ok, live and learn.  I changed it in GH site and it seems good now.

AaronDavidNewman:
Hi Mohit, is there anything else I need to do on this?  I think I addressed all the issues.

0xfe:
Thanks for following up, Aaron. I thought I replied a few weeks ago, but I don't see it here. Not sure what happened -- I may have forgotten to submit or something :-/

Anyhow, thanks for the changes -- it mostly looks good, and I'll merge. I think the next step is to make it pretty :-) Maybe we can get some examples of chord symbols in the wild that look good, and try to imitate them?

Also, it'd be nice if we had a default recommended font for text.

AaronDavidNewman:
I have loads of fake books, mostly of the 'Real Book' variety.  I can scan some images and compare to what we have. Music theory was long ago for me but I know that's a big use case.  Maybe others can point me to good role models.

>> Also, it'd be nice if we had a default recommended font for text.

I'm not sure what that means...I default to cursive with the Petaluma engraving font and Arial for the others.  And that can be overridden.  But I think you mean something else.

Maybe we also need a wiki?

0xfe:
Yeah, a wiki for chord symbols will be nice.

By recommended text font I meant, something from (e.g.,) fonts.google.com that works well for chord symbols. Default browser fonts like Arial are very bland, IMO. :-)

If we use a standard font for text, then it makes kerning, sizing, etc. a lot simpler and more consistent.

AaronDavidNewman:
so...many...fonts.  I am thinking about different recommended fonts for different purposes, with a 'general purpose' one that is the most consistently legible.

A few observations:
1. Capital letters and numbers should have the same baseline at the same font/weight size, e.g. F7
2. '9', '1', should be easily distinguishable at small font sizes
3. slab serif fonts tend to be the most legible in different contexts, kind of boring though
4. I preferred families that had multiple weights available, although there were exceptions (Sriracha).
5. Maybe we recommend different fonts for different engraving fonts, or for different purposes.  A handwriting one for Petaluma, a serif and a sans, and default to a blocky sans font.

You can see them all in action:
https://jsfiddle.net/AaronDavidNewman/vgsbfpxw/
https://jsfiddle.net/AaronDavidNewman/159gz640/
https://jsfiddle.net/AaronDavidNewman/62v8s1eu/

These all use my branch since this is not merged.  I think the most promising are:
Exo 2 (Petaluma, handwriting)
Roboto (Sans)
Jost (Sans)
Taviraj (serif)
Libre Baskerville (serif)
Roboto Slab (serif)
Sriracha (italic, sans, handwriting)
Concert One (1920's sheet music vibe).

Thoughts?

0xfe:
Wow, thanks for the demos, Aaron. Yes, def a lot of fonts, and we prob need some criteria before picking one. So far I have:

- Open license
- Internally consistent (kerning, metrics, glyphs, etc.)
  + glyphs: are the music notation glyphs (like sharp, flat) also part of the font (or do they fall back)
- Consistent with what's in public music notation (e.g., should they have serifs, etc.)
- Looks good :-)

Bringing in @mscuthbert in case he has any pointers.

0xfe:
Merging this now, let's continue on an issue. Thanks Aaron.

Issue #794, opened 2020-06-12
Map file not generated for debug version


AaronDavidNewman:
The minified monolith is generating a mapfile, but the debug version is not.  I think it's because these lines are transposed in the gruntfile.  I assume the intent here is not to generate a map file in production, but to generate one in debug.  If so I'll create a pull request.  Gruntfile.js

  devtool: (mode === 'production') ? 'source-map' : false,
s/b
devtool: (mode === 'production') ? false : 'source-map',

AaronDavidNewman:
Added a flag, closing...

PR #795, opened 2020-06-12
generate map file in debug mode

AaronDavidNewman:
#794 Map file not generated for debug build

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -39,7 +39,7 @@ module.exports = (grunt) => {
         libraryTarget: 'umd',
         libraryExport: 'default',
       },
-      devtool: (mode === 'production') ? false : 'source-map',
+      devtool: (process.env.VEX_GENMAP || mode === 'production') ? 'source-map' : false,
       module: {
         rules: [
           {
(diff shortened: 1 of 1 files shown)


0xfe:
This disables map files for production, where it is more important (IMO). Is that intentional?

AaronDavidNewman:
Hmm...a fair point.  So maybe we always want the map file, then?  Before, it was generating a map file for the minified version but not the debug version.  While it's possible to debug the monolith, it is not ideal

`devtool: 'source-map',...`

I think that's what it did prior to April.

0xfe:
How about an env setting for it instead,  e.g., calling `VEX_GENMAP=true grunt` will always generate a map file? So:

```
devtool: (process.env.VEX_GENMAP || mode === 'production') ? 'source-map' : false
```

AaronDavidNewman:
OK.

0xfe:
Thanks Aaron. Can you confirm that this works correctly for:

- with env flag:
  + dev: generates source map
  + prod: generates source map
- without env flag
  + dev: does not generate source map
  + prod: generates source map

AaronDavidNewman:
That is exactly how it works.

0xfe:
Thanks, merging.

PR #797, opened 2020-06-18
Merge fixes from smoosic branch

AaronDavidNewman:
Added metrics for text fonts, created 'preferred' fonts.  Updated all font and chord glyph metrics to be on the scale of glyph resolution for offsets, and scales.  Tune stacking, normalize scaling so all letters have correct, proportional width.

I also created a wiki page.  I don't know if I can just add it on the main vexflow repository,  but you can preview it here:
https://github.com/AaronDavidNewman/vexflow_smoosic/wiki/Chord-Symbols


0xfe:
Aaron, great change! I appreciate that you used the metric files to get all this to work.

The vexflow wiki is editable, so you can update that.

One question though: you added the fonts to the test CSS file, how reliant are we on those specific fonts, and how would a vexflow user know to load those fonts? (i.e., what parts of docs should be updated, and what warning/error messages can the user see if they don't have them loaded.)

AaronDavidNewman:
I cover the font issue in some detail in the wiki.  The current location of the fonts is OK for the short term.  Probably the most robust location for those fonts would be in vexflow itself.  They are not large, (woff ~ 40K for each font).  But in a nutshell:

-  If the fonts are not available, the browser fallback fonts will be used, using the metrics of the default fonts.  Works OK for formatting, but the fonts are not as nice
-  Other fonts can be substituted, and they will work well if the metrics are similar to the default fonts
-  If the metrics are far enough apart that using them becomes destructive, you can disable their use, and the result is basically what you had before I made this change.

afaik there is no way to detect if a font is not available.  Also, a font could be available but not contain a glyph that you reference, and fonts may be loaded when the music is first rendered.  So you don't get a warning, but gracefully degraded results.

I also included some instructions on how to create your own metrics for your favorite fonts.

AaronDavidNewman:
I updated the Wiki with chord change description.  Let me know if you want me to add the text fonts to the vexflow project instead of importing them.

0xfe:
Hey Aaron, Sorry, I've been out on vacation. Will finish the review as soon
as I'm back.

0xfe:
Yes, please add the fonts to the VexFlow repo -- better still, if there are authoritative repos or CDNs (e.g., Google Fonts), those would be preferred.

0xfe:
(Apologies for the slow review -- hands pretty full these days.)

AaronDavidNewman:
The petaluma font is not available in any CDN that I know of - so I'll add
that to the vexflow repo and get the Robo one from Google fonts.

0xfe:
Sounds good.

AaronDavidNewman:
OK, done.  I put the woff files in the releases/fonts directory, since the build directory doesn't persist.  And it seemed wrong somehow to pull it from tools or src.

PR #801, opened 2020-07-15
Jazz articulations/techniques

AaronDavidNewman:
I call them 'techniques' b/c that's what Sibelius and Steinberg call them.

image


0xfe:
Also, should there be some space between this ornament and the note?
Screen Shot 2020-07-15 at 2 20 21 PM

AaronDavidNewman:
>> Why not use ornaments or articulations for this?

ornaments are probably closest.   I could move the logic into that module, if you like.  The formatting logic is a little different, but they can co-exist.

Do you want me to put the codes in `Flow.ornamentCodes`?

>> should there be some space between this ornament and the note?

Agreed.  I'll fix that also.

0xfe:
On Wed., Jul. 15, 2020, 17:48 Aaron, <notifications@github.com> wrote:

> Why not use ornaments or articulations for this?
>
> ornaments are probably closest. I could move the logic into that module,
> if you like. The formatting logic is a little different, but they can
> co-exist.
>
> Do you want me to put the codes in Flow.ornamentCodes?
>

Yes please.

AaronDavidNewman:
Merged into ornaments module.  Tried to reduce dueling format logic as much as I could..  Cleanup up some formatting.

image

src/ornament.js
```diff
@@ -27,27 +27,100 @@ export class Ornament extends Modifier {
...
+    ];
+  }
+
+  // ### ornamentArticulation
+  // goes above/below the note based on space availablity
+  static get ornamentArticulation() {
+    return ['bend', 'plungerClosed', 'plungerOpen'];
+  }
+
+  static get glyphMetrics() {
```
0xfe, line 116: All VexFlow elements derive from `Element`, so you can use `getFontStack()` to get the current font stack.

src/ornament.js
```diff
@@ -27,27 +27,100 @@ export class Ornament extends Modifier {
...
-      } else {
-        ornament.setTextLine(state.text_line);
-        state.text_line += increment;
+        // articulations above/below the line can be stacked.
+        if (Ornament.ornamentArticulation.indexOf(ornament.type) >= 0) {
+          // Unfortunately we con't know the stem direction.  So we base it
+          // on the line number, but also allow it to be overridden.
+          if (ornament.note.getLineNumber() >= 3 || ornament.getPosition() === Modifier.Position.ABOVE) {
+            state.top_text_line += increment;
+            ornament.render_options.yOffset += yOffset;
```
0xfe, line 61: Ornaments derive from `Modifier` which already have `x_shift` and `y_shift`, and you can call `setXShift/setYShift`, would those work here instead of adding custom `render_options`?

src/ornament.js
```diff
@@ -27,27 +27,100 @@ export class Ornament extends Modifier {
...
+  // goes above/below the note based on space availablity
+  static get ornamentArticulation() {
+    return ['bend', 'plungerClosed', 'plungerOpen'];
+  }
+
+  static get glyphMetrics() {
+    return Vex.Flow.DEFAULT_FONT_STACK[0].metrics.glyphs.jazzOrnaments;
+  }
+
+  get metrics() {
```
0xfe, line 120: We don't use (non-static) metrics as convention, can you replace this with `getMetrics()` (this keeps the codebase consistent.)

AaronDavidNewman:
> is there a way to remove the if (ornament.jazzMetrics) and have all the formatting take the same code path?

I was able to remove this check everywhere except here in the constructor.  I am a little confused by this one, but the legacy ornaments have spacing issues without it. 

I should be able to push these other changes shortly.

``` javascript
    if (!jazzMetrics) {
      this.glyph.setOrigin(0.5, 1.0); // FIXME: SMuFL won't require a vertical origin shift
    }

```

0xfe:
I think this is fine, you can now rename jazzMetrics to something more
specific (resetOrigin?)

AaronDavidNewman:
ok done

0xfe:
Looks great, thanks for working on this! Merging. :-)

PR #804, opened 2020-08-03
add context/svg group for tabnote

sschmidTU:
This makes it so that a tabnote can be found as a group in the svg DOM.

The same was already possible for stavenote (shown as `<g class="vf-stavenote"...`).

image

`npm test` passes except for one visual diff in EasyScore.Draw_Options, which i don't think i affected.

### DIFF
--- a/src/tabnote.js
+++ b/src/tabnote.js
@@ -470,6 +470,7 @@ export class TabNote extends StemmableNote {
     this.setRendered();
     const render_stem = this.beam == null && this.render_options.draw_stem;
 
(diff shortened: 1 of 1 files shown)


0xfe:
Looks good, thanks! Merging.

dmgdwd:
On new install using `npm install --save vevflow` this update is not getting installed.
Should I be doing something different?

sschmidTU:
@dmgdwd there hasn't been a release since this merge yet, so it's not yet on npm, just merged to the master branch.

PR #806, opened 2020-08-30
Double thickness of Ledger Lines

mscuthbert:
Elaine Gould in "Behind Bars" notes: "Ledger lines are an extension of the stave. They are spaced the same distance apart as stave—lines, but *they are about twice as thick*"  Ledger lines in Vexflow have always been too thin for my tastes.

Makes default ledge lineWidth double the default staff width.

Obviously breaks many visual regression tests.  Here is a pretty normal change:

Blessed:
Accidental Automatic_Accidentals___Multi_Voice_Inline_Blessed

Current:
Accidental Automatic_Accidentals___Multi_Voice_Inline_Current

Diff:
Accidental Automatic_Accidentals___Multi_Voice_Inline

On some lineStyles and scales, the rounding in making the PNG makes the line change more obvious:

Blessed:
Annotation Bottom_Annotations_with_Beams_Blessed

Current:
Annotation Bottom_Annotations_with_Beams_Current

Diff:
Annotation Bottom_Annotations_with_Beams

This was the only test that explicitly called setLedgerLine style (on the second to last eighth note) -- its ledger line style is respected.  However the stave.lineWidth = 3 -- which is already almost ridiculously thick -- makes even more ridiculous ledger line widths.  Should we set a maximum width above which we do not double ledger line thickness?  Or do people get what they ask for?  I think it looks okay in context.

Blessed:
StaveNote Stave__Ledger_Line__Beam__Stem_and_Flag_Styles_Blessed

Current:
StaveNote Stave__Ledger_Line__Beam__Stem_and_Flag_Styles_Current

Diff:
StaveNote Stave__Ledger_Line__Beam__Stem_and_Flag_Styles

Other visual regression test failures have to do with placement of articulations which changed in a recent PR, not this one.

### DIFF
--- a/src/stavenote.js
+++ b/src/stavenote.js
@@ -1018,9 +1018,9 @@ export class StaveNote extends StemmableNote {
     // if lineWidth is not specified in getLedgerLineStyle will use
     // twice stave.getStyle() lineWidth
     if (ledger_line_style.lineWidth === undefined && style.lineWidth !== undefined) {
-      style.lineWidth *= 2;
+      style.lineWidth *= Flow.LEDGER_LINE_THICKNESS_MULTIPLIER;
     } else if (style.lineWidth === undefined) {
-      style.lineWidth = Flow.STAVE_LEDGER_LINE_THICKNESS;
+      style.lineWidth = Flow.STAVE_LINE_THICKNESS * Flow.LEDGER_LINE_THICKNESS_MULTIPLIER;
     }
 
     this.applyStyle(ctx, style);
--- a/src/stavenote.js
+++ b/src/stavenote.js
@@ -81,7 +81,7 @@ export class StaveNote extends StemmableNote {
       let minL = props[props.length - 1].line;
       const stemDirection = notes[i].getStemDirection();
       const stemMax = notes[i].getStemLength() / 10;
-      const stemMin = notes[i].getStemMinumumLength() / 10;
+      const stemMin = notes[i].getStemMinimumLength() / 10;
 
       let maxL;
       if (notes[i].isRest()) {
@@ -327,6 +327,9 @@ export class StaveNote extends StemmableNote {
     // Drawing
     this.note_heads = [];
     this.modifiers = [];
+
+    // Default ledger line style. If lineWidth not specified will set to double stave width
+    this.ledgerLineStyle = {};
 
     Vex.Merge(this.render_options, {
       // font size for note heads and rests
@@ -835,7 +838,7 @@ export class StaveNote extends StemmableNote {
   // Get the width of the note if it is displaced. Used for `Voice`
   // formatting
   getVoiceShiftWidth() {
-    // TODO: may need to accomodate for dot here.
+    // TODO: may need to accommodate for dot here.
     return this.getGlyphWidth() * (this.displaced ? 2 : 1);
   }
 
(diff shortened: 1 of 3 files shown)


src/tables.js
```diff
@@ -11,6 +11,7 @@ const Flow = {
   STEM_WIDTH: 1.5,
   STEM_HEIGHT: 35,
   STAVE_LINE_THICKNESS: 1,
+  STAVE_LEDGER_LINE_THICKNESS: 2,  // Gould, Behind Bars: "about twice as thick"
```
0xfe, line 14: What's `STAVE_LINE_THICKNESS` used for? (I thought I had used that for ledger line thickness.)

src/stavenote.js
```diff
@@ -81,7 +81,7 @@ export class StaveNote extends StemmableNote {
       let minL = props[props.length - 1].line;
```
0xfe, line 81: Would this work for tab notes too?

mscuthbert: I don't think they have ledger lines.  If you're thinking of the stem-length PR it might, but I they won't need it because of lack of ledger lines.

src/tables.js
mscuthbert: STAVE_LINE_THICKNESS is the default thickness of a normal staff line.  

I've renamed STAVE_LEDGER_LINE_THICKNESS to LEDGER_LINE_THICKNESS_MULTIPLIER and now use as a default STAVE_LINE_THICKNESS * LEDGER_LINE_THICKNESS_MULTIPLIER

mscuthbert:
Great eye -- I looked through all the changes before but looking again in fact there were a few that were missed:

TabNotes don't have ledger lines, so none of the tab examples changed and that's what was expected, but the C#s here in the staff should have changed and didn't:

Formatter Notes_with_Tab

We actually don't have any tests of grace notes on ledger lines -- that's a test I'll try to add -- but the two StaveNote Cs here should have thicker ledger lines and don't:

Grace_Notes Grace_Notes_Multiple_Voices_Multiple_Draws

I'll keep looking at this after teaching tomorrow and see if I can figure out why.  One uses EasyScore but the other does not so that can't be it.

mscuthbert:
Follow up -- actually the grace note example is changing the ledger length, it's just below the perceptual similarity threshold.

For some reason most of the Formatter examples are not generating images any more, do you know why?  They are being called with `runSVG` rather than `VF.Test.runTests` but they obviously generated images at some point.

0xfe:
Apologies for the extremely slow response.

Right, I think that's because `runSVGTest` never generated images in the first place. (`runTests` calls `runSVGTest`, `runCanvasTest`, and `runNodeTest`, the last one generates images on the commandline.)

If you think there are formatter tests that need to be in the visual test images, then feel free to add an equivalent `runNodeTest`.

Or maybe even replace `runSVGTest` with `runTests` wrapped by:

```
const runCanvasTests = VF.Test.RUN_CANVAS_TESTS;
VF.Test.RUN_CANVAS_TESTS = false;
runTests(...)
VF.Test.RUN_CANVAS_TESTS = runCanvasTests;
```

Outside of that, the only thing missing here is a test for grace notes on ledger lines, correct?

rvilarl:
@mscuthbert this is issue was mentioned by @ronyeh in #1030 Have a look at "Ledger Lines are darker (thicker?) on VexFlow.com"

rvilarl:
@0xfe this one is no longer required, you can close it

rvilarl:
@mscuthbert the double ledger lines are already merged. I suggest to close this PR.

0xfe:
Closing this. Thanks for working on it, everyone!

PR #807, opened 2020-08-30
Stem length to middle line

mscuthbert:
Behind Bars, p. 14: Stems for notes on more than one ledger line extend to the middle stave-line.

Subclasses getStemExtension on stavenote to account for the height of the note.  As notes get further from Mid line (b/4 in t.c.) stems get longer (unless they go the opposite direction).  Does nothing to non-stem notes like whole notes.

Adds new test and visual test:
StaveNote StaveNote_Draw___StaveNote_Stem_Lengths

Fixes behavior in about 18 tests.  Example:

Blessed:
Clef Clef_Change_Test_Blessed

New:
Clef Clef_Change_Test_Current

Diff:
Clef Clef_Change_Test

Improves beam angles in several tests:

Blessed:
Curve Simple_Curve_Blessed

New:
Curve Simple_Curve_Current

Diff:
Curve Simple_Curve

It makes more obvious an existing bug in a current test.  The accent mark and the fermata on the last note of the first measure here appear to be on top of each other, but actually they are separated by 1/2 line space, similar to the staccato and accent on blessed g/5 in m. 1.  Solution will come in another PR: while the first articulation can appear in a half-line-space, all additional ones on the same side need a minimum of 1 line space (or need to take glyph-size into account). 

Articulation Articulation___Inline_Multiple_Blessed
Articulation Articulation___Inline_Multiple_Current
Articulation Articulation___Inline_Multiple

I believe that this is a fundamental enough improvement to put in first and fix 3-auto-positioned articulations later?

### DIFF
--- a/src/stavenote.js
+++ b/src/stavenote.js
@@ -81,7 +81,7 @@ export class StaveNote extends StemmableNote {
       let minL = props[props.length - 1].line;
       const stemDirection = notes[i].getStemDirection();
       const stemMax = notes[i].getStemLength() / 10;
-      const stemMin = notes[i].getStemMinumumLength() / 10;
+      const stemMin = notes[i].getStemMinimumLength() / 10;
 
       let maxL;
       if (notes[i].isRest()) {
@@ -439,6 +439,10 @@ export class StaveNote extends StemmableNote {
 
   // Automatically sets the stem direction based on the keys in the note
   autoStem() {
+    this.setStemDirection(this.calculateOptimalStemDirection());
+  }
+
+  calculateOptimalStemDirection() {
     // Figure out optimal stem direction based on given notes
     this.minLine = this.keyProps[0].line;
     this.maxLine = this.keyProps[this.keyProps.length - 1].line;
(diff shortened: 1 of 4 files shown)


mscuthbert:
(last commit should have been: "No need for such a *tall* contextBuilder")

sschmidTU:
> Behind Bars, p. 14: Stems for notes on more than one ledger line extend to the middle stave-line.

@mscuthbert is that really mandated by music notation?

This new ultra long stem looks ugly to me:

image

i much prefer the previous usual length stem:

image

mscuthbert:
> > Behind Bars, p. 14: Stems for notes on more than one ledger line extend to the middle stave-line.
> 
> @mscuthbert is that really mandated by music notation?

Hi Simon -- yes this is a general style of music notation -- you'll see it in Finale and all other musical software and engraved music:

Screen Shot 2020-09-01 at 21 31 37

if you don't do it, you end up w/ music like in m. 2 where you have stems that don't match beam length which looks odd or you have beams the obscure the ledger lines which makes it extremely difficult to read the lines:

Screen Shot 2020-09-01 at 21 34 22

in isolation, you're right, a single note looks funny, but it helps to put the majority of the rhythmic information in the bar at about the same eye level.

mscuthbert:
I am noticing that our stems are propotionally thicker than other engraving systems'.  With the AMAZING default of light gray for staves in VexFlow (let's never change that!) we can probably even have thinner stems than other systems.  Gould says, "Stems should be thinner than the stave-line, but not so thin as to reproduce too faintly"

sschmidTU:
Thanks for the explanation @mscuthbert, makes sense!

PR #812, opened 2020-10-31
Text Font feature

AaronDavidNewman:
This is the base component for text font metrics.  I updated the chordsymbol.js module to use TextFont object as an example.  Other than chordsymbol using the new file, there are no other changes or new tests.

There are many places where modules use `this.context.measureText(...)` to get the bounding box of text elements.  In most cases we should be able to substitute the text font object and this should speed up rendering (and also allow applications to use their own fonts).  But I think we should do this on a per-module basis as we'll have to verify the visual regression tests.  Some may change slightly and we may have to create 'opt-in' logic for those.  Also, we may need to include metrics for built-in fonts (e.g. Arial, Times New Roman) since most text formatting in the other unit tests is done using those fonts.


src/fonts/bravura_metrics.js
```diff
@@ -317,13 +317,13 @@ export const BravuraMetrics = {
...
       csymDiminished: {
-        scale: 1,
+        scale: 0.8,
         leftSideBearing: -32,
         advanceWidth: 506,
         yOffset: 0
       },
       csymHalfDiminished: {
-        scale: 1,
+        scale: 0.8,
```
0xfe, line 326: Are these scale changes intentional?

AaronDavidNewman: Yes, sorry I forgot to mention that.  I tweaked the scale for some symbols that were oversized vs. the text.

src/factory.js
```diff
@@ -48,6 +48,7 @@ import { BarNote } from './barnote';
 import { TabNote } from './tabnote';
 import { TabStave } from './tabstave';
 import { TextNote } from './textnote';
+import { TextFont } from './textFont';
```
0xfe, line 51: Can we remove the camel case please? (to keep file naming style consistent.)

src/index.js
```diff
@@ -75,6 +75,10 @@ import { Registry } from './registry';
 import { StaveText } from './stavetext';
 import { GlyphNote } from './glyphnote';
 import { RepeatNote } from './repeatnote';
+import { TextFont } from './textFont';
+import { PetalumaScriptMetrics } from './fonts/petalumaScript_metrics';
+import { RobotoSlabMetrics } from './fonts/robotoSlab_metrics';
```
0xfe, line 80: So, looking at the content of these files, they look like glyphs, so should have the `_glyphs` suffix. The `_metrics` suffix is for customizing positioning/scale and other transformations. Also, let's find a way to separate the text glyphs from the music glyphs. Maybe `petalumascript_text_glyphs`?

(Sorry, I should've caught this in your previous PR.)

0xfe (commented):
Overall looks good. Some style comments.

src/index.js
AaronDavidNewman: Wouldn't the glyphs be in the actual font files, though?  I would say they are metrics, indexed by glyph, and I think advanceWidth/leftBearing would be regarded by many as (a subset of) font metrics.  But I will call them glyphs if you like :)

0xfe: Yes, you're right, however the `x_min`, `x_max`, `ha` etc. fields are typically part of the glyphs (because they capture bounding box info.) If you look at the other `_glyph` files, you'll see the actual glyphs and the same bounding box information.

Anyhow, since there are no glyphs here, maybe it's okay to call them metrics. Can you call them `_textmetrics` instead? Also do you have the code to generate these metrics?

AaronDavidNewman:
I added the font generator file.  It is a limited right now in that it uses `charCodeAt` so it is limited to the printable ASCII characters (32-127). I made the other changes you suggested, the text font files to be myFont_textmetrics.js, and the objects are like MyFontTextMetrics.

0xfe (approved):
Looks great, thanks for addressing. Merging.

PR #817, opened 2020-11-16
Add Sagittal accidentals

cmloegcmluin:
I've added accidentals for the Sagittal microtonal notation system.

More information on Sagittal is available here: http://sagittal.org/

Please let me know if I missed anything. I followed the instructions [here](https://github.com/0xfe/vexflow/wiki/VexFlow-Font-Rendering#glyphs-file-format) to the best of my ability, and included a new test which I made sure passed.


sschmidTU:
Could you add at least one screenshot of the sagittal tests?
Of course we can compile and run them ourselves, but it would be nice to see at a glance.

cmloegcmluin:
I would love to, but unfortunately I can't figure out how...

I haven't yet found any contributing guidelines for this repo, or instructions in the readme for how to develop or how to run tests. So far I've been able to pattern match to add my tests, and diagnose failures in the script for executing the test suite. But I didn't even realize I could see the tests. I now suspect I didn't even successfully register the test I added in the first commit, so it wasn't actually running. 

I've corrected that now, but even after having registered it appropriately, I'm not able to get it to generate an image like other tests that are very similar to it. I do see that images are being generated in `build/images` as a result of running the test suite, however, my new Sagittal test isn't generating anything. I've experimented with clearing out that folder completely and then commenting out the tests which are most similar to my Sagittal test, to verify that I could prevent them from being generated, as a first step toward process-of-elimination brute forcing my way to figuring out what I'm missing, but even when commented out, these tests still generate images, and I don't find any other place in the code base where they are referenced, so it's a complete mystery to me how that's happening.

FYI, I expect that once I am able to actually see what the test I wrote does, I will want to revise it, because I literally just copy-pasted the existing microtonal test and then randomly copy-pasted Sagittal accidentals over what was already there. Some of what appears on the page may be invalid or nonsense by the rules of Sagittal notation. But driving blind, I couldn't figure out what that test was actually testing. 

Thanks in advance for your support on this!

cmloegcmluin:
Update: I have now successfully generated a Sagittal test image, by commenting out part of my test. There must be something inside the specifics of the test itself which causes the image generation to fail, without the test suite failing.

cmloegcmluin:
Alright, I've got something for you. Sorry for the trouble.

oie_1705640qLh6nrnQ

sschmidTU:
No problem, thanks for the screenshot!

0xfe:
Thanks Douglas -- The code looks good. Can you add more tests please (including chords, and beams.)

cmloegcmluin:
Thanks for the nudge to improve the test, now that I can see what I'm doing!

But I am a little bit disappointed with the positioning of the accidentals in the chords. This is the example I am modeling after:

Fig6

and this is what I've managed to recreate in Vexflow at the moment:

oie_1805939YU07Zbi7

I only care about the positioning of the accidentals. I'm not concerned with any other differences between these two images.

If you look at the last chord, the issue is quite subtle. And in that case, I imagine the situation could be improved when I get around to adding [bounding box cut-outs](https://www.w3.org/2019/03/smufl13/specification/bbox-cut-outs.html) to the glyphs, as I already have plans to do. 

But the other discrepancies seem like they would not be resolved by bounding box cut-outs. Bounding box cut-outs only free up a bit of space from the corners, allowing things to slide slightly left or right to nestle up against each other more closely; they're a good solution for glyphs which are diagonal from each other, but because they don't change what the ultimate highest or lowest point in a glyph is, they don't really help with glyphs which are supposed to be horizontal or vertical from each other (in this case vertical).

So regarding these other discrepancies: I assume that Vexflow has positioned the accidentals for the lower pitches (below and) out to the side of the accidentals for the higher pitches because it considers the highest point of the lower accidental to surpass the lowest point of the higher accidentals. But I think that shouldn't be the case. It looks like they should be able to fit as they do in the example I'm modeling after. Is there anything I can do on Vexflow's end to make it so? 

I see here
https://github.com/0xfe/vexflow/blob/00c8a0d17092c24e40dd4519d4b8ec59c711923c/src/accidental.js#L305 some hardcoding of clearance in terms of staff spaces which boots accidentals to new columns. So perhaps it's not something I can accomplish without modifying one of the main modules. Please let me know if I'm on the right track.

Thanks in advance for your guidance. I am just learning Vexflow in the process of making this PR, so I apologize if I've missed something.

cmloegcmluin:
FWIW, the application I plan to use Vexflow for does not require chords for the MVP. So I'm okay with merging this as is, and treating chords as a separate issue to deal with later. Unless there's some complication with the visual regression suite where you prefer to get these tests exactly correct from the start, I think we can just go ahead and merge.

0xfe:
Thanks Douglas, this is good. The (suboptimal) behaviour you see applies to all accidentals, and needs to be addressed separately. Merging! :-)

PR #819, opened 2020-12-03
CI Replacement: use github actions instead of travis

wassertim:
This will replace obsolete travis with the modern and fancy github actions


.travis.yml
```diff
@@ -1,14 +0,0 @@
-language: node_js
-node_js:
-  - "lts/*"
-before_install:
-  - wget http://www.imagemagick.org/download/ImageMagick.tar.gz
-  - tar xf ImageMagick.tar.gz
-  - cd ImageMagick*
```
0xfe, line 7: Can you confirm that the regression tests run correctly? (i.e., ImageMagick is already installed in ubuntu-latest.)

wassertim: I don't know what is correct. It least the tests are all green and don't fail. You can see here in my fork:
https://github.com/wassertim/vexflow/pull/1/checks?check_run_id=1488734596

wassertim: Would be nice if you could give me a hint what exactly to check.

0xfe:
Thanks for the link, yep it's all good. The line "Success - All diffs under threshold!" is what we're looking for. :-)

Merging. Thanks for the change.

wassertim:
Yahoooo!!!! Finally it worked :)

PR #826, opened 2020-12-26
Put drawLedgerLines() within stavenotegroup

wassertim:
The call to drawLedgerLines method when put out of vf-stavenote group makes the drawn ledger lines to be not easy to remove from the stave. By keeping them together within the vf-stavenote group, one can remove a note and the corresponding ledgerline by doing just this:

`note.attrs.el.remove();`

### DIFF
--- a/src/beam.js
+++ b/src/beam.js
@@ -336,7 +336,6 @@ export class Beam extends Element {
       return uniqueTuplets;
     }
 
-
     // Using closures to store the variables throughout the various functions
     // IMO Keeps it this process lot cleaner - but not super consistent with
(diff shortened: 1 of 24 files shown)


0xfe:
Thanks, this is a good idea. Merging. (FYI, there's a bunch of unrelated files in this commit -- looks like they're just removing new lines.)

wassertim:
> FYI, there's a bunch of unrelated files in this commit -- looks like they're just removing new lines.

I removed thes LBs because eslint complained about them. I have no idea why it did not complain before on the CI side though.

PR #830, opened 2021-01-17
Fix preCalculateMinTotalWidth

deemaagog:


0xfe:
Thanks @deemaagog, can you confirm that the visual regression tests are all good?

sschmidTU:
Actually there seems to be a preexisting issue on the master branch, this is the master visual regression test:

Accidental.Cautionary_Accidental_Current
image

Ornament.Ornaments_Vertically_Shifted_diff
image

Same result for this PR.

Probably accidentally sneaked in during a previous commit.
Maybe make a binary search on which commit caused this.

sschmidTU:
I found the commit that added these issues to the regression tests:
39b4842c5daaceead48b901cae0af00da727a92f  from PR #817.
Before this commit, master didn't have the above diffs in the regression tests.

sschmidTU:
It's weird, the changes in #817 don't look like they could cause regressions,
but i've double checked, this merge caused the issues.

deemaagog:
@0xfe I ran visual regression tests and about 25 diff files were generated. 

image
image

What does that mean?

sschmidTU:
@deemaagog As I wrote above your comment, there were some preexisting issues on master causing most of the diffs, that were not caused by your pull request:
https://github.com/0xfe/vexflow/pull/830#issuecomment-762958005

0xfe:
Okay, merging this since it fixes an actual bug. I'll dig into those diffs separately.

0xfe:
Looks like `totalLeftPx` is already factored into `width`: Digging into context.getMetrics():

```
glyphPx: 11.934000000000001
leftDisplacedHeadPx: 0
modLeftPx: 12.892
modRightPx: 0
notePx: 11.934000000000001
rightDisplacedHeadPx: 0
totalLeftPx: 12.892
totalRightPx: 0
width: 24.826
```

Reverting this change to find the actual problem.

PR #831, opened 2021-01-21
Opt-out of reporting width for chord symbols

AaronDavidNewman:
When there are a few chord symbols and many notes, it is often desirable to ignore the width of the chord symbol and let the music flow normally.
With `reportWidth = true:` (default)
image

With `reportWidth = false:`
image

This change also uses the new TextFont class for text font metrics, vs reading it from the text fonts directly like I originally did.

### DIFF
--- a/src/chordsymbol.js
+++ b/src/chordsymbol.js
@@ -11,9 +11,8 @@
 import { Vex } from './vex';
 import { Flow } from './tables';
 import { Glyph } from './glyph';
+import { TextFont } from './textfont';
 import { Modifier } from './modifier';
-import { PetalumaScriptTextMetrics } from './fonts/petalumascript_textmetrics';
-import { RobotoSlabTextMetrics } from './fonts/robotoslab_textmetrics';
 
 // To enable logging for this class. Set `Vex.Flow.ChordSymbol.DEBUG` to `true`.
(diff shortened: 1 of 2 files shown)


0xfe:
Thanks Araron -- this looks fine to me, but seems like a bandaid. If the chord symbols report the correct width, justified over the center of the notehead, and are on a separate voice, then you should get the same result, no?

AaronDavidNewman:
It could be a band-aid but I think there's also a legit use case for chord symbols.  My example was not good, though.  

Sometimes you have justification of the notes, lyrics and chords all competing for space.  In that case I think you want to give space to the lyrics and space to format the notes, and it's OK for the chord to 'hang' over multiple notes because it applies to those notes, also:

This is with chord reporting width 0.  

image

The (b) of the Bb and the B dim can hang over the next note and it's fine because they apply to those notes, also.  And the notes are still well-proportioned.

If I have both annotations and chord reporting width I get this.  The measure takes up more space than it needs to.

image

My vex branch has similar logic for annotations.  I _do_ use that as a band-aid sometimes, because width estimates for annotations are not so good.  They don't use text metrics and so the estimates are too big.   In fact I'd like to address that next,  either by fixing estimates from annotations or creating a new text class that uses metrics in formatting, similar to what chords do.

0xfe:
Okay, yeah, those are good reasons. It would be good to try and figure out how to infer this stuff from context.

PR #838, opened 2021-02-19
chordsymbol factory fix, smufl fontgen skip missing glyphs

AaronDavidNewman:
Unrelated issues 822 and 837, but a small change so I included both.
822: syntax error in ChordSymbol factory.js methods.  I added factory calls to unit tests to verify they work.  
837: smufl_fontgen.js: when generating glyphs for a music font, only create glyphs for fonts present in the font file, so the font stack falls back to other fonts for missing glyphs


tools/smufl/smufl_fontgen.js
```diff
@@ -44,6 +44,8 @@ function toVFPath(glyph) {
     'y_min': bb.y1,
     'y_max': bb.y2,
     'ha': bb.y2 - bb.y1,
+    'leftSideBearing': glyph.leftSideBearing,
```
0xfe, line 47: This will significantly bloat the font files for non-chordsymbol glyphs, maybe we can parameterize the generation?

Also, to keep the files small, can we choose smaller names (e.g., `lsb` instead of `leftSideBearing`? (And add comments expanding on them.)

0xfe (changes requested):
Thanks, I like that there's kerning info in the fonts that we can use.

tools/smufl/smufl_fontgen.js
AaronDavidNewman: Those 2 lines are a mistake left over from early text font tool.  I can just remove them.  
Someday, I'd like to make a more general container class that can format glyphs and text in various ways, and then we can revisit where that font metrics information lives, but I don't think it's too useful for music engraving.

0xfe:
Looks good. Merging.

PR #843, opened 2021-03-02
add some formatter test cases

AaronDavidNewman:
Exhibit some problems with current formatter.  The first case is not really with the formatter, but I needed the accidental padding to be correct to test other alignment issues.

### DIFF
--- a/tests/formatter_tests.js
+++ b/tests/formatter_tests.js
@@ -157,6 +157,9 @@ VF.Test.Formatter = (function() {
       beams11.forEach(function(b) {
         b.setContext(ctx).draw();
       });
+      if (voice11.tickables[1].getX() >= voice21.tickables[1].getX()) {
+        console.warn('Second note of voice 1 is not to the right of the second note of voice 2');
+      }
       ok(true);
     },
     unalignedNoteDurations2: function(options) {
@@ -201,6 +204,10 @@ VF.Test.Formatter = (function() {
       stave2.setContext(context).draw();
       voice1.draw(context, stave1);
       voice2.draw(context, stave2);
+      if (voice1.tickables[1].getX() >= voice2.tickables[1].getX()) {
+        console.warn('Second note of voice 1 is not to the right of the second note of voice 2');
+      }
+
       ok(true);
     },
 
--- a/tests/formatter_tests.js
+++ b/tests/formatter_tests.js
@@ -157,7 +157,7 @@ VF.Test.Formatter = (function() {
       beams11.forEach(function(b) {
         b.setContext(ctx).draw();
       });
-      if (voice11.tickables[1].getX() >= voice21.tickables[1].getX()) {
+      if (voice11.tickables[1].getX() <= voice21.tickables[1].getX()) {
         console.warn('Second note of voice 1 is not to the right of the second note of voice 2');
       }
       ok(true);
@@ -204,7 +204,7 @@ VF.Test.Formatter = (function() {
       stave2.setContext(context).draw();
       voice1.draw(context, stave1);
       voice2.draw(context, stave2);
-      if (voice1.tickables[1].getX() >= voice2.tickables[1].getX()) {
+      if (voice1.tickables[1].getX() <= voice2.tickables[1].getX()) {
         console.warn('Second note of voice 1 is not to the right of the second note of voice 2');
       }
 
--- a/tests/formatter_tests.js
+++ b/tests/formatter_tests.js
@@ -11,7 +11,6 @@ VF.Test.Formatter = (function() {
     Start: function() {
       QUnit.module('Formatter');
       test('TickContext Building', Formatter.buildTickContexts);
-      runSVG('Accidental Padding', Formatter.formatAccidentalSpaces);
       runSVG('Justification and alignment with accidentals', Formatter.accidentalJustification);
       runSVG('Vertical alignment - few unaligned beats', Formatter.unalignedNoteDurations);
       runSVG('Vertical alignment - many unaligned beats', Formatter.unalignedNoteDurations2);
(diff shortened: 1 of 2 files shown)


tests/formatter_tests.js
```diff
@@ -79,6 +83,200 @@ VF.Test.Formatter = (function() {
...
+      stave21.setContext(ctx).draw();
+      voice11.draw(ctx, stave11);
+      voice21.draw(ctx, stave21);
+      beams21.forEach(function(b) {
+        b.setContext(ctx).draw();
+      });
+      beams11.forEach(function(b) {
+        b.setContext(ctx).draw();
+      });
+      ok(true);
```
rvilarl, line 234: I suggest to check the position substituting:
```js 
ok(true)
```
by:
```js
ok(voice1.tickables[1].getX() > voice2.tickables[1].getX(), 'Second note of voice 1 is not to the right of the second note of voice 2');
```

rvilarl (commented):
I have suggested to automatic tests

tests/formatter_tests.js
```diff
@@ -79,6 +83,200 @@ VF.Test.Formatter = (function() {
...
+      formatter.joinVoices([voice2]);
+      var width = formatter.preCalculateMinTotalWidth([voice1, voice2]);
+      formatter.format([voice1, voice2], width + 20);
+      var stave1 = new VF.Stave(10, 40, width + 30);
+      var stave2 = new VF.Stave(10, 100, width + 30);
+      stave1.setContext(context).draw();
+      stave2.setContext(context).draw();
+      voice1.draw(context, stave1);
+      voice2.draw(context, stave2);
+      ok(true);
```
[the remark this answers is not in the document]
0xfe: I like the suggestion, however, let's leave it as a log statement instead of a test, otherwise the failing test will break our build pipeline.

tests/formatter_tests.js
0xfe: Same comment as above.

tests/formatter_tests.js
```diff
@@ -79,6 +83,200 @@ VF.Test.Formatter = (function() {
       equal(tickables1[2].getX(), tickables2[2].getX(), 'Last notes of both voices have the same X');
       ok(tickables1[1].getX() < tickables2[1].getX(), 'Second note of voice 2 is to the right of the second note of voice 1');
     },
+    formatAccidentalSpaces: function(options) {
```
0xfe, line 86: Maybe we should move this to `accidentals_test.js`?

AaronDavidNewman: sure

tests/formatter_tests.js
rvilarl: @AaronDavidNewman the second note of voice11 should be to the right of second note of voice21. In the if condition checking the failure, it should be, I believe, <= and not >=.
```js
if (voice11.tickables[1].getX() <= voice21.tickables[1].getX())
```
@0xfe what is the prefered logging function?
@0xfe should the alignment between staves consider the value getWidth of the ticketables? I mean should the condition for the warning rather be:
```js
if (voice11.tickables[1].getX() < (voice21.tickables[1].getX() + voice21.tickables[1].getWidth()))
```

tests/formatter_tests.js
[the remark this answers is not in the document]
rvilarl: See comment as above

tests/formatter_tests.js
```diff
@@ -80,6 +83,134 @@ VF.Test.Formatter = (function() {
...
+      stave21.setContext(ctx).draw();
+      voice11.draw(ctx, stave11);
+      voice21.draw(ctx, stave21);
+      beams21.forEach(function(b) {
+        b.setContext(ctx).draw();
+      });
+      beams11.forEach(function(b) {
+        b.setContext(ctx).draw();
+      });
+      if (voice11.tickables[1].getX() >= voice21.tickables[1].getX()) {
```
rvilarl, line 160: @AaronDavidNewman the second note of voice11 should be to the right of second note of voice21. In the if condition checking the failure, it should be, I believe, <= and not >=.
```js
if (voice11.tickables[1].getX() <= voice21.tickables[1].getX())
```
@0xfe should the alignment between staves consider the value getWidth of the ticketables? I mean should the condition for the warning rather be:
```js
if (voice11.tickables[1].getX() < (voice21.tickables[1].getX() + voice21.tickables[1].getWidth()))
```

tests/formatter_tests.js
```diff
@@ -80,6 +83,134 @@ VF.Test.Formatter = (function() {
...
+        console.warn('Second note of voice 1 is not to the right of the second note of voice 2');
```
rvilarl, line 161: @0xfe what is the prefered logging function? I see the warning below and I did not see console.warn being used a lot in tests

tests/formatter_tests.js
```diff
@@ -80,6 +83,134 @@ VF.Test.Formatter = (function() {
...
+      formatter.joinVoices([voice2]);
+      var width = formatter.preCalculateMinTotalWidth([voice1, voice2]);
+      formatter.format([voice1, voice2], width + 20);
+      var stave1 = new VF.Stave(10, 40, width + 30);
+      var stave2 = new VF.Stave(10, 100, width + 30);
+      stave1.setContext(context).draw();
+      stave2.setContext(context).draw();
+      voice1.draw(context, stave1);
+      voice2.draw(context, stave2);
+      if (voice1.tickables[1].getX() >= voice2.tickables[1].getX()) {
```
rvilarl, line 207: @AaronDavidNewman here the second note of voice1 should also be to the right of second note of voice2. In the if condition checking the failure, it should be, I believe, <= and not >=.
```js
if (voice1.tickables[1].getX() <= voice2.tickables[1].getX())
```
depending on the answer from @0xfe prehaps:
```js
if (voice1.tickables[1].getX() < (voice2.tickables[1].getX() + voice2.tickables[1].getWidth()))
```

AaronDavidNewman: This is done.  I added the '=' because the X locations are currently the same (also why I didn't catch <> issue).

tests/formatter_tests.js
0xfe: `console.{log|warn}` is okay in tests. In `src/` it'll cause linter errors.

tests/formatter_tests.js
0xfe: No need to consider `getWidth()` if formatting is already done.

rvilarl (approved):
I have no more comments :+1:

0xfe:
Thanks folks. Merging.

Issue #848, opened 2021-03-10
Annotations widths are wrong


AaronDavidNewman:
Annotations assume each character is 7px wide, which is perfect if you are typing in a monospaced 10pt font using all caps :)  The result is usually too wide.

I'm going to be submitting a pull request for a change to use text metrics for annotation widths.   Here's an example, before:

![image](https://user-images.githubusercontent.com/5438280/110631516-37599180-816c-11eb-95c4-0b6e4729b9b2.png)

after:
![image](https://user-images.githubusercontent.com/5438280/110631564-46404400-816c-11eb-8438-31ca10dd4f80.png)

To my eye, these did not change any existing test cases, but all the same I'm wondering if we want to be able to opt out for people who depend on the existing width calculation. 

Also, I have metrics for Arial and Times fonts that I could include with the change (only the ~124 or so us-en ASCII glyphs) that I could include, if you don't think it will be too bloat-y.  If not, it will fall back to one of the existing fonts which will still be pretty close.

AaronDavidNewman:
This is merged, closing.

PR #849, opened 2021-03-10
Use text metrics to calculate annotation widths

AaronDavidNewman:
See my comments in issue #848 .  Use text metrics to calculate the correct width of annotatons for lyrics, etc.


src/annotation.js
```diff
@@ -66,8 +67,15 @@ export class Annotation extends Modifier {
...
     let width = 0;
     for (let i = 0; i < annotations.length; ++i) {
+      let testWidth = 0;
       const annotation = annotations[i];
-      width = Math.max(annotation.getWidth(), width);
+      const textFont = TextFont.getTextFontFromVexFontData({ family: annotation.font.family,
+        size: annotation.font.size, weight: 'normal' });
+      textFont.setFontSize(annotation.font.size);
+      for (let j = 0; j < annotation.text.length; ++j) {
+        testWidth += textFont.getWidthForCharacter(annotation.text[j]) * (72 / 96);
```
sschmidTU, line 76: why 72/96? this is 3/4. if you don't want to explain the magic number, you can just write * 0.75 ;)

sschmidTU: i'd at least explain where the number comes from in a comment, if not create named variables for the constituents.

AaronDavidNewman: Is it really just 3/4?  I feel silly :).  It's points to pixels, but really the font code should return pixels. And I don't like the way you have to set the font size after sending it in, let me clean this up a bit.

src/annotation.js
```diff
@@ -66,8 +67,15 @@ export class Annotation extends Modifier {
 
     let width = 0;
     for (let i = 0; i < annotations.length; ++i) {
```
0xfe, line 69: This could get very costly for large scores -- I think you could stick a cache in `TextFont` indexed by `hash(family+size+weight+text)`, which you can check before calculating the width.

src/annotation.js
0xfe: Yes, a comment here would be helpful.

tests/annotation_tests.js
```diff
@@ -18,8 +19,251 @@ VF.Test.Annotation = (function () {
...
+        stave.setContext(context).draw();
+        voice.draw(context, stave);
+      });
+      musicArray.forEach((music) => {
+        music.beamGroups.forEach((beam) => {
+          beam.setContext(context).draw();
+        });
+      });
+    },
+    lyrics: function(options) {
```
0xfe, line 94: This seems pretty verbose, why not use EasyScore instead (which is what I use for any new tests.)

src/annotation.js
AaronDavidNewman: great idea

tests/annotation_tests.js
AaronDavidNewman: Sure, is there a way to add Annotations?  Or should I just go `system.part.voices[0].tickables[i].addAnnotation...`.

0xfe: Yes, you can (by giving IDs to the notes):

```
      system = makeSystem(150);
      system.addStave({ voices: [voice(notes('D5/q[id="m2a"], G4, G4'))] });
      system.addStave({ voices: [voice(notes('B3/h.', { clef: 'bass' }))] });
      system.addConnector('singleRight');

      id('m2a').addModifier(0, vf.Articulation({ type: 'a.', position: 'above' }));
```

See `tests/bach_tests.js` for a full example.

src/annotation.js
AaronDavidNewman: 2 questions: 1) where is the distance between staves (10) defined?  I want to add a line to the annotation if the font height exceeds it.   Right now I added a constant in Annotation but I figure there's a better place. 2) Web fonts don't seem to work for canvas.  Is that expected?

AaronDavidNewman: OK, this is done.  Each glyph lookup is O(1) so it may not be such an improvement.

src/annotation.js
AaronDavidNewman: It was a mistake, copied from a smoosic calculation (which was also wrong).  I removed it.

AaronDavidNewman:
image

src/annotation.js
```diff
@@ -21,6 +22,11 @@ export class Annotation extends Modifier {
     return 'annotations';
   }
 
+  // Is there a better source for this?
+  static get distanceBetweenLines() {
```
0xfe, line 26: I'm surprised that there's no global constant here, I thought it was in `tables.js` or `stave.js`.

I suggest adding `STAVE_LINE_DISTANCE` into the top at `tables.js` and using that here, and also please update `stave.js` and maybe `tabstave.js` to use that constant.

src/textfont.js
```diff
@@ -118,12 +118,20 @@ export class TextFont {
...
+    return TextFont.fontTextHashInstance;
+  }
+
   // ### getTextFontFromVexFontData
   // Find the font that most closely matches the parameters from the given font data.
   // Primarily we look for font family, also bold and italic attributes.  This
   // method will always return a fallback font if there are no matches.
   static getTextFontFromVexFontData(fd) {
     let i = 0;
+    let rv = null;
```
0xfe, line 134: Can you use a more descriptive variable name here?

src/textfont.js
```diff
@@ -226,6 +246,18 @@ export class TextFont {
     return (metric.advanceWidth / this.resolution) * this.pointsToPixels;
   }
 
+  getWidthForString(s) {
+    const key = this.textHashBase + '-' + s;
```
0xfe, line 250: The key here can blow up memory. I suggest a two-level hash like `TextFont.widthCache[fontName][text]`.

src/textfont.js
```diff
@@ -226,6 +246,18 @@ export class TextFont {
...
+    let width = 0;
+    if (!TextFont.fontTextHash[key]) {
+      for (let j = 0; j < s.length; ++j) {
+        width += this.getWidthForCharacter(s[j]);
+      }
+      TextFont.fontTextHash[key] = width;
```
0xfe, line 256: I think the name `textWidthCache` is more descriptive here.

src/textfont.js
```diff
@@ -204,6 +217,13 @@ export class TextFont {
     if (!this.maxSizeGlyph) {
       this.maxSizeGlyph = 'H';
     }
+    this.weight = typeof this.weight === 'undefined' ? '' : this.weight;
+    this.style = typeof this.style === 'undefined' ? '' : this.style;
+    this.setHashBase();
+  }
+
+  setHashBase() {
```
0xfe, line 225: This method name reads like a setter. Maybe `initHashBase`?

0xfe (changes requested):
Thanks for the change. The examples look great! Some comments.

src/textfont.js
AaronDavidNewman: is a [n*m] hash less efficient in some way than 2 level [n][m]?  It seems like the memory usage would be the same.

0xfe: Yes, because you have huge key prefixes which are duplicated for each entry. With 2-level, you don't duplicate the 'font-size-weight' part of the key.

0xfe: An alternative would be to SHA160 hash the key, but I'm not sure that JS has a portable hashing library.

src/annotation.js
AaronDavidNewman: tabstave currently has `spacing_between_lines_px: 13`, you still want me to change it?

src/textfont.js
AaronDavidNewman: You are right!  Updated to use 2 level cache hash.

src/textfont.js
AaronDavidNewman: ok, I have `TextFont.textWidthCache[this.fontCacheKey] ` as the first level.

src/textfont.js
AaronDavidNewman: I called it 'update'...

src/annotation.js
AaronDavidNewman: Updated stave and annotation to use new constant.  I think tabstave change would break those test cases.

src/textfont.js
```diff
@@ -204,6 +219,15 @@ export class TextFont {
     if (!this.maxSizeGlyph) {
       this.maxSizeGlyph = 'H';
     }
+    this.weight = typeof this.weight === 'undefined' ? '' : this.weight;
+    this.style = typeof this.style === 'undefined' ? '' : this.style;
+    this.updateCacheKey();
+  }
+
+  // ### updateCacheKey
```
0xfe, line 227: nit: please remove this line

0xfe (approved):
Looks great! Thanks. One minor comment before merging.

src/textfont.js
AaronDavidNewman: Line 226?  I removed it.

0xfe:
Did the build break?

AaronDavidNewman:
I had to merge in the TS changes.  It builds now and the unit test cases pass.  It looks like you have to do `npm install` after merging in the latest changes (not sure if it's .ts or eslint changes that caused this).

mscuthbert:
> I had to merge in the TS changes. It builds now and the unit test cases pass. It looks like you have to do `npm install` after merging in the latest changes (not sure if it's .ts or eslint changes that caused this).

Both, almost certainly.

0xfe:
Great, merging.

Issue #850, opened 2021-03-13
Vexflow Typescript File by File Migration


wassertim:
After @wassertim and @rvilarl had a discussion we propose to migrate to typescript file by file, providing support for js and ts simultaneously.

First PR will contain infrastructure for ts and eslint.

And then we will create many PRs with ts files substituting js files.

@0xfe

0xfe:
Good plan. Much easier to review -- we can start with small chunks, and as we gain confidence, start adding bigger chunks to the PRs.

As part of the build infrastructure (first commit): add `src/dummy.ts` and confirm that it lints, and builds.

After the build infrastructure, lets do the test infrastructure. And add a `tests/dummy_tests.ts` that just does `ok(true)`.

Once that's good, we can try some independent source files (music.js, tuning.js, or fraction.js), one by one and use those as strawmen to decide on style, etc. To start, I suggest `src/music.js` and `tests/music_tests.js`. 

Please also tag @mscuthbert in the reviews.

rvilarl:
@0xfe the plan is to add the infractstructure to the current project, see that all the tests pass. And then on a second to change one file form js to ts.

0xfe:
Thanks @rvilarl -- that sounds good, just also requesting that the first PR includes a `dummy.ts` file or something (which we can remove later) so we know that the build/lint infrastructure does what it's supposed to do. Otherwise, the only signal we get is that tests pass, not that the build infrastructure works.

rvilarl:
@0xfe where do you want us to make the PR, on master? If you want to check that the infracstructure is running ok, I suggest to change element.js to element.ts in order to make it participant on the complete set of tests.

0xfe:
Yes, `master` is good, then we won't have to worry about keeping branches in sync. I don't think `element.js` is a good first class because it has a huge downstream effect. If you don't want to create a `dummy.ts`, then try migrating `music.js` or `tuning.js` which are less invasive?

rvilarl:
You are right. It is very invasive but proving that element with minor changes pass all the tests provides a lot of confidence. And being an abstract class we can be almost sure that we do not affect the client applications. Wouldn't you agree?

rvilarl:
@0xfe I am enjoying the experience a lot and I have two questions:
- does the package @types/vexflow belong to this repo? I would like to open issues for the topics above.
[...]

PR #856, opened 2021-03-22
Typescript: babel removal

rvilarl:
as described in #850


0xfe:
Thanks folks. This has more than airbnb + babel.

Also, not clear why you made `Music` static? This breaks the interface for external users, and prevents you from swapping it out in VexFlow.

rvilarl:
@0xfe the changes are related to **airbnc rules** (some examples):

**class-methods-use-this** does not allow to have static functions declared.
**no-param-reassign** does not allowed to change the value of a parameter within the function. ‘direction = 1’
**no-plusplus** does not allow to use this operator ‘i++’
…

The changes were minimal to pass airbnc. **I suggest we drop thr idea, to include airbnb.** I will leave it to eslint:recommended minimal rules as the rest.

**I will update the PR just to remove babel**.

rvilarl:
@0xfe I suggest to leave the change of Runtimeerror extending Error. Let me know if you agree with that.
I have left the configuration divided on TS files, JS files in src and JS files in folder tests, just in case we want to make a change in the future.

package.json
```diff
@@ -17,13 +17,8 @@
     "url": "https://github.com/0xfe/vexflow/issues"
   },
   "devDependencies": {
-    "@babel/core": "^7.8.4",
-    "@babel/plugin-transform-object-assign": "^7.8.3",
-    "@babel/preset-env": "^7.8.4",
```
mscuthbert, line 22: This might be a question from ignorance, but Isn't it premature to remove babel while there is still so much source in JS?  Does Typescript on JS sources really do all of the transformations that are needed?  If so, then it's a step I can remove in many of my sources.  But if we're still building to ES5 we will probably need the core.js features.

mscuthbert (commented):
Two questions that I'm unclear about on removing babel on a mixed JS/TS repository.

Gruntfile.js
```diff
@@ -175,8 +162,8 @@ module.exports = (grunt) => {
   grunt.loadNpmTasks('grunt-webpack');
 
   // Default task(s).
-  grunt.registerTask('default', ['eslint', 'webpack:buildDev', 'webpack:build', 'concat', 'docco']);
-  grunt.registerTask('test', 'Run qunit tests.', ['webpack:buildDev', 'concat', 'qunit']);
+  grunt.registerTask('default', ['clean', 'eslint', 'webpack:buildDev', 'webpack:build', 'concat', 'docco']);
+  grunt.registerTask('test', 'Run qunit tests.', ['clean', 'webpack:buildDev', 'concat', 'qunit']);
```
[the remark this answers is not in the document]
rvilarl: **clean** does not delete the releases folder, it deletes the build folder. With it we make sure that if the process is not properly finished, there are no remaining files from a previous build. I would not test / build on a live webpage there are many errors in the process that can impact on the availability.
This is just a suggestion anyway, it can be removed.

package.json
rvilarl: ts_loader is now compiling/transforming all the files
it is configured with javascript support and ES5 as target
https://github.com/0xfe/vexflow/blob/1924b2cc268de33fb19bbc8ba62c59b635614ea4/tsconfig.json#L1-L5

rvilarl:
@0xfe @mscuthbert @wassertim I have realised that we could still add `airbnb` and change the rules that imply a change in the interface from `error` to `warn`. With the one below the chnage to static is no longer required.

`"extends": ["eslint-config-airbnb-typescript/base"...`
` "rules": ["class-methods-use-this": "warn"...`

if you agree I will push this change.

package.json
0xfe: This is a bit scary to me too, but a good change if it does what it's supposed to do. Are you running the full visual regression tests?

0xfe:
I don't have a strong preference here, I would prefer `airbnb` with a small number of overrides. It's mostly style, since the TS compiler does a good job w.r.t. strictness, etc.

@rvilarl To directly answer your question, yes sure, let's do `airbnb` with some overriding rules.

Gruntfile.js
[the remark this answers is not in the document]
wassertim: I think clean makes sense here. This is how you would do in java with maven: `mvn clean package`, which also removes target folder.

wassertim:
We have to fix the CI tests

Gruntfile.js
```diff
@@ -42,20 +42,7 @@ module.exports = (grunt) => {
...
-                options: {
-                  presets: [preset],
-                  plugins: ['@babel/plugin-transform-object-assign'],
-                },
-              },
-            ],
-          },
-          {
-            test: /\.ts?$/,
+            test: /\.?s?$/,
```
wassertim, line 45: so this will also include css, scss ... do we need such generic regex?

wassertim (changes requested):

package.json
rvilarl: @0xfe  how do you that? I am doing grunt test & npm start.

0xfe: `npm test` will do it. See https://github.com/0xfe/vexflow/wiki/Visual-Regression-Tests.

rvilarl: Done, I get: **Success - All diffs under threshold!**

Gruntfile.js
rvilarl: Done changed to `test: /(\.ts?$|\.js?$)/`

rvilarl:
> We have to fix the CI tests

@wassertim how do we do that?

rvilarl:
@0xfe @mscuthbert @wassertim Wow! CI Test also fixed! I believe everything is done!
Let us merge it and continue to progress on the migration. Thanks for your comments.

package.json
```diff
@@ -63,7 +63,7 @@
     "generate:blessed": "node ./tools/generate_png_images.js ../releases ./build/images/blessed",
     "generate": "npm run generate:current && npm run generate:blessed",
     "diff": "./tools/visual_regression.sh",
-    "test": "npm run lint && npm run qunit && npm run generate && npm run diff"
+    "test": "npm run start && npm run generate && npm run diff"
```
wassertim, line 66: I'm not sure this is a way to go, since `npm start` will copy files from build to release it was supposed to be checked in. By doing this you are not longer forcing to check in releases.

wassertim (changes requested):

wassertim:
> > We have to fix the CI tests
> 
> @wassertim how do we do that?

I suppose you just have to check in compiled releases here

src/music.ts
```diff
@@ -202,11 +202,11 @@ export class Music {
 
   getNoteParts(noteString: string): NoteParts {
     if (!noteString || noteString.length < 1) {
-      throw new Vex.RERR('BadArguments', 'Invalid note name: ' + noteString);
+      throw new Vex.RERR('BadArguments', `Invalid note name: ${noteString}`);
```
wassertim, line 205: How does this relate to babel removal? This PR is too many things

rvilarl: All the changes in Music relate to Airbnb. They are done automatically by eslint.

rvilarl:
> > > We have to fix the CI tests
> > 
> > 
> > @wassertim how do we do that?
> 
> I suppose you just have to check in compiled releases here

You are right npm test requires compiled releases in the releases directory. I change package.json to make npm test generate releases.

package.json
rvilarl: I would suggest just checkin the releases on the formal tagged builds.

rvilarl: In order to achieve that we have to produce the releases to pass the CI tests. @0xfe  what do you think?

wassertim: should we do it now in this PR? It is about to remove babel

rvilarl: @0xfe you were concerned about making sure that the visual tests were passing. I have realised that there are merges where the tests were not running ie.: 1924b2cc268de33fb19bbc8ba62c59b635614ea4

Now we have the following actions:
`'clean', 'eslint', 'webpack:buildDev', 'webpack:build', 'concat', 'docco', 'qunit', 'copy:release', 'generate', 'diff'`

in comparison with:
`'clean', 'webpack:buildDev', 'concat', 'qunit'`

with this approach we also get the eslint warnings displayed in the code:
image

@0xfe do you like it or you prefer me to revert the error as suggested by @wassertim? With your answer I will proceed and resolve this topic. I can also remove some of the steps if you prefer.

0xfe: I don't recommend changing this line either -- or we're making too many changes. Let's set `class-methods-use-this` to `off` for now.

src/music.ts
0xfe: eslint fixes are okay with me -- I've rarely found them to break things.

package.json
0xfe: Actually, this is only a problem in a small number of files. Let's instead add a comment to the top of `music.ts` disabling the rule just for the file.

rvilarl: > @0xfe you were concerned about making sure that the visual tests were passing. I have realised that there are merges where the tests were not running ie.: [1924b2c](https://github.com/0xfe/vexflow/commit/1924b2cc268de33fb19bbc8ba62c59b635614ea4)
> 
> Now we have the following actions:
> `'clean', 'eslint', 'webpack:buildDev', 'webpack:build', 'concat', 'docco', 'qunit', 'copy:release', 'generate', 'diff'`
> 
> in comparison with:
> `'clean', 'webpack:buildDev', 'concat', 'qunit'`
> 
@0xfe what about the change in package.json?

rvilarl: @wassertim hightlighted that this alreay answer "I don't recommend changing this line either -- or we're making too many changes."
I will make the comment in music.ts roll back package and checking the releases dir

src/music.ts
```diff
@@ -331,11 +329,11 @@ export class Music {
 
     let relativeNoteName = parts.root;
     if (interval > 0) {
-      for (let i = 1; i <= interval; ++i) {
+      for (let i = 1; i <= interval; i += 1) {
```
wassertim, line 334: why changing this?

src/music.ts
```diff
@@ -368,17 +366,16 @@ export class Music {
    * E.g., Given the scale C, and the note E, returns M3
    */
   getIntervalBetween(note1: number, note2: number, direction?: number): number {
-    if (direction == null) direction = 1;
+    const dir = direction == null ? 1 : direction;
```
wassertim, line 371: btw, dir is not a good name. If you don't look up to this line you ask yourself ist a directory or maybe one of these: https://www.thefreedictionary.com/words-that-start-with-dir?

wassertim (changes requested):

wassertim:
I have just only one remark, could we maybe just remove babel here? And introduce airbnb linting in another PR? These seem to be completely unrelated things for me.

src/music.ts
rvilarl: aribnb rule no use of ++ or --

src/music.ts
rvilarl: airbnb rule do not modify parameters

src/music.ts
wassertim: Ok, they have their conventions. But why we should follow exactly this one?

src/music.ts
wassertim: I'm commenting more on name here.

rvilarl:
for completeness the following **airbnb rules** affect this file:
1) **class-methods-use-this** it has been disabled by a comment to avoid the change in interface with `static`
2) **prefer-template** reason for the changes like `'Invalid note name: ${noteString}'`
3) **no-param-reassign** reason for `const dir = direction == null ? 1 : direction;`
4) **no-plusplus** reason for `i +=1`  or `i -=  1`
5) **@typescript-eslint/naming-convention** reason for `reverserInterval`

rvilarl:
@wassertim they are not unrelated it is setting the basis for the infracstructure typescript with airbnb style

src/music.ts
rvilarl: I am happy to change dir if  you provide a proposal

wassertim:
> @wassertim they are not unrelated it is setting the basis for the infracstructure typescript with airbnb style

Ok with babel, everyone agrees it has to be deleted. But airbnb linting raises a lot of question. Why mix them in on PR?

src/music.ts
wassertim: for simplicity sake you could rename the parameter to directionParam or something, but the const make direction. So you don't have to change so many lines after

rvilarl:
My oppinion is that we should decide now how we want to proceed. We will not see many opportunities like this one where all the files have to be updated.
Mi preference would be to leave it on and on a file basis decide is any of the rules have to be disabled with a comment like we already did for `static`
@0xfe how would you like to proceed?

0xfe:
Okay, I think @wassertim is right -- this change is way to muddy. Let's just do babel right now.

Re: airbnb, maybe we shouldn't use that at all, I like the `++` and and `?` -- they make the code very readable, especially in the busier sections. Maybe just the typescript recommended is okay?

Either way, let's just update babel in this PR, and then decide the eslint immediately after. @rvilarl I'm sure this is frustrating, and I appreciate the effort -- this is just the unfortunate nature of managing risk in such a fundamental change. :-)

rvilarl:
@0xfe @wassertim I think it is a pitty but I do accept it. There are reasons behind those rules and we could have conciously disabled them in a per file basis like I did for the static topic with:
`/* eslint-disable class-methods-use-this */`
or globally in the .eslintrc

As an example for the reasons:
https://eslint.org/docs/rules/no-plusplus
https://eslint.org/docs/rules/no-param-reassign

src/music.ts
rvilarl: That was a good idea...

src/music.ts
rvilarl: see answer below, the reasons are documented

src/vex.js
```diff
@@ -30,14 +30,10 @@ Vex.MakeException = (name) => {
 };
 
 // Default runtime exception.
-class RuntimeError {
+class RuntimeError extends Error {
```
wassertim, line 33: You just forgot to revert it or it has some reason to stay?

Gruntfile.js
```diff
@@ -94,7 +77,7 @@ module.exports = (grunt) => {
       },
     },
     eslint: {
-      target: SOURCES.concat('./tests'),
+      target: SOURCES.concat('./tests/**/*.?s'),
```
wassertim, line 80: too generic regex

Gruntfile.js
```diff
@@ -36,40 +36,23 @@ module.exports = (grunt) => {
         libraryExport: 'default',
       },
       resolve: {
-        extensions: ['.ts', '.js', '.json']
+        extensions: ['.ts', '.js', '.json'],
```
wassertim, line 39: comma is not needed

wassertim (changes requested):

src/music.ts
wassertim: Not really, this one is just a default value, which can be done more elegantly in typescript: https://www.typescripttutorial.net/typescript-tutorial/typescript-default-parameters/

wassertim:
> @0xfe @wassertim I think it is a pitty but I do accept it. There are reasons behind those rules and we could have conciously disabled them in a per file basis like I did for the static topic with:
> `/* eslint-disable class-methods-use-this */`
> or globally in the .eslintrc
> 
> As an example for the reasons:
> https://eslint.org/docs/rules/no-plusplus
> https://eslint.org/docs/rules/no-param-reassign
```js
i
++
j
```
is not really a good reason to disable ++ and ---
Typescript compiler will force you to put semicolon anyway.

src/vex.js
rvilarl: "It is considered good practice to only throw the Error object itself or an object using the Error object as base objects for user-defined exceptions. **The fundamental benefit of Error objects is that they automatically keep track of where they were built and originated.**"
It helps future debug (I consider that part of the infraesctructure) and does not impact on the external interface.

Gruntfile.js
rvilarl: Please propose a regex

rvilarl: Please note that the original one was leaving the content of the tests directory unchecked.

rvilarl:
> > @0xfe @wassertim I think it is a pitty but I do accept it. There are reasons behind those rules and we could have conciously disabled them in a per file basis like I did for the static topic with:
> > `/* eslint-disable class-methods-use-this */`
> > or globally in the .eslintrc
> > As an example for the reasons:
> > https://eslint.org/docs/rules/no-plusplus
> > https://eslint.org/docs/rules/no-param-reassign
> 
> ```js
> i
> ++
> j
> ```
> 
> is not really a good reason to disable ++ and ---
> Typescript compiler will force you to put semicolon anyway.

@wassertim and do you agree with?

_"Assignment to variables declared as function parameters can be misleading and lead to confusing behavior, as modifying function parameters will also mutate the arguments object. Often, assignment to function parameters is unintended and indicative of a mistake or programmer error."_

0xfe:
Hi folks -- there's an expectation of reasonableness here :-) That example is quite contrived and I don't think we should be disabling (IMO) useful syntax like `++` -- I'm okay with a small risk of error for the benefit of certain kinds of terseness.

`no-param-reassign` is fine, but overall let's move style to a separate discussion. For now, let's just get removing babel working with the minimal amount of changes.

Gruntfile.js
0xfe: I don't think this is a regex, this is some file globbing syntax. How about: `./tests/**/*.(js|ts)`

src/vex.js
0xfe: This is good, let it stay.

Gruntfile.js
0xfe: If that doesn't work, then this existing one is fine -- but good to try and find one that is more tightly scoped.

0xfe:
Overall, this looks a lot better. Can you verify that the regression tests run, and remove `releases/*` from this PR please?

rvilarl:
@0xfe the is regression tests run, if I remove the releases we will get here "all checks have passed" red. Perhaps you can assist on the open questions:
- The comma I will delete
- Runtimeerror extending Error, background in the conversation above
- regex to use on tests directory, background in the conversation above

I would love to get this one closed, I have already done 8 commits going back and forward :)

Gruntfile.js
rvilarl: You are right it is not a regex, it is a list. Both .js and .ts added to the list

Gruntfile.js
rvilarl: I remove it to reduce changes, however all the other instances in the file have the comma.

rvilarl:
@0xfe @wassertim this should now be ready! Please confirm :)

package.json
mscuthbert: Oh, I understand why I was confused -- I've never used typescript alone to target a specific ECMAScript target but generally used TypeScript with babel/browserlist/core-js to ensure that I'm outputting to a level of compliance that allows using features like WeakMap and native classes (for custom elements); if we're going to pure ES5 then there's no need.

mscuthbert (approved):
This looks good to me.

wassertim (approved):
Great work @rvilarl!

0xfe:
Thanks everyone. Merging! Good job, @rivalri.

PR #857, opened 2021-03-23
fix new keyword mess

wassertim:
This PR fixes new keyword mess:
```typescript
// new keyword is not needed. contextBuilder is just a function, not a constructor
var ctx = contextBuilder(options.elementId, 600, 200); 

// new keyword is not needed. Formatter() is just a factory method in Factory class, not a constructor
var vf = VF.Test.makeFactory(options, 300);
vf.Formatter().joinVoices([voice, voice2]).formatToStave([voice, voice2], stave); 
```
@0xfe I ask for a quick review. It is a really straightforward and small PR, but it will help me to continue my work on ts migration

### DIFF
--- a/tests/gracenote_tests.js
+++ b/tests/gracenote_tests.js
@@ -175,7 +175,7 @@ VF.Test.GraceNote = (function () {
       voice.addTickables(createNoteBlock(['g/4'], 1));
       voice.addTickables(createNoteBlock(['d/5'], -1));
 
-      new vf.Formatter().joinVoices([voice]).formatToStave([voice], stave);
+      vf.Formatter().joinVoices([voice]).formatToStave([voice], stave);
 
       vf.draw();
 
@@ -220,7 +220,7 @@ VF.Test.GraceNote = (function () {
       voice.addTickables(createBeamdNoteBlock(['g/4'], 1, beams));
       voice.addTickables(createBeamdNoteBlock(['d/5'], -1, beams));
 
-      new vf.Formatter().joinVoices([voice]).formatToStave([voice], stave);
+      vf.Formatter().joinVoices([voice]).formatToStave([voice], stave);
 
       vf.draw();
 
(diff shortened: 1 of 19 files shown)


tests/gracenote_tests.js
```diff
@@ -275,7 +275,7 @@ VF.Test.GraceNote = (function () {
       voice.addTickables(createNoteBlock(['d/4', 'a/4'], 1));
       voice.addTickables(createNoteBlock(['d/4', 'a/4'], -1));
 
-      new vf.Formatter().joinVoices([voice]).formatToStave([voice], stave);
+      vf.Formatter().joinVoices([voice]).formatToStave([voice], stave);
```
rvilarl, line 278: I think that this one requires new. I have included this call in the migrated music.ts and I get the following

image

On this test I have also realised that the line 274 above also requires new.

image

rvilarl (changes requested):

tests/gracenote_tests.js
sschmidTU: yeah, shouldn't running all tests tell you all the errors? (at least if there are none)

wassertim: Guys, please read the comment to this PR. vf is an object of class Factory, which has Formatter method. So it does not require new

wassertim:
```typescript
var vf = VF.Test.makeFactory(options, 300);
vf.Formatter().joinVoices([voice, voice2]).formatToStave([voice, voice2], stave); 
```

wassertim:
```typescript
export class Factory {
  Formatter(options) {
    return new Formatter(options);
  }
}
```

rvilarl: you are right @wassertim it is a factory class, please ignore my comment.

rvilarl (approved):

wassertim:
@0xfe would be nice if you merge this. This fixes code mistakes with `new` keyword

0xfe:
Wow, surprised that all this even works! :-)

Thanks for fixing.

PR #861, opened 2021-03-28
typescript: tuning & fraction

rvilarl:
Migration of tuning and fraction


rvilarl:
@0xfe @wassertim @mscuthbert the next two classes are there!

I gave it a try at copy:releases in Gruntfile.js, if it does not fix the tests I will remove it.

src/fraction.ts
```diff
@@ -32,39 +41,44 @@ export class Fraction {
...
     }
+    const arg0 = args[0];
+    args.shift();
+    return Fraction.LCM(arg0, Fraction.LCMM(args));
   }
 
-  constructor(numerator, denominator) {
-    this.set(numerator, denominator);
+  constructor(numerator?: number, denominator?: number) {
+    this.numerator = numerator === undefined ? 1 : numerator;
```
0xfe, line 67: Why not call `this.set` instead of duplicating code?

src/fraction.ts
```diff
@@ -6,14 +6,23 @@
...
-/* eslint-disable no-underscore-dangle */
-
 import { Vex } from './vex';
+
 export class Fraction {
+  numerator: number;
+
+  denominator: number;
+
+  private static __compareA = new Fraction();
```
0xfe, line 16: Can you add a comment here: `Cached objects for comparisons`

src/fraction.ts
rvilarl: To avoid eslint complaining about not initialising member variables in the constructor. I also do not like the duplication.

src/fraction.ts
rvilarl: Ok.
 
Would you change the names to remove the `__`? `fractionCompareA` for example rather than `__compareA`

src/fraction.ts
mscuthbert: Can't we instead use:

```
numerator: number = 1;
denominator: number = 1;

constructor(numerator?: number, denominator?: number) {
     this.set(numerator, denominator);
}

set(numerator: number = 1, denominator: number = 1) {
    this.numerator = numerator;
    this.denominator = denominator;
}
```

and if eslint is complaining, let's disable the check?  It seems like we're using a code paradigm that no one likes for the sake of an automated program.  My colleagues use paradigms like this suggested one everywhere and we've never ran into a problem.

src/fraction.ts
```diff
@@ -134,9 +150,10 @@ export class Fraction {
     const u = this.numerator * a - otherNumerator * b;
     return this.set(u, lcm);
   }
-  multiply(param1, param2) {
-    let otherNumerator;
-    let otherDenominator;
+
+  multiply(param1?: Fraction | number, param2?: number): this {
```
mscuthbert, line 154: Can these be given default values of 1 and 1 also?  I feel it just helps document what happens if they're omitted.

src/tables.js
```diff
@@ -30,6 +30,8 @@ const Flow = {
 
   /* Kerning (DEPRECATED) */
   IsKerned: true,
+  keyProperties: keyProperties,
+  integerToNote: integerToNote,
```
mscuthbert, line 34: This is good for now, but I'd suggest that we tackle tables.js => tables.ts (or maybe finally the time to rename to flow.ts -- I always forget where Flow is?) and make them `export function...` for each of the properties with a `export default const Flow = { }` at the end?

src/tuning.ts
```diff
@@ -5,9 +5,14 @@
...
 import { Vex } from './vex';
 import { Flow } from './tables';
-
 export class Tuning {
-  static get names() {
+  private numStrings: number;
+
+  private tuningString: string;
+
+  private tuningValues: number[];
```
mscuthbert, line 13: can we please use `protected` over `private` so that developers can subclass Tuning to get at these attributes?

src/tuning.ts
```diff
@@ -19,59 +24,63 @@ export class Tuning {
...
-      this.tuningValues[i] = this.noteToInteger(keys[i]);
+    for (let i = 0; i < this.numStrings; i += 1) {
+      this.tuningValues[i] = Tuning.noteToInteger(keys[i]);
     }
   }
 
-  getValueForString(stringNum) {
-    const s = parseInt(stringNum, 10);
+  getValueForString(stringNum: string | number): number {
+    const s = typeof stringNum === 'string' ? parseInt(stringNum, 10) : stringNum;
```
mscuthbert, line 59: checked that `parseInt()` on possible int only occurs about 15 times in the code, so no need for a helper function, it won't be everywhere.

src/tuning.ts
```diff
@@ -19,59 +24,63 @@ export class Tuning {
...
 
-  getNoteForFret(fretNum, stringNum) {
+  getNoteForFret(fretNum: string | number, stringNum: string | number): string {
     const noteValue = this.getValueForFret(fretNum, stringNum);
 
     const octave = Math.floor(noteValue / 12);
     const value = noteValue % 12;
 
-    return Flow.integerToNote(value) + '/' + octave;
+    return `${Flow.integerToNote(value)}/${octave}`;
```
mscuthbert, line 84: Here's where I much prefer the clarity of `+` and strings -- we add six extra `${}` control characters to get rid of the space-separated elegance on seeing exactly what is concatenated, what is a function call, and what is not.  Even `return Flow.integerToNote(value) + '/' + octave.toString();` seems clearer to me

+1 on disabling this lint rule.

mscuthbert (commented):
Looks good -- a few code comments, mostly for community discussion as we start doing more with typescript + the new linter at the same time.

My overall feeling is that the new linter is encouraging as many new bad habits as it's removing old ones.

src/fraction.ts
rvilarl: @mscuthbert very good idea, much clearer

src/fraction.ts
rvilarl: @mscuthbert any proposal on static variable naming convention? This `__` is unique in the program. Would you change to `fractionCompareA`?

src/tables.js
rvilarl: I agree that Flow location is not obvious. I consider flow.ts much better but it would change the interface to apps using the code.

We will see a change here anyway if we port `tables` because TS does not like nested definitions:
`Flow.keyProperties` with a function
`Flow.keyProperties.note_values` with the values.

src/tuning.ts
rvilarl: agree

rvilarl: @mscuthbert @0xfe both `private` and `protected` change the interface anyway. Should  we leave them `public`?

src/tuning.ts
rvilarl: @mscuthbert @0xfe this is an interface change. Should we do it?

src/tuning.ts
rvilarl: This one is a style topic. In the music.ts I believe that @0xfe said that he was fine with it.

Personally I like it, it is compacter and I do not like the `+` on cases with multiple concatenations.
Setting the rule also ensures that only templates will be used and not a mix of both approaches.

@0xfe how would you like to proceed?

src/fraction.ts
rvilarl: I agree it also simplifies the code.

src/tuning.ts
rvilarl: My preference is as @mscuthbert to make them protected. If we go this way, I would also: 
- delete `tuningStrings`which is not used
- delete `numStrings` whose content is not obvious and use `tuningValues.length` instead

src/tuning.ts
rvilarl: I agree that this is used with string nowhere. Probably the best is to change the parameter to number. All those checks also consume processing time.

src/fraction.ts
```diff
@@ -134,127 +132,111 @@ export class Fraction {
...
-  // Returns the fraction component when reduced to a mixed number
-  fraction() {
+  //** Returns the fraction component when reduced to a mixed number */
+  fraction(): number {
     return this.numerator % this.denominator;
   }
 
-  // Returns the absolute value
-  abs() {
+  //** Returns the absolute value */
```
rvilarl, line 221: This comment is misleading. It does not only return the absolute value, it modifies the object as well.
What is right the comment or the implementation?

src/tuning.ts
0xfe: I think `protected` is good. I'd prefer not to change any implementation details in this PR.

src/tuning.ts
0xfe: Although I see what Michael is saying for this instance, I prefer this approach overall, because templates are easier to read, and `+` gets really messy and can mask bugs. So with the intent of keeping the style consistent (within reason), let's leave the template here.

src/fraction.ts
rvilarl: The `__` contradicts the camel naming convention. I like your suggestion to add `static`, but this one I believe `__` can go away. Would you agree?

src/fraction.ts
rvilarl: This comment has raised the issue. I suggest to add comments to the functions without them. @0xfe @mscuthbert I will start with the obvious ones but I will need your help on those for `tuning`.

src/fraction.ts
rvilarl: I was wrong, it does not contradict camelCase. They are ignored at the beginning.
https://eslint.org/docs/rules/camelcase

I will leave them.

src/tables.js
rvilarl: This is not a `eslint` it is a `tsc` import topic. Import at compilation time will only recognize the elements declared as part of the structure, not the ones added on run-time.

I suggest to leave it.

src/tuning.ts
rvilarl: @0xfe ok, how should I proceed with the other topics. One issue collecting all the issues found in the migration or one issue per topic?

image

src/fraction.ts
rvilarl: @0xfe in reality what was wrong is the comment, not the implementation. I will leave a note in the code to discuss about it. This abs function is only used in this file.

src/tables.js
rvilarl: I agree with @mscuthbert this is good for now

src/fraction.ts
0xfe: Right, it's just unexpected behavior -- I'd suggest renaming to `makeAbs` and updating references.

src/tables.js
0xfe: okay

src/tuning.ts
rvilarl: ok

src/fraction.ts
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  /** Returns string representation in mixed form */
+  toMixedString(): string {
+    let s = '';
+    const q = this.quotient();
+    const f = Fraction.__staticFractionTmp.copy(this);
+
+    if (q < 0) {
+      f.makeAbs().fraction();
```
rvilarl, line 253: @0xfe this is the only location where abs was called. I do not understand why fraction is called, do you? The same applies to line 255.

I do not understand why this is now presented as a new file and not a renamed one.

src/tuning.ts
```diff
@@ -19,59 +24,63 @@ export class Tuning {
...
   constructor(tuningString = 'E/5,B/4,G/4,D/4,A/3,E/3,B/2,E/2') {
     // Default to standard tuning.
+    this.tuningString = '';
+    this.tuningValues = [];
+    this.numStrings = 0;
     this.setTuning(tuningString);
   }
 
-  noteToInteger(noteString) {
+  static noteToInteger(noteString: string): number {
```
rvilarl, line 33: @0xfe @mscuthbert proposals on comments of functions in this file is appreciated.

src/tuning.ts
mscuthbert: Sorry, I didn't mean to remove "| number" from parameter type.

mscuthbert: I just was wondering if we needed something like this to be written once so that we don't need to rewrite it every time it appears in the code:

```
export function possibleNumberAsString(stringOrNum: string | number): number {
    return typeof stringNum === 'string' ? parseInt(stringNum, 10) : stringNum;
}

const s = possibleNumberAsString(stringNum);
```

but the particular `parseInt(stringNum, 10)` code only appears about 12 times in the entire codebase, so we wouldn't need it too often.  (but it would make it clearer what the `const s =` declaration is doing.

src/tuning.ts
mscuthbert: sounds good.  Will keep that in mind.

src/fraction.ts
mscuthbert: I'm afraid I don't know the ins and outs of the fraction module to know why it calls fraction.  The guts of this file were stable since before I started using Vexflow, so I've never dived deeply into it.

> I do not understand why this is now presented as a new file and not a renamed one.

This happens far too often on javscript -> typescript conversions I've found.  When I remembered in my own work I tried to make the first commit just the rename with nothing else.  Even a few changes after the rename tends to make git have a hard time figuring it out.

src/fraction.ts
```diff
@@ -0,0 +1,281 @@
...
+    return this.set(copy.numerator, copy.denominator);
+  }
+
+  /** Returns the integer component eg. (5/2) == 2 */
+  quotient(): number {
+    return Math.floor(this.numerator / this.denominator);
+  }
+
+  /** Returns the fraction component when reduced to a mixed number */
+  fraction(): number {
```
0xfe, line 225: This should be renamed to `remainder`

src/fraction.ts
0xfe: `fraction()` should be renamed to `remainder()`, but even so, I think the call is unnecessary. The call to `makeAbs` should be enough.

Re: `.js` -> `.ts`, you can use `git mv` to tell it that it's a renamed file. Sometimes it's heuristics for detecting this fail.

src/tuning.ts
0xfe: Pro-tip: `const s = Number(stringNum)` 🥇

src/tuning.ts
```diff
@@ -7,7 +7,13 @@ import { Vex } from './vex';
 import { Flow } from './tables';
 
 export class Tuning {
-  static get names() {
+  protected numStrings: number;
+
```
wassertim, line 11: what these line breaks for? Do you want to emphasise something?

src/tuning.ts
```diff
@@ -46,32 +59,35 @@ export class Tuning {
...
 
     return this.tuningValues[s - 1];
   }
 
-  getValueForFret(fretNum, stringNum) {
+  /** Returns the number representing the note associated with a string and a fret */
+  getValueForFret(fretNum: string | number, stringNum: string | number): number {
     const stringValue = this.getValueForString(stringNum);
-    const f = parseInt(fretNum, 10);
+    const f = typeof fretNum === 'string' ? parseInt(fretNum, 10) : fretNum;
```
wassertim, line 75: why not use `getValueForString`?

wassertim (changes requested):

rvilarl:
@0xfe this one is running for a while. There are a lot of conversations but **only two open topics**:
1) **stringNum should not be string | number**. It should be only number. String is related to instrument string and not with string type.
2) **replace the dependencies with Flow with dependencies with Music**. Music has already those functions and we could even delete one function from Flow that no other class use.

Would it be ok that I proceed with both changes? I am convinced that they make a lot of sense. Then you can review the final state.

0xfe:
1. `stringNum` should be a number, but we can't change the interface, so let's leave it as is. I would say that you can generally replace 'const s = typeof stringNum === 'string' ? parseInt(stringNum, 10) : stringNum', with the much more readable `const s = Number(stringNum)`.

2. I don't understand this one, did I miss the conversation on it? (In general, I'd suggest not changing any implementation details as part of this PR, unless it's really trivial.)

rvilarl:
> 1. `stringNum` should be a number, but we can't change the interface, so let's leave it as is. I would say that you can generally replace 'const s = typeof stringNum === 'string' ? parseInt(stringNum, 10) : stringNum', with the much more readable `const s = Number(stringNum)`.

@0xfe In the javascript library this will not change the interface regardless what we decide. Once compiled the type definitions are away.
The group defining @type/vexflow assumed string, wich is wrong, but I based my initial decision on their guess. However all the instances in this library use numbers, this is why I had to do string | number. My failure. 
Once this is ported, @types/vexflow is no longer needed and I believe that leaving a dual type due to the failure in @types/vexflow and my falure making it a dual type is not right.

> 2. I don't understand this one, did I miss the conversation on it? (In general, I'd suggest not changing any implementation details as part of this PR, unless it's really trivial.)

@0xfe We have concerns about tables.js and flow, you said you spent years moving things to the right place. Based on this comment I realised that music.ts, already ported, has two functions providing the functionality required in tuning.ts. The proposal is to use the functions in music.ts and remove the dependency to flow. With this approach two elements in table.js are also no longer needed. For the changes please look at the comments on the code on line 35 & 91 of the new version of tuning.js identified with FLOW_REMOVAL.

0xfe:
1. I see what you're saying, and I see that all the internal use is `number`, but the original code always did the type conversion, it means we supported `number | string` (since `parseInt` accepts `number | string`). So I think we're kinda stuck with `number | string` here.

2. How about we do this in a separate follow-up PR?

rvilarl:
This is fine

rvilarl:
@0xfe @mscuthbert all comments addressed. **Ready for merge!**

0xfe:
Thanks @rvilarl. Can you try `git mv fraction.js fraction.ts` so Git knows it's the same file?

rvilarl:
@0xfe I already followed your advise and rebased with a git mv in the front, but it also did not work. Look in the commit history:
`tuning and fraction reanmed`

rvilarl:
@0xfe  the mv is seen in the history:
image

rvilarl:
@0xfe there is a lot of discussion in the web but it seems to be a design decission.
https://git.wiki.kernel.org/index.php/GitFaq#Why_does_Git_not_.22track.22_renames.3F

0xfe:
Great, thank you. Merging! :-)

PR #863, opened 2021-03-29
Dependencies updated to avoid checks fail on PR (fixes #859)

### LINKED ISSUE #859: Avoid unsuccessful checks on PRs
image

rvilarl:
Dependencies updated on package.json.
A comment in stringnumber.js to disable eslint error.

I have not include releases and package-lock.json in .gitignore. @0xfe let me know if you would find this good.


mscuthbert:
I have found in other projects that grunt-webpack v4 w/ typescript v4 puts a deprecation warning on that there is no callback argument to watch. -- we're just as deprecated with older webpack/grunt versions, but now we know about it.

adding package-lock.json to .gitignore is fine by me -- it's only really useful in the Git repository if someone will be deploying a server just to run Vexflow and wants to always have the same installed modules as our dev environments.  Generally, someone will deploy her or his own code that packages vexflow, in which case they'll have their own package-lock.json file.

The releases folder though I would want to keep in the repo.   There are presumably some users who are still depending on downloading and linking to it and not fully in the `npm install` ecosystem.

rvilarl:
> I have found in other projects that grunt-webpack v4 w/ typescript v4 puts a deprecation warning on that there is no callback argument to watch. -- we're just as deprecated with older webpack/grunt versions, but now we know about it.
> 
@mscuthbert what is your proposal here, should I change any version?

> adding package-lock.json to .gitignore is fine by me -- it's only really useful in the Git repository if someone will be deploying a server just to run Vexflow and wants to always have the same installed modules as our dev environments. Generally, someone will deploy her or his own code that packages vexflow, in which case they'll have their own package-lock.json file.
>
Ok
 
> The releases folder though I would want to keep in the repo. There are presumably some users who are still depending on downloading and linking to it and not fully in the `npm install` ecosystem.
>
I propose to place releases in .gitignore in order to avoid that developers add those files on every commit. @0xfe already asked me to remove them on #851. I believe that having them ignored does not prevent to add them with `git add`. Probably it would make sense just to add them on the tagged versions or in another vexflow-bin repo managed internally. The reality is that now there are a lot of commits where code and releases files are inconsistent.

0xfe:
Hi Rodrigo, Please don't remove either:

- I'd like to keep `package-lock.json` for reproducible builds. It's been a big help when debugging regressions.
- The releases folder is important to keep too. They are generated as part of `grunt publish`, and are used by the regression tests to detect how far we are from the last release. (Usually they don't show up with every commit, only when you run the regression tests, which is not what people do on every commit, but before sending a PR.)

0xfe:
This looks good. Approving! Thanks for fixing!

mscuthbert:
> > I have found in other projects that grunt-webpack v4 w/ typescript v4 puts a deprecation warning on that there is no callback argument to watch. -- we're just as deprecated with older webpack/grunt versions, but now we know about it.
> 
> @mscuthbert what is your proposal here, should I change any version?

No.  Just that people might be aware that there might be this message shown:

```
[DEP_WEBPACK_WATCH_WITHOUT_CALLBACK] DeprecationWarning: A 'callback' argument need to be provided to the 'webpack(options, callback)' function when the 'watch' option is set. There is no way to handle the 'watch' option without a callback.
```

Hopefully someone is maintaining `grunt-webpack` and will fix this soon.  It would be super helpful because then we can run tests after a successful build while maintaining the webpack cache that makes builds so fast.

PR #864, opened 2021-03-29
map generation with typescript (fixes #862)

### LINKED ISSUE #862: How to debug typescript
Second Edit:
I am getting this on _all_ javascript files also, not just the .ts.  So maybe there is some configuration change?  In other words, I see the webpack output for all javascript under /src, not just the .ts files.  Files under /tests still look like javascript (from the source map).

I am cheering you guys on in your typescript conversion!  And I'm sorry for bringing up a question that does not directly relate to Vex.  I am apparently one of the few web developers who has never used Angular and so I am just learning ts.

When I open music.ts, I see the output of webpack.  I know I've seen other screenshots of people with TS highlighting, public/interface and all that.  How do we get that?   Do we need to install Visual Studio Code?  That is where all the TS question about angular lead.  Is there some Chrome add-on?

If I can figure it out, I will add it to the wiki so other TS newbs won't have to ask the same question.

Edit: I should mention that I am generating the debug build so I have a source map.  The .ts shows up in the sidebar but I see the transpiled .js.

image
image

rvilarl:
Enables map generation to debug typescript.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -38,7 +38,7 @@ module.exports = (grunt) => {
       resolve: {
         extensions: ['.ts', '.js', '.json']
       },
-      devtool: process.env.VEX_GENMAP || mode === 'production' ? 'source-map' : false,
+      devtool: process.env.VEX_GENMAP || mode === 'production' ? 'source-map' : 'source-map',
       module: {
         rules: [
           {
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,6 +3,7 @@
     "target": "es5",                                /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
     "module": "commonjs",                           /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
     "allowJs": true,                                /* Allow javascript files to be compiled. */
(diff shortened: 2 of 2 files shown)


Gruntfile.js
```diff
@@ -38,7 +38,7 @@ module.exports = (grunt) => {
       resolve: {
         extensions: ['.ts', '.js', '.json']
       },
-      devtool: process.env.VEX_GENMAP || mode === 'production' ? 'source-map' : false,
+      devtool: process.env.VEX_GENMAP || mode === 'production' ? 'source-map' : 'source-map',
```
mscuthbert, line 41: can remove the mode step ? : tests since they evaluate to the same.

Agreed that generating source-maps is just as crucial in development as in implementation.  Takes a little longer to generate but so worth it.

rvilarl: I left it in order to decide if a different approach was required between `production` and `development`
if this is not the case I would simply suggest
`devtool: 'source-map'`

0xfe: Thinking about this more, this is really going to slow down the build during the dev workflow. We already have `env.VEX_GENMAP` for folks to use, so maybe this change is unnecessary. You can `VEX_GENMAP=true grunt build` on the commandline.

rvilarl: Gruntfile.js reverted

rvilarl:
@0xfe ready to merge

0xfe:
Is this conditional on production builds (or VEX_GENMAP) only?

rvilarl:
> Is this conditional on production builds (or VEX_GENMAP) only?

what do you mean? Now webpack will produce a map if mode is production OR VEX_GENMAP is true.

PR #868, opened 2021-04-01
remove annoying font warning in unit tests

AaronDavidNewman:
I added the Petaluma text font from Steinberg in woff and woff2 format and used it in the chord change unit tests (it is not available as a web font afaik).  The woff2 file wasn't loading, which generated a warning.  The unit tests pass anyway, but it was causing some confusion so I removed the woff2 file and the css reference to it.  Woff is supported by all major browsers, so it's fine.


0xfe:
Thanks. Yes, this was a red herring, but the message was annoying.

rvilarl:
@AaronDavidNewman Thank you! That was indeed annoying :1st_place_medal:

PR #872, opened 2021-04-02
Typescript: music_test migrated

rvilarl:
For discussion on tests migration approach

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -14,6 +14,7 @@ module.exports = (grunt) => {
   const BUILD_DIR = path.join(BASE_DIR, 'build');
   const RELEASE_DIR = path.join(BASE_DIR, 'releases');
   const MODULE_ENTRY = path.join(BASE_DIR, 'src/index.js');
+  const TEST_TS_ENTRY = path.join(BASE_DIR, 'tests/run_ts.ts')
   const TARGET_RAW = 'vexflow-debug.js';
   const TARGET_MIN = 'vexflow-min.js';
 
@@ -24,10 +25,10 @@ module.exports = (grunt) => {
   const TARGET_TESTS = path.join(BUILD_DIR, 'vexflow-tests.js');
   const TEST_SOURCES = ['tests/vexflow_test_helpers.js', 'tests/mocks.js', 'tests/*_tests.js', 'tests/run.js'];
 
-  function webpackConfig(target, mode) {
+  function webpackConfig(target, entry, mode) {
     return {
       mode: mode,
-      entry: MODULE_ENTRY,
+      entry: entry,
       output: {
         path: BUILD_DIR,
         filename: target,
@@ -51,8 +52,9 @@ module.exports = (grunt) => {
     };
   }
 
-  const webpackProd = webpackConfig(TARGET_MIN, 'production');
-  const webpackDev = webpackConfig(TARGET_RAW, 'development');
+  const webpackProd = webpackConfig(TARGET_MIN, MODULE_ENTRY, 'production');
+  const webpackDev = webpackConfig(TARGET_RAW, MODULE_ENTRY, 'development');
+  const webpackTest = webpackConfig('vexflow-tests-ts.js', TEST_TS_ENTRY, 'development');
 
   grunt.initConfig({
     pkg: grunt.file.readJSON('package.json'),
@@ -69,6 +71,7 @@ module.exports = (grunt) => {
     webpack: {
       build: webpackProd,
       buildDev: webpackDev,
+      buildTest: webpackTest,
       watch: {
         ...webpackDev,
         watch: true,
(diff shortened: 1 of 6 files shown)


rvilarl:
@0xfe @mscuthbert @wassertim I would like to discuss on the migration approach for the tests I will make the questions/proposals directly on the code of music_test.cpp

tests/music_tests.ts
```diff
@@ -2,27 +2,34 @@
...
+      test('Canonical Intervals', this.canonicalNotes);
+      test('Scale Tones', this.scaleTones);
+      test('Scale Intervals', this.scaleIntervals);
     },
 
     validNotes: function () {
       expect(10);
 
-      var music = new VF.Music();
+      var music = new Music();
```
0xfe, line 32: While we're at it I'd change the `var` to `let` or `const`.

0xfe:
Overall, seems good. I'd certainly like to see what `tests/vexflow_test_helpers.js` would look like.

@mscuthbert and @wassertim for thoughts

rvilarl (commented):
> Overall, seems good. I'd certainly like to see what `tests/vexflow_test_helpers.js` would look like.
> 
> @mscuthbert and @wassertim for thoughts

do you mean migrating `test/vexflow_test_helpers.js` to typescript? I was plannig to leave that to the end in order to avoid two variants of this file one for js and one for ts

tests/music_tests.ts
rvilarl: Done

Gruntfile.js
```diff
@@ -81,7 +84,7 @@ module.exports = (grunt) => {
       options: { fix: true },
     },
     qunit: {
-      files: ['tests/flow.html'],
+      files: ['tests/flow.html', 'tests/flow_ts.html'],
```
wassertim, line 87: I personally don't see a benefit of having 2 flow html files

tests/music_tests.ts
```diff
@@ -2,27 +2,34 @@
...
+      test('Canonical Notes', this.canonicalNotes);
+      test('Canonical Intervals', this.canonicalNotes);
+      test('Scale Tones', this.scaleTones);
+      test('Scale Intervals', this.scaleIntervals);
     },
 
     validNotes: function () {
       expect(10);
 
-      var music = new VF.Music();
```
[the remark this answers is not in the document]
wassertim: VF is not a factory, but an Object where Vex.Flow is put. `const VF = Vex.Flow`;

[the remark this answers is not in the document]
rvilarl: I mentioned the factory because there are cases where it is indeed a factory. I refer to the discussion in #857.
I think using `Music` directly is clearer than `VF.Music` or `vf.Music`. Do you agree?

Gruntfile.js
rvilarl: It is very simple and does not have drawbacks. One benefit might be that it provides a _**nice visual feedback on the progress**_ of the migration and in the end when everything is migrated we will get one file anyway. 
But now you managed one file. Let us review it!

wassertim: A drawback is a code repetition, because we have 2 identical files.

tests/music_tests.ts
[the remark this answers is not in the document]
wassertim: I just wanted to specify, that VF.Music is different from vf.Music(). vf.Music() does not need to be replaced because it return new Music() and is perfectly fine with typescript. What is making type checking possible is `import { Music } from '../src/music'`. vf.Music() does not stop you from type checking.

Gruntfile.js
rvilarl: After reviewing #873, my oppinion is this approach(#872) has the following:

**benefits**:
- it is **very simple**, simpler than the one file approach
- it provides **visual feedback on the progress of the migration**

**drawbacks**:
- code repetition on the **two html files**, but one will no longer be needed at the end of the migration.

Personally I prefer this approach but I also made it :)

@0xfe @mscuthbert you decide, in the end of the day it is a matter of taste and both approaches do work.

wassertim: it is not only `flow_ts.html`, which makes a confusion, you also have `vexflow-test-ts.js` and `run_ts.ts` - these are just ugly

0xfe: Thanks Rodrigo and Tim. Let's go with the single file please :-)

rvilarl: @0xfe this is fine, I will leave this open until the other approach is merged should there be a show stopper. 

@0xfe I still do not understand the relation between `vexflow-tests.js` `vexflow-tests-node.js` **please review `Gruntfile.js`**, I am definetively confused :).

rvilarl: @wassertim after the review #873 I believe that there are no longer show stoppers, I close this one.

PR #873, opened 2021-04-02
Typescript migration: compilation of tests with webpack

wassertim:
This is basically the same what @rvilarl proposed, but without concatenation and flow_ts.html file. Everything is done by webpack.


Gruntfile.js
```diff
@@ -77,8 +94,8 @@ module.exports = (grunt) => {
       },
     },
     eslint: {
-      target: SOURCES.concat(['./tests/**/*.ts', './tests/**/*.js']),
-      options: { fix: true },
+      target: SOURCES.concat('./tests'),
+      options: { fix: false },
```
rvilarl, line 98: Why don't you want the issues to be fixed? Eslint is quite good at it for example with templates on strings.

Gruntfile.js
```diff
@@ -99,7 +116,7 @@ module.exports = (grunt) => {
             expand: true,
             dest: RELEASE_DIR,
             cwd: BUILD_DIR,
-            src: ['*.ts', '*.js', 'docs/**', '*.map'],
+            src: ['*.js', 'docs/**', '*.map'],
```
rvilarl, line 119: There are no ts files?

tests/music_tests.ts
```diff
@@ -0,0 +1,311 @@
...
+    test('Valid Notes', this.validNotes);
+    test('Valid Keys', this.validKeys);
+    test('Note Values', this.noteValue);
+    test('Interval Values', this.intervalValue);
+    test('Relative Notes', this.relativeNotes);
+    test('Relative Note Names', this.relativeNoteNames);
+    test('Canonical Notes', this.canonicalNotes);
+    test('Canonical Intervals', this.canonicalNotes);
+    test('Scale Tones', this.scaleTones);
+    test('Scale Intervals', this.scaleIntervals);
```
rvilarl, line 27: giving it a second thought, I believe that `this` should be replaced by `MusicTests` because this is not the instantiation of a class, would you agree?

tests/flow.html
```diff
@@ -9,7 +9,7 @@
 
   <!-- Test Dependencies: jQuery, QUnit -->
   <link rel="stylesheet" href="support/qunit.css" type="text/css" media="screen" />
-<link href="https://fonts.googleapis.com/css2?family=Roboto+Slab:wght@500&display=swap" rel="stylesheet">
+  <link href="https://fonts.googleapis.com/css2?family=Roboto+Slab:wght@500&display=swap" rel="stylesheet">
```
rvilarl, line 12: Why did you changed that?

rvilarl (changes requested):

.eslintrc
```diff
@@ -11,19 +11,16 @@
   ],
   "rules": {
     "no-console": 1, // Means warning
-    "prettier/prettier": 2 // Means error
+    "prettier/prettier": 1, // Means error
```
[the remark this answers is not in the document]
wassertim: Did not find any mentions about it. This should be an error, it is just formatting, a warning should be fine

.eslintrc
```diff
@@ -11,19 +11,16 @@
...
+    "prettier/trailingComma": 0
```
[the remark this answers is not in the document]
wassertim: I did not notice that this project uses trailing commas. I personally hate them. I will change this back

tests/flow.html
wassertim: indentation is wrong

wassertim: I will revert, it does not actually belong here :)

tests/music_tests.ts
wassertim: `this` references to the MusicTests and is same. Why do you think it should be replaced?

Gruntfile.js
```diff
@@ -77,8 +94,8 @@ module.exports = (grunt) => {
       },
     },
     eslint: {
-      target: SOURCES.concat(['./tests/**/*.ts', './tests/**/*.js']),
-      options: { fix: true },
+      target: SOURCES.concat('./tests'),
```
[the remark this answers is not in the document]
wassertim: What do you mean does not match correctly? It did for me. Do you have any examples?

Gruntfile.js
wassertim: This makes unnecessary changes which I do not plan to introduce. Github and console shows you all the warnings. They can be fixed separately.

Gruntfile.js
wassertim: in build dir? No I don't think so

package.json
```diff
@@ -23,7 +23,6 @@
     "canvas": "^2.7.0",
     "docco": "^0.8.1",
     "eslint": "^7.23.0",
-    "eslint-config-airbnb-base": "^14.2.1",
```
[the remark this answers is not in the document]
wassertim: because it is not used

Gruntfile.js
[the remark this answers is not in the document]
rvilarl: Look in the code diff wih maste. You have writen the options in a complete different way and they do not match. It is not easy to see the changes. It seems that you had not used the version in master.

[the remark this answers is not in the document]
wassertim: I rebased. This one I changed on purpose. So what is your concern here?

tests/music_tests.ts
rvilarl: because I have already migrated `tuning_tests.ts` and I  have seen that with `this` it does not work when you try to call `checkStandard` and `checkStandardBanjo` :)

I decided to make today a second step including `Fraction` and `Tuning` to prove the general concept.

Gruntfile.js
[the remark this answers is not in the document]
rvilarl: I made the changed because it was not matching the files I had the discussion in the past about the regex with you and @0xfe in the end it was not a regex but a list.

tests/music_tests.ts
wassertim: It is because you have to bind this to standard function. That is how js works.

Gruntfile.js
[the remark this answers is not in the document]
wassertim: I just checked. And it worked, it matched ts and js files.
Screenshot 2021-04-02 at 20 39 04

tsconfig.json
```diff
@@ -1,12 +1,15 @@
...
+    "strict": true,
+    "module": "CommonJS",
+    "moduleResolution": "Node",
+    "allowSyntheticDefaultImports": true,
+    "target": "ES5",
+    "baseUrl": ".",
+    "outDir": "build",
+    "esModuleInterop": true,
+    "declaration": true,
+    "noImplicitAny": false,
```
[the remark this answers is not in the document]
wassertim: it is actually false by default, I can remove it

.eslintrc
```diff
@@ -11,19 +11,15 @@
   ],
   "rules": {
     "no-console": 1, // Means warning
-    "prettier/prettier": 2 // Means error
+    "prettier/prettier": 1 // Means warning
```
0xfe, line 14: Would def like this to remain error. We can use per-line or per-file overrides for where we disagree.

Gruntfile.js
0xfe, line 98: Please leave this as `true`. We should use overrides for where we disagree.

tools/generate_png_images.js
```diff
@@ -13,8 +13,9 @@ document = dom.window.document;
 const fs = require('fs');
 const [scriptDir, imageDir] = process.argv.slice(2, 4);
 
-const Vex = require(`${ scriptDir }/vexflow-debug.js`);
-Vex.Flow.Test = require(`${ scriptDir }/vexflow-tests.js`);
+global['Vex'] = require(`${scriptDir}/vexflow-debug.js`);
+
+require(`${scriptDir}/vexflow-tests-node.js`);
```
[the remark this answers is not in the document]
0xfe: Thanks, this is much better. Glad we're using webpack instead of concatenation here.

Gruntfile.js
wassertim: This will fix formatter.js, which does not have anything to do with this PR ;)

src/formatter.js
```diff
@@ -216,9 +216,7 @@ export class Formatter {
     const beams = options.auto_beam ? Beam.applyAndGetBeams(voice) : [];
 
     // Instantiate a `Formatter` and format the notes.
-    new Formatter()
-      .joinVoices([voice])
-      .formatToStave([voice], stave, { align_rests: options.align_rests, stave });
+    new Formatter().joinVoices([voice]).formatToStave([voice], stave, { align_rests: options.align_rests, stave });
```
wassertim, line 219: This change is made by es-lint autofix. This is really a strange rule to have this kind of chain in one line. IMO unreadable.

wassertim:
I fixed your remarks @0xfe. Merging this will increase ts migration speed greatly.

Gruntfile.js
```diff
@@ -18,22 +20,29 @@ module.exports = (grunt) => {
...
+  const SOURCES = ['./src/*.ts', './src/*.js', '!./src/header.js'];
 
-  // Take all test files in 'tests/' and build TARGET_TESTS
-  const TARGET_TESTS = path.join(BUILD_DIR, 'vexflow-tests.js');
-  const TEST_SOURCES = ['tests/vexflow_test_helpers.js', 'tests/mocks.js', 'tests/*_tests.js', 'tests/run.js'];
+  // Take all test files in 'tests/' and build TARGET_TESTS_BROWSER
+  const TARGET_TESTS_BROWSER = 'vexflow-tests.js';
+  const TARGET_TESTS_NODE = 'vexflow-tests-node.js';
+  const TEST_SOURCES = [
+    './tests/vexflow_test_helpers.js',
```
rvilarl, line 29: @0xfe you commented in #872

> I'd certainly like to see what tests/vexflow_test_helpers.js would look like.

do mean migrating to ts? in the one file approach it should be possible without duplicating the file.

@wassertim what do you think?

.eslintrc
```diff
@@ -19,11 +19,7 @@
       "extends": [
         "eslint:recommended",
         "prettier"
-      ],
-      "rules": {  
-        "no-console": 1, // Means warning
-        "prettier/prettier": 2, // Means error
-      }
+      ]
```
rvilarl, line 22: @wassertim have you tested that the rules are inhereted from the level above?

rvilarl:
@0xfe I believe that you asked me not ot include the files in the releases directory in my first PR. Is that the way to proceed or I have it wrong in mind?

tests/music_tests.ts
```diff
@@ -0,0 +1,296 @@
...
+    test('Valid Notes', MusicTests.validNotes);
+    test('Valid Keys', this.validKeys);
+    test('Note Values', this.noteValue);
+    test('Interval Values', this.intervalValue);
+    test('Relative Notes', this.relativeNotes);
+    test('Relative Note Names', this.relativeNoteNames);
+    test('Canonical Notes', this.canonicalNotes);
+    test('Canonical Intervals', this.canonicalNotes);
+    test('Scale Tones', this.scaleTones);
+    test('Scale Intervals', this.scaleIntervals);
```
rvilarl, line 27: All these instances of `this` should be changed to `MusicTests` as well.

tests/music_tests.ts
```diff
@@ -0,0 +1,296 @@
...
+declare const QUnit: any;
+declare const ok: any;
+declare const test: any;
+declare const expect: any;
+declare const equal: any;
+
+import { Music } from '../src/music';
+import { KeyManager } from '../src/keymanager';
+
+export const MusicTests = {
```
rvilarl, line 15: I know that this change is **ugly** but it will help all of us a lot going forward becuase it allows git to detect that the file music_tests.ts` was renamed and not deleted.
Change:
```
export const MusicTests = {
```
to:
``` 
export const MusicTests = (function () {
  return {
```
and add: 
```
})();
```
at the end of the definition after line 296.

At the end of the migration we can remove this unnecessary level, if we want, but during the migration it will allow us to see renamed files decently.

Gruntfile.js
rvilarl, line 29: @0xfe you commented in #872

> I'd certainly like to see what tests/vexflow_test_helpers.js would look like.

do mean migrating to ts? in the one file approach it should be possible without duplicating the file.

@wassertim what do you think?

rvilarl (changes requested):

tests/music_tests.ts
rvilarl: @wassertim you only changed the first one

tests/music_tests.ts
rvilarl: @0xfe @wassertim I believe that this one is **very important to facilitate the reviews**. 

I realised that in #872 the diff was present and I did look for the difference. I have also checked in my fork that making the change in this PR fix the issue. 

Look at `music_tests` in: https://github.com/rvilarl/vexflow/pull/13/files

tests/music_tests.ts
rvilarl: My fault then, but all have to be changed :)

Gruntfile.js
wassertim: Not here anyway. File by file it should be

tests/music_tests.ts
wassertim: Since we have seen this approach has worked on a test file, we don't need it anymore. I reverted back to music_tests.js, we can migrate this one later

wassertim:
Thanks for your review @0xfe, @rvilarl. I reverted back to music_tests.js, so it is now even simpler.

rvilarl:
It is a pity to discard now the changes and review effort on music_tests.ts. I think.

wassertim:
> It is a pity to discard now the changes and review effort on music_tests.ts. I think.

It is alright, nothing is lost. We can always use git history and take the source code into our next PR

0xfe:
Hey Tim, I think we should have at least one `.ts` test in this commit. (Think of it also as a test of this new infrastructure.) I'm okay with leaving `music_tests.ts` in (or something simpler is fine too, but something.)

wassertim:
> Hey Tim, I think we should have at least one `.ts` test in this commit. (Think of it also as a test of this new infrastructure.) I'm okay with leaving `music_tests.ts` in (or something simpler is fine too, but something.)

But we have it in different commit, so you can check it works if it is what you want. I just don't want to mix these things up

0xfe:
Sorry, different commits are okay. But same PR is important please.

wassertim:
> Sorry, different commits are okay. But same PR is important please.

done

wassertim:
@0xfe would be nice if you could restart the CI

wassertim:
It did run in my fork
Screenshot 2021-04-03 at 14 25 39

rvilarl:
@wassertim please take into account the review comments I made to this version:
- 'this' topic 
- change to allow git recognize that music_tests.js was renamed
Thanks

wassertim:
> @wassertim please take into account the review comments I made to this version:
> 
> * 'this' topic
> * change to allow git recognize that music_tests.js was renamed
>   Thanks

- I thought we already discussed `this`. You can use bind method for assigning it to any function.
- I don't think it is a good idea to adjust code to some diff tool (github is apparently buggy). If you need a diff tool which works, you can use IntelliJ IDEA for example

wassertim:
> @wassertim please take into account the review comments I made to this version:
> 
> * 'this' topic
> * change to allow git recognize that music_tests.js was renamed
>   Thanks

That is why I did not want music_tests.ts here. It involves unnecessary discussion :)

Gruntfile.js
```diff
@@ -18,22 +20,29 @@ module.exports = (grunt) => {
...
   // Used for eslint and docco
-  const SOURCES = ['src/*.ts', 'src/*.js', '!src/header.js'];
+  const SOURCES = ['./src/*.ts', './src/*.js', '!./src/header.js'];
 
-  // Take all test files in 'tests/' and build TARGET_TESTS
-  const TARGET_TESTS = path.join(BUILD_DIR, 'vexflow-tests.js');
-  const TEST_SOURCES = ['tests/vexflow_test_helpers.js', 'tests/mocks.js', 'tests/*_tests.js', 'tests/run.js'];
+  // Take all test files in 'tests/' and build TARGET_TESTS_BROWSER
+  const TARGET_TESTS_BROWSER = 'vexflow-tests.js';
+  const TARGET_TESTS_NODE = 'vexflow-tests-node.js';
```
0xfe, line 27: Why a separate target for node tests here? How does it interact with the regression tests?

0xfe:
Can you remove `releases/*` from this PR please?

0xfe:
Overall, looks good. Is there a way to make this work without the extra `-node` file?

We don't need to build everything for the commandline tests and CI, just the node tests.

Gruntfile.js
```diff
@@ -18,22 +20,29 @@ module.exports = (grunt) => {
...
 
   // Used for eslint and docco
-  const SOURCES = ['src/*.ts', 'src/*.js', '!src/header.js'];
+  const SOURCES = ['./src/*.ts', './src/*.js', '!./src/header.js'];
 
-  // Take all test files in 'tests/' and build TARGET_TESTS
-  const TARGET_TESTS = path.join(BUILD_DIR, 'vexflow-tests.js');
-  const TEST_SOURCES = ['tests/vexflow_test_helpers.js', 'tests/mocks.js', 'tests/*_tests.js', 'tests/run.js'];
+  // Take all test files in 'tests/' and build TARGET_TESTS_BROWSER
+  const TARGET_TESTS_BROWSER = 'vexflow-tests.js';
```
0xfe, line 26: In this case, maybe we should be clearer: `vexflow-tests-browser.js` and `vexflow-tests.node.js`.

wassertim:
> Can you remove `releases/*` from this PR please?

If we remove releases, regression tests will run on the version in master. Is it supposed to be so? The output will not be fair

Gruntfile.js
wassertim: I will try node bundle in the browser, this might solve the problem with 2 files. If no, I will change the names accordingly.

rvilarl:
> > @wassertim please take into account the review comments I made to this version:
> > 
> > * 'this' topic
> > * change to allow git recognize that music_tests.js was renamed
> >   Thanks
> 
> @0xfe That is why I did not want music_tests.ts here. It involves unnecessary discussion :)

@wassertim discussion improves the output :) 

1) I suggested to use `MusicTests` rather than `this` because I saw problems calling `checkStandard` in `tuning_tests` (give it a try yourself). You left a mix using `MusicTests` on the first line and `this` on the others (please decide one way or the other, your choice):
```
 test('Valid Notes', MusicTests.validNotes);
 test('Valid Keys', this.validKeys);
 test('Note Values', this.noteValue);
 test('Interval Values', this.intervalValue);
 test('Relative Notes', this.relativeNotes);
 test('Relative Note Names', this.relativeNoteNames);
 test('Canonical Notes', this.canonicalNotes);
 test('Canonical Intervals', this.canonicalNotes);
 test('Scale Tones', this.scaleTones);
 test('Scale Intervals', this.scaleIntervals);
```
2) In #861 @0xfe had some discussions and trials trying to figure out how to fix the fact that `fraction.ts` was presented as new file. In this PR we have the same problem however this problem was not there in #872. I figured out the difference, which is minor, and I proposed the change to facilitate reviews with github.

@0xfe would you agree on 2)?

0xfe:
> > Can you remove `releases/*` from this PR please?
> 
> If we remove releases, regression tests will run on the version in master. Is it supposed to be so? The output will not be fair

Yes, that's intended. We're not changing any implementation here, so it's fine. I'll cut a new versioned release to bring the regression tests up to speed.

Gruntfile.js
0xfe: That would be great. Alternatively, we can also trying to build just one `vexflow-tests.js` and adding a small shim for the node version.

wassertim: I tried using node bundle and it did not work. I will try a node shim though.

Gruntfile.js
```diff
@@ -42,17 +51,23 @@ module.exports = (grunt) => {
       module: {
         rules: [
           {
-            test: /(\.ts?$|\.js?$)/,
+            test: /(\.ts?$)/,
```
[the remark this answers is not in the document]
wassertim: ts-loader is for typescript files. It is not needed for javascript files. Webpack will only compile typescript files with typescript compiler

wassertim:
> > > @wassertim please take into account the review comments I made to this version:
> > > 
> > > * 'this' topic
> > > * change to allow git recognize that music_tests.js was renamed
> > >   Thanks
> > 
> > 
> > @0xfe That is why I did not want music_tests.ts here. It involves unnecessary discussion :)
> 
> @wassertim discussion improves the output :)
> 
> 1. I suggested to use `MusicTests` rather than `this` because I saw problems calling `checkStandard` in `tuning_tests` (give it a try yourself). You left a mix using `MusicTests` on the first line and `this` on the others (please decide one way or the other, your choice):
> 
> ```
>  test('Valid Notes', MusicTests.validNotes);
>  test('Valid Keys', this.validKeys);
>  test('Note Values', this.noteValue);
>  test('Interval Values', this.intervalValue);
>  test('Relative Notes', this.relativeNotes);
>  test('Relative Note Names', this.relativeNoteNames);
>  test('Canonical Notes', this.canonicalNotes);
>  test('Canonical Intervals', this.canonicalNotes);
>  test('Scale Tones', this.scaleTones);
>  test('Scale Intervals', this.scaleIntervals);
> ```
> 
> 1. In #861 @0xfe and I had some discussions and trials trying to figure out how to fix the fact that `fraction.ts` was presented as new file. In this PR we have the same problem however this problem was not there in #872. I figured out the difference, which is minor, and I think we should make the change to facilitate reviews with github.
> 
> @0xfe would you agree on 2)?

Yes I agree, that discussion improves an output and I gave you my opinion on these 2.

wassertim:
> > > Can you remove `releases/*` from this PR please?
> > 
> > 
> > If we remove releases, regression tests will run on the version in master. Is it supposed to be so? The output will not be fair
> 
> Yes, that's intended. We're not changing any implementation here, so it's fine. I'll cut a new versioned release to bring the regression tests up to speed.

Ok, I managed to do it without node bundle. I made some shim as you suggested. Release files a reverted.

0xfe:
> > > > Can you remove `releases/*` from this PR please?
> > > 
> > > 
> > > If we remove releases, regression tests will run on the version in master. Is it supposed to be so? The output will not be fair
> > 
> > 
> > Yes, that's intended. We're not changing any implementation here, so it's fine. I'll cut a new versioned release to bring the regression tests up to speed.
> 
> Ok, I managed to do it without node bundle. I made some shim as you @0xfe suggested. Release files a reverted.

Great, reviewing.

Gruntfile.js
```diff
@@ -18,41 +20,53 @@ module.exports = (grunt) => {
...
   // Used for eslint and docco
-  const SOURCES = ['src/*.ts', 'src/*.js', '!src/header.js'];
+  const SOURCES = ['./src/*.ts', './src/*.js', '!./src/header.js'];
 
-  // Take all test files in 'tests/' and build TARGET_TESTS
-  const TARGET_TESTS = path.join(BUILD_DIR, 'vexflow-tests.js');
-  const TEST_SOURCES = ['tests/vexflow_test_helpers.js', 'tests/mocks.js', 'tests/*_tests.js', 'tests/run.js'];
+  // Take all test files in 'tests/' and build TARGET_TESTS_BROWSER
+  const TARGET_TESTS_BROWSER = 'vexflow-tests.js';
+  const TARGET_TESTS_NODE = 'vexflow-tests-node.js';
```
0xfe, line 27: I guess we can remove this now, right? (And make `TARGET_TESTS_BROWSER` -> `TARGET_TESTS`)

Gruntfile.js
```diff
@@ -18,41 +20,53 @@ module.exports = (grunt) => {
...
+  ];
 
-  function webpackConfig(target, mode) {
+  function webpackConfig(target, moduleEntry, mode, libraryName, compilationTarget) {
     return {
+      node: compilationTarget !== 'node' ? { fs: 'empty' } : undefined,
       mode: mode,
-      entry: MODULE_ENTRY,
+      entry: moduleEntry,
+      target: compilationTarget,
```
0xfe, line 39: No need for this either, I'm guessing.

tests/vexflow_test_helpers.js
```diff
@@ -213,7 +209,7 @@ VF.Test = (function () {
     },
 
     runNodeTest: function (name, func, params) {
-      var fs = require('fs');
+      var fs = VF['fs'];
```
0xfe, line 212: Can we scope this to `VF['shims']['fs']` and also add a comment about it?

tools/generate_png_images.js
```diff
@@ -13,10 +13,13 @@ document = dom.window.document;
...
 
-const Vex = require(`${ scriptDir }/vexflow-debug.js`);
-Vex.Flow.Test = require(`${ scriptDir }/vexflow-tests.js`);
+global['Vex'] = require(`${scriptDir}/vexflow-debug.js`);
+
+require(`${scriptDir}/vexflow-tests.js`);
 
 const VF = Vex.Flow;
+VF['fs'] = fs;
+VF['process'] = process;
```
0xfe, line 22: Note above re scoping these into `VF['shims']`

0xfe:
FYI, this is much better than having two bundles -- thanks for figuring it out @wassertim.

tests/vexflow_test_helpers.js
wassertim: done

rvilarl:
I agree, it is much better! And thank you for attending the request to make git see music_test as renamed it also helps a lot!

rvilarl (approved):

rvilarl:
I think that this is now ready, let us merge it!

rvilarl:
@0xfe there is a minor error in the test `const` and `let` are back to `var`. I suggest that we include the fix in the next PR together with the test of fraction and tuning which are minor chnages and ready. Let us merge and move on, this PR is very goood already.

0xfe:
Looks great! Thanks @wassertim! Now we can move a lot faster! :-)

Issue #869, opened 2021-04-01
Typescript migration: Documentation


rvilarl:
I suggest to start the generation of documentation using typedoc for the migrated files

https://typedoc.org/

rvilarl:
![image](https://user-images.githubusercontent.com/22865285/113249083-691ecd80-92be-11eb-87c7-5939e7220b6c.png)

![image](https://user-images.githubusercontent.com/22865285/113249198-95d2e500-92be-11eb-973a-fe7769c63259.png)

wassertim:
Looks nice. But I would rather concentrate on typescript migration. Otherwise we start too many topics and finish to little

rvilarl:
@0xfe @mscuthbert @wassertim 

Why am I proposing to use typedoc for typescript migration?

**Target**: Help on review activities

**How**:
- typedoc allows us to **check if we are naming the differenct elements consistently**. This is the overview of the elements so far.

![image](https://user-images.githubusercontent.com/22865285/113390746-9dfc5480-9392-11eb-8af4-062c824b12f8.png)

- typedoc helps to **check that the comments make sense**. They are presented in the right context.

![image](https://user-images.githubusercontent.com/22865285/113390973-08ad9000-9393-11eb-904f-b9a35d7cf931.png)

**Do not worry!!!**:
- It is automatically generated.
- It does not interfere with docco
- It does not force to provide the comments, if you don't, you still get a very good  representation (llok at the Type Aliases of the first capture of this comment)

rvilarl:
@0xfe @mscuthbert your thoughts?

rvilarl:
I will make a discussions PR. I believe that it helps to show concepts. I will include also the automatic generated files so that you can look into them without compilation.

rvilarl:
Merged in #874

PR #874, opened 2021-04-03
Typescript: typedoc documentation (proposal on #869)

rvilarl:
As suggested in #869.

I have created `./typedoc.ts` to put all the elements in a single module in order to facilitate the review activities during the migration. With this approach all the classes, interfaces & type aliases are accesible  in a single page under `exports`.


rvilarl:
@0xfe do you find this documentation approach of interest? As I said I believe it should help as minimum with the reviews. But I think that it the answer to many comments on lack of API documentation.

0xfe:
This is fine with me. I'm okay with removing `docco` too. Let's do that in a separate PR.

rvilarl:
@0xfe if this is fine you can simply merge this PR. We will remove `docco`in a separate PR but I suggest we wait until the migration is over. Typedoc only documents the migrated files.

0xfe:
Okay, looks good. Merging. Thanks!

PR #882, opened 2021-04-07
Migration/element

rvilarl:
Let us continue with element!


src/annotation.js
```diff
@@ -156,7 +156,7 @@ export class Annotation extends Modifier {
 
     // We're changing context parameters. Save current state.
     this.context.save();
-    const classString = Object.keys(this.getAttribute('classes')).join(' ');
+    const classString = Object.keys(this.classes).join(' ');
```
0xfe, line 159: Is this intentional? Does this work on SVG elements?

rvilarl:
Later today I will try to make diff work. I did not know that git was so problematic with that. :)

src/element.ts
```diff
@@ -0,0 +1,275 @@
...
+
+/** Element attributes. */
+export interface ElementAttributes {
+  [name: string]: string;
+  id: string;
+  type: string;
+}
+
+/** Contexts common interface */
+export interface RenderContext {
```
0xfe, line 23: Maybe a separate `rendercontext.ts` for this.

src/element.ts
```diff
@@ -0,0 +1,275 @@
...
+  closeGroup(): void;
+
+  /**
+   * canvas returns TextMetrics, SVG returns SVGRect, Raphael returns {width : number, height : number}. Only width is used throughout VexFlow.
+   */
+  measureText(text: string): { width: number };
+}
+
+/** Element style */
+export interface Style {
```
0xfe, line 69: `ElementStyle`, so we're consistent.

rvilarl:
@0xfe I have been investigating how to get the Element diff to work. It is possible doing the migration in two PRs:
1) Sofware changes with all the interfaces in types/common
2) Modifications of Comments and movement of interfaces to the element.ts file
Should I proceed like that? Or should we consider as @wassertim said using another tool to look into the changes.

src/element.ts
rvilarl: I will move it to types/common if you agree.

src/ornament.js
```diff
@@ -226,7 +226,7 @@ export class Ornament extends Modifier {
     const stemDir = this.note.getStemDirection();
     const stave = this.note.getStave();
 
-    const classString = Object.keys(this.getAttribute('classes')).join(' ');
+    const classString = Object.keys(this.classes).join(' ');
```
wassertim, line 229: I don't understand this change

typedoc.ts
```diff
@@ -1,3 +1,4 @@
+export * from './src/element';
```
wassertim, line 1: somehow I missed this. Would you mind to explain this?

src/annotation.js
rvilarl: This is one refactoring that I propose in this PR: `this.attrs` 

It is intentional but I do not know if it works with SVG, how can I check that?

Reasons for the change:
- `this.attrs` contains three different types of information: string (`id`, `type`, ...), SVGSVGElement (`el`), Record<string, boolean> (`classes`)
- `el`  is not used in the software, it is only initialised to null
- there are two methods to access attributes `getAttribute` and `setAttribute`
- there are three methods to access classes `hasClass`, `addClass`, `removeClass`

My proposal in this PR:
- Remove `el`
- Leave string information in `this.attr` with its associated acces functions `getAttribute` and `setAttribute`.
- Place the classes in `this.attr` with its associated access functions `hasClass`, `addClass`, `removeClass`.
- The objects that require the complete list of classes are `annotation`, `chordsymbol` and `ornament` and they can use the protected variable `this.classes` directly.

src/element.ts
```diff
@@ -0,0 +1,275 @@
+// [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
```
wassertim, line 1: Why does not it have diff?

src/element.ts
rvilarl: ok

src/annotation.js
wassertim, line 159: Why not create a separate PR for this? This has nothing to do with migration to typescript, Ricardo

sschmidTU: If the result of `this.getAttribute('classes')` is the same as `this.classes`, it looks like an obvious simplification.

src/ornament.js
rvilarl: > This is one refactoring that I propose in this PR: `this.attrs`
> 
> Reasons for the change:
> 
> * `this.attrs` contains three different types of information: string (`id`, `type`, ...), SVGSVGElement (`el`), Record<string, boolean> (`classes`)
> * `el`  is not used in the software, it is only initialised to null
> * there are two methods to access attributes `getAttribute` and `setAttribute`
> * there are three methods to access classes `hasClass`, `addClass`, `removeClass`
> 
> My proposal in this PR:
> 
> * Remove `el`
> * Leave string information in `this.attr` with its associated acces functions `getAttribute` and `setAttribute`.
> * Place the classes in `this.classes` with its associated access functions `hasClass`, `addClass`, `removeClass`.
> * The objects that require the complete list of classes are `annotation`, `chordsymbol` and `ornament` and they can use the protected variable `this.classes` directly.

src/element.ts
rvilarl: > @0xfe I have been investigating how to get the Element diff to work. It is possible doing the migration in two PRs:
> 
> 1. Sofware changes with all the interfaces in types/common
> 2. Modifications of Comments and movement of interfaces to the element.ts file
>    Should I proceed like that? Or should we consider as @wassertim said using another tool to look into the changes.

src/annotation.js
rvilarl: I think that we have to as far as possible avoid the usage of `any` in order to get the advantages of TypeScript type checking and double types in order to simplify the code.

With that we should improve maintainability and we will also have the possibility to find bugs in this migration journey.

sschmidTU: @rvilarl did you comment on the wrong change? there is no `any` in this change.

typedoc.ts
rvilarl: > I have created `./typedoc.ts` to put all the elements in a single module in order to facilitate the review activities during the migration. With this approach all the classes, interfaces & type aliases are accesible in a single page under `exports`.

from #874 this file includes the files to document with typedoc

src/annotation.js
rvilarl: @sschmidTU we were writing at the same time, I have put the quote :)

rvilarl: @sschmidTU do you have an oppinion on this one? How is that related to SVG Elements in OSMD?

sschmidTU: Still, i don't see what changing `this.getAttribute('classes')` to `this.classes` has to do with `any`. But i'm not against the change anyways.

rvilarl: @sschmidTU have a look where I left you a question. I will expand on the topic any editing it.

src/element.ts
0xfe: Okay, that's fine for now. I want to be careful not to bloat `types/common`.

This might be a good opportunity to add some directory structure. Create a `renderer` directory with it's own `.d.ts` file and move all the rendering code there: `svgcontext`, `canvascontext`, `raphaelcontext`, and `renderer.js`.

Maybe in the next PR.

src/ornament.js
0xfe: Thanks for the explanation Rodrigo. I agree with @wassertim that there's too much logic refactoring going on here. I think we should separate it out and focus on TypeScript.

src/element.ts
0xfe: @rvilarl Looks like you can use the `-M` flag to tell git what diff threshold to use for detecting renames. You can try it with a small value like `-M0.4` (<40% file change.)

rvilarl: @0xfe is this one thing that we could configure in github as well or just locally

src/annotation.js
sschmidTU: i'm not sure, OSMD doesn't use the Annotation class. If you haven't verified that this works on SVG elements, i guess it shouldn't be changed. Otherwise the `classes` change looks fine to me on a first look.

src/element.ts
rvilarl: @0xfe this is not working for me even locally, I suggest to follow the two PRs approach, is that fine?

src/annotation.js
rvilarl: I suggest to close this conversation to leave have all the discussion in the first instance of `this.classes` in `annotation`

src/ornament.js
rvilarl: I suggest to close this conversation to leave have all the discussion in the first instance of `this.classes` in `annotation`

src/annotation.js
rvilarl: > Thanks for the explanation Rodrigo. I agree with @wassertim that there's too much logic refactoring going on here. I think we should separate it out and focus on TypeScript.

@0xfe my focus is on TypeScript I think that we have to as far as possible avoid the usage of `any` in order to get the advantages of TypeScript type checking and and to double types in order to simplify the code.

The explanation is long, but the change related to this is minor this why I put it in this CR to avoid two `any`. Here is the related change, if you still want to remove it I will take it out. :)

In `constructor`:
```diff
     this.attrs = {
      ...
-      classes: {},
     };
+    this.classes = {};
```
In `hasClass`:
```diff
-    return this.attrs.classes[className] === true;
+    return this.classes[className] === true;
```
In `addClass`
```diff
-    this.attrs.classes[className] = true;
+    this.classes[className] = true;
```
In `removeClass`:
```diff
-    delete this.attrs.classes[className];
+    delete this.classes[className];
```

wassertim (changes requested):
I will review again when you put out the changes with Object.keys(this.getAttribute('classes')). For now it is a bit too much to understand what is going on

src/annotation.js
0xfe: I guess I don't understand why we need to change the interface (the `getAttribute` call) to avoid `any`. Either way, attributes are generic key/value pairs, and it's one of the few places `any` is useful, because it allows clients to attach arbitrary objects to elements. So I think (as an exception), `any` is useful here.

src/element.ts
0xfe: Yes, it's fine. Thanks for trying. I'd suggest calling out (in github comments) parts where the TS diverges a lot from the JS.

0xfe: Actually for the next PR, can we try the following to make the diffs work? 1) send PR with just rename of file, and add an eslint-diasble for the whole file, 2) send PR with actual changes and eslint-disable removed.

We can even try it for this PR if you're up for it.

src/element.ts
```diff
@@ -9,22 +9,42 @@
...
+import { RenderContext, Style, ElementAttributes } from './types/common';
 
-export class Element {
-  static newID() {
-    return 'auto' + Element.ID++;
+export abstract class Element {
+  protected static ID: number = 1000;
+
+  protected context?: RenderContext;
+
```
wassertim, line 20: do we need all these line breaks among properties?

src/element.ts
```diff
@@ -36,27 +56,28 @@ export class Element {
...
 
   // set music font
-  setFontStack(fontStack) {
+  setFontStack(fontStack: Font[]): this {
     this.fontStack = fontStack;
     this.musicFont = fontStack[0];
     return this;
   }
-  getFontStack() {
+  getFontStack(): Font[] {
```
wassertim, line 64: Line break before would be consistent with the rest of the file

src/element.ts
rvilarl: @0xfe This is a very good idea. I have also decided to go for the 2 steps approach because I believe two other changes where overseen due to the lack of diff. I will set up the questions directly on the code and also a question related on how to proceed on annotations,...

src/element.ts
```diff
@@ -67,26 +88,29 @@ export class Element {
     return this;
   }
 
-  restoreStyle(context = this.context, style = this.getStyle()) {
+  restoreStyle(context: RenderContext | undefined = this.context, style: Style | undefined = this.getStyle()): this {
```
wassertim, line 91:
```ts
context?: RenderContext
```

src/element.ts
rvilarl: done

src/annotation.js
rvilarl: @0xfe I believe this was an easy instance to avoid `any` but accepted, my biggest concern now is your question _Does this work on SVG elements?_ I will open an issue on this one. 

Just to finish this, do you want to leave the access to the protected variable through the function `getAttribute`:
```js
this.getAttribute('classes')
```
or directly (my preference with protected variables)
```js
this.attr.classes
```

src/element.ts
rvilarl: Do not worry, this is included in the 2nd PR with the comments.

src/element.ts
```diff
@@ -9,22 +9,42 @@
...
+
+  protected musicFont: Font;
+
+  protected registry?: Registry;
+
+  static newID(): string {
+    return `auto${Element.ID++}`;
   }
 
-  constructor({ type } = {}) {
```
rvilarl, line 18: @0xfe I removed the parameters because they are not used on any of the references. It is also misleading to accept an `ElementAttributes` and ignore all the other fields. Is that fine?

src/element.ts
```diff
@@ -9,22 +9,42 @@
...
+  constructor() {
     this.attrs = {
       id: Element.newID(),
-      el: null,
```
rvilarl, line 21: @0xfe I also removed `el` because it is not used in the code or the tests. Is this ok with you?

src/annotation.js
0xfe: `getAttribute` is part of the API, so I would like all accesses to go through there. Part of the idea was also that classes could be inherited (it's not implemented yet), but `getAttribute` lets me change the lookup logic more consistently.

src/element.ts
0xfe: My preference is no line breaks. (Breaks between groups of related members is okay.)

wassertim:
This can not be removed. This can be used by end users as a reference to a note in dom

src/annotation.js
rvilarl: `Annotation` is a child of `Element` and has access to its protected function. Just to be sure I leave `attr` `protected` in `Element`, no?

src/element.ts
0xfe: The downstream effect of this is not clear to me -- there are other users that extend `Element`, so it could be called with `super({type: "FooNote"})`. I'd prefer to keep this (as a string.)

src/annotation.js
0xfe: I think that this would also break `registry.js` which relies on `getAttribute(...)` to build indexes.

0xfe: Please let's keep `getAttribute` -- there's a lot of things hanging on this :-)

src/element.ts
0xfe: This is used to associate DOM elements with vexflow elements. (We should have tests for it, but we don't unfortunately.)

src/annotation.js
rvilarl: @0xfe this is understood my question is: do you want to change `attr` to `private` in the base class (Element) or we leave it `protected` like all the others.

src/element.ts
rvilarl: Ok I change to string then.

src/element.ts
rvilarl: This is not possible, I believe,  with a default parameter ( = this.context ).

src/element.ts
rvilarl: ok

src/annotation.js
0xfe: Ah, gotcha. Yes, `private` please, so we force all accesses through `getAttribute`.

src/element.ts
0xfe: Maybe `context: RenderContext = this.context` is enough, no? It breaks if there is no context. (i.e., no need for `undefined`, let the compiler verify that.)

rvilarl: @wassertim _"Parameter cannot have question mark and initializer.ts(1015)"_
@0xfe `this.context` can also be undefined, it is not initilized in the constructor: _"Type 'undefined' is not assignable to type 'RenderContext'.ts(2322)"_

rvilarl:
@0xfe I believe that all the comments are addressed, ready to merge and I will make the second PR with comments!

src/element.ts
```diff
@@ -9,22 +9,33 @@
...
+  protected boundingBox?: BoundingBox;
+  protected fontStack: Font[];
+  protected musicFont: Font;
+  protected registry?: Registry;
+
+  static newID(): string {
+    return `auto${Element.ID++}`;
+  }
+
+  constructor(type: string) {
```
0xfe, line 31: This should be `constructor({type: string} = {})` -- it's a destructured object param.

rvilarl: Ok after some googling to get it to compile I found the way to declare that with TypeScript:
`constructor({ type }: { type?: string } = {})` not very obvious in the first instance :)

src/types/common.d.ts
```diff
@@ -0,0 +1,62 @@
+/** Element attributes. */
+export interface ElementAttributes {
```
0xfe, line 2: Sorry, I didn't notice this in the last change, but let's keep `ElementAttributes` and `ElementStyle` in `element.ts` please.

`RenderContext` is okay here, but I'd like to reorganize it later.

0xfe:
Overall looks good, it's pretty much ready to merge, except for that last comment.

typedoc.ts
wassertim: but docco is still used. Should not it be deleted then?

wassertim:
Looks good for me too. Just one remark about documentation. I missed how it happened to be in the project, but since you have done it, should we remove docco then? A separate issue with a separate PR would be nice.

src/types/common.d.ts
rvilarl: This comes in the 2nd PR with the comments. If I do the change here we lose the diff. I had not forgotten it :)

rvilarl:
> Overall looks good, it's pretty much ready to merge, except for that last comment.

@0xfe ready to merge, the interfaces will be made in the next PR with the comments.

rvilarl:
> Looks good for me too. Just one remark about documentation. I missed how it happened to be in the project, but since you have done it, should we remove docco then? A separate issue with a separate PR would be nice.

I suggested to leave docco until we finish the migration. It could also be removed.

src/types/common.d.ts
0xfe: Great, thanks!

0xfe:
Merging! Thanks!

mscuthbert:
Ok after some googling to get it to compile I found the way to declare that with TypeScript:
constructor({ type }: { type?: string } = {}) not very obvious in the first instance :)

Yeah, it’s REALLY annoying.  I find it more helpful and less confusing just to create an Interface for every instance like this, even if it just has one optional variable since otherwise the sea of curly brackets breaks one’s mind.

interface ElementParams {
  type?: string,
}

constructor({ type }: ElementParams = {})

It is way better though to use this style than positional params IMHO once there are more than two or three parameters.

PR #888, opened 2021-04-07
Add js files to ts-loader

wassertim:
See the corresponding issue for an explanation. Here is also a stackoverflow problem:

https://stackoverflow.com/questions/51860043/javascript-es6-typeerror-class-constructor-client-cannot-be-invoked-without-ne

@rvilarl FYI

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -48,7 +48,7 @@ module.exports = (grunt) => {
       module: {
         rules: [
           {
-            test: /(\.ts?$)/,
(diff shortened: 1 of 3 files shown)


wassertim:
@0xfe this one is important for us to continue the typescript migration. Otherwise tests would fail on creating instances with a class which inherits some not migrated class (for example: Modifier)

0xfe:
Looks good, kinda surprised that this worked before.

PR #877, opened 2021-04-05
Migration/tuning fraction test

rvilarl:
Tests for Tuning and Fraction migrated

PR #889, opened 2021-04-08
Migration2/tuning fraction music tests

rvilarl:
As discussed in #877 second step to simplify tests

### DIFF
--- a/tests/fraction_tests.ts
+++ b/tests/fraction_tests.ts
@@ -7,12 +7,12 @@ import { Fraction } from '../src/fraction';
 import { QUnit, ok, test, notOk, notEqual, strictEqual, deepEqual, notDeepEqual, notStrictEqual } from './declarations';
 
 export const FractionTests = {
-  Start: function () {
+  Start(): void {
     QUnit.module('Fraction');
     test('Basic', FractionTests.basic);
   },
 
-  basic: function () {
+  basic(): void {
     const f_1_2 = new Fraction(1, 2);
     ok(f_1_2.equals(0.5), 'Fraction: 1/2 equals 0.5');
     ok(f_1_2.equals(new Fraction(1, 2)), 'Fraction: 1/2 equals 1/2');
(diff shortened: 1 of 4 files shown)


rvilarl:
The diffs still do not help much, do we really want to do that? Or just leave them like in #877

rvilarl:
Personally I would leave it unchanged.

tests/fraction_tests.ts
```diff
@@ -6,56 +6,52 @@
...
-      ok(tF_n.equals(2), 'Fraction: 4 / 2 equals 2');
-
-      // TODO: Add more detailed tests.
-    },
-  };
-
-  return FractionTests;
-})();
+export const FractionTests = {
+  Start: function () {
```
0xfe, line 10: Can we add return types, etc here?

Are these Github check warnings coming from es-lint warnings? (If so we should either address or silence them.)

rvilarl: This is an es-lint warning. Yes we can add it. This means that you prefer to go for the change? I asked before:

> The diffs still do not help much, do we really want to do that? Or just leave them like in #877

0xfe: Sorry, I don't really understand the proposal. Would the intent be to remove those warnings in the next PR? (If so, that's okay.)

Also, I had another idea, can you break up the file rename and the changes to two different commits in the same PR? Maybe GitHub will recognize that? (Not sure if you already tried that.)

rvilarl: My question is: do we want to go with this second step or just forget it and leave it as it is cancelling this PR?

rvilarl: The question is related to how to proceed. Options:
1 - we do not do this second step in the migration and we leave it with 
```js
export default (function () {
  const TuningTests = {
    Start: function () {
```
2 - we go forward then we had two options
2a -
```js
export const TuningTests = {
    Start: function (): void {
```
2b -
```js
export const TuningTests = {
    Start(): void {
```
What is your preference 1, 2a or 2b? I was doubting due to the fact that diff is not helping but I am open.
If you choose 1 I will cancel this PR.
If you choose 2a or 2b I will make the modification including the return types of the functions. Here I would prefer 2b.

0xfe: 2b definitely.

rvilarl: Done

0xfe:
Looks good. Merging.

PR #891, opened 2021-04-08
Migration/tickable

rvilarl:
This one is a very straight forward one. :)


src/tickable.ts
```diff
@@ -8,8 +8,52 @@ import { Vex } from './vex';
...
+  right: number;
+}
+export interface Space {
+  used: number;
+  mean: number;
+  deviation: number;
+}
+export interface FormatterMetrics {
+  duration: string;
+  freedom: LeftRight;
```
0xfe, line 28: Do we need a new type name `LeftRight` here? Can we type it inline?

src/tickable.ts
```diff
@@ -8,8 +8,52 @@ import { Vex } from './vex';
...
+import { ModifierContext } from './modifiercontext';
+import { Tuplet } from './tuplet';
+import { Voice } from './voice';
+import { Modifier } from './modifier';
+
+export interface LeftRight {
+  left: number;
+  right: number;
+}
+export interface Space {
```
0xfe, line 21: How about `Spacing`?

0xfe:
Looks good overall. There are some unmodified files with lint warnings (for stuff that we're okay with), can you add `eslint-disable-next-line` comments around them in this PR please?

src/tickable.ts
rvilarl: Not really, I change it.

src/tickable.ts
rvilarl: I would leave it as Space. If you search in the code, you will find many references to `space` but none to `spacing`. Your choice anyway.

src/tickable.ts
rvilarl: Now inline.

rvilarl:
> Looks good overall. There are some unmodified files with lint warnings (for stuff that we're okay with), can you add `eslint-disable-next-line` comments around them in this PR please?

@0xfe I would suggest to leave the `warnings` visivisble until the end of the migration. Perhaps we get ideas on the way on how to resolve these `any`

0xfe:
> > Looks good overall. There are some unmodified files with lint warnings (for stuff that we're okay with), can you add `eslint-disable-next-line` comments around them in this PR please?
> 
> @0xfe I would suggest to leave the `warnings` visivisble until the end of the migration. Perhaps we get ideas on the way on how to resolve these `any`

The warnings start to collect and pollute the PRs, and reviews get messier over item. But worse, they start to mask real warnings that we might want to pay attention to.

I see your point, and I think an alternate strategy is `grep eslint-disable-next-line *.ts`, or grep `TODO *.ts`.and reviews. (It's how I typically handle these types of issues.)

Alternatively, we can create a meta issue around `any` and collect instances that we can review at the end.

src/tickable.ts
0xfe: Okay, it's fine.

rvilarl:
Ready!

0xfe:
Great, thanks!

PR #892, opened 2021-04-08
Migration/note

rvilarl:
This one is little more challenging... tests are running

I have not commented all the functions to avoid that diff stops working.


src/note.ts
```diff
@@ -218,30 +464,30 @@ export class Note extends Tickable {
...
   // Don't play notes by default, call them rests. This is also used by things like
   // beams and dots for positioning.
-  isRest() {
+  isRest(): boolean {
     return false;
   }
 
   // TODO(0xfe): Why is this method here?
-  addStroke(index, stroke) {
+  addStroke(index: number, stroke: Stroke): this {
```
rvilarl, line 490: @0xfe what is the reason for the TODO?

src/element.ts
```diff
@@ -37,7 +37,7 @@ export interface ElementStyle {
  */
 export abstract class Element {
   protected static ID: number = 1000;
-  protected context?: RenderContext;
+  public context?: RenderContext;
```
0xfe, line 40: Is there a `context` reference somewhere in the code? (Can it be replaced with `getContext/setContext`?

src/note.ts
```diff
@@ -262,59 +508,59 @@ export class Note extends Tickable {
...
   // Get the glyph associated with this note.
-  getGlyph() {
+  getGlyph(): Glyph {
     return this.glyph;
   }
 
-  getGlyphWidth() {
+  getGlyphWidth(): number {
     // TODO: FIXME (multiple potential values for this.glyph)
     if (this.glyph) {
```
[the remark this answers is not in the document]
0xfe: Yes, this is a mess, the type here is pretty overloaded for historical reasons. For now, let's allow `this.glyph` to have either of these types. (Or `any` is fine here too if there are too many variations. After TS, it'll be easier to clean up.)

src/note.ts
0xfe: Leave it for now. It was around whether it belongs in `Note` vs `StemmableNote`, because it's used just by `StaveNote` and `TabNote`.

src/note.ts
```diff
@@ -250,10 +496,10 @@ export class Note extends Tickable {
...
   // Get and set the target stave.
-  getStave() {
+  getStave(): Stave | undefined {
     return this.stave;
   }
-  setStave(stave) {
+  setStave(stave: Stave): this {
     this.stave = stave;
     this.setYs([stave.getYForLine(0)]); // Update Y values if the stave is changed.
     this.context = this.stave.context;
```
rvilarl, line 505: @0xfe this is the reference to context.

src/element.ts
rvilarl: I have left a comment below in note.js L505 where the reference is.

src/note.ts
0xfe: So if you have `context` as `private`, then you can `this.setContext(this.stave.getContext())`.

src/element.ts
rvilarl: you are right

src/note.ts
rvilarl: I have proceeded with your proposal, context is now private. It was protected.

src/note.ts
```diff
@@ -250,71 +494,65 @@ export class Note extends Tickable {
...
     // TODO: FIXME (multiple potential values for this.glyph)
     if (this.glyph) {
       if (this.glyph.getMetrics) {
         return this.glyph.getMetrics().width;
-      } else if (this.glyph.getWidth) {
-        return this.glyph.getWidth(this.render_options.glyph_font_scale);
+        //} else if (this.glyph.getWidth) {
+        //  return this.glyph.getWidth(this.render_options.glyph_font_scale);
       }
     }
```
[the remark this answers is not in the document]
rvilarl: @0xfe I guess that I should just use any, should I?

 But this is a topic worth investigating. Do you know anyone that could take over this issue?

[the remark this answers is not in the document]
0xfe: We need that `else` for sure. What was the compiler error? Doesn't `getWidth` return `number`?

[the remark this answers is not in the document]
0xfe: Either way, okay to use `any` here with `eslint-disable` and maybe a TODO because we know we need to fix it.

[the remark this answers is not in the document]
rvilarl:
```
ERROR in /home/rodrigo/git/vexflow/src/note.ts
./src/note.ts 554:28-36
[tsl] ERROR in /home/rodrigo/git/vexflow/src/note.ts(554,29)
      TS2339: Property 'getWidth' does not exist on type 'Glyph | GlyphProps'.
  Property 'getWidth' does not exist on type 'Glyph'.

ERROR in /home/rodrigo/git/vexflow/src/note.ts
./src/note.ts 555:26-34
[tsl] ERROR in /home/rodrigo/git/vexflow/src/note.ts(555,27)
      TS2339: Property 'getWidth' does not exist on type 'Glyph | GlyphProps'.
  Property 'getWidth' does not exist on type 'Glyph'.
```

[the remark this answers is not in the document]
0xfe: Okay, yeah, you need to do a type check, something like (off the top of my head, not sure if it compiles, but this is the general idea):

```typescript
else if (typeof this.glyph == 'Glyph') {
  return (this.glyph as Glyph).getWidth(this.render_options.glyph_font_scale);
}
```

0xfe:
This is probably the dirtiest class in VexFlow :-)

src/note.ts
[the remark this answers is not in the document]
rvilarl: The error is complaing about the fact that one of the members in the double type, `Glyth` does not have the function `getWidth` I have defined now the function member in `Glyth`. 

But I have also realised that both types implement `getMetric` so I believe that this else will never be executed.

Now it is compiling and running the tests correctly.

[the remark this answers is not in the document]
0xfe: If you look at `tables.js` there are some data objects that implement a custom `getWidth`, so I think this may not be sufficient. I think we should create an interface:

```typescript
interface WidthGetter {
  getWidth(): number
}
```

Type `glyph` as:

```typescript
glyph: Glyph | GlyphProps | WidthGetter
```

And make this line:

```typescript
else if ('getWidth' in glyph as any) {
  return (this.glyph as WidthGetter).getWidth(this.render_options.glyph_font_scale);
}
```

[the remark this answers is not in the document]
rvilarl: The compiler will then complain about WidthGetter not implementing `getMetrics`. 

I do not know, I do not think that we should twist TS to comply with the code, we should untwist the code with the help of TS. Wouldn't you agree? I would rather leave any on the things we do not figure out how to simplify (this also applies to formatter)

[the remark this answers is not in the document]
0xfe: Are you sure? It shouldn't complain about that because `glyph` is a union type that includes `Glyph | GlyphProps | WidthGetter`.

I'm fine with using `any` here too.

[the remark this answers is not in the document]
rvilarl: The error was:
```
[tsl] ERROR in /home/rodrigo/git/vexflow/src/note.ts(555,27)
      TS2339: Property 'getWidth' does not exist on type 'Glyph | GlyphProps'.
  Property 'getWidth' does not exist on type 'Glyph'.
```
the compound type does not support `getWidth` and then the detail `Glyph` does not have it.

I have included `getWidth` in `Glyph` and now all the tests are runing.

I believe leaving it like it is now will force a check on the next ts migrated files that it is either `Glyph | GlyphProps` and we will see if something else is popping up. For the js file it does not matter because there is no type checking on run time.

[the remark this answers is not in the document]
0xfe: Thanks, right, I understand why the error is happening. We can either use the interface or use `any` types, both which are permissive. But I'm not very comfortable with this approach because I know for certain that it doesn't cover all cases. Also, we shouldn't need to add `getWidth` to `Glyph` to make this work -- it's difficult to reason about that implementation change.

Let's leave `any` and deal with this separately, it's a reasonable place for it.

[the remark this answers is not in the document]
rvilarl: Left as any

rvilarl:
I agree the Glyph topic is drity. We will clean it in the way. 
Ready to merge. I will fix the comments in a migration2 step. :)

rvilarl:
I found two bugs with the comments:
- constructor requires a NoteStruct
- SetBeam has a parameter on child classes.
Both fixed, ready to merge.

src/note.ts
```diff
@@ -13,15 +13,254 @@
...
+import { BoundingBox } from './boundingbox';
+import { Voice } from './voice';
+import { TickContext } from './tickcontext';
+import { ModifierContext } from './modifiercontext';
+import { Modifier } from './modifier';
+import { RenderContext } from './types/common';
+import { Fraction } from './fraction';
+import { Beam } from './beam';
+
+export const GLYPH_PROPS_VALID_TYPES: Record<string, Record<string, string>> = {
```
0xfe, line 27: Does this need to be exported? I'd prefer to move this into `parseNoteStruct` below where it's used.

src/note.ts
```diff
@@ -321,95 +546,117 @@ export class Note extends Tickable {
...
     modifier.setIndex(index);
     this.modifiers.push(modifier);
     this.setPreFormatted(false);
     return this;
   }
 
-  // Get the coordinates for where modifiers begin.
-  getModifierStartXY() {
+  /** Get the coordinates for where modifiers begin. */
+  getModifierStartXY(): Coordinates {
```
0xfe, line 659: I'd prefer not to have a new type here. `Coordinates` is too generic. Let's use something like:

```
getModifierStartXY(): {x: number, y: number}
```

src/note.ts
```diff
@@ -13,15 +13,254 @@
...
+
+export const GLYPH_PROPS_VALID_TYPES: Record<string, Record<string, string>> = {
+  n: { name: 'note' },
+  r: { name: 'rest' },
+  h: { name: 'harmonic' },
+  m: { name: 'muted' },
+  s: { name: 'slash' },
+};
+
+export interface Coordinates {
```
0xfe, line 35: See comment below about removing this.

src/note.ts
```diff
@@ -13,15 +13,254 @@
...
+  boundingBox: { yMin: number; xMin: number; yMax: number; xMax: number };
+  resolution: number;
+  descender: number;
+  familyName: string;
+  lineHeight: number;
+  underlineThickness: number;
+  /**
+   * This property is missing in vexflow_font.js, but present in gonville_original.js and gonville_all.js.
+   */
+  original_font_information?: {
```
0xfe, line 55: This is a good place for `any`, I don't want to tie it too much to the TTF exported fields.

0xfe: Or an interface with just the fields that we use.

0xfe: Perhaps for this whole `Font` interface actually. It's noise in this file.

src/note.ts
```diff
@@ -13,15 +13,254 @@
...
+
+  getMetrics(): GlyphMetrics;
+}
+
+export interface GlyphOptions {
+  fontStack: Font[];
+  category: string;
+}
+
+export interface GlyphMetrics {
```
0xfe, line 106: This can move to `glyph.ts`.

0xfe: Also `GlyphOptions`.

src/note.ts
```diff
@@ -13,15 +13,254 @@
...
+  dots: number;
+  type: string;
+}
+
+export interface DurationCode {
+  common: Type;
+  type: Record<string, Type>;
+}
+
+export interface Type extends KeyProps {
```
0xfe, line 149: Is there a better name than `Type` here?

src/note.ts
```diff
@@ -13,15 +13,254 @@
...
+  code_flag_upstem: string;
+  code_flag_downstem: string;
+  flag: boolean;
+  width: number;
+  text: string;
+  tabnote_stem_down_extension: number;
+  tabnote_stem_up_extension: number;
+  beam_count: number;
+  duration_codes: Record<string, DurationCode>;
+  validTypes: Record<string, NameValue>;
```
0xfe, line 93: Instead of `NameValue`, let's use `Record<string, {name: string}>`.

src/note.ts
```diff
@@ -13,15 +13,254 @@
...
+ *
+ * Some notes have stems, heads, dots, etc. Most notational elements that
+ * surround a note are called *modifiers*, and every note has an associated
+ * array of them. All notes also have a rendering context and belong to a stave.
+ */
+export abstract class Note extends Tickable {
+  protected stave?: Stave;
+  protected render_options: NoteRenderOptions;
+  protected duration: string;
+  protected positions?: TabNotePositon[];
```
0xfe, line 243: Seems like we could actually move this to `TabNote`. It's not used anywhere here, no?

src/note.ts
rvilarl: glyph.ts is still not there. Two options:
1) We move it once it is there
2) we place it common

0xfe: Okay, I think put it in common with a TODO to move it to glyph.ts.

src/note.ts
rvilarl: It does not need to be exported it is a intermediate solution until `table.js` is migrated. This type is also defibed there.

src/note.ts
rvilarl: do you want to move it to 'types/common'? The group from `@types/vexflow`made a great job figuring out the contents of the type. It is a pitty to lose it.

src/note.ts
rvilarl: I agree I never liked that

src/note.ts
0xfe: Yeah, I understand, but I like to actively eliminate unnecessary code. We can always get it back if we need.

src/note.ts
0xfe: Yeah, I see. Is there a better way to do this? I'm not a fan of this duplication. Maybe we can have a separate file for this intermediate stuff and reference that?

rvilarl: `common` with `TODO`?

src/note.ts
rvilarl: ok

src/note.ts
rvilarl: `common` with `TODO`

src/note.ts
rvilarl: Being child of `KeyProps`, I changed to `TypeProps`, better than `Type` anyway.

src/note.ts
rvilarl: deleted

rvilarl:
@0xfe this one should be ready to merge

src/types/common.ts
```diff
@@ -0,0 +1,174 @@
+//** TODO: Resolve duplication with Flow */
+export const GLYPH_PROPS_VALID_TYPES: Record<string, Record<string, string>> = {
```
0xfe, line 2: Can you not do this:

```
export const GLYPH_PROPS_VALID_TYPES: Record<string, Record<string, string>> = Flow.getGlyphProps.validTypes
```

0xfe: If this works, then I'd say we can put it in `Note`.

Either way, this is just a comment, we can do this in a followup PR.

rvilarl: I would wait to see other things that come out of table.

src/types/common.d.ts
```diff
@@ -1,45 +0,0 @@
-/** Contexts common interface */
-export interface RenderContext {
```
0xfe, line 2: I think we'll need both `.d.ts` and `.ts`, no? The `.d.ts` should only have interfaces with no code.

I would suggest leaving this file as `src/types/common.d.ts` and adding a `src/common.ts` with just the assignment (for the glyph props.)

0xfe:
Thanks Rodrigo. Looks good. Two minor comments left.

rvilarl:
Let us merge and I will PR some minor quickwins! :)

src/types/common.d.ts
rvilarl: I really do not know. I tried the declarations on tests with .d.ts and I did not get it to work. Then @wassertim showed that it work with .ts. If we have .d.ts and .ts they have to be in diferent directories otherwise #import does not tell the difference.
We also decided to put the exports on the normal files, I mean we did not do .d.ts for element with the interfaces like @wassertim did in his first migration. We had a discussion about that and you said that you prefered them self contained.

rvilarl: It is your choice, I just wanted to higlight that it is inconsistent with the others.

rvilarl: Thinking more about that I find it inconvenient to have them two imports one from `./common` and one from `./types/common` Then perhaps `types/common.d.ts` with interfaces and `flow.ts`where we start porting little by little out of `table.ts`?

rvilarl: forget about it `common.ts` is fine

rvilarl:
@0xfe Hopefully ready :)

src/accidental.js
```diff
@@ -56,7 +56,7 @@ export class Accidental extends Modifier {
         }
         prevNote = note;
       }
-      if (stave !== null) {
+      if (stave) {
```
wassertim, line 59: that is not the same as:
```javascript
if (stave !== null)
```
I would be more careful with these changes

src/element.ts
```diff
@@ -37,7 +37,7 @@ export interface ElementStyle {
  */
 export abstract class Element {
   protected static ID: number = 1000;
-  protected context?: RenderContext;
+  private context?: RenderContext;
```
wassertim, line 40: you should order the fields. I would suggest
- public
- protected
- private

Otherwise it is a mess

src/common.ts
```diff
@@ -0,0 +1,8 @@
+//** TODO: Resolve duplication with Flow */
+export const GLYPH_PROPS_VALID_TYPES: Record<string, Record<string, string>> = {
```
wassertim, line 2: What is this needed for? Looks really strange. Better not to touch at all, than having it half done

src/accidental.js
rvilarl: That was done on purpose. I believe that this is the right way to check that an `object` is not `null` or `undefined`

src/element.ts
0xfe: Agreed with this. We can fix in the next PR with comments. (Also maybe group it in some sane way.)

src/common.ts
0xfe: I think let's remove this in the next PR too. I don't like it either.

src/accidental.js
0xfe: Yes, this is fine. Thanks.

0xfe:
Thanks Rodrigo. This was a tough one. Looking forward to some of the follow ups for this file.

PR #894, opened 2021-04-10
Migration/tickcontext

rvilarl:
Quick win :)


src/tickcontext.ts
```diff
@@ -7,25 +7,72 @@
 import { Vex } from './vex';
 import { Tickable } from './tickable';
 import { Fraction } from './fraction';
+import { Note } from './note';
 
+export interface Metrics {
```
0xfe, line 12: Didn't we just create a `Metrics` for `Note` -- is this different?

rvilarl: This is right. It had a minor impact in `note`.

PR #895, opened 2021-04-10
Migration/voicegroup

rvilarl:
Quick win! :)


src/voicegroup.ts
```diff
@@ -1,24 +1,32 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 
+import { ModifierContext } from './modifiercontext';
 import { Vex } from './vex';
+import { Voice } from './voice';
 
 /** @constructor */
```
wassertim, line 7: This comment does not make sense

src/voicegroup.ts
```diff
@@ -1,24 +1,32 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 
+import { ModifierContext } from './modifiercontext';
```
wassertim, line 3: since you don't use a constructor of ModifierContext, I think it would be wiser just only import a type:
```typescript
import type { ModifierContext } from './modifiercontext';
```

Same for Voice

src/voicegroup.ts
rvilarl: It was already there this is jsdoc. I wanted to check what is produces with typedoc. The comments I normally rearrange on a second step to avoid that diff stops working.

0xfe: Yes, this is from the pre-ES6 JS days when there was no notion of a constructor. You can remove this.

src/voicegroup.ts
0xfe: Good idea.

src/voicegroup.ts
rvilarl: @0xfe removing the comment I have just realised that this class `voicegroup` has no parent. Is that ok or should it extend `element`?

0xfe: This is intentional. They're just containers for formatting logic.

0xfe:
Thanks!

src/voicegroup.ts
rvilarl: Then this one is ready.

rvilarl:
Ready to merge

PR #896, opened 2021-04-10
Migration/gracenotegroup

rvilarl:
Quick win! :)


src/gracenotegroup.ts
```diff
@@ -14,19 +14,40 @@ import { Beam } from './beam';
 import { StaveTie } from './stavetie';
 import { TabTie } from './tabtie';
 import { StaveNote } from './stavenote';
+import { Note } from './note';
 
-// To enable logging for this class. Set `Vex.Flow.GraceNoteGroup.DEBUG` to `true`.
-function L(...args) {
+export interface GraceNoteGroupFormatOptions {
```
wassertim, line 19: if it is for internal use only, should it be exported?

src/gracenotegroup.ts
```diff
@@ -90,7 +108,7 @@ export class GraceNoteGroup extends Modifier {
     this.preFormatted = false;
 
     this.show_slur = show_slur;
-    this.slur = null;
+    this.slur = undefined;
```
wassertim, line 111: changing null to undefined can be a breaking change. I've seen a lot of places with:
```javascript
if (a === null) ...
```
which is not the same as:
```javascript
if (a === undefined) ...
```

src/gracenotegroup.ts
rvilarl: Hi good to see you! I believe that anyone using the `format` will require this definition.

src/gracenotegroup.ts
rvilarl: I agree I have already changed 
```
if (a === null)
```
by 
```
if (!a)
```
If we also leave `null` the declarations, parameters and return values will get very complex.

wassertim: same problem here. I would not change that:
```javascript
if (a === null)
```
is not the same as:
```javascript
if (!a)
```

wassertim: Maybe it was not intentional @0xfe? Then it is a fine change

src/gracenotegroup.ts
0xfe: I'm not a fan of these single-field structures, it makes it harder to read the code as you jump around through types. Let's just use inline types here.

src/gracenotegroup.ts
```diff
@@ -14,19 +14,40 @@ import { Beam } from './beam';
...
+  protected beams: Beam[];
+
+  static get CATEGORY(): string {
     return 'gracenotegroups';
   }
 
-  // Arrange groups inside a `ModifierContext`
-  static format(gracenote_groups, state) {
+  /** Arranges groups inside a `ModifierContext`. */
+  static format(gracenote_groups: GraceNoteGroup[], state: GraceNoteGroupFormatOptions): boolean {
```
0xfe, line 50: Prefer: `state: {  left_shift: number }`

src/gracenotegroup.ts
0xfe: This case is okay (though I agree with @wassertim that we should be careful with these.) `undefined` is semantically more correct here than `null`.

rvilarl:
This one should be now ready for merge.

src/gracenotegroup.ts
```diff
@@ -14,19 +14,33 @@ import { Beam } from './beam';
 import { StaveTie } from './stavetie';
 import { TabTie } from './tabtie';
 import { StaveNote } from './stavenote';
+import { Note } from './note';
 
-// To enable logging for this class. Set `Vex.Flow.GraceNoteGroup.DEBUG` to `true`.
-function L(...args) {
+// To enable logging for this class. Set `GraceNoteGroup.DEBUG` to `true`.
+function L(...args: string[]) {
```
0xfe, line 20: This should be `any`. It goes right into `console.log`.

rvilarl: ok, done

rvilarl:
Ready to merge.

0xfe:
Thanks!

PR #898, opened 2021-04-11
Migration/stemmablenote

rvilarl:
I have decided to follow this path to get to StaveNote. :)


src/stemmablenote.ts
```diff
@@ -178,66 +200,60 @@ export class StemmableNote extends Note {
...
     } else {
-      return this.stave.getYForBottomText(textLine);
+      return this.stave?.getYForBottomText(textLine) || 0;
     }
   }
 
-  hasFlag() {
-    return Flow.getGlyphProps(this.duration).flag && !this.beam;
+  hasFlag(): boolean {
+    return Flow.getGlyphProps(this.duration).flag && !this.beam && this.flag;
```
rvilarl, line 240: @0xfe I find it strange that this.flag was not in the condition. I good also say that this.flag nevers get initialised. Is this a bug?

src/stemmablenote.ts
```diff
@@ -141,15 +163,15 @@ export class StemmableNote extends Note {
     if (this.flag) {
       this.buildFlag();
     }
```
rvilarl, line 165: @0xfe I would say that flag is never initialised. Is this a bug?

src/stemmablenote.ts
0xfe: I'm not exactly sure -- I have to dig in. I'd say add a TODO, but don't change the logic.

src/stemmablenote.ts
```diff
@@ -178,66 +200,60 @@ export class StemmableNote extends Note {
...
 
-  getYForBottomText(textLine) {
+  /** Gets the `y` value for the bottom modifiers at a specific `textLine`. */
+  getYForBottomText(textLine: number): number {
     const extents = this.getStemExtents();
     if (this.hasStem()) {
       return Math.max(
-        this.stave.getYForTopText(textLine),
-        extents.baseY + this.render_options.annotation_spacing * textLine
+        this.stave?.getYForTopText(textLine) || 0,
```
0xfe, line 231: These `|| 0`s are dangerous.

src/stemmablenote.ts
rvilarl: Without this check stavenote is now crashing. There must be a difference that I have introduced. But the reality is that stavenote uses hasFlag to safely access the flag.

rvilarl: This is once again this getGlyphProp. Could you provide a description of the aim of this function?

src/stemmablenote.ts
rvilarl: || 0 applied to a number will result in 0 if the number is undefined, null, 0, NaN.
Should I go for ?? 0 that only makes the change is number is undefined or null? Or you prefer a different approach?

0xfe: Yes, I understand, but it's an implementation change and it's hard to reason about what happens. Somtimes a different layer of code deals with undefined/null. Or sometimes, it should just crash (with console error/exception) instead of rendering something incorrect.

This applies to `stave?` too -- it changes the behavior and expectations of the program. Those should stay `stave`. (You can use `stave?` if you see something like `if (stave) ... else undefined` or `a = f ? c : null`

rvilarl: @0xfe explanation:
- `stave` is a member of `Element`
- `stave` is not assigned a value in the `constructor` of `Element`. For this reason it is declared `stave?: Stave;`
- due to the fact that TS knows that `stave` can be `undefined` it forces us to say how do we want to handle that. This is not the case in JS.

The approach that I have selected is:
`this.stave?.getYForTopText(textLine) || 0`
basicly if stave is undefined or getYFortTopText does not return a valid number then 0. I will change it to the mos restrictive 
`this.stave?.getYForTopText(textLine) ?? 0`

src/stemmablenote.ts
rvilarl: Ok I will revert the change. I have seen that the condition used in `StaveNote` to see if there is a `flag` is `hasFlag`. I have replaced the condition is this class to the same check `hasFlag` and the tests pass correctly.

rvilarl:
I would say ready to merge.

src/stemmablenote.ts
rvilarl: I have checked that this is not a bug, `buildFlag` is called in `StaveNote` however `this.flag` could be replaced by `this.hasFlag()` for consistency with the other checks, your choice.

src/stemmablenote.ts
0xfe: If it forces to handle that `stave` can be undefined then I would prefer `this.stave?.getYForTopText(textLine) ?? undefined` as opposed `?? 0` because then it fails loudly instead of silently returning 0 with a weird formatting problem that is difficult to diagnose. It's the zero-default which is the problem here.

Alternatively, we can:

- make `stave` non-optional, and initialize to `null`.
- add an assertion before the line, something like, `assertDefined(this.stave)` (we'll have to build some standard assertions.)

src/stemmablenote.ts
rvilarl: Modified to use consitently the same check.

rvilarl:
Ready to merge

rvilarl:
@0xfe are you fine with this class now?

src/stemmablenote.ts
```diff
@@ -178,66 +199,60 @@ export class StemmableNote extends Note {
...
-  // Get the `y` value for the top/bottom modifiers at a specific `textLine`
-  getYForTopText(textLine) {
+  /** Gets the `y` value for the top modifiers at a specific `textLine`. */
+  getYForTopText(textLine: number): number {
     const extents = this.getStemExtents();
     if (this.hasStem()) {
       return Math.min(
-        this.stave.getYForTopText(textLine),
-        extents.topY - this.render_options.annotation_spacing * (textLine + 1)
+        this.stave?.getYForTopText(textLine) ?? 0,
```
0xfe, line 217: These `?? 0` are still a problem. The following is better:

``` typescript
getYForTopText(textLine: number) : number {
  if (this.stave === null) throw new Vex.RERR('Uninitialized', 'Stave not set');

  // No need for ?? anymore
  if (this.hasStem()) {
    const extents = this.getStemExtents();
    if (extents === null) throw new Vex.RERR('Uninitialized', 'No stem extents available')
    return Math.min(this.stave.getYForTopText(textLine),
      extents.topY - this.render_options.annotation_spacing * (textLine + 1))
  } else {
    return this.stave.getYForTopText(textLine);
  }
}
```

Again, failing loud is better than silently doing the wrong thing.

rvilarl: Solved I have used the same message used in note.

rvilarl:
Ready to merge hopefully.

0xfe:
Looks good, thanks!

PR #899, opened 2021-04-13
Migration/glyph

rvilarl:
I decided to go for glyph.
Surprisignly the results of few pedal tests pop up at the beginning of the output of flow.html

image


src/glyph.ts
```diff
@@ -5,20 +5,133 @@ import { Flow } from './tables';
 import { Element } from './element';
 import { BoundingBoxComputation } from './boundingboxcomputation';
 import { BoundingBox } from './boundingbox';
+import { Font } from './smufl';
+import { RenderContext } from './types/common';
+import { Stave } from './stave';
+
+export interface KeyProps {
```
0xfe, line 12: Only `GlyphProps` belongs here. Let's leave 'KeyProps` and `TypeProps` in common for now.

src/glyph.ts
```diff
@@ -5,20 +5,133 @@ import { Flow } from './tables';
...
-  let x;
-  let y;
+export interface GlyphProps {
+  code_head: string;
+  dot_shiftY: number;
+  position: string;
+  rest: boolean;
+  line_below: number;
+  line_above: number;
+  stem_up_extension: never;
```
0xfe, line 59: There's a lot of shared state with `TypeProps` -- why not inherit?

src/glyph.ts
```diff
@@ -5,20 +5,133 @@ import { Flow } from './tables';
...
+  y_min: number;
+  y_max: number;
+  ha: number;
+  o: string;
+  leftSideBearing?: number;
+  advanceWidth?: number;
+  cached_outline?: string[];
+}
+
+export interface GlyphLookup {
```
0xfe, line 108: This can be inline, no need for a separate type.

src/glyph.ts
rvilarl: There is a lot of shared but also differences. I would not go into that now, I would first integrate `getGlyphProps` from flow into this class. Wouldn´t you agree?

image

src/glyph.ts
rvilarl: ok

src/glyph.ts
rvilarl: ok

rvilarl:
Ready to merge unless you want to merge `getGlyphProps` from `Flow` in `Glyph`. This we could also do in a second PR.

src/glyph.ts
```diff
@@ -5,20 +5,94 @@ import { Flow } from './tables';
...
+  x_max: number;
+  x_shift: number;
+  y_shift: number;
+  scale: number;
+  ha?: number;
+  outline: string[];
+  font?: Font;
+}
+
+export interface FontGlyph {
```
0xfe, line 62: Is there a way to share this state? e.g., one derives from the other? (or one contains the other.)

rvilarl: @0xfe if you want to clarify the font topic I suggest to include the Font class in this PR, it is not very big and it would be the best way to check that both classes interact fine together. I would rename `smufl.js`to `font.ts`. Is that fine?

rvilarl: I have gone for it :)

rvilarl: Fonts types simplified

src/font.ts
```diff
@@ -8,37 +7,59 @@ import { PetalumaMetrics } from './fonts/petaluma_metrics';
 import { CustomFont } from './fonts/custom_glyphs';
 import { CustomMetrics } from './fonts/custom_metrics';
 
+export interface FontData {
```
0xfe, line 10: +1 this is a much better place for this.

src/glyph.ts
0xfe: This is way better, thanks!

0xfe:
Merging!

PR #905, opened 2021-04-16
Migration/stavenote

rvilarl:
The addModifier inconsistency generates a lot of output on the testst


rvilarl:
I have adapted all src files and all tests files. Not light!
@0xfe your turn to check

src/element.ts
```diff
@@ -131,6 +132,17 @@ export abstract class Element {
   /** Draws an element. */
   abstract draw(element?: Element, x_shift?: number): void;
 
+  /** Gets SVGElement.  */
+  getSvgElement(): SVGElement | undefined {
```
0xfe, line 136: Unclear why we have this getter/setter here. We have `setAttribute` and `getAttribute` that maintains the indexes too.

src/note.ts
```diff
@@ -261,6 +262,7 @@ export abstract class Note extends Tickable {
     }
 
     // Set note properties from parameters.
+    this.keys = noteStruct.keys;
```
0xfe, line 265: Why do we need `keys` in `Note` (which is more of an abstract concept, unlike `StaveNote`.)

src/stavenote.ts
```diff
@@ -14,22 +14,95 @@ import { Flow } from './tables';
...
+  isrest: boolean;
+  stemDirection: number;
+  stemMax: number;
+  stemMin: number;
+  voice_shift: number;
+  is_displaced: boolean;
+  note: StaveNote;
+}
+
+export interface StaveNoteStruct {
```
0xfe, line 57: Wow, this is a lot. Are all of these really the options for `StaveNote`? I can't imagine why we have `font`, `subscript`, `superscript`, etc. here.

src/stavenote.ts
```diff
@@ -14,22 +14,95 @@ import { Flow } from './tables';
...
+  ignore_ticks: boolean;
+  smooth: boolean;
+  glyph: string;
+  font: Font;
+  subscript: string;
+  superscript: string;
+  text: string;
+  positions?: { str: number; fret: number }[];
```
0xfe, line 65: This is a `TabNote` feature -- I don't think it should be here.

0xfe:
Oof! :-)

Let's start by resolving `StaveNoteStruct` -- seems like it can be slimmed down considerably.

(Tests look reasonable, BTW.)

src/note.ts
rvilarl: Because `keys` are being used by several `Modifier` wich have only acces to `Note`:
- `Accidental`
- `FretHandFinger`
- `GraceNoteGroup`
- `StringNumber`

src/element.ts
rvilarl: I wanted to avoid the usage of `any` in `setAttribute` `getAttribute`. I believe also that we have a mixture of things in `attrs` that not fit together. Normally `attrs` are only strings.

0xfe: Gotcha, it's okay for `attrs` to be `string -> any` -- it's intentional and allows users to attach arbitrary objects to elements (which they already do.)

src/stavenote.ts
rvilarl: I have left only the elements used and I have also extended it from `NoteStruct` to avoid duplication. If we need some element more in the migration we can always add it :)

src/note.ts
0xfe: Right -- all those modifiers only apply to `StaveNote`, they shouldn't be using `Note` as their interface.

src/stavenote.ts
rvilarl: removed

src/stavenote.ts
0xfe: Thanks, looks way better -- can you eyeball the tests and see if everything renders correctly?

rvilarl:
Ready to continue, let us get it through! Yes modifying all the tests was tedious.

0xfe:
@rvilarl I think you missed my comment about `get/setSVGElement`. We should remove it please.

src/note.ts
rvilarl: `Modifier` has an attribute which is an array `note` this attribute has to be of type `Note` because not all the elements in this array are of type `StaveNote`. Then the classes which derive from `Modifier` use this `note` to access the `keys`.
Moving `keys` to `Note` avoid making more complicated handling.
@0xfe how would you prefer to resolve the issue? Do you want me to include `Modifier` in this PR to see the issue?

src/element.ts
rvilarl: However the `attrs.classes` have their our methods: `addClass`, `removeClass`, `hasClass`.
Actions:
1 Remove `get/setSvgElement`
2 Change `get/setAttribute` to `any`
@0xfe do you want both actions or only 2?

rvilarl:
@0xfe OK then it is action 1 & 2, remove and `any`, what about `keys`

src/element.ts
rvilarl: done

rvilarl:
@0xfe only the `keys` conversation is open.

src/note.ts
rvilarl: @0xfe is my explanation above clear?

0xfe: Can we do a separate PR for Modifier first, and work on it, and then come back to this. It may simplify the end result.

rvilarl: I think that it is not so easy to see the interaction in isolation. If we forget the tests this is not so complex. I will include Modifier here.

rvilarl: @0xfe Modifier is now there

rvilarl: @0xfe look at:

https://github.com/0xfe/vexflow/blob/6527cb7777d185a246b022ff3e9e10e0bac4af7f/src/gracenotegroup.ts#L63-L65
 
this `note` variable type is `Note`

src/modifier.ts
```diff
@@ -51,126 +69,134 @@ export class Modifier extends Element {
...
+
+  getXShift(): number {
     return this.x_shift;
   }
 
   // Render the modifier onto the canvas.
-  draw() {
-    this.checkContext();
+  draw(): void {
     throw new Vex.RERR('MethodNotImplemented', 'draw() not implemented for this modifier.');
```
0xfe, line 185: Not sure why `checkContext` was removed here.

src/note.ts
0xfe: Beams can apply to `TabNotes` too that's why there's no check in `Beam`.

For `GraceNoteGroup` the check is there for head displacement, but it can also apply to `TabNotes`.

0xfe: (TabNotes don't have heads.)

src/modifier.ts
rvilarl: To reduce size of code, I thought it makes no sense checking that the context is there before throwing a `RERR`.
I also spent some time asking myself why we would do that.

0xfe: This will cause a crash for logical elements that don't render anything, but just reserve space. So they will have a context, but not draw anything. In general, better not to change logic in these commits -- it's very hard to predict the downstream effects.

rvilarl: ok

src/note.ts
rvilarl: `Beam` is showing that this is a difficult one. Thoughts:
- `NoteStruct`contains an attribute `keys` and `Note` is using it already before the change. (this would speak for storing keys in Note already)
- I just realised in the migration that `Beam`uses `keyProps`. The piece of code above does not compile due to the fact that `note` is of type `Note`
-  `keys` and `keyProps` should be in the same class. It is wrong now to have them in different classes.

Could not `Note` contain `keys` and `keyProps` arrays with just one element on each?
Would you model that in a different way?

0xfe: Okay, I agree this is complicated to tease apart. Let's keep `keys` here then, and figure it out later.

src/stavenote.ts
```diff
@@ -486,7 +552,7 @@ export class StaveNote extends StemmableNote {
 
       // Calculate displacement of this note
       const line = props.line;
-      if (lastLine === null) {
+      if (!lastLine) {
```
0xfe, line 555: This is bad because `lastline` can be `0`. Please check the other false tests too for `number`.

src/note.ts
rvilarl: Then I will put `keyProps` together with it. Otherwise we will have the same problem in `Beam`.

0xfe: Okay, let's see how that looks.

rvilarl: I agree, we can move them back to `StaveNote` if we figure something different out later.
Done!

src/stavenote.ts
rvilarl: That is right, checked againt undefined now.

rvilarl:
Hopefully ready! :)

src/element.ts
```diff
@@ -18,6 +18,7 @@ export interface ElementAttributes {
   // eslint-disable-next-line
   [name: string]: any;
   id: string;
+  el?: SVGElement;
```
0xfe, line 21: Let's make this `any`. I'm sure that we use this for more than just `SVGElement`.

src/stavenote.ts
```diff
@@ -589,44 +653,46 @@ export class StaveNote extends StemmableNote {
...
 
-  getStemX() {
+  getStemX(): number {
     if (this.noteType === 'r') {
       return this.getCenterGlyphX();
     } else {
       // We adjust the origin of the stem because we want the stem left-aligned
       // with the notehead if stemmed-down, and right-aligned if stemmed-up
-      return super.getStemX() + getStemAdjustment(this);
+      return super.getStemX() + (this.stem_direction ? Stem.WIDTH / (2 * -this.stem_direction) : 0);
```
0xfe, line 680: Why did you replace `getStemAdjustment`?

0xfe (commented):
Hey -- whew, okay, overall it looks good! Thanks for working on this! Just a few minor comments left.

But: this change is really hairy. So can you do a side-by-side visual comparison of the tests please?

The way I do it:

- Create two separate directories and clone vexflow into them
- Git checkout a pre-typescript version on one of them
- Run `grunt` on both directories
- Open two tabs in your browser with `DIR/tests/flow.html` for each directory
- Scroll down and eyeball differences

If you spot any, there's probably some bug in the PR. If you can't find them you can post screenshots here, and I can help you track them down.

src/stavenote.ts
rvilarl: This was a global non exported const function in this file used only once here. Within this function that did not belonged to the class it was also not possible to check if `this.stem_direction` was not `undefined`: 

```
const getStemAdjustment = (note) => Stem.WIDTH / (2 * -note.getStemDirection());
```

0xfe: Okay, thanks, makes sense.

src/element.ts
rvilarl: done

rvilarl:
> Hey -- whew, okay, overall it looks good! Thanks for working on this! Just a few minor comments left.
> 
> But: this change is really hairy. So can you do a side-by-side visual comparison of the tests please?
> 
> The way I do it:
> 
> * Create two separate directories and clone vexflow into them
> * Git checkout a pre-typescript version on one of them
> * Run `grunt` on both directories
> * Open two tabs in your browser with `DIR/tests/flow.html` for each directory
> * Scroll down and eyeball differences
> 
> If you spot any, there's probably some bug in the PR. If you can't find them you can post screenshots here, and I can help you track them down.

@0xfe hi Mohit, I have the visual test running. First the good news! I have the visual test running!
Then the bad news it seems that it has not been used for a while!

rvilarl:
I have decided to base my investigation in e9209f6 and this one reports 36 errors already comparing its own blessed to current.

rvilarl:
I have compared e9209f6 with migration/stavenote with master and there are only 6 differences introduced in the migration process. The same ones that I realised as significant in the comparison between  e9209f6 and migration/stavenote:

- some vibrato, some location of text and part of the partiture missing on lyrics tests.

I am going to look into the shift so that we can finally check in this PR. Then we can investigate the 6 particular topics.

@0xfe Do you consider  e9209f6 a good commit for pre Typescript migration reference?

rvilarl:
@0xfe I will create smaller PRs based on this one in order to identify where the visual shift is comming from.

0xfe:
> @0xfe I will create smaller PRs based on this one in order to identify where the visual shift is comming from.

Sounds good.

rvilarl:
No longer required

PR #906, opened 2021-04-16
Migration/textfont

rvilarl:
Let us finish with fonts
@0xfe I can add the removal of legacy fonts to this PR


rvilarl:
Help with `textWidthCacheInstance` type appreciated!

AaronDavidNewman:
Do you mean which type should we use?  It is a 2-level hash, whose value is will be a number (float).  Does that help?  Sorry my typescript knowledge is shallow.

rvilarl:
> Do you mean which type should we use? It is a 2-level hash, whose value is will be a number (float). Does that help? Sorry my typescript knowledge is shallow.

Thanks yes this helps and the two levels are indexed by string?

AaronDavidNewman:
Yes, correct.

rvilarl:
@AaronDavidNewman  any resolved, thank you!

src/textfont.ts
```diff
@@ -9,33 +9,93 @@
...
+  protected maxSizeGlyph: string;
+  protected weight: string;
+  protected style: string;
+  protected fontCacheKey: string = '';
+
+  protected static registryInstance: TextFontRegistry[];
+  protected static textWidthCacheInstance?: Record<string, Record<string, number>>;
+
+  protected size: number;
+  protected attrs: TextFontAttributes;
```
0xfe, line 76: Let's inline this.

src/textfont.ts
```diff
@@ -9,33 +9,93 @@
...
   // Getter of an array of available fonts.  Applications may register their
   // own fonts and the metrics for those fonts will be available to the
   // application.
-  static get fontRegistry() {
+  static get fontRegistry(): TextFontRegistry[] {
     if (!TextFont.registryInstance) {
       TextFont.registryInstance = [];
       TextFont.registryInstance.push({
-        name: 'Roboto Slab',
+        name: 'RobotoSlab',
```
0xfe, line 98: Is this intentional? (The space is missing.)

src/textfont.ts
```diff
@@ -196,59 +256,55 @@ export class TextFont {
...
-    }
-    this.weight = typeof this.weight === 'undefined' ? '' : this.weight;
-    this.style = typeof this.style === 'undefined' ? '' : this.style;
     this.updateCacheKey();
   }
   // Create a hash with the current font data, so we can cache computed widths
-  updateCacheKey() {
-    this.fontCacheKey = this.family + '-' + this.size + '-' + this.weight + '-' + this.style;
+  updateCacheKey(): void {
+    this.fontCacheKey = this.family + '-' + this.size.toString() + '-' + this.weight + '-' + this.style;
```
0xfe, line 284: Better to use a template string here.

src/textfont.ts
```diff
@@ -9,33 +9,93 @@
 import { Vex } from './vex';
 import { PetalumaScriptTextMetrics } from './fonts/petalumascript_textmetrics';
 import { RobotoSlabTextMetrics } from './fonts/robotoslab_textmetrics';
+export interface TextFontMetrics {
+  advanceWidth: number;
+  ha: number;
+}
+
+export interface TextFontGlyph {
```
0xfe, line 17: This looks just like the other glyph metrics. Can we share?

src/textfont.ts
AaronDavidNewman: Hi, the space is in the font name, probably a mistake:

`<link href="https://fonts.googleapis.com/css2?family=Roboto+Slab:wght@500&display=swap" rel="stylesheet">`

src/textfont.ts
rvilarl: Yes, with minor changes in `FontGlyph` done.

src/textfont.ts
rvilarl: done

src/textfont.ts
rvilarl: Not really. Reversed.

src/textfont.ts
rvilarl: Agree

rvilarl:
Ready

0xfe:
Looks good!

PR #907, opened 2021-04-16
Migration/stem

rvilarl:
Another quick win!


src/stem.ts
```diff
@@ -7,116 +7,156 @@
...
 
 // To enable logging for this class. Set `Vex.Flow.Stem.DEBUG` to `true`.
-function L(...args) {
+function L(
+  // eslint-disable-next-line
+  ...args: any[]) {
   if (Stem.DEBUG) Vex.L('Vex.Flow.Stem', args);
 }
 
+export interface StemStruct {
```
0xfe, line 19: Let's call all options parameters `*Options` so 'StemOptions' here. (`NoteStruct` is a legacy wart to clean up.)

0xfe:
Looks good. Just rename `StemStruct` please, and I can merge.

src/stem.ts
rvilarl: ok

rvilarl:
Ready.

0xfe:
Looks good.

PR #910, opened 2021-04-17
Migration/fixFlags

rvilarl:
Fix flags bug


rvilarl:
@0xfe this fixes the flag topic. It is related to `undefined` and `null`

0xfe:
Thanks for fixing so quickly. Away from laptop now, but will review in the
morning.

0xfe:
Thanks. Merging this, however that `null -> undefined` change makes me nervous because of the downstream implications.

PR #915, opened 2021-04-20
Migration/stavenote1

rvilarl:
First step omitting migration.
The all the changes in this PR were already review in #905 except the changes included now in stavenote.js:
- addMofidier parameters
- keys and keyProps moved to Note

There is no visual change when compared to master.


rvilarl:
@0xfe this is the "small" PR. A lot of files changes (most of them due to the parameter order topic in addModifier).

You have already checked everything except `stavenote.js` I think it would make sense to merge it and then try to figure out afterwards what is causing the shift of drawing in #905 

This change produces no visual error. **Let us merge it once you check `stavenote.js`!**

0xfe:
Looks good! Much simpler to review :-)

PR #916, opened 2021-04-20
Migration/ref images

rvilarl:
@0xfe Checking against reference images. Instructions:
- checkout master
- run `npm run reference`
- checkout the branch you are working on
- run `npm run test:reference` to check that you did not introduce any visual change compared to master

You can use other commits than master as reference.
The script now also checks that ImageMagick is installed. Failing if it is not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+images-ref
+images-ref/
 build
 build/
 *.lock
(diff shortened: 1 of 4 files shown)


0xfe:
Looks great. Can you update the Wiki too with these new instructions please?

PR #917, opened 2021-04-20
fix getYForTopText

rvilarl:
Migration visual issues resolved. @0xfe you were right a proble with `getYForTopText` but in `StemmableNote`

### DIFF
--- a/src/stemmablenote.ts
+++ b/src/stemmablenote.ts
@@ -171,7 +171,7 @@ export abstract class StemmableNote extends Note {
   getStemExtension(): number {
     const glyph = this.getGlyph();
 
-    if (this.stem_extension_override != null) {
+    if (this.stem_extension_override != undefined) {
       return this.stem_extension_override;
     }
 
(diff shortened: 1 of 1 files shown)


src/stemmablenote.ts
```diff
@@ -194,29 +194,27 @@ export abstract class StemmableNote extends Note {
 
   /** Gets the `y` value for the top modifiers at a specific `textLine`. */
   getYForTopText(textLine: number): number {
-    if (!this.stave) throw new Vex.RERR('NoStave', 'No stave attached to this note.');
```
0xfe, line 197: Let's keep this so we don't need the `?` below.

src/stemmablenote.ts
```diff
@@ -194,29 +194,27 @@ export abstract class StemmableNote extends Note {
...
     const extents = this.getStemExtents();
-    if (extents) {
+    if (this.hasStem()) {
       return Math.min(
-        this.stave.getYForTopText(textLine),
-        extents.topY - this.render_options.annotation_spacing * (textLine + 1)
+        this.stave?.getYForTopText(textLine) ?? 0,
```
0xfe, line 200: If you keep the `if (!this.stave)` above, you don't need the `stave?` or the `?? 0` -- these are silent failures.

src/stemmablenote.ts
```diff
@@ -194,29 +194,27 @@ export abstract class StemmableNote extends Note {
...
+        extents?.topY ?? 0 - this.render_options.annotation_spacing * (textLine + 1)
       );
     } else {
-      return this.stave.getYForTopText(textLine);
+      return this.stave?.getYForTopText(textLine) ?? 0;
```
0xfe, line 204: Same here.

src/stemmablenote.ts
```diff
@@ -194,29 +194,27 @@ export abstract class StemmableNote extends Note {
 
   /** Gets the `y` value for the top modifiers at a specific `textLine`. */
   getYForTopText(textLine: number): number {
-    if (!this.stave) throw new Vex.RERR('NoStave', 'No stave attached to this note.');
     const extents = this.getStemExtents();
-    if (extents) {
+    if (this.hasStem()) {
```
0xfe, line 198: `if (!extents) throw new Vex.RERR('InvalidState', 'Stem does not have extents.');`

then we can avoid `extents?` and `??` below.

src/stemmablenote.ts
rvilarl: ok

src/stemmablenote.ts
rvilarl: done

src/stemmablenote.ts
rvilarl: Ok I have also moved `getStemExtents` inside the `if`.

src/stemmablenote.ts
rvilarl: done

rvilarl:
Done, ready to checkin!

0xfe:
Great!

PR #923, opened 2021-04-22
Migration/keysignote

rvilarl:
keysignote is there!


src/keysignote.ts
```diff
@@ -15,28 +18,32 @@ export class KeySigNote extends Note {
     this.ignore_ticks = true;
   }
 
-  getBoundingBox() {
+  getBoundingBox(): BoundingBox | undefined {
```
0xfe, line 21: I think we can remove this whole method, now?. It just calls the parent.

0xfe (commented):
Looks good. Minor comment.

src/keysignote.ts
rvilarl: done

0xfe:
👍

Issue #918, opened 2021-04-20
Possible missing offsetYBaseStemUp: -2,


ronyeh:
https://github.com/0xfe/vexflow/blob/a61b6f59bcd9376c00bd853573cb3a487bc8264f/src/fonts/bravura_metrics.ts#L149

On my simple tests with EasyScore, I notice that my quarter notes with the stem going up look different from my quarter notes with the stem going down.

I assumed they should be just 180 degree rotations of one another, but the quarter note with the stem going up has a sharp edge where the stem meets the note head. 

When you zoom in on the web page, you see that the stem pokes out a little beyond the edge of the note head. It's not a smooth transition.

I traced it to the bravura_metrics.ts file, and when I add an extra line:

```
      noteheadBlack: {
        offsetYBaseStemUp: -2,    // <--- I added this line
        offsetYBaseStemDown: 2,
      },

```

Then the notes look great and symmetrical. They also match how it looks when I use the Bravura font in Affinity Designer and insert a quarter note with stem going upward.

So should we add this line? Or does it break other things? I am new to VexFlow so I am not sure if adding this line will break your visual regression tests.

Thanks!

0xfe:
I think this is right. We may have to fix this for Petaluma too.

Can you attach a before/after screenshot please?

ronyeh:
I'll investigate and report back. I see some small issues with the half notes too, both in Bravura and Petaluma. Are you phasing out support for Gonville, or should I investigate that font too?

The only reason I noticed is because I'm making a sight-reading tool targeted at beginners and older adults with poor vision. Thus, I scaled up the SVG context by calling:

```
const renderer = new VF.Renderer(targetDIV, VF.Renderer.Backends.SVG);
const context = renderer.getContext();
context.scale(2, 2);
```

As a result, I noticed the small issues with the note stem poking through the note heads.

ronyeh:
I'll try my best. :-) I'll work on a PR with fixes for Bravura, Petaluma, and Gonville.

0xfe:
Thank you -- :-)

h-sug1no:
Hi!

smufl seems to define the position of the stem connection for each glyph:

https://w3c.github.io/smufl/latest/specification/glyph-registration-notes-flags.html

[...]

PR #926, opened 2021-04-23
Fix small issue where parts of the stem are visible around the note head. This patch affects the Bravura font.

ronyeh:
The visual diffs are very minimal, and only noticeable at context scale 2 or higher.

However, I did notice one visual diff that is noticeable at 1x scale. Since the `Annotation.TabNote_Annotations` test does not include round note heads, it is more obvious that the height of the note stems has changed. Please see images below. Perhaps this is something that would require a special case, to make sure **stems without note heads** have identical heights whether they are going up or down.

Annotation TabNote_Annotations_Current

Annotation TabNote_Annotations_Reference

Annotation TabNote_Annotations

### DIFF
--- a/src/fonts/bravura_metrics.ts
+++ b/src/fonts/bravura_metrics.ts
@@ -145,7 +145,12 @@ export const BravuraMetrics = {
         offsetYBaseStemUp: -4,
         offsetYBaseStemDown: 4,
       },
+      noteheadHalf: {
+        offsetYBaseStemUp: -2.5,
+        offsetYBaseStemDown: 2.6,
+      },
       noteheadBlack: {
+        offsetYBaseStemUp: -2,
         offsetYBaseStemDown: 2,
       },
       noteheadSquareWhite: {


rvilarl:
@ronyeh the number of visual changes compared to the blessed images goes from 36 to 61 and visual_regression.sh is also reporting a lot of errors. Have you looked into that?

ronyeh:
I made a small fix to `tools/visual_regression.sh` and would like to send that in as a separate pull request, but apparently I can only send one PR at a time from my fork. I suppose I would need to create a separate branch just to submit the 3 characters I edited? I am clearly not a git expert :-). Anyways, feel free to make those fixes manually in your master repo.

See https://github.com/0xfe/vexflow/commit/d49d7c13e6cddc1a517d0121c4a1c8c5c1a9e314

I am currently improving the testing functions so that I can provide an in depth report of what has changed visually.

My tweaks to the note stems **will change almost every output image** (by a tiny bit), so Mohit & Rodrigo will have to decide whether it makes sense to 1) prioritize stability of the current look at 1x scale, or 2) include the changes so that the visual bugs are not as apparent at 2x, 4x, 8x scale. I understand 1x scale is the most common usage, so I totally understand if you don't want to include these tweaks at the moment. Maybe we can figure out a bullet-proof solution in the future that makes it look great at all common context sizes.

I created a tool for myself to compare the before and after: https://vexflow-demo.surge.sh/fonts/
I referenced this here: https://github.com/0xfe/vexflow/issues/918#issuecomment-825128518

I am going off of a font sample that I created with the most recent versions of Bravura and Petaluma and Gonville-18.
https://vexflow-demo.surge.sh/images/font_sample.svg  I used Affinity Designer and installed the most recent fonts and used 400 pt.

I'm still tweaking the values of Petaluma and Gonville, and once I'm done I'll send the changes in for your review.

From my early tests, it looks like improving Petaluma will have to be an iterative process, since the way we currently render stems and flags doesn't quite result in the same look and feel as the Petaluma font sample. The Petaluma stems are not straight lines, so we might need to render out shapes (!) or at the very least use linecaps to approximate the look. Petaluma flags are also quite varied. :-)

Anyways, more to come! I'm happy to incorporate any feedback from @rvilarl or @0xfe on this matter.

Next up, I'll try to send in a PR (from a different branch?) for my improvements to `runNodeTest` in `vexflow_test_helpers.js`

ronyeh:
Pardon the extra noise as I learn more about git. I just rebased & force-pushed so that this PR includes only the Bravura tweaks.

My modifications to the testing code is now in a separate PR: https://github.com/0xfe/vexflow/pull/932

I will create separate PRs for my Gonville tweaks and also for the Petaluma tweaks. That way, you can decide whether to merge on a case-by-case basis.

0xfe:
This looks good. Thanks. Merging! (And thanks for the thorough testing.)

Re: git, yes, you'll need to create separate branches for different PRs -- it's actually not so bad, the branches are quite lightweight and easy to switch between.

Re: stems without noteheads, I think you're right about a special code path for them. Seems like a good next PR. The difference is barely visible, so I'm merging this in. :-)

PR #931, opened 2021-04-23
fix/note

rvilarl:
@0xfe getBoundingRect is already availabe in Element, we removed from KeySigNote and the implementation in Note was just reurning undefined.


0xfe (changes requested):
Can you remove the `import { BoundingBox }` line too? (fixes the lint error.)

rvilarl:
Done, good catch!

PR #932, opened 2021-04-23
Test All Fonts

ronyeh:
This change allows us to render visual diff images for all three fonts.

runNodeTest() now tests all three fonts, and outputs images files with names like:
```
  Accidental.Basic.Bravura.png
  Accidental.Basic.Gonville.png
  Accidental.Basic.Petaluma.png
  ...
  Vibrato.Harsh_Vibrato.Bravura.png
  Vibrato.Harsh_Vibrato.Gonville.png
  Vibrato.Harsh_Vibrato.Petaluma.png
```
This allows us to tweak individual fonts and use `npm run test:reference` to see the changes via visual diff.

### DIFF
--- a/tests/vexflow_test_helpers.js
+++ b/tests/vexflow_test_helpers.js
@@ -52,8 +52,6 @@ if (!global.QUnit) {
 
 global['VF'] = Vex.Flow;
 VF.Test = (function () {
-  var DEFAULT_FONTS_TO_TEST = [VF.Fonts.Bravura, VF.Fonts.Gonville, VF.Fonts.Petaluma];
-
   var Test = {
     // Test Options.
     RUN_CANVAS_TESTS: true,
@@ -67,8 +65,8 @@ VF.Test = (function () {
     // Default font properties for tests.
     Font: { size: 10 },
 
-    // Test all three fonts by default. Customize the FONTS_TO_TEST array to test fewer fonts.
-    FONTS_TO_TEST: DEFAULT_FONTS_TO_TEST,
+    // Customize this array to test more fonts (e.g., ['Bravura', 'Gonville', 'Petaluma']).
+    FONT_STACKS_TO_TEST: ['Bravura'],
 
     FONT_STACKS: {
       Bravura: [VF.Fonts.Bravura, VF.Fonts.Gonville, VF.Fonts.Custom],
@@ -242,7 +240,16 @@ VF.Test = (function () {
         if (VF.Renderer.lastContext !== null) {
           var moduleName = sanitizeName(QUnit.current_module);
           var testName = sanitizeName(QUnit.current_test);
-          var fileName = `${VF.Test.NODE_IMAGEDIR}/${moduleName}.${testName}.${fontName}.png`;
+          var fileName;
+          if (fontName === 'Bravura' && VF.Test.FONT_STACKS_TO_TEST.length === 1) {
+            // If we are only testing Bravura, we do not add the font name
+            // to the output image file's name, which allows visual diffs against
+            // the previous release: version 3.0.9. In the future, if we decide
+            // to test all fonts by default, we can remove this check.
+            fileName = `${VF.Test.NODE_IMAGEDIR}/${moduleName}.${testName}.png`;
+          } else {
+            fileName = `${VF.Test.NODE_IMAGEDIR}/${moduleName}.${testName}.${fontName}.png`;
+          }
 
           var imageData = canvas.toDataURL().split(';base64,').pop();
           var image = Buffer.from(imageData, 'base64');
@@ -254,14 +261,10 @@ VF.Test = (function () {
       VF.Test.runTestWithFonts(name, testFunc);
     },
 
-    // Run QUnit.test() for each font that is included in VF.Test.FONTS_TO_TEST.
+    // Run QUnit.test() for each font that is included in VF.Test.FONT_STACKS_TO_TEST.
     runTestWithFonts: function (name, func) {
-      if (!Array.isArray(VF.Test.FONTS_TO_TEST)) {
-        VF.Test.FONTS_TO_TEST = DEFAULT_FONTS_TO_TEST;
-      }
-
-      VF.Test.FONTS_TO_TEST.forEach((font) => {
-        QUnit.test(name, func(font.getName()));
+      VF.Test.FONT_STACKS_TO_TEST.forEach((fontName) => {
+        QUnit.test(name, func(fontName));
       });
     },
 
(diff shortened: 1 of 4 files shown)


0xfe:
Thanks for this change -- I thought we were already testing the different fonts.

Two questions:

- How much longer does it take to test all fonts vs. a single font?
- Can you put this behind a flag? (Default to all fonts.)

ronyeh:
When I run the tests from the command line, it goes through `runNodeTest()` which only generated PNG images for the Bravura font. I copied your approach from `runSVGTest()` to call `QUnit.test(name, testFunc(...))` 3X, each time with a different font stack.

In bash, I ran `time npm run test:reference` with a current build identical to the reference build (so no expected diffs), and got:

```
real	2m52.817s
user	8m29.607s
sys	0m47.694s
```

This took under 3 minutes and generated 948 images in `vexflow/build/images/current`and also 948 images in `vexflow/build/images/reference`.

My machine is a MacBook Pro (15-inch, Late 2016), 2.9 GHz Intel Core i7.

I tested it a second time with arbitrary changes to all 3 font metrics (to trigger diffs in every image). This resulted in 948 PNG files in each of the `current` and the `reference` folders, and 1427 files in the `diff` folder.

```
real	3m13.222s
user	8m58.910s
sys	0m56.378s
```

So also about three minutes on my 2016 era MBP.

ronyeh:
I am currently timing a run with a single font, and will add a flag.

ronyeh:
I ran `time npm run test:reference` on just the Petaluma font, and it generated 316 files in current, 316 files in reference, and 605 files in diff.

```
real	1m54.410s
user	3m54.010s
sys	0m23.482s
```

About 2 minutes to test one font. (I tested this twice.) 

At the moment, it defaults to testing all three fonts.

To test only one or two fonts, the developer needs to uncomment this line in `generate_png_images.js`:

```
// VF.Test.FONTS_TO_TEST = [VF.Fonts.Petaluma];
```

In the future, we could add support for a `fonts` argument via `process.argv` if we want to add more scripts to the package.json, beyond the current set of `generate:current`, `generate:reference`, `generate:blessed`.

ronyeh:
`time npm run generate:current`
> 23.712s for all three fonts.
> 9.860s for one font. 

`time npm run generate:reference`
> 25.017s for all three fonts.

`time npm run diff:reference`
> 1m12.682s for all three fonts with zero diffs.
> 1m20.851s for all three fonts with max diffs (I changed the color of the stave lines).

ronyeh:
OK I pushed my changes https://github.com/0xfe/vexflow/pull/932/commits/134338f9d021a0d42abe59ee166124ee7a2fe36a

Feel free to review and suggest improvements.

The only issue I see is that when we compare the newest version against the previous "blessed" release, the tests won't work because `generate:blessed`will create files with names like 

```
Accidental.Basic.png
```

but `generate:current`will create files with names like: 

```
Accidental.Basic.Bravura.png
Accidental.Basic.Gonville.png
Accidental.Basic.Petaluma.png
```

So we could either:

A) manually patch line 262 of the blessed `vexflow-tests.js` https://github.com/0xfe/vexflow/blob/b373637eb2ae700a6c6cbe2606448fdea461b317/releases/vexflow-tests.js#L262 to insert `.Bravura` into the file names,

OR

B) add a special case into `vexflow_test_helpers.js` to output an extra set of image files for Bravura, so that `Accidental.Basic.Bravura.png` and `Accidental.Basic.png` are identical images.

I think option A would be simpler, but it does feel hacky to manually patch a "blessed" version. But at least it's only one line of the testing library!

0xfe:
Thanks for doing this. I missed your responses. Will review.

tools/generate_png_images.js
```diff
@@ -31,6 +31,13 @@ VF.Test.RUN_RAPHAEL_TESTS = false;
 VF.Test.RUN_NODE_TESTS = true;
 VF.Test.NODE_IMAGEDIR = imageDir;
 
+// By default we test all fonts.
+// Modify VF.Test.FONTS_TO_TEST to test fewer fonts at a time.
+// In the future, we can allow customization via process.argv.
+// VF.Test.FONTS_TO_TEST = [VF.Fonts.Petaluma];
+// VF.Test.FONTS_TO_TEST = [VF.Fonts.Gonville];
+// VF.Test.FONTS_TO_TEST = [VF.Fonts.Bravura];
```
0xfe, line 39: Okay, let's just do Bravura for now as the default. This should have no effect on the existing repo, correct?

tests/vexflow_test_helpers.js
```diff
@@ -65,6 +67,15 @@ VF.Test = (function () {
     // Default font properties for tests.
     Font: { size: 10 },
 
+    // Test all three fonts by default. Customize the FONTS_TO_TEST array to test fewer fonts.
+    FONTS_TO_TEST: DEFAULT_FONTS_TO_TEST,
```
0xfe, line 71: We should probably keep the semantics consistent and say `FONT_STACKS_TO_TEST: ['Bravura', 'Goneville', 'Petaluma']`. It's better to index by font stack than fonts.

ronyeh: Sounds good.

tools/generate_png_images.js
ronyeh: Will do. Default to Bravura only. Will add a flag to specify which fonts to test.

ronyeh:
Please see the current diff: https://github.com/0xfe/vexflow/compare/master...ronyeh:vexflow_test_helpers

It now defaults to only generating images for Bravura.

For backward compatibility (visual diffing against version 3.0.9), when Bravura is the only font specified, we will not add the font name to the output test PNG's file name. The file names will still look like `Accidental.Basic.png`. If Petaluma and Gonville are also specified, we will add the font name and the output files will look like:

```
  Accidental.Basic.Bravura.png
  Accidental.Basic.Gonville.png
  Accidental.Basic.Petaluma.png
```

I added one npm script:

```
npm run generate:current:all-fonts
```

Since it defaults to Bravura only (with everything else hidden behind flags and an alternate npm script), this shouldn't change any tests. I've verified on my own local copy.

Please review and comment!

tests/vexflow_test_helpers.js
```diff
@@ -230,17 +235,36 @@ VF.Test = (function () {
...
         func(testOptions, VF.Renderer.getCanvasContext);
+        VF.DEFAULT_FONT_STACK = defaultFontStack;
 
         if (VF.Renderer.lastContext !== null) {
           var moduleName = sanitizeName(QUnit.current_module);
           var testName = sanitizeName(QUnit.current_test);
-          var fileName = `${VF.Test.NODE_IMAGEDIR}/${moduleName}.${testName}.png`;
+          var fileName;
+          if (fontName === 'Bravura' && VF.Test.FONT_STACKS_TO_TEST.length === 1) {
+            // If we are only testing Bravura, we do not add the font name
```
0xfe, line 245: Good catch!

0xfe:
Looks great! Merging. Thanks for the PR!

PR #933, opened 2021-04-24
Migration/boundingbox

rvilarl:
BoundingBox and BoundingBoxComputation


src/boundingbox.ts
```diff
@@ -3,58 +3,76 @@
 //
 // Copyright Mohit Muthanna 2010
 
+import { RenderContext } from './types/common';
+
 // Bounding boxes for interactive notation
 
 export class BoundingBox {
-  static copy(that) {
+  protected y: number;
```
0xfe, line 11: Can you order this `x, y, w, h`?

src/boundingbox.ts
```diff
@@ -3,58 +3,76 @@
...
+  protected x: number;
+  protected w: number;
+  protected h: number;
```
0xfe, line 14: Why not private for all these?

src/boundingboxcomputation.ts
```diff
@@ -10,7 +10,12 @@
 // nvg.js#L449
 
 export class BoundingBoxComputation {
-  constructor(x1, y1, x2, y2) {
+  x1: number;
+  y1: number;
+  protected x2: number;
+  protected y2: number;
```
0xfe, line 16: Why mixed visibility here? Can't all be private?

src/boundingbox.ts
rvilarl: I thought we have decided to use `protected` in general. `private` only when `get/set` functions make a special handling like with `context`. This is here not the case and the class has also no child.

src/boundingboxcomputation.ts
rvilarl: Glyph is accessing these attributes:

https://github.com/0xfe/vexflow/blob/b373637eb2ae700a6c6cbe2606448fdea461b317/src/glyph.ts#L281

src/boundingbox.ts
rvilarl: ok

src/boundingbox.ts
0xfe: Okay, that's fine.

src/boundingboxcomputation.ts
rvilarl: I made to access functions in order to make them protected.

rvilarl:
Ready to merge!

0xfe:
Looks great!

PR #934, opened 2021-04-24
Migration/curve

rvilarl:
Curve is there.
@0xfe I am not particulary happy with the ammount of checks introduced, how do you see it? Should we use `as` instead in those  cases were the variable is obvoiusly defined?


src/curve.ts
```diff
@@ -5,16 +5,32 @@
...
 import { Element } from './element';
+import { StaveNote } from './stavenote';
+import { StaveOptions } from './stave';
+
+export enum Position {
+  NEAR_HEAD = 1,
+  NEAR_TOP = 2,
+}
+
+export interface CurveRenderParams {
```
0xfe, line 16: This can be inlined.

src/curve.ts
```diff
@@ -62,13 +79,23 @@ export class Curve extends Element {
...
+  isPartial(): boolean {
     return !this.from || !this.to;
   }
 
-  renderCurve(params) {
-    const ctx = this.context;
+  renderCurve(params: CurveRenderParams): void {
+    const ctx = this.checkContext();
     const cps = this.render_options.cps;
+    if (
```
0xfe, line 89: I'd move all these checks into an internal `check()` method and call it form `renderCurve` and from `draw`.

```typescript
check() {
  if cps == undefined || cps[0] == undefined || ... throw new Vex.RERR("InvalidParams", "No curve control points");
  if ths.render_options.x_shift == undefined || y_shift == undefied ... throw new Vex.RERR("...
  ...
}
```

src/curve.ts
```diff
@@ -27,7 +43,7 @@ export class Curve extends Element {
   //    cps: List of control points
   //    x_shift: pixels to shift
   //    y_shift: pixels to shift
-  constructor(from, to, options) {
+  constructor(from: StaveNote, to: StaveNote, options: StaveOptions) {
```
0xfe, line 46: `StaveOptions` here seems wrong, particularly since we're extending `element`.

Also `from:` and `to:` should be `Note`, you can have curves on other elements (`GraceNoteGroup`, `TabNote`, etc.)

rvilarl: A few attributes from `options` are used: `x_shift`, `y_shift`, thickness`, ... but not all. It would make sense to **create a `CurveOptions`** . It is used a parameter and as attribute from `Curve`.

`to` and `from` as `Note` would not have the following methods used by `Curve`:
From `StaveNote`: `getTieRightX` & `getTieLeftX`
From `StemmableNote`: `getStemDirection` & `getStemExtents`
Due to the fact that we do not want to merge `Note`, `StemmableNote` & `StaveNote`, it would make sense to **define those methods in `Note` returning an appropriated value**.

@0xfe do you agree?

src/curve.ts
rvilarl: If we do not make the check inside the method using the variable and we do it in another method that we call the compiler does not realise it and we still get an error. I believe that the checks will no longer be so complex defining `CurveOptions` as suggested below.

src/curve.ts
0xfe: Okay, let's create a new `CurveOptions` with just what we need.

Re: `Note`, good catch, we should do two things:

- Move `getTieRightX()` and `getTieLeftX()` out of `TabNote` and into `Note` (the default implementation there will work everywhere.)
- Add a new curve option `direction?: enum = { -1: DOWN, 0: DEFAULT, 1: UP }` default to `0`. If the note is not a `StemmableNote` then use the `direction`.

src/curve.ts
rvilarl: done

src/curve.ts
rvilarl: As expected check no longer required

src/curve.ts
rvilarl: Done `getStemDirection` `getStemExtents` also implemented in `Note` throwing an error.

rvilarl:
Ready to merge!

src/curve.ts
```diff
@@ -5,16 +5,33 @@
 
 import { Vex } from './vex';
 import { Element } from './element';
+import { Note } from './note';
+export interface CurveOptions {
```
0xfe, line 9: New line before export. (in other files too.)

src/curve.ts
```diff
@@ -5,16 +5,33 @@
...
+  //[name: string]: any;
```
0xfe, line 10: Is this comment intentional?

src/stave.ts
```diff
@@ -21,10 +21,11 @@ export interface StaveLineConfig {
...
   //[name: string]: any;
-  spacing: number;
-  thickness: number;
-  x_shift: number;
-  y_shift: number;
+  spacing?: number;
+  thickness?: number;
+  x_shift?: number;
+  y_shift?: number;
+  position?: number;
```
0xfe, line 28: Can we remove these changes?

src/curve.ts
rvilarl: removed

src/stave.ts
rvilarl: yes

rvilarl:
Done

src/note.ts
```diff
@@ -599,4 +599,34 @@ export abstract class Note extends Tickable {
   setPreFormatted(value: boolean): void {
     this.preFormatted = value;
   }
+
+  // Get the direction of the stem
+  getStemDirection(): number {
```
0xfe, line 604: This is causing a merge conflict because it was also in Migration/tremolo. You can remove it.

rvilarl: done

rvilarl:
Ready

0xfe:
Great!

Issue #937, opened 2021-04-27
Migration issues on voice


rvilarl:
1) `Draw` parameters do not correspond to those in `Element` and there is an error not allowing to overload.
2) constructor uses `StaveOptions` requiring an additional attribute `softmaxFactor`

@0xfe any tip on these issues?

rvilarl:
> 1. `Draw` parameters do not correspond to those in `Element` and there is an error not allowing to overload.

Solution: `draw` is now allowing `any` in Element.

> 2. constructor uses `StaveOptions` requiring an additional attribute `softmaxFactor`

Solution: `StaveOptions` no longer used. `softmaxFactor` is inline

3. `Beam` uses `getTickables` from `Voice` which returns `Note`. Looking in details, the implementation of `Beam` check if the notes have stems. 

Solution: `Beam` adapted to allow `Note`

0xfe:
First two solutions seem good.

Re 3: it seems far more reasonable to downcast from `Note` to `StemmableNote` than to add `*Stem*` methods to `Note` or change beam's interface.

PR #938, opened 2021-04-28
Migration/voice

rvilarl:
Voice is there with considerations from #937


src/note.ts
```diff
@@ -599,4 +600,60 @@ export abstract class Note extends Tickable {
   setPreFormatted(value: boolean): void {
     this.preFormatted = value;
   }
+
+  // Get the number of beams for this duration
+  getBeamCount(): number {
```
0xfe, line 605: Not a fan of these changes -- they pollute `Note` which is already quite polluted. :-)

src/beam.ts
```diff
@@ -11,11 +11,10 @@ import { Fraction } from './fraction';
 import { Tuplet } from './tuplet';
 import { Stem } from './stem';
 import { Note } from './note';
-import { StemmableNote } from './stemmablenote';
 import { Voice } from './voice';
 import { RenderContext } from './types/common';
 
-function calculateStemDirection(notes: StemmableNote[]) {
+function calculateStemDirection(notes: Note[]) {
```
0xfe, line 17: These should remain `StemmableNote` -- beams don't make sense unless there are stems.

rvilarl: Sure but:
- the code is getting the contents as `Note` from `Voice` 
https://github.com/0xfe/vexflow/blob/ec03fb3cf194c48b67c95f6798b0e9bec5491c2e/src/beam.ts#L134
- the code is expecting the notes not to be `StemmableNote`:
https://github.com/0xfe/vexflow/blob/ec03fb3cf194c48b67c95f6798b0e9bec5491c2e/src/beam.ts#L682-L683

0xfe: So for `applyAndGetBeams`:

```typescript
static applyAndGetBeams(voice: Voice, stem_direction?: number, groups?: Fraction[]): Beam[] {
    const stemmables = (voice.getTickables() as Array<StemmableNote>);
    return Beam.generateBeams(stemmables, {
      groups,
      stem_direction,
    });
  }
```

The second part is effectively a run-time type check for `StemmableNote` and throws an exception if there's no stem -- if it was all typescript, you can just remove that check.

rvilarl: This `as` approach bypasses the TypeScripts checks. The current approach with the functions in `Note` calling RERR make sure that we do not make an error. I think that this is better, is not it?

0xfe: Not really -- getTickables() returns `Note` which is type checked, and `Note` is explicitly typecast to `StemmableNote` -- so the window of vulnerability is pretty small. This is not bypassing checks entirely like using `any`.

You can close the vulnerability with a typeguard: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

This approach is way cleaner than absorbing `StemmableNote` methods into `Note`, and then accepting `Note` in `Voice`.

rvilarl: No doubt that this is not as bad as `any`!
Are you proposing to use `as` and then `instanceof` to check if the assumption is good? Have a look at the fiddle:

https://jsfiddle.net/zorjyg8L/

rvilarl: Well not really needed. If there is an error it will be catched in the accepted changes in Note where we ar both not really fans. :)

src/note.ts
rvilarl: agree

src/beam.ts
rvilarl: done.

rvilarl:
Ready

src/note.ts
```diff
@@ -600,6 +601,32 @@ export abstract class Note extends Tickable {
...
+
+  // Get StemX
+  getStemX(): number {
+    throw new Vex.RERR('NoStem', 'No stem attached to this note.');
+  }
+
+  // Set the direction of the stem
+  setStemDirection(direction: number): this {
+    throw new Vex.RERR('NoStem', `Calling getStemDirection(${direction}) in Note.`);
+  }
```
0xfe, line 628: Can we remove all these calls? I don't think we need them anymore.

rvilarl: These calls will be used if we assign by error a `Note` to `Beam` (`Note` included in `Voice`) producing a `RERR`. Otherwise it seems that we will get a TypeError specific error at run-time. The jsfiddle below show how `instanceof` and functions get called when using `as`:
 
https://jsfiddle.net/o8yav0er/

If you want to go the TypeError specific error and not RERR it seems that we can delete them.

0xfe: Thanks for the explanation. `TypeError` is definitely better -- let's remove this and have TS deal with it. :-)

rvilarl: Changes on `note` reverted to master.

rvilarl:
Ready

src/voice.ts
```diff
@@ -9,23 +9,52 @@ import { Vex } from './vex';
...
+import { Tickable } from './tickable';
+import { RenderContext } from './types/common';
+
+export interface VoiceTime {
+  num_beats: number;
+  beat_value: number;
+  resolution: number;
+}
+
+export enum Mode {
```
0xfe, line 25: This can be `VoiceMode`

src/voice.ts
```diff
@@ -257,11 +283,11 @@ export class Voice extends Element {
   }
 
   // Preformats the voice by applying the voice's stave to each note.
-  preFormat() {
+  preFormat(): this {
     if (this.preFormatted) return this;
 
     this.tickables.forEach((tickable) => {
-      if (!tickable.getStave()) {
+      if (!tickable.getStave() && this.stave) {
```
0xfe, line 290: This will silently fail, maybe throw an exception in the beginning of `preFormat` if `stave` is not defined.

0xfe (commented):
Few more

src/voice.ts
rvilarl: done

src/voice.ts
rvilarl: done

rvilarl:
Ready!

0xfe:
Looks good.

PR #939, opened 2021-04-28
Migration/bend

rvilarl:
Bend is there!


src/bend.ts
```diff
@@ -7,6 +7,29 @@
...
+
+export interface Phrase {
+  x?: number;
+  type: number;
+  text: string;
+  width?: number;
+  draw_width?: number;
+}
+
+export interface State {
```
0xfe, line 19: This can be inline.

src/bend.ts
```diff
@@ -7,6 +7,29 @@
 import { Vex } from './vex';
 import { Flow } from './tables';
 import { Modifier } from './modifier';
+
+export interface Phrase {
```
0xfe, line 11: Better to call this `BendPhrase`

src/bend.ts
```diff
@@ -134,7 +170,7 @@ export class Bend extends Modifier {
     let total_width = 0;
     for (let i = 0; i < this.phrase.length; ++i) {
       const bend = this.phrase[i];
-      if ('width' in bend) {
+      if (bend.width) {
```
0xfe, line 173: This is wrong because `bend.width` could be 0.

src/bend.ts
rvilarl: ok

src/bend.ts
rvilarl: ok, checked with `undefined`

src/bend.ts
rvilarl: ok

rvilarl:
ready!

0xfe:
Good stuff!

rvilarl:
@0xfe anything to be done?

0xfe:
Sorry, I missed this. Looks good! Thanks. :-)

PR #940, opened 2021-04-28
Visual fixes for all fonts: Bravura, Petaluma, Gonville.

ronyeh:
This pull request includes many subtle fixes for all three fonts. Some of the issues were visible at 1x scale in version 3.0.9, but most of the fixes are only noticeable when the context scale is set to 2x or 4x. I tested all of these fixes at up to 32x scale.

- note stems no longer protrude through the flags
- note stems no longer protrude through note heads

There are still some tiny issues, that I will address in a future PR. These issues were also present in 3.0.9, and relate to parts of the stem being visible outside the note head on half notes (in Bravura). Additionally, note stems might be too tall, when compared to samples manually typeset with the Bravura otf.

I have already run the visual regression tests, and they look good to me.

I will also attach some sample GIFs with before & afters so you can inspect.

I created a [test website where you can compare the changes](https://vexflow-demo.vercel.app/fonts/) against 3.0.9 (loaded from [unpkg](https://unpkg.com/vexflow@3.0.9/releases/vexflow-debug.js)).

https://vexflow-demo.vercel.app/fonts/

For example, here is Petaluma at 8x scale:

https://vexflow-demo.vercel.app/fonts/offsets.html?vex_version=current&font=petaluma&scale=8&backend=svg&visual_debug=true&rot=0&offset_x=0&offset_y=0

Here is version 3.0.9 at 8x scale:

https://vexflow-demo.vercel.app/fonts/offsets.html?vex_version=3.0.9&font=petaluma&scale=8&backend=svg&visual_debug=true&rot=0&offset_x=0&offset_y=0

In 3.0.9, notice how the half note with upward stem has an issue where the note head doesn't flow smoothly into the stem. Additionally, the stems stick out slightly beyond the flags, so there are sharp corners. These types of issues have been fixed (for all fonts) in the version proposed by this PR.

Comments & suggestions are appreciated! Thanks.


ronyeh:
## Gonville Before & After
G1
G2

ronyeh:
## Bravura Before & After (@ 16x scale)

B1

B2

ronyeh:
## Petaluma Before & After (@ 16x scale)

P1

P2

P3

src/stavenote.ts
```diff
@@ -1207,6 +1211,12 @@ export class StaveNote extends StemmableNote {
       this.setStem(new Stem(stemOptions));
     }
 
+    // If we will render a flag, we shorten the stem so that the tip
+    // does not poke through the flag.
+    if (this.shouldDrawFlag() && this.stem) {
+      this.stem.renderHeightAdjustment = -3;
```
0xfe, line 1217: Can we move this constant into a font metrics file?

0xfe (commented):
These are great adjustments! Thanks for fine-tuning.

src/stavenote.ts
ronyeh: I was just sticking to the convention of keeping magic numbers (like +/- 2) in the rendering functions. 😂

But yes, I'll find a way to move these adjustments out to the metrics files. This will future proof the code in case we add more fonts.

ronyeh:
I moved the constant out into the three metrics files.

4 visual diffs fail the threshold, but the diffs look like this:

TextNote TextNote_Formatting_2

TextNote TextNote_Formatting

src/stavenote.ts
ronyeh: Done.

src/fonts/bravura_metrics.ts
```diff
@@ -114,10 +114,11 @@ export const BravuraMetrics = {
       offsetYStemDown: 5,
       offsetXStemUp: 7,
       offsetXStemDown: 1,
-    }
+    },
   },
 
   stem: {
+    renderHeightAdjustmentForFlag: -3,
```
0xfe, line 121: Let's just call this `heightAdjustmentForFlag`

src/stavenote.ts
```diff
@@ -1154,28 +1154,32 @@ export class StaveNote extends StemmableNote {
     ctx.closeGroup();
   }
 
+  shouldDrawFlag(): boolean {
```
0xfe, line 1157: Thanks for this refactoring. 👍

src/stavenote.ts
```diff
@@ -1207,6 +1211,12 @@ export class StaveNote extends StemmableNote {
       this.setStem(new Stem(stemOptions));
     }
 
+    // If we will render a flag, we shorten the stem so that the tip
+    // does not poke through the flag.
+    if (this.shouldDrawFlag() && this.stem) {
```
0xfe, line 1216: It seems like `shouldDrawFlag` should also check for `this.stem` -- then you can eliminate the `this.stem? || 0` conditionals above too, no?

src/stem.ts
```diff
@@ -179,6 +177,14 @@ export class Stem extends Element {
     return this;
   }
 
+  adjustHeightForFlag() {
```
0xfe, line 180: Missing types.

src/stem.ts
```diff
@@ -179,6 +177,14 @@ export class Stem extends Element {
...
+    this.renderHeightAdjustment = this.musicFont.lookupMetric('stem.renderHeightAdjustmentForFlag', -3);
+  }
+
+  adjustHeightForBeam() {
```
0xfe, line 184: Missing types.

0xfe (commented):
Thanks for doing this. Approach looks good -- some comments.

src/stavenote.ts
ronyeh: I made these changes but am running into this issue:

TS compiler complains that this.stem can still be undefined (even though I checked in shouldDrawFlag()). So I add a non-null assertion (!). 

```
    if (this.shouldDrawFlag()) {
      const { y_top, y_bottom } = this.getNoteHeadBounds();
      const noteStemHeight = this.stem!.getHeight();

```

But then our eslint rules warn that:

```
vexflow/src/stavenote.ts
  1172:30  warning  Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion
```

I see that `// eslint-disable-next-line` is used in some places. Is this a good place to disable eslint? We know `this.stem` is defined at this point, due to `shouldDrawFlag()`.

Anyways, I'll push my changes for your review.

ronyeh:
Ready for review. 

The only diff that looks "bad" is StaveNote.StaveNote_Draw___StaveNote_Styles.png probably because it uses a shadowBlur.

StaveNote StaveNote_Draw___StaveNote_Styles

src/stavenote.ts
0xfe: Okay, let's leave it like this.

0xfe:
Looks great, merging!

PR #942, opened 2021-04-29
Migration/stavemodifier

rvilarl:
Stavemodifier is there!


src/stavemodifier.ts
```diff
@@ -0,0 +1,152 @@
...
+import { Stave } from './stave';
+
+export interface LayoutMetrics {
+  xMin: number;
+  xMax: number;
+  paddingLeft: number;
+  paddingRight: number;
+}
+
+export interface Metrics {
```
0xfe, line 17: I don't think we need this -- this looks like the metrics we use for `StaveNote`

rvilarl:
Ready

rvilarl:
@0xfe anything pending here?

src/stave.ts
```diff
@@ -189,9 +189,7 @@ export class Stave extends Element {
     this.end_x += shift;
     for (let i = 0; i < this.modifiers.length; i++) {
       const mod = this.modifiers[i];
-      if (mod.x !== undefined) {
-        mod.x += shift;
-      }
+      mod.setX(mod.getX() + shift);
```
0xfe, line 192: What if `mod.x` is `undefined`? Seems like we should catch that and throw an exception earlier on.

src/stavemodifier.ts
```diff
@@ -0,0 +1,147 @@
+// [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
+//
+// ## Description
+// A base class for stave modifiers (e.g. clefs, key signatures)
+
+import { Element } from './element';
+import { Glyph } from './glyph';
+import { Stave } from './stave';
+import { Metrics } from './note';
+export interface LayoutMetrics {
```
0xfe, line 10: Newline before export please.

src/stavemodifier.ts
```diff
@@ -0,0 +1,147 @@
...
+  setX(x: number): this {
+    this.x = x;
+    return this;
+  }
+
+  getCategory(): string {
+    return '';
+  }
+
+  makeSpacer(padding: number): StaveModifierSpacer {
```
0xfe, line 93: Let's remove this  method and the `StaveModifierSpacer` -- it's not used anywhere.

src/stavemodifier.ts
rvilarl: done

src/stavemodifier.ts
rvilarl: done

src/stave.ts
rvilarl: mod.getX() returns a number, undefined is not possible.

rvilarl:
Ready, hopefully :)

0xfe:
Looks good.

PR #944, opened 2021-04-30
Discuss: Migrate Parser

ronyeh:
**Work in Progress**

I am integrating feedback from Rodrigo, and am also awaiting comments from Mohit before I create a new PR from another branch (for the purposes of having a cleaner git repo history). I messed up this PR because my rebase squashed the file rename, so the diff looks ugly.

***

The tests all pass.

My simple test page also works fine with the new parser.ts.
https://vexflow-demo.vercel.app/easyscore/test.html?vex_version=current-easyscore&test_number=3
https://vexflow-demo.vercel.app/easyscore/test.html?vex_version=current-easyscore&test_number=2

Here is the diff:
https://gist.github.com/ronyeh/6b9572d31c81e6eb7989fbebcf1f0cd5

Here is the current `parser.ts`:
https://github.com/ronyeh/vexflow/blob/migration/parser/src/parser.ts

I left several TODOs in the code that we can address in a future pull request (e.g., remove parameters that are not referenced anywhere in the code base).

I have one question:
https://github.com/ronyeh/vexflow/blob/06d7a677e7959bf6ca2f113cea5c8d0ac19c2f99/src/parser.ts#L187

TypeScript complained about that line because we were trying to push a `Result[]` into a `Result[]`... but only `Result` objects can be pushed into a `Result[]`. So I used a spread operator `...` to flatten the array into individual `Result` objects before pushing all of them onto the array.

It doesn't seem to break anything, and in my tests, it actually results in fewer calls to `flattenMatches(...)` because it doesn't have to recurse as much.

I just wanted to check that using the spread operator in this instance is OK, and that we aren't changing anything fundamental about how we collect results from parsing.


src/parser.ts
```diff
@@ -0,0 +1,261 @@
...
+  // Parser Rules
+  expect?: RuleFunction[];
+  zeroOrMore?: boolean;
+  oneOrMore?: boolean;
+  maybe?: boolean;
+  or?: boolean;
+  run?: TriggerFunction;
+}
+// END: Move to easyscore.ts
+//////////////////////////////////////////////////////////////////////////////////////////////////
```
rvilarl, line 26: Move these to `types/common`, you will see there other elements that need to be moved. With this we will also probably get git to realise that this is the new version of `parser.js` to facilitate the review.
Did you use `git mv src/parser.js src/parser.ts`? This should also help to realise that this is a new version.

src/parser.ts
```diff
@@ -0,0 +1,261 @@
...
+  // Parser Results
+  matches?: Match[];
+  numMatches?: number;
+  results?: Result[];
+  errorPos?: number; // Set to NO_ERROR if successful. N if there is an error in the string.
+}
+
+// Converts parser results into an easy to reference list that can be used in triggers.
+// eventually returns: string (including empty strings) | nested array | null
+type Match = null | string | Array<Match>;
```
rvilarl: We are moving from `null` to `undefined`. Change to:
```
type Match = string | Array<Match>
```
Would it be possible simply:
```
type Match = string | string[]
```
?

src/parser.ts
```diff
@@ -0,0 +1,261 @@
...
+function flattenMatches(r: Result | Result[]): Match {
```
rvilarl, line 58: Change to:
```
function flattenMatches(r: Result | Result[]): Match | undefined
```

src/parser.ts
```diff
@@ -0,0 +1,261 @@
...
+  if ('matchedString' in r) return r.matchedString as string; // string
+  if ('results' in r) return flattenMatches(r.results as Result[]);
+  const results = r as Result[];
+  if (results.length === 1) return flattenMatches(results[0]);
+  if (results.length === 0) return null; // null
```
rvilarl, line 63:
```
return undefined:
```

src/parser.ts
```diff
@@ -0,0 +1,261 @@
...
+// This is the base parser class. Given an arbitrary context-free grammar, it
+// can parse any line and execute code when specific rules are met (e.g.,
+// when a string is terminated.)
+export class Parser {
+  public static DEBUG: boolean = false;
+
+  private grammar: Grammar;
+  private line: string;
+  private pos: number;
+  private errorPos: number;
```
rvilarl, line 76: Please change to `protected` we are using that in general. `private` only if we want to force the child to use access functions making some extra required processing.

src/parser.ts
```diff
@@ -0,0 +1,261 @@
...
+  public parse(line: string): Result {
+    this.line = line;
+    this.pos = 0;
+    this.errorPos = NO_ERROR_POS;
+    const result = this.expect(this.grammar.begin());
+    result.errorPos = this.errorPos;
+    return result;
+  }
+
+  private matchFail(returnPos: number): void {
```
rvilarl, line 99: We are not declaring visibility of methods leaving all of them public. This is better but it is difficult to know who is using what.
@0xfe what do you think we should do going forward?

rvilarl (changes requested):
**Thanks a lot!**
I would suggest that you used branches so that you can pull several changes in parallel not having to wait for merges. I do it like that:
```
git checkout master
npm run reference // to generate the reference test images from master
git pull upstream master // to get the latest changes, upstream is configured to be Mohit's repo
git checkout -b migration/parser
git mv src/parser.js src/parser.ts
git commit -m "parser renamed"
... make the changes
npm run test:reference // to check that the test results do not differ from master
git add src/*
git commit -m "parser migrated"
git push // the first one is different but git will tell you how to do it
```
then you can go back to master and continue with the next one
```
git checkout master
npm run reference // to generate the reference test images from master
git pull upstream master // to get the latest changes, upstream is configured to be Mohit's repo
git checkout -b migration/XXX
...
```

ronyeh:
Thanks for all of your comments Rodrigo. Very helpful!

Yes, I did use a branch for the migration (`migration/parser`). I committed the file rename before actually editing it, but I believe I goofed up my rebase and clobbered the history. 

I will go ahead and apply this diff to a new branch (`migration/parser2`) so that we preserve a clean history in the git repo. Once that's done, I'll submit a new pull request and then close this one.

I will also integrate your code review & feedback into the new pull request.

Thanks.

src/parser.ts
ronyeh: Sounds good. The reason I kept it in parser.ts was because I figured I could migrate easyscore.ts next, and it would be a simple edit to bring the types over. Plus, these types would be used only by these two classes. I also tried to avoid editing too many files because I know there are others working on the migration. I don't want to cause too many ugly merges!

src/parser.ts
ronyeh: Hmm... we should consult with @0xfe on this. I like the idea of being consistent throughout the code base, and if we are using `undefined` instead of `null` everywhere else, that's fine by me.

However, for this TS migration I tried to stick as much as possible to the original code's behavior. My only goal was to apply typings and enforce type checks to capture bugs at compile time. I didn't want to change the actual output of the method calls.

In the original parser, the `flattenMatches(...)` function returns a nested array in which the leaf elements are either string or null. This array is actually returned to an external class (e.g., Grammar in easyscore.js). The `run: (state) => ...` triggers take this input and calls code to build up the VexFlow objects. It is possible that someone 

I haven't looked deep enough to know if changing null to undefined will break anything in this case. It is possible that those builder functions check for null instead of undefined.

In any case, I'll wait for Mohit to to chime in before I start changing how the code behaves. Maybe this should be a separate PR that happens after the migration for parser.ts. That PR can tackle the conversion from `null` to `undefined` and make sure it doesn't break any external code.

src/parser.ts
ronyeh: Sounds good. Will do.

I actually like when libraries minimize use of `private` methods and instance variables because it allows a developer to subclass and override behaviors when necessary.

I initially chose `private` in this case, because the variables like line, pos, and errorPos are used mostly like temporary internal state for the Parser. They shouldn't really be accessed by anyone outside, because their state continually changes as we parse the grammar.

If the Parser were built in a more functional / closure-heavy fashion, these variables might have been hidden local vars that were captured by the closure.

Also, during migration, setting everything to private has a nice side effect. The typescript compiler errors will tell me which external classes are trying to access the class I'm migrating! :-)

src/parser.ts
ronyeh: OK, I'm happy to not declare visibility for methods. During migration it helps to set everything to private and only make public the functions that are accessed by external classes. 

In this case, I discovered that only two fields of the Parser class were ever accessed by external classes:

```
public parse(line: string): Result { ... }

public static DEBUG: boolean = false;
```

However, I'm happy to remove the public / private modifiers to be consistent with the rest of the code. I will keep instance variables protected as you mentioned earlier.

src/parser.ts
rvilarl: You are right:
```
type Match = string | string[]
```
makes no sense. The `undefined` was suggested to go for ? in parameters and attributes when they ar not mandatory.

src/parser.ts
rvilarl: You are right this is nice :)

ronyeh:
Close this discussion as it has been superseded by https://github.com/0xfe/vexflow/pull/945

PR #945, opened 2021-04-30
Migrate Parser

ronyeh:
Ready for review by @0xfe. I integrated comments by @rvilarl in the earlier discussion thread: https://github.com/0xfe/vexflow/pull/944.

The tests all pass with zero visual diffs. Here is the current [parser.ts](https://github.com/ronyeh/vexflow/blob/migration/parser2/src/parser.ts).

The only open questions are:

1. Can I use the spread operator here: https://github.com/ronyeh/vexflow/blob/aedc11121f9ceb4108c60d061412313d528242fb/src/parser.ts#L171  It flattens the results a bit earlier, and reduces the number of calls to flattenMatches(). I did this because the TypeScript compiler wouldn't let me push a Result[] into a Result[]. :-)
2. Rodrigo suggested that we should change `null` to `undefined` here: https://github.com/ronyeh/vexflow/blob/aedc11121f9ceb4108c60d061412313d528242fb/src/parser.ts#L44  I left it alone because the results of this function are passed to external classes, so I was worried about breaking an "API".... i.e., if the EasyScore Builder functions expect `null` instead of `undefined`. Perhaps we can leave this to a future PR.
3. I left a TODO in the code that we can address in a future pull request (remove a parameter that is not referenced anywhere in the code base).


src/types/common.d.ts
```diff
@@ -109,3 +109,25 @@ export interface ModifierContextState {
   text_line: number;
   top_text_line: number;
 }
+
+
+/* TODO: Move to EasyScore / Grammar / Parser */
+export type Match = string | Match[] | null;
```
0xfe, line 115: This can be in `parser.ts`

src/types/common.d.ts
```diff
@@ -109,3 +109,25 @@ export interface ModifierContextState {
...
+export interface Rule {
```
0xfe, line 116: This also belongs in `parser.ts`

src/types/common.d.ts
```diff
@@ -109,3 +109,25 @@ export interface ModifierContextState {
...
+
+  // Parser Rules
+  expect?: RuleFunction[];
+  zeroOrMore?: boolean;
+  oneOrMore?: boolean;
+  maybe?: boolean;
+  or?: boolean;
+  run?: TriggerFunction;
+}
+export interface TriggerState {
```
0xfe, line 129: This can be inlined in `TriggerFunction`

src/types/common.d.ts
```diff
@@ -109,3 +109,25 @@ export interface ModifierContextState {
...
+    matches: Match[];
+}
+export type RuleFunction = () => Rule;
+export type TriggerFunction = (state?: TriggerState) => void;
```
0xfe, line 133: Actually, all of these belong in `parser.ts` :-)

src/types/common.d.ts
0xfe: (i.e., no need for a separate named interface `TriggerState`.

src/types/common.d.ts
rvilarl: @ronyeh this is my failure I understood from the comment that all of this should be in EasyScore, sorry

src/types/common.d.ts
rvilarl: @0xfe I also commented on the topic that we are moving from `null` to `undefined`. Can you comment on that? @ronyeh 's concern is the change on the interface. Can you comment on that? That was already discussed in the past.

0xfe: Good question. In general, `undefined` for uninitialized values, and `null` when we specifically mean "none, empty set, not found, etc." -- so here, `null` is good.

rvilarl: This is a normal approach agree. But I believe we discussed that we wanted to use either `undefined` or `null` and moved to all using `undefined` in order not to have to include `null` in types like the example above and in order to use ? in optional attributes and params as @wassertim suggested. This is just a convention, which is also normal, but if we change it now we will be inconsistent with the rest. We decided to move to:

```
class A {
  reqString: string;
  optString?: string;
};

functionA(  reqString: string; optString?: string;): string | undefined
```
and not use:
```
class A {
  reqString: string;
  optString: string | null;
};

functionA(  reqString: string; optString: string | null;): string | null
```
I started with `null` and I changed it. :)

0xfe: Right, your example is fine because semantically `optString` can be defined or undefined. That is different from `null`, which, semantically, means defined and has some kind of value. It's like `undefined` vs `0` for `number`, or `undefined` vs `""` for `string`.

In this code `Match` as `null` means no match, as opposed to (not defined, don't know, etc.) The way I see it, `null` should only be used when it means something specific (typically [zero value](https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful)) -- `undefined` when it doesn't have a meaning, or the meaning is not known.

src/types/common.d.ts
ronyeh: Yes, I agree that they should be moved back into parser.ts, haha. I think I was working late and a bit sleepy when trying to decide where the interfaces should live.

I think in the end, most of these interfaces should actually be internal to or exported by Parser, as needed. Even the Grammar should actually be an interface (or abstract class) exported by Parser. It might only have one method called `begin()`.... but I think it makes sense that the Parser shouldn't be **importing** any classes **from** EasyScore. EasyScore is just a particular implementation that provides a particular grammar to a generic parser.

Some day, we could conceivably create a EasierScore or a HarderScore class, which provides a different grammar to the same generic parser. Thus, Parser shouldn't be importing anything from the specific implementations.

Once I migrate EasyScore, I'll make sure to remove the `import { Grammar } from './easyscore';` from the top of Parser.ts.

I will make these changes and update this PR.

ronyeh:
@0xfe  can you comment on this remaining question?

Is it fine to use the spread operator here: https://github.com/ronyeh/vexflow/blob/aedc11121f9ceb4108c60d061412313d528242fb/src/parser.ts#L171  It flattens the results a bit earlier, and reduces the number of calls to flattenMatches(). I did this because the TypeScript compiler wouldn't let me push a Result[] into a Result[]. :-)

src/types/common.d.ts
ronyeh: Sounds good. In the future if we make the TriggerState object more complicated, Parser might have to export it as a named interface, so that external classes (i.e. EasyScore) can import the type. For now it's a simple struct, so inlining makes sense.

ronyeh:
`npm run test:reference` results in zero diffs.

I think it's ready to be merged.

src/types/common.d.ts
0xfe: @rvilarl I mostly disagree with that Internet comment -- I think it completely disregards the nuances :-)

> Null and undefined are essentially two different values that mean the same thing.

Null and undefined are different zero values with different semantics. The former is an explicit semantic, the latter is not. See my example below.

> The only difference is in the conventions of how you use them in your system.

Javascript has already defined a semantic for `undefined` as variables that are not initialized or don't exist, and we can't change it. So we have no choice but to use `null` for explicit zero values.

> I never have to check if x === undefined || x === null.

I agree with that -- this happens when the two semantics are conflated, and we have a lot of this mess in Vexflow, because we don't have a clear convention. IMO, you should always know what to do with `null`, but you may not know what to do with `undefined`.

```typescript
if (x === null) {
  // defined predictable behavior, do something with it, or return recoverable error
}

if (x === undefined) {
  // unpredictable behaviour, can't continue, throw exception or return irrecoverable error
}

// this is bad, because it's conflating the two
if (x === null || x === undefined) { ... }
```

I think in general it's okay to use `undefined` when we don't know the zero value, and use `null` sparingly when we have a clear semantic for it. It makes a lot of sense to use `null` here.

0xfe:
> @0xfe can you comment on this remaining question?
> 
> Is it fine to use the spread operator here: https://github.com/ronyeh/vexflow/blob/aedc11121f9ceb4108c60d061412313d528242fb/src/parser.ts#L171 It flattens the results a bit earlier, and reduces the number of calls to flattenMatches(). I did this because the TypeScript compiler wouldn't let me push a Result[] into a Result[]. :-)

Yes, please use the spread operator if it does the same thing!

src/parser.ts
```diff
@@ -0,0 +1,264 @@
...
+    const results: Result[] = [];
+    const pos = this.pos;
+    let numMatches = 0;
+    let more = true;
+
+    do {
+      const result = this.expectOne(rule);
+      if (result.success && result.results) {
+        numMatches++;
+        // TODO: Is it okay to use the spread operator here to flatten the results?
```
0xfe, line 186: Yes, this is fine I think. I'd suggest adding a few more tests to `tests/parser_tests.js` to exercise this path, and make sure the before/after behavior is correct.

src/types/common.d.ts
rvilarl: That's ok. However it does go into the differences:

>That said, there is one major difference - variables that aren't initialized (including function parameters where no argument was passed, among other things) are always undefined.

We are not using 'undefined' for unpredictible, can´t continue... only we use it as well for optional parameters in functions.

As I said, that's ok, as you said we should avoid the `if (x === null || x === undefined)`

src/parser.ts
ronyeh: Sounds good. Will write some tests.

0xfe: Hi @ronyeh -- did you manage to figure out tests? Can I help in any way?

ronyeh: Is there a way to just run a single test suite, via npm / grunt options? I'm working on a parser test now and it seems easier for me to write it separately on my own demo site, and then copy it into `parser_tests.js` when I'm done.

To run a single QUnit test, I've been doing this in `run.js`:

```
VF.Test.run = function () {
  VF.Test.Parser.Start(); // Only run this test!
  // skips everything else, but doesn't trigger any eslint / TS compiler errors.
  if (Math.random() > -1) {
    return;
  }
 ... 
```

0xfe: Not sure about the command line, but my workflow includes keeping the test page open in the browser and using the dropdown to select a specific test category (e.g., parser). I use `grunt watch` to rebuild as I change code, and just refresh the page. It's quick and effective.

ronyeh: Thanks for the workflow tips! I didn't even think to open flow.html locally in my browser, and I also did not notice that dropdown in the upper right.

I have been using my own test pages running on a localhost webserver, and I also watch the vexflow library with:
`grunt webpack:watch`

`grunt watch` seems to update the tests, but if I change the vexflow core library, it doesn't update the build files (even though it triggers the grunt watch and concat:tests task since src/* is monitored).

Is grunt watch supposed to also recompile the vexflow library + vexflow tests?

ronyeh: Also, flow.html has an outdated title:
**VexFlow 2** - JavaScript Music Notation and Guitar Tab

We could change it to:
VexFlow - Music Notation and Guitar Tab

...and it'll always be up to date, even when you upgrade to VexFlow 4!

I won't submit a PR for this, since you can just change it yourself on master, haha.

ronyeh: @rvilarl  have you been using `grunt watch` successfully? I noticed that it concats & outputs `vexflow-tests.js` and 
`vexflow-tests.js.map` into the vexflow root directory, which seems incorrect.

So I modified Gruntfile.js to look like this:
```
const TARGET_TESTS_BROWSER = 'build/vexflow-tests.js'; // I changed it to output to the build folder
```
previously it looked like this:
```
const TARGET_TESTS_BROWSER = 'vexflow-tests.js'; // outputs to the vexflow root dir
```

So now when I load flow.html, it loads the correct vexflow-tests.js file.

However, I get this error:
```
vexflow-tests.js:16989 Uncaught SyntaxError: Cannot use import statement outside a module
flow.html:26 Uncaught ReferenceError: VF is not defined
    at HTMLDocument.<anonymous> (flow.html:26)

```

I think it is because we are using grunt-contrib-concat to combine the test files. However, some of the test files (e.g. music_tests.ts) now have import statements because they are in TypeScript.

I think we need to modify the watch tasks to use webpack to compile the TS..... alas I'm no expert in grunt or webpack.

ronyeh: I am now able to dev and test quickly by refreshing flow.html in the browser.

I added a watch task that will watch over tests/, and trigger the webpack build.

To make it faster, I run the webpack task for src/ and the webpack task for tests/ in parallel. So if you only change a file in tests/, it won't rebuild the vexflow core library.

See:
https://github.com/0xfe/vexflow/pull/961

rvilarl: @ronyeh I have never used `grunt watch` :) Now I just use:
- `npm run reference` on master to produce the reference images
- `npm run test:reference` on the branches to make sure that there are no diff on images compared to master
- and `flow.html` with Chrome to debug.

ronyeh:
I did some extensive testing....

I added a new test case for the parser, with a different grammar to test: https://github.com/0xfe/vexflow/pull/945/commits/8318a4113f3439af1e23fd23b61f049a89231e80

Parser and EasyScore tests pass (along with all other tests).

ronyeh:
I noticed a discrepancy between this build and 3.0.9. I'm putting together a demo page to illustrate.

ronyeh:
parser_flatten

See the below two test pages. Open the developer console and drill down to inspect the results.

3.0.9's Parser implementation
https://vexflow-demo.vercel.app/parser/test?vex_version=3.0.9&test_number=0

Current Parser implementation (with spread operator)
https://vexflow-demo.vercel.app/parser/test?vex_version=current&test_number=0

Basically the spread operator does not affect whether the parsing results in success or failure. All tests pass.

However, the depth of the nested array structure in `state.matches` that is provided to the run triggers is now shallower (due to the spread operator pre-flattening the results).

This doesn't affect EasyScore, as far as I can tell.

However, if someone out there implemented their own custom grammar and relied on the particular index/positioning of items in `state.matches`, then this pull request could be a breaking change.

src/parser.ts
0xfe: It's possible that `grunt watch` broke during the typescript infra change. Thanks for the PR, looking now.

0xfe:
Thanks for the test -- I had a hunch that this would affect the depth of the matches :-) In this case we should keep the original mechanism.

ronyeh:
Done! To get around the TypeScript compiler complaints, I introduced a new type `GroupedResults` which allows the mixing of Result and Result[]. My tests confirm that this version's parser works just like ver 3.0.9's parser.

0xfe:
Looks great! And good job on specifying and testing a new grammar! :-)

PR #949, opened 2021-05-01
Add releases/ to vscode's search.exclude.

ronyeh:
Makes it easier to search in VS Code, because we now ignore build files placed in `releases/`.


0xfe:
Thanks!

PR #950, opened 2021-05-01
Migration/clefnote

rvilarl:
Clef & Clefnote are there!


src/clef.ts
```diff
@@ -114,7 +135,7 @@ export class Clef extends StaveModifier {
       this.annotation = { code, point, line, x_shift };
 
       this.attachment = new Glyph(this.annotation.code, this.annotation.point);
-      this.attachment.metrics.x_max = 0;
+      if (this.attachment.metrics) this.attachment.metrics.x_max = 0;
```
0xfe, line 138: Not sure why we have the `if` here, seems like the previous line always set `x_max` to 0.

src/types/common.d.ts
```diff
@@ -111,3 +111,11 @@ export interface ModifierContextState {
   text_line: number;
   top_text_line: number;
 }
+
+/** TODO: Move to Annotation */
+export interface AnnotationInfo {
```
0xfe, line 116: This is a different type of annotation from `annotation.js`, you can call this `ClefAnnotation` and move it back to `clef.ts`.

src/clef.ts
rvilarl: The compiler complains about this.attachment.metrics due to the fact that it might be `undefined` in `Glyph`.

src/types/common.d.ts
rvilarl: Done

rvilarl:
Ready

src/clef.ts
0xfe: Right -- I think the previous version would have caused a failure.

I was going to suggest checking for `metrics` and throwing an exception, but I just learned about this: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator. We could use this in a lot of places.

So, how about changing this line to `this.attachment.metrics!.x_max = 0`.

rvilarl: This can save indeed a lot of changes. It is a warning now in the eslint rules that we have but we could disable it and use it. I have done a lot to try to avoid the non-null-assertion and perhaps it is not worth it. Perhaps we should look into the eslint rules why this is not recommended.
 
@ronyeh FYI you also had a warning on this one

0xfe: The rationale for the default is that it cancels the benefit of strict null-checking. It makes sense if the operator is overused, and I think we can manage that in our code. By using `!.` we're still being explicit about it, so I'm okay with removing that rule.

ronyeh: I'm not an expert in the vexflow codebase, so I can only comment on the idea of strict null checking and use of the ! operator.

In general, I am a fan of stricter checks by the TS compiler. The more checks we have, the more chances we can catch a bug before it gets released. Run time bugs are harder to track down.

The question is how often we have to add the `// eslint-disable-next-line` into the code base. I think I had to add it twice during my migration of the Parse. And one of them was just the logging function. ESLint complained about using any, `function L(...args: any[]) {`

So if we **KNOW** something is not null, I am happy to use the ! operator. But I am also happy to use the // eslint-disable-next-line for now, because it is evidence in the code base that perhaps it could either be 1) better designed in the future (via refactor), or 2) it's just a compiler limitation (e.g., I had a conditional check if the stavenote should draw the flag, and that function verified that this.stem was not null, but then later I had to use ! to tell the TS compiler that this.stem was indeed not null.)

rvilarl: I removed the rule already and chahnged a few on the other files :(. If we want to enable it agin I suggest to do it in another PR

rvilarl:
Ready

0xfe:
Great.

PR #951, opened 2021-05-01
Migration/vibratobracket

rvilarl:
vibratobracket!


src/vibratobracket.ts
```diff
@@ -9,21 +9,45 @@
 import { Vex } from './vex';
 import { Element } from './element';
 import { Vibrato } from './vibrato';
+import { Note } from './note';
+
+export interface VibratoBracketData {
```
0xfe, line 14: This can be inline.

src/vibratobracket.ts
```diff
@@ -9,21 +9,45 @@
...
+  stop: Note;
+  start: Note;
+}
+
+export interface VibratoBracketRenderOptions {
```
0xfe, line 19: This can also be inline.

src/vibratobracket.ts
```diff
@@ -37,37 +61,40 @@ export class VibratoBracket extends Element {
...
 
-    const y = this.start
-      ? this.start.getStave().getYForTopText(this.line)
-      : this.stop.getStave().getYForTopText(this.line);
+    const y = startStave ? startStave.getYForTopText(this.line) : stopStave ? stopStave.getYForTopText(this.line) : 0;
 
     // If start note is not set then vibrato will be drawn
     // from the beginning of the stave
-    const start_x = this.start ? this.start.getAbsoluteX() : this.stop.getStave().getTieStartX();
+    const start_x = this.start ? this.start.getAbsoluteX() : stopStave ? stopStave.getTieStartX() : 0;
```
0xfe, line 89: This seems to change the logic to silently fail. If `this.start` is not set, then `stopStave` must be set.

rvilarl: changed to assertion

src/vibratobracket.ts
rvilarl: done

src/vibratobracket.ts
rvilarl: done

rvilarl:
Ready

0xfe:
Looks good.

PR #955, opened 2021-05-02
Migration/modifier context

rvilarl:
ModifierContext is there!

I was not expecting so many changes. Reason the mix between `StaveNote` and Modifiers in `ModifierClass` this is worth a refactoring.


src/modifiercontext.ts
```diff
@@ -21,12 +21,74 @@ import { ChordSymbol } from './chordsymbol';
...
+  | typeof Articulation
+  | typeof Ornament
+  | typeof Annotation
+  | typeof ChordSymbol
+  | typeof Bend
+  | typeof Vibrato;
+
+export type PostformatModifierType = typeof StaveNote;
+
+export type ModifierClass =
```
0xfe, line 54: There must be a better approach here -- maybe there's a required method that all of them implement? (format / preformat?)

0xfe: I think better to create an interface in common that has `getCategory` and `setModifierContext`

rvilarl: As I said above:
>Reason the mix between StaveNote and Modifiers in ModifierClass this is worth a refactoring.

0xfe: Sorry -- I don't understand what you mean by "mix between StaveNote and Modifiers". Can't we have:

```typescript
export type ModifierClass = Modifer | StaveNote
interface Preformattable {
  format(...)
}

interface Postformattable {
  postformat(...)
}
```

rvilarl (commented):
Of course we can compile almost the mix we want. It is more a conceptual and maintinability topic. Perhaps I am missing something, but I understand that `StaveNote` is not a `Modifier`. 
1) My concerns related to `StaveNote`:
- If I am right `StaveNote` should not be a `ModifierClass`
- Using `addModifier` and `getModifiers` with `StaveNote` does not make a lot of sense.
2) The mix you propose will not work due to the fact the modifier does not provide the `format` so you have to make the complete list: `accidental, annotation` surprisingly (for me) `Tremolo` is a `Modifier` but it is not in the list and does not provide the static method `format`
3) `Modifier` does not provide `postFormat`

We can leave it as it is as well, this is your choice. :)

src/modifiercontext.ts
rvilarl: Of course we can compile almost the mix we want. It is more a conceptual and maintinability topic. Perhaps I am missing something, but I understand that `StaveNote` is not a `Modifier`. 
1) My concerns related to `StaveNote`:
- If I am right `StaveNote` should not be a `ModifierClass`
- Using `addModifier` and `getModifiers` with `StaveNote` does not make a lot of sense.
2) The mix you propose will not work due to the fact the modifier does not provide the `format` so you have to make the complete list: `accidental, annotation` surprisingly (for me) `Tremolo` is a `Modifier` but it is not in the list and does not provide the static method `format`
3) `Modifier` does not provide `postFormat`

We can leave it as it is as well, this is your choice. :)

0xfe: Thanks, rereading this, and going through the rest of the code:

* `ModifierContext` is not just modifiers, it's a container for all elements that belong within the inner formatting contexts within the tick contexts, so it includes tickables (like `StaveNote`, `TabNote`, etc.)
* the `format()` calls here refer to the static class methods for the specific elements, which all modifiers in `PREFORMAT` should have. Since interfaces apply to class instances, not the classes themselves, we can't really create a generic type for them.
* `postFormat` is also static, so we can't really create a generic type for it too.

So I'd say leave this class as an exception and use `any[]` for `PREFORMAT`, `POSTFORMAT`.

Re: `ModifierClass`, I don't understand why it's there, why can't we just use `Modifier` here? Do we have any code that adds `StaveNote` as a modifier?

(FYI: https://github.com/0xfe/vexflow/wiki/How-Formatting-Works)

rvilarl: I was wrong `ModifierClass = Modifier | StaveNote` works and does not produce a failure because `Tremolo` is not in `PreformatModifierType`. This together with `as` in `StaveNote.getAccidentals` and `StaveNote.getDots`makes the change more sensible.
> Re: `ModifierClass`, I don't understand why it's there, why can't we just use `Modifier` here? Do we have any code that adds `StaveNote` as a modifier?

`StaveNote` does it:
https://github.com/0xfe/vexflow/blob/ba717d830c8a37d25f4e690f9b2983adc852d400/src/stavenote.ts#L877
 
> (FYI: https://github.com/0xfe/vexflow/wiki/How-Formatting-Works)

According to the wikie: "`Modifiers` do not, but instead have a static `<Modifier>.format()` method."
However `Tremolo` is a `Modifier` and does not have the static method `format`. **Why is that?**

rvilarl:
Independently of the `StaveNote` and `Tremolo` issue in our conversation above. I believe that this is from a migration perspective ready and now can be merged!

src/modifiercontext.ts
0xfe: Okay, so that code does not add `StaveNote` as a modifier, just as an element in the `ModifierContext` -- this is the same for `TabNote` and anything that has `setModifierContext`.

So two things, for `ModifierClass`, let's rename it to something more descriptive, maybe `ModifierContextMember` as so:

```typescript
export interface ModifierContextMember {
  getCategory(...) : string;
  setModifierContext(...);
}
```

Then lets fix some of the internal comments and variable names: `modifier -> member` so it's clear that these are not just modifiers.

> However `Tremolo` is a `Modifier` and does not have the static method `format`.

This is probably because no additional reformatting is required for `Tremolo` --  notice that it's also not in `PREFORMAT`.

0xfe: This is a tricky one -- that's for digging in.

rvilarl: Done

rvilarl:
I think it is more consistent with members. Ready to merge!

src/modifiercontext.ts
```diff
@@ -0,0 +1,178 @@
...
+    // Update width of this member context
+    this.width = this.state.left_shift + this.state.right_shift;
+    this.preFormatted = true;
+  }
+
+  postFormat(): void {
+    if (this.postFormatted) return;
+    this.POSTFORMAT.forEach((member) => {
+      L('Postformatting ModifierContext: ', member.CATEGORY);
+      member.postFormat(this.getMembers(member.CATEGORY) as StaveNote[]);
```
0xfe, line 175: The `as StaveNote` here should not exist, because it restricts `this.POSTFORMAT` -- `as Tickable` is better here.

0xfe: Also in `stavenote.ts`, `postFormat` can take `Tickable[]` as a parameter.

src/modifiercontext.ts
```diff
@@ -0,0 +1,178 @@
...
+  text_line: number;
+  top_text_line: number;
+}
+
+export interface ModifierContextMetrics {
+  width: number;
+  spacing: number;
+}
+
+export type PreformatModifierType =
```
0xfe, line 38: This can be `any` and moved inline.

src/modifiercontext.ts
```diff
@@ -0,0 +1,178 @@
...
+  | typeof NoteSubGroup
+  | typeof StringNumber
+  | typeof Articulation
+  | typeof Ornament
+  | typeof Annotation
+  | typeof ChordSymbol
+  | typeof Bend
+  | typeof Vibrato;
+
+export type PostformatModifierType = typeof StaveNote;
```
0xfe, line 54: This also can be `any` and moved inline.

src/modifiercontext.ts
```diff
@@ -0,0 +1,178 @@
...
+  addMember(member: ModifierContextMember): this {
+    const type = member.getCategory();
+    if (!this.members[type]) this.members[type] = [];
+    this.members[type].push(member);
+    member.setModifierContext(this);
+    this.preFormatted = false;
+    return this;
+  }
+
+  getMembers(type: string): ModifierContextMember[] {
```
0xfe, line 128: These break public interfaces, so make some deprecated aliases with a console warning.

```typescript
addModifier(...) {
  LW("addModifier is deprecated, use addMember instead.")
  return addMember(...);
}
```

0xfe (commented):
Thanks -- getting there.

src/modifiercontext.ts
rvilarl: I changed to `Note` it makes more sense with the variables used in postFormat and it shoudl be enough as generalization. Do you agree?

src/modifiercontext.ts
rvilarl: done

src/modifiercontext.ts
rvilarl: done

rvilarl:
Ready

0xfe:
Looks good! Merging. Whew :-)

0xfe:
FYI @rvilarl -- this is probably a good checkpoint to run the visual tests against the last release and see how much we've diverged. IIRC, you had tracked the number of divergences since the TS migration started I think.

PR #956, opened 2021-05-02
Migration/stavetempo

rvilarl:
StaveTempo!


src/stavetempo.ts
```diff
@@ -4,13 +4,38 @@
...
+import { Stave } from './stave';
+
+export interface StaveTempoOptions {
+  bpm: number;
+  dots: number;
+  duration: string;
+  name: string;
+}
+
+export interface Tempo {
```
0xfe, line 17: Both `StaveTempoOptions` and `Tempo` are the same -- let's just use `StaveTempoOptions`.

rvilarl: done

rvilarl:
Ready

0xfe:
Looks good!

PR #957, opened 2021-05-02
Migration/stavetext

rvilarl:
StaveText!


src/stavetext.ts
```diff
@@ -31,36 +52,41 @@ export class StaveText extends StaveModifier {
...
-  setShiftY(y) {
+
+  setShiftY(y: number): this {
     this.shift_y = y;
     return this;
   }
 
-  setFont(font) {
+  setFont(font: FontInfo): this {
     Vex.Merge(this.font, font);
```
0xfe, line 75: We can use the spread operator here.

0xfe (commented):
Looks good. Minor comment.

src/stavetext.ts
rvilarl: `FontInfo` has no optional member. Equal is enough.

src/stavetext.ts
```diff
@@ -31,36 +52,41 @@ export class StaveText extends StaveModifier {
...
+
+  setShiftY(y: number): this {
     this.shift_y = y;
     return this;
   }
 
-  setFont(font) {
-    Vex.Merge(this.font, font);
+  setFont(font: FontInfo): this {
+    this.font = font;
```
0xfe, line 75: This changes the interface because `this.font` is already initialized. So if I do `setFont({size: 12}`, then `family` and `weight` will be clobbered.

This should be `this.font = {...this.font, ...font}`.

rvilarl: Now none of the attributes (`family`, `size` or `weight`) is optional. 
```js
export interface FontInfo {
  size: number;
  weight: string;
  family: string;
}
```
In order this to work we have to make at least one attibute optional. Do you want all of them optional? 
```js
export interface FontInfo {
  size?: number;
  weight?: string;
  family?: string;
}
```

0xfe: Right, I think it's good to have them as non-optional. But for JS compatibility, we don't want to change the behavior.

rvilarl: Understood, done.

rvilarl:
Ready to merge

rvilarl:
@0xfe anything pending here?

0xfe:
Looks good, merging!

PR #959, opened 2021-05-04
Migrate Registry

ronyeh:
Ready for code review. Tests pass. No visual diffs.


src/registry.ts
```diff
@@ -0,0 +1,139 @@
...
+    return this.getElementsByAttribute('type', type);
+  }
+
+  getElementsByClass(className: string): Element[] {
+    return this.getElementsByAttribute('class', className);
+  }
+
+  // This is called by the element when an attribute value changes. If an indexed
+  // attribute changes, then update the local index.
+  onUpdate(id: string, name: string, value: string | undefined, oldValue: string | undefined): Registry {
```
0xfe, line 132: This changes the interface, and can break existing uses. We should take an object parameter (like the prior implementation) for backward compatibility.

rvilarl: @0xfe do you refer to `id` ?

0xfe: yes, correct -- I mean the original interface took a single object parameter.

ronyeh: Will do!

I originally assumed onUpdate() was a private API, since it is only used by element.ts, and does not show up in the tests. Changing it to separate parameters with explicit types reduces the number of TypeScript warnings and required conditional checks.... **but** if it has been used by external projects and consumed as a public API, then we definitely need to not break the API.

I'll make the changes and update this PR.

src/element.ts
```diff
@@ -66,8 +66,9 @@ export abstract class Element {
     this.musicFont = Flow.DEFAULT_FONT_STACK[0];
 
     // If a default registry exist, then register with it right away.
-    if (Registry.getDefaultRegistry()) {
-      Registry.getDefaultRegistry().register(this);
+    const defaultRegistry = Registry.getDefaultRegistry();
+    if (defaultRegistry) {
+      defaultRegistry.register(this);
```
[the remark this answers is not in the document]
ronyeh: Done!

src/registry.ts
ronyeh: Done!

0xfe:
Looks good. Merging.

PR #960, opened 2021-05-05
Migration/timesignote

rvilarl:
TimeSignature & TimeSignote migrated.

New class TimeSignatureGlyph to refactor the glyph creation out of TimeSignature. No change in any existent interface.


src/glyph.ts
```diff
@@ -371,10 +371,10 @@ export class Glyph extends Element {
...
       width: this.bbox.getW(),
       height: this.bbox.getH(),
-      scale: 1,
-      x_shift: 0,
-      y_shift: 0,
-      outline: [],
+      scale: this.metrics.scale,
+      x_shift: this.metrics.x_shift,
+      y_shift: this.metrics.y_shift,
+      outline: this.metrics.outline,
```
mscuthbert, line 377: should this be a copy of the Array, so it doesn't mutate?

src/timesigglyph.ts
```diff
@@ -0,0 +1,85 @@
...
+    let botWidth = 0;
+    for (let i = 0; i < botDigits.length; ++i) {
+      const num = botDigits[i];
+      const botGlyph = new Glyph('timeSig' + num, this.timeSignature.point!);
+
+      this.botGlyphs.push(botGlyph);
+      botWidth += botGlyph.getMetrics().width!;
+    }
+
+    this.width = topWidth > botWidth ? topWidth : botWidth;
```
mscuthbert, line 42: clearer perhaps as `this.width = Math.max(topWidth, botWidth)` ?

src/timesigglyph.ts
```diff
@@ -0,0 +1,85 @@
...
+    this.xMin = this.getMetrics().x_min;
```
mscuthbert, line 43: this just returns the current value of xMin, so it's the same as `this.xMin = this.xMin`.

src/timesignature.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    if (!this.stave) {
+      throw new Vex.RERR('TimeSignatureError', "Can't draw time signature without stave.");
+    }
+
+    this.setRendered();
+    this.timeSig!.glyph.setStave(this.stave);
+    this.timeSig!.glyph.setContext(this.stave.getContext());
+    this.placeGlyphOnLine(this.timeSig!.glyph, this.stave, this.timeSig!.line!);
```
mscuthbert, line 139: Why is there an "!" after .line"!"?   For numeric time signatures,  this.timeSig does not have a line.

src/timesignote.ts
```diff
@@ -1,46 +1,46 @@
...
     this.setRendered();
 
     if (!this.timeSig.glyph.getContext()) {
-      this.timeSig.glyph.setContext(this.context);
+      this.timeSig.glyph.setContext(ctx);
     }
 
     this.timeSig.glyph.setStave(this.stave);
-    this.timeSig.glyph.setYShift(this.stave.getYForLine(this.timeSig.line) - this.stave.getYForGlyphs());
+    this.timeSig.glyph.setYShift(this.stave.getYForLine(this.timeSig.line!) - this.stave.getYForGlyphs());
```
mscuthbert, line 43: as above

mscuthbert (commented):
Looks very good to me.  Some points that could be looked at.

src/glyph.ts
rvilarl: But the attributes above are not one to one copies of metric:
```js
return {
      x_min: this.metrics.x_min * this.scale * this.metrics.scale,
      x_max: this.metrics.x_max * this.scale * this.metrics.scale,
      width: this.bbox.getW(),
      height: this.bbox.getH(),
      scale: this.metrics.scale,
      x_shift: this.metrics.x_shift,
      y_shift: this.metrics.y_shift,
      outline: this.metrics.outline,
    };
```
I did not went into this investigation. I find specially strange the two scales: `this.scale` and `this.metrics.scale`
@mscuthbert do you know why is that so?

src/timesigglyph.ts
rvilarl: done

src/timesigglyph.ts
rvilarl: `this.xMin` removed, now using `this.getMetrics().x_min`

rvilarl: I had to undo this change and leave as it was. Somehow `this.getMetrics().x_min` changes its value in excution time.

src/timesignature.ts
rvilarl: You are right! This was also the case in the original code. Do you know where this line should come from?

rvilarl: @0xfe @ronyeh this example shows that disabling `@typescript-eslint/no-non-null-assertion` is not a good idea. I agree with @ronyeh we should enable it again and use `// eslint-disable-next-line` with a lot of care.

ronyeh: I am not a fan of using the non-null assertion (!) operator for two reasons:

1) Its presence in the code suggests that the code could be improved in the future through better type information. We are using TypeScript, so we should strive to benefit from type checking.

2) It's kind of difficult to spot while reading code, and it's also difficult to search for via CTRL/CMD+F, because the ! is also the not operator. While `// eslint-disable-next-line` doesn't look as elegant, it's a big red flag to future readers of the code. When we revisit those lines post TS migration, we could explore ways to refactor to avoid the `// eslint-disable-next-line`.

I am also not a fan of over-using optional properties (? operator) for a similar reason. When I was migrating the Parser, I realized that the Results interface has a lot of optional properties. https://github.com/ronyeh/vexflow/blob/8318a4113f3439af1e23fd23b61f049a89231e80/src/parser.ts#L36  Some of the properties are only used in the tokenizer/lexer stage, and other properties are only used in the parser stage. In the future, it might be nice to break it up into two: LexerResult vs. ParserResult. This might clean up some of the code later on, where we have to check for the presence of properties before proceeding.

0xfe: I generally agree with all of this, and it's good practice to be explicit about things, have sane defaults, minimize optional properties etc. But we're migrating a relatively large decade-old code base with lots of warts (JS itself evolved significantly over that period). This is already tiresome, so I'm hesitant to put up too many barriers and increase friction.

I've also noticed that without `!`, there's a tendency to use things like `?? 0` or `|| default_value`, which IMO is a lot worse because they lead to silent hard-to-debug incorrect behavior. I'd rather have the code fail early and loud with a clear stack trace than the alternative, and lean on heavy unit testing to force as many code paths as possible.

My suggestion would be to minimize the use of things like `any`, `!`, `?` wherever possible. If the pure approach incurs too much gymnastics, then use a shortcut (like `any` or `!`), and deal with it later (add a TODO, file an issue, etc.) We're still gaining a ton from the other stuff.

src/glyph.ts
rvilarl: @0xfe can you answer the question?

src/timesignature.ts
rvilarl: @0xfe Do you know where this line should come from?

src/timesignature.ts
```diff
@@ -0,0 +1,142 @@
...
+// ## Description
+// Implements time signatures glyphs for staffs
+// See tables.js for the internal time signatures
+// representation
+
+import { Vex } from './vex';
+import { Glyph } from './glyph';
+import { StaveModifier } from './stavemodifier';
+import { TimeSignatureGlyph } from './timesigglyph';
+export interface TimeSignatureInfo {
```
0xfe, line 12: New line before export.

src/timesignature.ts
```diff
@@ -0,0 +1,142 @@
...
+  glyph: Glyph;
+  line?: number;
+  num: boolean;
+}
+
+export interface TimeSignatureGlyphInfo {
```
0xfe, line 18: This can be inline.

src/timesignature.ts
0xfe: This is effectively a y-shift, so it can default to `0`. You can make this non-optional.

src/timesignature.ts
rvilarl: done

src/timesignature.ts
rvilarl: done

src/timesignature.ts
rvilarl: Done

rvilarl:
@0xfe I had to change it to 3 to get 'C' & `C|` placed in the right line.
Ready to merge. I would be good to understand the open conversations though.

src/glyph.ts
0xfe: `scale` is a default scaling factor for the class, and used by the rendering context, and `this.metrics.scale` is a font-specific override loaded from the font metrics files. E.g., this allows two `flat` symbols in different fonts with different sizes to scale consistently within the context.

0xfe: For `scale` it may be better to return `this.metrics.scale * this.scale` -- either way, how does this affect the visual tests? Unclear to me what the downstream effects of this change are.

src/timesignature.ts
```diff
@@ -0,0 +1,137 @@
+// [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
+//
+// ## Description
+// Implements time signatures glyphs for staffs
+// See tables.js for the internal time signatures
+// representation
+
+import { Vex } from './vex';
```
0xfe, line 8: Tough to review this file, any way to make it show up as a diff in here?

src/glyph.ts
rvilarl: No, the number of change go beyond the threshold

rvilarl: // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 //
 // ## Description
 // Implements time signatures glyphs for staffs
 // See tables.js for the internal time signatures
 // representation
 
 import { Vex } from './vex';
 import { StaveModifier } from './stavemodifier';
+import { TimeSignatureGlyph } from './timesigglyph';
 
-const assertIsValidFraction = (timeSpec) => {
+export interface TimeSignatureInfo {
+  glyph: Glyph;
+  line?: number;
+  num: boolean;
+}
+
+const assertIsValidFraction = (timeSpec: string) => {
   const numbers = timeSpec.split('/').filter((number) => number !== '');
 
   if (numbers.length !== 2) {
     throw new Vex.RERR(
       'BadTimeSignature',
       `Invalid time spec: ${timeSpec}. Must be in the form "<numerator>/<denominator>"`
     );
   }
 
   numbers.forEach((number) => {
     if (isNaN(Number(number))) {
       throw new Vex.RERR('BadTimeSignature', `Invalid time spec: ${timeSpec}. Must contain two valid numbers.`);
     }
   });
 };
 
 export class TimeSignature extends StaveModifier {
-  static get CATEGORY() {
+  point?: number;
+  bottomLine?: number;
+  protected timeSig?: TimeSignatureInfo;
+
+  protected validate_args: boolean;
+
+  topLine?: number;
+
     return 'timesignatures';
   }
 
-  static get glyphs() {
+  static get glyphs(): Record<string, { code: string; point: number; line: number }> {
     return {
       C: {
         code: 'timeSigCommon',
         point: 40,
         line: 2,
       },
       'C|': {
         code: 'timeSigCutCommon',
         point: 40,
         line: 2,
       },
     };
   }
 
-  constructor(timeSpec = null, customPadding = 15, validate_args = true) {
+  constructor(timeSpec?: string, customPadding = 15, validate_args = true) {
     super();
     this.setAttribute('type', 'TimeSignature');
     this.validate_args = validate_args;
 
-    if (timeSpec === null) return;
-
+    if (!timeSpec) return;
     const padding = customPadding;
 
     this.point = this.musicFont.lookupMetric('digits.point');
     const fontLineShift = this.musicFont.lookupMetric('digits.shiftLine', 0);
     this.topLine = 2 + fontLineShift;
     this.bottomLine = 4 + fontLineShift;
     this.setPosition(StaveModifier.Position.BEGIN);
-    this.setTimeSig(timeSpec);
-    this.setWidth(this.timeSig.glyph.getMetrics().width);
+    this.timeSig = this.parseTimeSpec(timeSpec);
+    this.setWidth(this.getTimeSig().glyph.getMetrics().width!);
     this.setPadding(padding);
   }
 
-  getCategory() {
+  getCategory(): string {
     return TimeSignature.CATEGORY;
   }
 
-  parseTimeSpec(timeSpec) {
+  parseTimeSpec(timeSpec: string): TimeSignatureInfo {
     if (timeSpec === 'C' || timeSpec === 'C|') {
       const { line, code, point } = TimeSignature.glyphs[timeSpec];
       return {
         line,
         num: false,
         glyph: new Glyph(code, point),
       };
     }
 
     if (this.validate_args) {
       assertIsValidFraction(timeSpec);
     }
 
     const [topDigits, botDigits] = timeSpec.split('/').map((number) => number.split(''));
 
     return {
       num: true,
       glyph: this.makeTimeSignatureGlyph(topDigits, botDigits),
     };
   }
 
-  makeTimeSignatureGlyph(topDigits, botDigits) {
-    const glyph = new Glyph('timeSig0', this.point);
-    glyph.topGlyphs = [];
-    glyph.botGlyphs = [];
-
-    let topWidth = 0;
-    for (let i = 0; i < topDigits.length; ++i) {
-      const num = topDigits[i];
-      const topGlyph = new Glyph('timeSig' + num, this.point);
-
-      glyph.topGlyphs.push(topGlyph);
-      topWidth += topGlyph.getMetrics().width;
-    }
-
-    let botWidth = 0;
-    for (let i = 0; i < botDigits.length; ++i) {
-      const num = botDigits[i];
-      const botGlyph = new Glyph('timeSig' + num, this.point);
-
-      glyph.botGlyphs.push(botGlyph);
-      botWidth += botGlyph.getMetrics().width;
-    }
-
-    const width = topWidth > botWidth ? topWidth : botWidth;
-    const xMin = glyph.getMetrics().x_min;
-
-    glyph.getMetrics = () => ({
-      x_min: xMin,
-      x_max: xMin + width,
-      width,
-    });
-
-    const topStartX = (width - topWidth) / 2.0;
-    const botStartX = (width - botWidth) / 2.0;
-
-    const that = this;
-    glyph.renderToStave = function renderToStave(x) {
-      let start_x = x + topStartX;
-      for (let i = 0; i < this.topGlyphs.length; ++i) {
-        const glyph = this.topGlyphs[i];
-        Glyph.renderOutline(
-          this.context,
-          glyph.metrics.outline,
-          glyph.scale,
-          start_x + glyph.x_shift,
-          this.stave.getYForLine(that.topLine)
-        );
-        start_x += glyph.getMetrics().width;
-      }
-
-      start_x = x + botStartX;
-      for (let i = 0; i < this.botGlyphs.length; ++i) {
-        const glyph = this.botGlyphs[i];
-        that.placeGlyphOnLine(glyph, this.stave, glyph.line);
-        Glyph.renderOutline(
-          this.context,
-          glyph.metrics.outline,
-          glyph.scale,
-          start_x + glyph.x_shift,
-          this.stave.getYForLine(that.bottomLine)
-        );
-        start_x += glyph.getMetrics().width;
-      }
-    };
-
+  makeTimeSignatureGlyph(topDigits: string[], botDigits: string[]): Glyph {
+    const glyph = new TimeSignatureGlyph(this, topDigits, botDigits, 'timeSig0', this.point!);
     return glyph;
   }
 
-  getTimeSig() {
-    return this.timeSig;
+  getTimeSig(): TimeSignatureInfo {
+    return this.timeSig!;
   }
 
-  setTimeSig(timeSpec) {
+  setTimeSig(timeSpec: string): this {
     this.timeSig = this.parseTimeSpec(timeSpec);
     return this;
   }
 
-  draw() {
+  draw(): void {
     if (!this.x) {
       throw new Vex.RERR('TimeSignatureError', "Can't draw time signature without x.");
     }
 
     if (!this.stave) {
       throw new Vex.RERR('TimeSignatureError', "Can't draw time signature without stave.");
     }
 
     this.setRendered();
-    this.timeSig.glyph.setStave(this.stave);
-    this.timeSig.glyph.setContext(this.stave.context);
-    this.placeGlyphOnLine(this.timeSig.glyph, this.stave, this.timeSig.line);
-    this.timeSig.glyph.renderToStave(this.x);
+    this.timeSig!.glyph.setStave(this.stave);
+    this.timeSig!.glyph.setContext(this.stave.getContext());
+    this.placeGlyphOnLine(this.timeSig!.glyph, this.stave, 2);
+    this.timeSig!.glyph.renderToStave(this.x);
   }
 }

rvilarl: @0xfe there you are, generated with diff.

src/timesignote.ts
rvilarl: done. Also `2`

src/glyph.ts
rvilarl: > For `scale` it may be better to return `this.metrics.scale * this.scale` -- either way, how does this affect the visual tests? Unclear to me what the downstream effects of this change are.

Done, no change on images.

src/timesignature.ts
rvilarl: @0xfe You got the diff above :)

rvilarl:
Review comments addressed.

src/glyph.ts
0xfe: Thanks for the diff!

src/timesignature.ts
```diff
@@ -0,0 +1,137 @@
...
+    };
+  }
+
+  makeTimeSignatureGlyph(topDigits: string[], botDigits: string[]): Glyph {
+    const glyph = new TimeSignatureGlyph(this, topDigits, botDigits, 'timeSig0', this.point!);
+    return glyph;
+  }
+
+  getTimeSig(): TimeSignatureInfo {
+    return this.timeSig!;
```
0xfe, line 114: Better if this returns `TimeSignatureInfo?` and remove `!` from body. The callers can do the explicit check.

0xfe: Actually, looks like `this.timeSig` is set in the constructor, so no need for the protected instance variable to be optional, no?

src/timesignature.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    if (!this.stave) {
+      throw new Vex.RERR('TimeSignatureError', "Can't draw time signature without stave.");
+    }
+
+    this.setRendered();
+    this.timeSig!.glyph.setStave(this.stave);
+    this.timeSig!.glyph.setContext(this.stave.getContext());
+    this.placeGlyphOnLine(this.timeSig!.glyph, this.stave, 2);
```
0xfe, line 134: Where is the `2` coming from?

rvilarl: It is the value by try and change that makes the `C` & `C|` time signatures stay in the middle. The line has no influence on the position but it does have one on the two mentioned before. `2` is the middle line.

0xfe: Okay, thanks, can you add a comment stating 2 == middle line, and it affects only the `C` and `C|`.

src/timesignature.ts
rvilarl: Changed to return `TimeSignature | undefined`, I agree that this is better.
`this.timeSig` is not always set in the constructor due to the `if (!timeSpec) return;`

rvilarl:
@0xfe ready to merge

0xfe:
Great.

PR #961, opened 2021-05-05
Add 'grunt dev' task, which watches src/ and tests/ concurrently.

ronyeh:
When a file changes, it triggers the appropriate webpack task: `webpack:watchDev` or `webpack:watchTest`.

This allows for quick development and testing by simply refreshing flow.html in the local browser.

***** DETAILS BELOW *****

Due to the migration to TypeScript, it looks like the `grunt watch` task no longer works as intended. When we concatenate TS files together, the web browser complains about import statements being outside of a JS module.

To fix this, I did the following:

1) `npm install --save-dev grunt-concurrent`

2) Modify the Gruntfile.js to include:

```
grunt.loadNpmTasks('grunt-concurrent');

... 

grunt.registerTask('dev', 'Watch src/ and tests/ concurrently', ['clean', 'eslint', 'concurrent']);
```

3) Rename the `webpack:watch` task to `webpack:watchDev`. Add a new task `webpack:watchTest` which does the same thing, except with the test/ sources.

4) Add a `concurrent` task, which runs `webpack:watchDev` and `webpack:watchTest` in parallel.

Now when we type `grunt dev` on the command line, it outputs something like this:

```
....
Running "concurrent:tasks" (concurrent) task
    Running "webpack:watchDev" (webpack) task
    Running "webpack:watchTest" (webpack) task
```

We can now open flow.html in a local browser, and modifying files in `src/` or `tests/` will compile only the
`vexflow-debug.js` and `vexflow-tests.js` as appropriate. Then refresh flow.html in the local browser to see the tests run, just like before.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -82,12 +82,24 @@ module.exports = (grunt) => {
       build: webpackProd,
       buildDev: webpackDev,
       buildTest: webpackTest,
-      watch: {
+      watchDev: {
         ...webpackDev,
         watch: true,
         keepalive: true,
         failOnError: false,
       },
+      watchTest: {
+        ...webpackTest,
+        watch: true,
+        keepalive: true,
+        failOnError: false,
+      },
+    },
+    concurrent: {
+      options: {
+        logConcurrentOutput: true,
+      },
+      tasks: ['webpack:watchDev', 'webpack:watchTest'],
     },
     eslint: {
       target: SOURCES.concat('./tests'),
@@ -140,8 +152,8 @@ module.exports = (grunt) => {
           out: 'build/typedocs',
           name: 'vexflow',
         },
-        src: ['./typedoc.ts']
-      }
+        src: ['./typedoc.ts'],
+      },
     },
     gitcommit: {
       releases: {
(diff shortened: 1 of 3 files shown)


ronyeh:
@0xfe if you can verify that 'grunt watch' no longer does what you want it to do... perhaps I can rename 'grunt dev' to 'grunt watch'

ronyeh:
I actually prefer the alternate approach suggested in this PR: https://github.com/0xfe/vexflow/pull/964

Gruntfile.js
```diff
@@ -189,9 +201,19 @@ module.exports = (grunt) => {
...
+  grunt.registerTask('default', [
+    'clean',
+    'eslint',
+    'webpack:build',
+    'webpack:buildDev',
+    'webpack:buildTest',
+    'docco',
+    'typedoc',
+  ]);
+  grunt.registerTask('dev', 'Watch src/ and tests/ concurrently', ['clean', 'eslint', 'concurrent']);
```
0xfe, line 216: This is great, thanks! Let's call this `watch`.

0xfe:
Merged the other one. Closing this.

PR #964, opened 2021-05-06
Update 'grunt watch' to watch src/ and tests/ concurrently.

ronyeh:
It will trigger the correct webpack task when it notices a changed file.

This is an alternative to an earlier PR proposal: https://github.com/0xfe/vexflow/pull/961

I actually prefer this PR, because it updates 'grunt watch' to use the webpack tasks, so it handles TypeScript files gracefully.

I decided to not include the `package-lock.json` in the PR, so that the diff is easy to understand.

If Mohit decides to merge, he can do a npm install on his machine and then check in the new `package-lock.json`.

I've tested this version and am happy with it. I drag flow.html into my local browser, and edit either src/* files or tests/* files and it recompiles quickly. Then I just CMD+R my browser to run the tests again.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -82,12 +82,24 @@ module.exports = (grunt) => {
       build: webpackProd,
       buildDev: webpackDev,
       buildTest: webpackTest,
-      watch: {
+      watchDev: {
         ...webpackDev,
         watch: true,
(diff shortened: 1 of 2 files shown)


0xfe:
This is great, merging!

PR #965, opened 2021-05-06
Registry: methods that return this should use the 'this' return type.

ronyeh:
Small change.
Improving the Registry class with more correct typings.


0xfe:
Good catch.

PR #967, opened 2021-05-07
Migration/tabnote

rvilarl:
TabNote


src/tabnote.ts
```diff
@@ -228,21 +240,21 @@ export class TabNote extends StemmableNote {
...
+  setStave(stave: Stave): this {
     super.setStave(stave);
-    this.context = stave.context;
+    this.setContext(stave.getContext());
 
     // Calculate the fret number width based on font used
     let i;
-    if (this.context) {
-      const ctx = this.context;
+    const ctx = this.getContext();
```
0xfe, line 249: You can remove a `getContext` call with:

```typescript
setStave(..) {
  super.setStave(stave);
  const ctx = stave.getContext();
  this.setContext(ctx);

  let i;
  ...
}
```

src/tabnote.ts
```diff
@@ -228,21 +240,21 @@ export class TabNote extends StemmableNote {
...
 
   // Set the `stave` to the note
-  setStave(stave) {
+  setStave(stave: Stave): this {
     super.setStave(stave);
-    this.context = stave.context;
+    this.setContext(stave.getContext());
 
     // Calculate the fret number width based on font used
     let i;
```
0xfe, line 248: I think you can remove this, and move the `let` to the beginning of the loop below.

src/tabnote.ts
```diff
@@ -266,13 +278,13 @@ export class TabNote extends StemmableNote {
...
-  getPositions() {
+  getPositions(): { str: string; fret: string }[] {
     return this.positions;
   }
 
   // Add self to the provided modifier context `mc`
-  addToModifierContext(mc) {
-    this.setModifierContext(mc);
+  addToModifierContext(mc: ModifierContext): this {
+    this.modifierContext = mc;
```
0xfe, line 287: Why not `this.setModifierContext`?

rvilarl: To let the compiler realize that `this.modifierContext` is not `undefined` avoiding the usage of `!`. 
It is a `protected` member and we save a function call which simply does the assignment.

src/tabnote.ts
rvilarl: done

src/tabnote.ts
rvilarl: done

rvilarl:
Ready to merge!

rvilarl:
@0xfe anything pending here?

src/tabnote.ts
0xfe: Okay, that's a good reason. Looking at the codebase, it seems like `setModifierContext` doesn't do anything special, and it's only called in three places.

Can you remove `setModifierContext` entirely from `note.ts` and `modifier.ts`, and the references in `stavenote.ts`, `tabnote.ts`, and `modifiercontext.ts` too in this PR please? It's a good cleanup.

0xfe (commented):
One minor comment, rest of the change looks good.

src/tabnote.ts
rvilarl: I have modified `StaveNote` to use the protected member. We cannot remove `setModifierContext` it is used by `ModifierContext` in `addMember`

rvilarl:
I have modified `StaveNote` to use the protected member. We cannot remove `setModifierContext` it is used by `ModifierContext` in `addMember`
@0xfe ready to merge

src/tabnote.ts
0xfe: Okay, yes, thanks.

src/tabnote.ts
```diff
@@ -431,33 +449,40 @@ export class TabNote extends StemmableNote {
...
-  draw() {
-    this.checkContext();
+  draw(): void {
+    const ctx = this.checkContext();
 
     if (!this.stave) {
       throw new Vex.RERR('NoStave', "Can't draw without a stave.");
     }
 
+    if (!this.stem) {
```
0xfe, line 474: This seems wrong -- you can have tab notes without stems. Does this affect the tests?

src/tabnote.ts
```diff
@@ -466,13 +491,13 @@ export class TabNote extends StemmableNote {
     this.stem.setNoteHeadXBounds(stem_x, stem_x);
 
     if (render_stem) {
```
0xfe, line 493: This is a better place to put a check for `this.stem`.

src/tabnote.ts
rvilarl: No impact on tests. However I agree with you. I have changed the implementation and it also pass all the tests.

src/tabnote.ts
rvilarl: Done this way, I am happy to see we agree! :)

rvilarl:
@0xfe ready to merge!

0xfe:
lgtm

PR #968, opened 2021-05-07
Use webpack to bundle tests, instead of concatenating the JS files.

ronyeh:
I kept the globals as is for now (e.g., VF.Test.Dot, Vex.Flow.Test.Accidental). As we port the tests over to TypeScript, the compiler will complain and then we can either remove the globals (assuming they aren't used elsewhere), or find a TS approved way to introduce globals. (I assume if we have a global declaration for Vex.Flow.Test as `any`, then TS won't complain.) But I'm not sure we need those Vex.Flow.Test.* anymore, once everything is in typescript.

All the test scripts work for me, and `flow.html` reports: Tests completed in 10699 milliseconds. 3305 assertions of 3305 passed, 0 failed.


0xfe:
This is really excellent. Glad to finally be rid of `concat`! :-)

PR #970, opened 2021-05-08
migration/typedoc

rvilarl:
documentation update

I had to resolve overlaps on `X` and `Position`

@ronyeh I removed the export from 'X' typedoc was complaining about two different implementations of the same element.


0xfe:
Thanks @rvilarl --- looks good. `X` can be internal.

PR #972, opened 2021-05-08
Migration/dot

rvilarl:
Dot!


src/dot.ts
```diff
@@ -5,30 +5,44 @@
...
+
+  static get CATEGORY(): string {
     return 'dots';
   }
 
   // Arrange dots inside a ModifierContext.
-  static format(dots, state) {
+  static format(
+    dots: Dot[],
+    state: {
```
0xfe, line 23: I think this state is common to all the static formatter methods. Maybe it's a good idea to define an interface for this in `common`, perhaps `ModiferContextState`?

src/dot.ts
```diff
@@ -5,30 +5,44 @@
...
       let props;
       let shift;
 
-      // Only StaveNote has .getKeyProps()
-      if (typeof note.getKeyProps === 'function') {
+      // If it's a StaveNote
+      if (note instanceof StaveNote) {
         props = note.getKeyProps()[dot.getIndex()];
         shift = note.getRightDisplacedHeadPx();
       } else {
```
0xfe, line 48: Good place to explicitly check for `TabNote` and throw an exception in a default `else`.

src/dot.ts
```diff
@@ -125,20 +139,20 @@ export class Dot extends Modifier {
...
+    const ctx = this.checkContext();
     this.setRendered();
 
-    if (!this.note || this.index === null) {
+    if (!this.note || this.index === undefined) {
       throw new Vex.RERR('NoAttachedNote', "Can't draw dot without a note and index.");
     }
 
-    const lineSpace = this.note.stave.options.spacing_between_lines_px;
+    const lineSpace = this.note.getStave()?.getOptions().spacing_between_lines_px ?? 0;
```
0xfe, line 155: This is a potential silent failure, let's use `getStave()!` instead (or explicitly check for stave above.)

rvilarl: `getStave()!` is also silent you get in the end `undefined` and that's it. In **migration/tmisignote** we got even `this.timeSig!.line!` when `line` was mo known attribute.

rvilarl: Done

src/dot.ts
rvilarl: done

src/dot.ts
0xfe: Is it really silent, I thought the `!` causes an exception if the value is undefined, no?

src/dot.ts
rvilarl: It was already available in `ModifierContext`. :)
Changed.

rvilarl:
@0xfe Ready to merge!

src/dot.ts
rvilarl: No compilation failure and no runtime failure. This is what we observed in **migration/timesignote** with `this.timeSig!.line!`
@ronyeh can you confirm?
I believe this is whay this a deprecated feature in eslint.

0xfe: Oh, no, I completely misunderstood the `!` operator -- I thought it would be a runtime assertion, but it doesn't emit any code at all. (The term "non-null assertion operator" make me think it was an actual assertion.)

In that case, yes 100% this should be completely avoided, and we should use explicit checks like `if (something not defined) { throw exception };`

Apologies for this, can we re add the es-lint rule and replace all the `!` operators with explicit checks (where reasonable), maybe in a separate issue/PR?

src/dot.ts
```diff
@@ -125,20 +135,24 @@ export class Dot extends Modifier {
...
 
-  draw() {
-    this.checkContext();
+  draw(): void {
+    const ctx = this.checkContext();
     this.setRendered();
-
-    if (!this.note || this.index === null) {
+    if (!this.note || !this.note.getStave() || this.index === undefined) {
       throw new Vex.RERR('NoAttachedNote', "Can't draw dot without a note and index.");
```
0xfe, line 147: Message "Drawing a dot requires a stave, a note, and an index".

0xfe (commented):
Minor comment about error message.

src/dot.ts
rvilarl: I will do that once we closed the open PRs. I will open no new one and then I will remove the allowance and go though all the cases. I also understood it wrong at the beginning, it is very misleading!

rvilarl:
@0xfe ready to merge!

src/dot.ts
rvilarl: Done

rvilarl:
@0xfe ready to merge.

src/dot.ts
```diff
@@ -125,20 +135,24 @@ export class Dot extends Modifier {
...
+  draw(): void {
+    const ctx = this.checkContext();
     this.setRendered();
+    if (!this.note || !this.note.getStave() || this.index === undefined) {
+      throw new Vex.RERR('NoStaveNoteIndex', 'Drawing a dot requires a stave, a note, and an index.');
+    }
 
-    if (!this.note || this.index === null) {
-      throw new Vex.RERR('NoAttachedNote', "Can't draw dot without a note and index.");
+    const stave = this.note.getStave();
```
0xfe, line 150: We're already checking for stave and throw an exception above. Maybe:

```typescript
draw() : void {
  const ctx = ...;
  const stave = this.note.getStave();
  if (!stave || !this.note !! this.index === undefined) {
     throw ...
  }
```

  this.setRendered()

rvilarl: Good catch! But we cannot place `this.note.getStave()` before checking the for the existence of `this.note`
I have removed the duplication anyway, thanks.
Done

rvilarl:
@0xfe hopefully ready!

ronyeh:
Hi @rvilarl and @0xfe 

Sorry I missed the conversation, but thank you for clarifying the `?` vs `!` operators, because they are very different.

I did an experiment to see what code is emitted by the TS compiler.

Input:
```
  console.log(c.stave?.hello());
  console.log(c.stave!.goodbye());
```

If c.stave is undefined, the first line prints `undefined` because we are telling TypeScript that we don't know the current status of c.stave. So TypeScript will do the check for us and will short circuit at `stave?` and never ever call `hello()`.

The second line is a runtime error when c.stave is undefined. `Uncaught TypeError: Cannot read property 'goodbye' of undefined.` The `!` operator actually doesn't do anything other than tell TypeScript to stop complaining. In this instance, we are telling the TypeScript compiler that we **_know_** c.stave was set properly, so trust us, it'll work fine.

This is the actual code that is produced by the TypeScript compiler:
```
console.log((_a = c.stave) === null || _a === void 0 ? void 0 : _a.hello());
console.log(c.stave.goodbye());
```

So yeah, the `?` operator produces extra code to help us check if c.stave is either null or undefined (`void 0` === undefined, as I learned today). The `!` operator, on the other hand, does absolutely nothing. It just silences the TS compiler for that particular line of code.

rvilarl:
Hi @ronyeh.
thanks for that! I still have one question about the runtime error. `Uncaught TypeError: Cannot read property 'goodbye' of undefined.` **is this the printout of the console or the execution being stopped with the uncaught exeption?**
I am asking that because on `TimeSignature` on the call to `placeGlyphOnLine` there was a parameter `this.timeSig!.line` where `timeSig` had no `line` attribute. At runtime there was no exception and debugging I could see that the parameter was simply `undefined`.
**Could you make a similar test with a non existent attribute?**

rvilarl:
@ronyeh well revisiting `TimeSignature`, in reallity `this` defines `timeSig?` where `line?` is defined. So I was wrong and `undefined`makes sense.

However I do not get to compile your examples above, I have tried in `draw` at `Dot`:
```
console.log(stave?.hello());
console.log(stave!.goodbye());
```
@0xfe perhaps you are right and `!` is good for runtime. And I have to changed `TimeSignature` the `2` makes no sense.

rvilarl:
@0xfe @ronyeh I have been doing several tries. I used `this.timeSig.line` in `draw` from `TimeSignature` for the investigation.
In `this.timeSig.line` both `timeSig` and `line` can be `undefined`
For the test I have defined two functions:
```
function optional(num?: number)
function notoptional(num: number)
```
**Observations:**
1) The following compile and produce no runtime error if `timeSig` is not `undefined`:
`this.timeSig, this.timeSig!.line!, typeof this.timeSig!.line, typeof this.timeSig?.line`
`notoptional(this.timeSig!.line!)` however the function can get a not expected `undefined` at run time.
`optional(this.timeSig!.line!), optional(this.timeSig!.line), optional(this.timeSig?.line)`

2) The following compile and produce no runtime error if `timeSig` is `undefined`
`this.timeSig?.line`
`notoptional(this.timeSig?.line??0)`
`optional(this.timeSig?.line)`

**Conclusion:**
- we should avoid `!` and only use it when we are 100% sure that the element will not be `undefined`

**Solution for TimeSignature**
Taking into consideration 1) and 2) there are two possibilities:
A)
```
placeGlyphOnLine(glyph: Glyph, stave: Stave, line: number, customShift = 0): void {
...
this.placeGlyphOnLine(this.timeSig!.glyph, this.stave, typeof this.timeSig?.line ?? 0);
```
B)
```
placeGlyphOnLine(glyph: Glyph, stave: Stave, line?: number, customShift = 0): void {
   line = line ?? 0;
...
this.placeGlyphOnLine(this.timeSig!.glyph, this.stave, typeof this.timeSig?.line);
```

**What is your preference A) or B)?**

rvilarl:
@0xfe this has no longer to do with Dot and I think that Dot is ready to merge. Is that right?

PR #973, opened 2021-05-08
Migration/crescendo

rvilarl:
Crescendo!


src/crescendo.ts
```diff
@@ -55,58 +72,62 @@ export class Crescendo extends Note {
...
     this.height = 15;
 
-    Vex.Merge(this.render_options, {
-      // Extensions to the length of the crescendo on either side
-      extend_left: 0,
-      extend_right: 0,
-      // Vertical shift
-      y_shift: 0,
-    });
+    this.render_options = {
```
0xfe, line 75: It seems like we don't use the parent's `render_options`, so we can remove this merge entirely.

src/crescendo.ts
```diff
@@ -55,58 +72,62 @@ export class Crescendo extends Note {
...
     const y = this.stave.getYForLine(this.line + -3) + 1;
 
     L('Drawing ', this.decrescendo ? 'decrescendo ' : 'crescendo ', this.height, 'x', begin_x - end_x);
 
-    renderHairpin(this.context, {
-      begin_x: begin_x - this.render_options.extend_left,
-      end_x: end_x + this.render_options.extend_right,
-      y: y + this.render_options.y_shift,
+    renderHairpin(ctx, {
+      begin_x: begin_x - this.render_options.extend_left!,
```
0xfe, line 128: If we remove the merge above, these fields can be non-optional, and you can remove the `!` checks.

src/crescendo.ts
rvilarl: We cannot do that the compiler complains about missing properties.

src/crescendo.ts
rvilarl: `!` is not a check at all! We are telling the compiler not to check, that we are sure that the properties exist and this indeed the case here. They are defined in the constructor.
I will add the eslint comment for the future.

rvilarl:
Ready to merge

src/crescendo.ts
0xfe: Ah, because `render_options` is already typed in the parent. In that case, let's rename this to `options`:

```typescript
this.options = {
  extend_left: 0,
  ...
}
```

src/crescendo.ts
0xfe: Right, it's not a compile-time check, but it's a runtime check, which we can avoid. Seems like we don't really need to override `render_options` here. See my comment above about renaming to `options`.

src/crescendo.ts
rvilarl: done

rvilarl:
@0xfe ready to merge

src/crescendo.ts
```diff
@@ -54,59 +78,55 @@ export class Crescendo extends Note {
...
     const y = this.stave.getYForLine(this.line + -3) + 1;
 
     L('Drawing ', this.decrescendo ? 'decrescendo ' : 'crescendo ', this.height, 'x', begin_x - end_x);
 
-    renderHairpin(this.context, {
-      begin_x: begin_x - this.render_options.extend_left,
-      end_x: end_x + this.render_options.extend_right,
-      y: y + this.render_options.y_shift,
+    renderHairpin(ctx, {
+      // eslint-disable-next-line
```
0xfe, line 124: Unclear why you have these eslint comments?

rvilarl: I forgot to remve them when I removed the `!` operator.

Done!

rvilarl:
@0xfe hopefully ready!

rvilarl:
@0xfe anything pending here?

0xfe:
lgtm

PR #975, opened 2021-05-09
Migration/gracenote

rvilarl:
GraceNote!


src/gracenote.ts
```diff
@@ -1,24 +1,35 @@
...
+  static get SCALE(): number {
     return 0.66;
   }
 
-  constructor(note_struct) {
+  constructor(note_struct: GraceNoteStruct) {
     super({
-      glyph_font_scale: Flow.DEFAULT_NOTATION_FONT_SCALE * GraceNote.SCALE,
-      stroke_px: GraceNote.LEDGER_LINE_OFFSET,
+      ...{
```
0xfe, line 29: This looks strange, what's wrong with:

```typescript
super({
   glyph_font_scale: Flow.DEF...,
   stroke_px: GraceNote.LE...,
   ...note_struct,
});

0xfe (commented):
Looks good overall, minor syntactic question.

src/gracenote.ts
rvilarl: The compiler complains because the attributes are not optional. The compilers thinks that this makes no sense. You explained that we still have to do it due to javascript. If you use two spread operators the compiler does not complain anymore.

rvilarl: @0xfe is that fine?

0xfe: Ah, okay, that's good.

Issue #974, opened 2021-05-08
Typescript migration: formatter


rvilarl:
I have completed the migration and I am getting some differences in the images.
I will share them here to get some tips to fix bugs.

rvilarl:
![Auto_Beaming Break_Beams_on_Middle_Rests_Only](https://user-images.githubusercontent.com/22865285/117550553-51670180-b041-11eb-92c4-2334fd37040f.png)
**Migrated (third rest is lower):**
![Auto_Beaming Break_Beams_on_Middle_Rests_Only_Current](https://user-images.githubusercontent.com/22865285/117550516-18c72800-b041-11eb-9607-9a0e19e4d26c.png)
**Reference(master):**
![Auto_Beaming Break_Beams_on_Middle_Rests_Only_Reference](https://user-images.githubusercontent.com/22865285/117550517-1c5aaf00-b041-11eb-882b-12dc04768537.png)

rvilarl:
![Bend Double_Bends](https://user-images.githubusercontent.com/22865285/117550651-c0dcf100-b041-11eb-9c77-2c9587654476.png)
**Migrated (yellow bar):**
![Bend Double_Bends_Current](https://user-images.githubusercontent.com/22865285/117550656-c5090e80-b041-11eb-8d08-fbbc5630259f.png)
**Reference(master)**
![Bend Double_Bends_Reference](https://user-images.githubusercontent.com/22865285/117550659-c89c9580-b041-11eb-8b69-fdccf6761f62.png)

rvilarl:
![Three_Voice_Rests Auto_Adjust_Rest_Positions___Two_Voices](https://user-images.githubusercontent.com/22865285/117550792-7d36b700-b042-11eb-9c51-71884df8eaa4.png)
**Migrated(Default rest positions are different, but I would say better located)**
![Three_Voice_Rests Auto_Adjust_Rest_Positions___Two_Voices_Current](https://user-images.githubusercontent.com/22865285/117550793-8162d480-b042-11eb-93c8-a1bac99c614f.png)
**Reference(Master)**
![Three_Voice_Rests Auto_Adjust_Rest_Positions___Two_Voices_Reference](https://user-images.githubusercontent.com/22865285/117550798-858ef200-b042-11eb-801f-6d254c805ccf.png)

AaronDavidNewman:
Regarding the last example (3 voice rests), it looks like the default setting to `align_rests` was false but now it's true.

[...]

PR #979, opened 2021-05-11
Migration/formatter2

rvilarl:
formatter in order to further discuss #974 

@AaronDavidNewman @0xfe any tip on the bugs?


src/formatter.ts
```diff
@@ -556,11 +666,10 @@ export class Formatter {
 
               // Calculate the expected distance of the current context from the last matching tickable. The
               // distance is scaled down by the softmax for the voice.
-              if (options.globalSoftmax) {
+              if (formatterOptions.globalSoftmax) {
                 const t = totalTicks;
-                const exp = (v) => Math.pow(softmaxFactor, v / t);
-                expectedDistance = (exp(maxTicks) / expTicksUsed) * adjustedJustifyWidth;
-              } else {
+                expectedDistance = (softmaxFactor ** maxTicks / t / expTicksUsed) * adjustedJustifyWidth;
```
0xfe, line 671: This is wrong too, precedence of `**` is higher than `/`.

src/formatter.ts
```diff
@@ -575,19 +684,20 @@ export class Formatter {
...
       let spaceAccum = 0;
 
       contextList.forEach((tick, index) => {
         const context = contextMap[tick];
         if (index > 0) {
-          const x = context.getX();
+          const contextX = context.getX();
           const ideal = idealDistances[index];
-          const errorPx = ideal.fromTickable.getX() + ideal.expectedDistance - (x + spaceAccum);
+          const errorPx = (ideal.fromTickable?.getX() ?? 0) + ideal.expectedDistance - (contextX + spaceAccum);
```
0xfe, line 700: This is a silent failure, we should use an assertion instead.

src/formatter.ts
```diff
@@ -772,29 +886,29 @@ export class Formatter {
         }
       }
 
-      shift *= options.alpha;
+      if (options) shift *= options.alpha;
```
0xfe, line 889: Silent failure here -- `options.alpha` should always exist.

src/formatter.ts
rvilarl: done

src/formatter.ts
```diff
@@ -479,10 +590,10 @@ export class Formatter {
     });
 
     // Use softmax based on all notes across all staves. (options.globalSoftmax)
-    const options = this.options;
-    const softmaxFactor = options.softmaxFactor || 100;
-    const exp = (tick) => Math.pow(softmaxFactor, contextMap[tick].getMaxTicks().value() / totalTicks);
-    const expTicksUsed = contextList.map(exp).reduce((a, b) => a + b);
+    const formatterOptions = this.formatterOptions;
+    const softmaxFactor = formatterOptions.softmaxFactor || 100;
+    const exp = (tick: number) => softmaxFactor ** contextMap[tick].getMaxTicks().value() / totalTicks;
```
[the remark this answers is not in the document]
rvilarl: done

src/formatter.ts
rvilarl: done

src/formatter.ts
rvilarl: done

src/formatter.ts
rvilarl: done

rvilarl:
@0xfe thanks for the review. However the bugs described in #974 are still there (rest alignment and yellow lines)

rvilarl:
@0xfe @AaronDavidNewman rest alignment fix. The yellow lines are however still there.

rvilarl:
Accidental Accidental_Padding_Current
Bend Double_Bends_Current
Bend Double_Bends_With_Release_Current
Formatter Proportional_Formatting___No_Tuning_Current
GlyphNote GlyphNote_No_Stave_Padding_Current
NoteSubGroup Basic___ClefNote__TimeSigNote_and_BarNote_Current
NoteSubGroup Multi_Voice_Current
NoteSubGroup Multi_Voice_Multiple_Draws_Current
TextNote TextNote_Formatting_2_Current

rvilarl:
I have rebased with master and I get the same.

> We should not need these here, they're set by `tickcontext`. Is current `HEAD` broken (i.e., before this PR?)

@0xfe I do not understand what you mean. I will put some comment in the code to let you know where this is coming from.

src/note.ts
```diff
@@ -574,6 +582,8 @@ export abstract class Note extends Tickable {
       // Modifier spacing.
       modLeftPx,
       modRightPx,
+      totalLeftPx,
+      totalRightPx,
```
rvilarl, line 586: @0xfe the yellow lines appear because `totalLeftPx` and `totalRightPx` before they were undefined causing a NaN in other part of the code I will also highlight.

0xfe:
On Thu., May 13, 2021, 08:46 Rodrigo Vilar, ***@***.***>
wrote:

> I have rebased with master and I get the same.
>
> We should not need these here, they're set by tickcontext. Is current HEAD
> broken (i.e., before this PR?)
>
> @0xfe <https://github.com/0xfe> I do not understand what you mean. I will
> put some comment in the code to let you know where this is coming from.
>

Thanks. I'm asking if this issue is currently in master or is it just in
this formatter PR?

src/note.ts
0xfe: I think the problem is in `evaluate` -- it's the only place where there's a connection between `note.metrics.totalRightPx` and `formatterMetrics`.

What happens if you initialize `totalLeftPx` and `totalRightPx` here to `0`?

Alternatively, you might just be able to remove the references to `note.metrics.totalLeft/RightPx` from `evaluate()` (note: not the `tickContext` metrics properties, just the `note` ones.)

rvilarl:
The yellow lines are not in master but master is doing calculations with undefined which I believe is wrong.

rvilarl:
@0xfe What is the meaning of the yellow line?

0xfe:
> The yellow lines are not in master but master is doing calculations with undefined which I believe is wrong.

Yes, you are correct -- I'm trying to understand if `undefined` numbers are treated as `0` values, or the results are `NaN` values, which don't end up getting rendered.

src/note.ts
rvilarl: They were 0 and they would also be there but overlapping with the other lines. Now they are, I believe, in the right position.

0xfe:
> @0xfe What is the meaning of the yellow line?

They represent the amount of freedom available on the right of the note that the formatter can use to squeeze the right notes into. In these examples, the long lines are correct.

src/note.ts
0xfe: Okay, I think we should remove it completely from here, because the names are not representative of the values. Instead, we should update `evaluate` to derive these values (just like you did here.)

rvilarl:
In master they are not present due to the calculation with NaN, undefined.

rvilarl:
`console.log(xStart, xPre1, xPre1, xAbs, xPost1, xPost2, xEnd, xFreedomRight);` in `Note.plotMetrics` produce the following in master:

Accidental :: Accidental Padding
27.000000000000004 40.74016 40.74016 40.74016 52.67416 52.67416 58.67416 58.67416
58.67416 71.56616 71.56616 71.56616 83.50016 83.50016 83.50016 **83.50016**
88.29980009854792 88.29980009854792 88.29980009854792 88.29980009854792 100.23380009854792 100.23380009854792 100.23380009854792 **100.23380009854792**
110.61446780806918 110.61446780806918 110.61446780806918 110.61446780806918 122.54846780806918 122.54846780806918 122.54846780806918 122.54846780806918
122.54846780806918 161.94270780806917 161.94270780806917 161.94270780806917 173.87670780806917 173.87670780806917 173.87670780806917 173.87670780806917
178.6763479066171 178.6763479066171 178.6763479066171 178.6763479066171 190.6103479066171 190.6103479066171 190.6103479066171 190.6103479066171
190.6103479066171 217.2548279066171 217.2548279066171 217.2548279066171 229.1888279066171 229.1888279066171 229.1888279066171 229.1888279066171
233.98846800516503 233.98846800516503 233.98846800516503 233.98846800516503 245.92246800516503 245.92246800516503 245.92246800516503 245.92246800516503
248.47920333045494 248.47920333045494 248.47920333045494 248.47920333045494 260.41320333045496 260.41320333045496 260.41320333045496 260.41320333045496
262.96993865574484 262.96993865574484 262.96993865574484 262.96993865574484 274.90393865574487 274.90393865574487 274.90393865574487 274.90393865574487
279.7035787542928 279.7035787542928 279.7035787542928 279.7035787542928 291.6375787542928 291.6375787542928 291.6375787542928 291.6375787542928

and in this branch:
   
Accidental :: Accidental Padding
27.000000000000004 40.74016 40.74016 40.74016 52.67416 52.67416 58.67416 58.67416
58.67416 71.56616 71.56616 71.56616 83.50016 83.50016 83.50016 **88.29980009854792**
88.29980009854792 88.29980009854792 88.29980009854792 88.29980009854792 100.23380009854792 100.23380009854792 100.23380009854792 **110.61446780806918**
110.61446780806918 110.61446780806918 110.61446780806918 110.61446780806918 122.54846780806918 122.54846780806918 122.54846780806918 122.54846780806918
122.54846780806918 161.94270780806917 161.94270780806917 161.94270780806917 173.87670780806917 173.87670780806917 173.87670780806917 178.6763479066171
178.6763479066171 178.6763479066171 178.6763479066171 178.6763479066171 190.6103479066171 190.6103479066171 190.6103479066171 190.6103479066171
190.6103479066171 217.2548279066171 217.2548279066171 217.2548279066171 229.1888279066171 229.1888279066171 229.1888279066171 233.98846800516503
233.98846800516503 233.98846800516503 233.98846800516503 233.98846800516503 245.92246800516503 245.92246800516503 245.92246800516503 248.47920333045494
248.47920333045494 248.47920333045494 248.47920333045494 248.47920333045494 260.41320333045496 260.41320333045496 260.41320333045496 262.96993865574484
262.96993865574484 262.96993865574484 262.96993865574484 262.96993865574484 274.90393865574487 274.90393865574487 274.90393865574487 279.70357875429283
279.7035787542928 279.7035787542928 279.7035787542928 279.7035787542928 291.6375787542928 291.6375787542928 291.6375787542928 278.94487999999996

The question is what is right?

src/note.ts
rvilarl: I copied the formula from:

https://github.com/0xfe/vexflow/blob/adda50f0d90212d5b8238f772d62b14a40288a17/src/tickcontext.ts#L245-L248

As you see above, they could be greater values

0xfe:
Okay, if `Note` derives from `Tickable` and `this.totalLeft/RightPx` is protected, what happens if we assign `this.totalLeft/RightPx` to it?

0xfe:
> Okay, if `Note` derives from `Tickable` and `this.totalLeft/RightPx` is protected, what happens if we assign `this.totalLeft/RightPx` to it?

Actually, ignore that, `this.totalLeft/RightPx` is only set in `TickContext`, so not really needed here.

src/note.ts
0xfe: This is `TickContext` which has different metrics from `Note`. Those operations are not run here.

0xfe: Okay, for the formatter, all that matters are the variables that are defined here -- we don't need `totalLeft/RightPx` here. We should remove them from here derive them in `evaluate` (they're only used for the debugging plots, AFACT.)

rvilarl:
@0xfe is there anything to do in the end with the yellow lines or we leave it as it is?

0xfe:
@rvilarl Yes, please remove `total*` from `Note` and fix the calculations in `evaluate` to derive the yellow lines. So remove the references in `evaluate` to `note.totalLeft/RightPx` (not `tickcontext.totalLeft/RightPx`.) The yellow lines are correct here, because they're no longer derived from `undefined`.

0xfe:
To be clear, this is a bug in the original code (using `undefined` for calculations.)

So I'm suggesting a PR just to fix that code by removing the references to `totalLeftRightPx` (members of `Note.metrics` instances, not `TickContext.metrics`) inside `evaluate`, and deriving `freedom.right` (the yellow line) from the existing members (e.g., `modRightPx`.)

rvilarl:
0xfe I have tried to follow your suggestion. Please check if I understood correctly.

0xfe:
> 0xfe I have tried to follow your suggestion. Please check if I understood correctly.

Yes, this is exactly what I meant. Can we put that fix in a separate PR?

This PR is very complicated, maybe we should break it up.

Also, we need to think about how to rationalize the metrics between `Note` and `TickContext` -- maybe have a single structure in `common`?

rvilarl:
@0xfe how would you do that now? I mean the migration to typescript showed a bug and, I belive, we resolved it. On the Metrics side we are trying to remove the optional attributes. The total attributes are only now in TickContextMetrics extending the NoteMetrics.

What would you propose? I can break the calculation of the yellow line to get exactly the same image output. Should I do that?

rvilarl:
@0xfe calculation broken:

https://github.com/rvilarl/vexflow/blob/db43e2cb777cf91e1d34bb4417991eb9eaac6d85/src/note.ts#L130

Now all the images are identical to master. This should give confidence to merge. :)

0xfe:
Trying to figure out all the issues here -- don't try to get the same image output for the yellow lines, the long lines are correct.

But, what about the other note/rest formatting issues, were they resolved (by fixing the exponents?)

rvilarl:
Yes everything else is fixed. The only differences are the nine images that I added in this PR. I have removed the last commit breaking the yellow line calculation.

src/formatter.ts
```diff
@@ -23,17 +23,69 @@ import { Flow } from './tables';
...
+import { Stave } from './stave';
+import { StemmableNote } from './stemmablenote';
+import { Tickable } from './tickable';
+import { TabStave } from './tabstave';
+import { TabNote } from './tabnote';
+import { BoundingBox } from './boundingbox';
+import { StaveNote } from './stavenote';
+import { check } from './common';
+
+interface ContextGaps {
```
0xfe, line 39: this can be inline

src/formatter.ts
```diff
@@ -23,17 +23,69 @@ import { Flow } from './tables';
...
+  total: number;
+  gaps: { x1: number; x2: number }[];
+}
+
+export interface Distance {
```
0xfe, line 44: Does this need to be exported?

src/formatter.ts
```diff
@@ -23,17 +23,69 @@ import { Flow } from './tables';
...
+}
+
+export interface TickContextsStruct {
+  list: number[];
+  map: Record<number, TickContext>;
+  array: TickContext[];
+  resolutionMultiplier: number;
+}
+
+export interface ModifierContextsStruct {
```
0xfe, line 73: You can collapse the two:

```typescript
export interface FormatterContexts<T> {
  list: number[];
  map: Record<number, T>;
  array: T[];
  resolutionMultiplier: number;
}
```

src/formatter.ts
```diff
@@ -404,48 +417,137 @@ export class Formatter {
...
+    }, 1);
+    return resolutionMultiplier;
+  }
+
   // Create `ModifierContext`s for each tick in `voices`.
-  createModifierContexts(voices) {
-    const contexts = createContexts(voices, ModifierContext, (tickable, context) =>
-      tickable.addToModifierContext(context)
-    );
+  createModifierContexts(voices: Voice[]): ModifierContextsStruct {
```
0xfe, line 441: Can you share the logic with a template method?

```typescript
function createContexts<T>(voices: Voice[]) : FormatterContexts<T> {
   ...
}
```

rvilarl: I will give it a try.

rvilarl: The compiler complains in three places:
 `const newContext = new T({ tickID: integerTicks });`it seems that new is not allowed with T. There is also a different number of parameters on both constructors.
`tickToContextMap[integerTicks].addTickable(tickable, voiceIndex);` T does not have this method
```
contexts.forEach((context) => {
      context.tContexts = contexts;
    });
```

0xfe: Right, seems like TS doesn't like that, and you'll need to pass a builder parameter. I was able to get this prototype to work on the TypeScript playground (see the code after HERE):

```typescript

// Mock class for playground
class ModifierContext {}

// Mock class for playground
class TickContext {
    constructor(tick?: {tickID: number}) {}
    addTickable(n: number): void {}
}

// Mock class for playground
class Tickable {
    addToModifierContext(c: ModifierContext): void {}
}

// New formatter prototype begins HERE:

interface AlignmentContexts<T> {
    map: Record<number, T>,
    array: T[],
    list: T[]
    resolutionMultiplier: number
}

type addToContextFn<T> = (tickable: Tickable, context: T, voiceIndex?: number) => void
type makeContextFn<T> = (tick?: {tickID: number}) => T

function createContexts<T>(voices: string[], makeContext: makeContextFn<T>, addToContext: addToContextFn<T>): AlignmentContexts<T> {
    const c = makeContext();
    return {
        map: {},
        array: [],
        list: [],
        resolutionMultiplier: 1,
    };
}

function createModifierContexts(voices: string[]): AlignmentContexts<ModifierContext> {
    const fn: addToContextFn<ModifierContext> = (tickable: Tickable, context: ModifierContext, voiceIndex?: number) => tickable.addToModifierContext(context);
    return createContexts(voices, () => new ModifierContext(), fn);
}

function createTickContexts(voices: string[]): AlignmentContexts<TickContext> {
    const fn: addToContextFn<TickContext> = (tickable: Tickable, context: TickContext, voiceIndex?: number) => (context as TickContext).addTickable(voiceIndex as number);
    return createContexts(voices, (tick?: {tickID: number}) => new TickContext(tick), fn);
}
```

rvilarl: done

src/formatter.ts
rvilarl: Done

src/formatter.ts
rvilarl: done

rvilarl:
Ready to merge!

rvilarl:
@0xfe this one is complete, it would be good to merge after all the work we invested :)

0xfe:
LGTM. You're right, this was a lot of work. Thanks for tackling this one -- I think it's all downhill from here! :-)

PR #982, opened 2021-05-13
Migration/annotation

rvilarl:
Annotation!


src/textfont.ts
```diff
@@ -197,7 +198,7 @@ export class TextFont {
       selectedFont = new TextFont(candidates[0]);
     } else {
       const bold = TextFont.fontWeightToBold(fd.weight);
-      const italic = TextFont.fontStyleToItalic(fd.style);
+      const italic = TextFont.fontStyleToItalic(fd.weight);
```
rvilarl, line 201: @0xfe is that right? is `style` the same as `weight` or the same as `family`?

I do not remember :)

AaronDavidNewman:
italic is a style, bold is a weight. Family is like 'Arial', 'Times', etc.

rvilarl:
> italic is a style, bold is a weight. Family is like 'Arial', 'Times', etc.

Hi @AaronDavidNewman thanks, I would think so, but I remember that FontInfo used to have more members and @0xfe asked me to reduce it due to duplication. I will try to find the quote

src/textfont.ts
rvilarl: fixed, thank @AaronDavidNewman

AaronDavidNewman:
I see, font face and name doesn't really mean anything here - fonts are always specified using 'font-family'.  

In vexflow, (and maybe this is a smufl convention)  'size' is always 'size in points', a point being 1/72 of an inch, so they are the same, and size for TextFont is also assumed to be 'size in points', to be consistent. 

In fontInfo and CSS, font weight can be specified as a number (e.g. 700) or a string (bold, bolder).  Style is always a string.  

In TextFont, bold and italic are both boolean because they are used to a search/map for a set of metrics.  Italic and normal might have different metrics, for instance. (and now that I am explaining it, I realize it is a bit confusing).

rvilarl:
@0xfe any comment here?

0xfe:
Okay, we need to rationalize this across the codebase, and fortunately there isn't a lot of instances of this. So how about we have a single structure for font info, like this:

```typescript
interface FontInfo {
  family: string;
  style: string;
  weight: string;
  pointSize: number;
}
```

And a compatibility functions to support the older formats, sort of like:

```typescript
interface DeprecatedFontInfo {
  size?: number;
  face?: string;
  // ... other possible fields as optional
}

function BuildFontInfo(fontStruct: any) : FontInfo {
  let fontInfo : FontInfo;

  if (fontStruct.size !== undefined) {
    fontInfo.pointSize = fontStruct.size;
  }

  // ... other similar code ...

  return fontInfo;
}
```

Does this  make sense? If so, @AaronDavidNewman, is the proposed interface ideal (or should there be anything else?)

Alternatively, we can encapsulate this logic into `TextFont` or something, and make it a primitive for other classes?

rvilarl:
@0xfe I think we should start an issue with the information above and resolve it in another PR, don't you agree?

rvilarl:
I would not change the attibute names they are now in line with MusicXML

http://usermanuals.musicxml.com/MusicXML/MusicXML.htm#AT-MusicXML-font-family.htm%3FTocPath%3DMusicXML%2520Reference%7CScore%2520Schema%2520(XSD)%7CAttributes%7Ccommon%7C_____21

AaronDavidNewman:
Hi all, sorry been kind of a busy week, I hope I'm not holding you back.  Regarding this:

>> 0xfe  So how about we have a single structure for font info, like this:

I think this is good.  I don't know that we need to specify `pointSize` though, I like having `size`, because then we have all the major 'font-' css attributes.  vexflow seems pretty consistent in referring to font sizes in points, so people have been using that all along.  And I think we should force the user to specify font size as a number b/c font sizes in css cascade (like `em`) and that doesn't make sense in our context.  MusicXML allows the css strings, but (like many things in that spec) it's not clear what you do with that information, so I don't think we should follow that.

For the attributes in textfont registry that have the same names as fontInfo (e.g. bold), I will change them to `isBold` etc, to make it clear this is a read-only property and not a definition.

`function BuildFontInfo(fontStruct: any) `, sounds good, with reasonable defaults for everything.

0xfe:
> @0xfe I think we should start an issue with the information above and resolve it in another PR, don't you agree?

Yes, agreed.

src/annotation.ts
```diff
@@ -11,28 +11,50 @@ import { Vex } from './vex';
...
 
   // Text annotations can be positioned and justified relative to the note.
-  static get Justify() {
-    return {
-      LEFT: 1,
-      CENTER: 2,
-      RIGHT: 3,
-      CENTER_STEM: 4,
-    };
+  static get Justify(): typeof Justify {
```
0xfe, line 53: I think you can just do this.

```typescript
static Justify = Justify;
```

src/annotation.ts
```diff
@@ -11,28 +11,50 @@ import { Vex } from './vex';
...
 
 // To enable logging for this class. Set `Vex.Flow.Annotation.DEBUG` to `true`.
-function L(...args) {
+function L(
+  // eslint-disable-next-line
+  ...args: any []) {
   if (Annotation.DEBUG) Vex.L('Vex.Flow.Annotation', args);
 }
 
+export enum Justify {
```
0xfe, line 25: No need to export.

src/annotation.ts
```diff
@@ -11,28 +11,50 @@ import { Vex } from './vex';
...
+  LEFT = 1,
+  CENTER = 2,
+  RIGHT = 3,
+  CENTER_STEM = 4,
+}
+
+export enum VerticalJustify {
```
0xfe, line 32: No need to export this either.

src/annotation.ts
```diff
@@ -11,28 +11,50 @@ import { Vex } from './vex';
...
+  TOP = 1,
+  CENTER = 2,
+  BOTTOM = 3,
+  CENTER_STEM = 4,
+}
```
0xfe, line 37: New line.

src/annotation.ts
```diff
@@ -11,28 +11,50 @@ import { Vex } from './vex';
...
+  static DEBUG: boolean;
+
+  note?: StemmableNote;
+
+  protected justification: Justify;
+  protected vert_justification: VerticalJustify;
+  protected text: string;
+  protected font: FontInfo;
+
+  static get CATEGORY(): string {
```
0xfe, line 48: We should fix these everywhere, since TS supports static properties: `static CATEGORY = 'annotations'`

src/annotation.ts
```diff
@@ -116,36 +131,36 @@ export class Annotation extends Modifier {
...
-  setFont(family, size, weight) {
+  setFont(family: string, size: number, weight: string): this {
     this.font = { family, size, weight };
     return this;
   }
 
   // Set vertical position of text (above or below stave). `just` must be
   // a value in `Annotation.VerticalJustify`.
-  setVerticalJustification(just) {
+  setVerticalJustification(just: string | number): this {
```
mscuthbert, line 146: Should this be typed as `just: string | Justify` ?

src/annotation.ts
```diff
@@ -116,36 +131,36 @@ export class Annotation extends Modifier {
...
   }
 
   // Get and set horizontal justification. `justification` is a value in
   // `Annotation.Justify`.
-  getJustification() {
+  getJustification(): Justify {
     return this.justification;
   }
-  setJustification(just) {
+  setJustification(just: string | number): this {
```
mscuthbert, line 156: as above `string | Justify`

mscuthbert (commented):
Two comments on typing for setXXXJustify

src/annotation.ts
rvilarl: done

src/annotation.ts
rvilarl: done

src/annotation.ts
rvilarl: This would have to be done in another PR changing all of them. If I only change here I get:

```
Uncaught TypeError: Cannot set property CATEGORY of function Modifier() {
        var _this = _super.call(this) || this;
        _this.setAttribute('type', 'Modifi...<omitted>... } which has only a getter
```

src/annotation.ts
rvilarl: done

src/annotation.ts
rvilarl: done

src/annotation.ts
rvilarl: done

src/annotation.ts
rvilarl: done

rvilarl:
Ready to merge

0xfe:
Looks good.

PR #984, opened 2021-05-14
Migration/articulation

rvilarl:
Articulation!


src/articulation.ts
```diff
@@ -14,20 +14,40 @@ import { Flow } from './tables';
 import { Modifier } from './modifier';
 import { Glyph } from './glyph';
 import { Stem } from './stem';
+import { Note } from './note';
+import { StaveNote } from './stavenote';
+import { ModifierContextState } from './modifiercontext';
+import { check } from './common';
+import { Builder } from './easyscore';
+import { TabNote } from './tabnote';
+import { Stave } from './stave';
```
0xfe, line 23: Lint error: you can remove this import.

src/stavenote.ts
```diff
@@ -779,7 +779,7 @@ export class StaveNote extends StemmableNote {
   getModifierStartXY(
     position: number,
     index: number,
-    options: {
+    options?: {
```
0xfe, line 782: We can make this non-optional with `options: { ... } = {}`, and remove the `options = options || {}` in the body.

src/articulation.ts
rvilarl: done

src/stavenote.ts
rvilarl: done

rvilarl:
Ready to merge!

rvilarl:
@0xfe any comment beyond the one cleared with @AaronDavidNewman ?

rvilarl:
@0xfe anything pending here?

src/articulation.ts
```diff
@@ -143,13 +160,24 @@ const getInitialOffset = (note, position) => {
...
 export class Articulation extends Modifier {
-  static get CATEGORY() {
+  note?: Note;
+  readonly type: string;
+
+  protected render_options: { font_scale: number };
+  protected articulation?: ArticulationStruct;
+  // glyph defined calling reset in constructor
+  // eslint-disable-next-line
+  protected glyph!: Glyph;
```
0xfe, line 173: Why `!` here?

rvilarl: Because I can tell the compiler that it is surely defined in the construtor. I tried to explain it in the comment above. The `constructor` calls `reset` which defines `this.glyph`

rvilarl: @0xfe would you like a different approach? `!` is thought to tell the compiler: "forget it I know better"

0xfe: Ah, okay I see.

src/articulation.ts
```diff
@@ -143,13 +160,24 @@ const getInitialOffset = (note, position) => {
...
+  static DEBUG: boolean;
+
+  static get CATEGORY(): string {
     return 'articulations';
   }
-  static get INITIAL_OFFSET() {
+
+  static get INITIAL_OFFSET(): number {
```
0xfe, line 180: Can you make this a static property instead?

0xfe (commented):
Minor comments.

src/articulation.ts
rvilarl: done

rvilarl:
Ready to merge

0xfe:
looks great!

PR #993, opened 2021-05-23
Migration/stavehairpin

rvilarl:
StaveHairpin!


rvilarl:
@0xfe wow a lot of progress! Did you see this little one?

src/stavehairpin.ts
```diff
@@ -9,14 +9,39 @@
 import { Vex } from './vex';
 import { Element } from './element';
 import { Modifier } from './modifier';
+import { Note } from './note';
+import { RenderContext } from './types/common';
+
+export interface StaveHairpinRenderParams {
```
0xfe, line 15: This can be inline.

src/stavehairpin.ts
```diff
@@ -165,23 +203,23 @@ export class StaveHairpin extends Element {
...
     const start = firstNote.getModifierStartXY(this.position, 0);
     const end = lastNote.getModifierStartXY(this.position, 0);
 
     this.renderHairpin({
       first_x: start.x,
       last_x: end.x,
-      first_y: firstNote.getStave().y + firstNote.getStave().height,
-      last_y: lastNote.getStave().y + lastNote.getStave().height,
-      staff_height: firstNote.getStave().height,
+      first_y: firstNote.checkStave().getY() + (firstNote.checkStave().getHeight() ?? 0),
```
0xfe, line 220: This `?? 0` here is a silent failure. It seems like `stave.height` should not be optional, all staves must have a height. Can we make it (and the result of `getHeight`) non-optional?

src/stavehairpin.ts
rvilarl: done

src/stavehairpin.ts
rvilarl: yes it is always initialised in constructor, done

rvilarl:
Ready to merge

0xfe:
LGTM!

PR #995, opened 2021-05-23
Allow modifiers like ChordSymbol to be attached to GlyphNotes.

AaronDavidNewman:
This change allows modifiers to be attached to GlyphNotes.  This is used to notate chord changes in an improvised solo.

image

### DIFF
--- a/src/glyphnote.ts
+++ b/src/glyphnote.ts
@@ -14,9 +14,7 @@ export interface GlyphNoteOptions {
 
 export class GlyphNote extends Note {
   protected options: GlyphNoteOptions;
-  static format(notes: GlyphNote[], state: ModifierContextState): boolean {
-    return false;
-  }
+
   constructor(glyph: Glyph | undefined, noteStruct: NoteStruct, options?: GlyphNoteOptions) {
     super(noteStruct);
     this.options = {
@@ -50,7 +48,6 @@ export class GlyphNote extends Note {
     for (let i = 0; i < this.modifiers.length; ++i) {
       this.modifierContext.addMember(this.modifiers[i]);
     }
-    this.modifierContext.addMember(this);
     this.setPreFormatted(false);
     return this;
   }
@@ -62,7 +59,6 @@ export class GlyphNote extends Note {
     this.setPreFormatted(true);
     return this;
   }
-  // Draw all key modifiers
   drawModifiers(): void {
     const ctx = this.checkContext();
     ctx.openGroup('modifiers');
--- a/src/glyphnote.ts
+++ b/src/glyphnote.ts
@@ -1,6 +1,5 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 
-import { Vex } from './vex';
 import { BoundingBox } from './boundingbox';
 import { Glyph } from './glyph';
 import { Note, NoteStruct } from './note';
@@ -58,6 +57,7 @@ export class GlyphNote extends Note {
     this.setPreFormatted(true);
     return this;
   }
+
   drawModifiers(): void {
     const ctx = this.checkContext();
     ctx.openGroup('modifiers');
--- a/src/glyphnote.ts
+++ b/src/glyphnote.ts
@@ -1,8 +1,12 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 
+import { Vex } from './vex';
 import { BoundingBox } from './boundingbox';
 import { Glyph } from './glyph';
 import { Note, NoteStruct } from './note';
+import { ModifierContext } from './modifiercontext';
+import { ModifierContextState } from './modifiercontext';
+
 export interface GlyphNoteOptions {
   ignoreTicks?: boolean;
   line?: number;
@@ -10,6 +14,9 @@ export interface GlyphNoteOptions {
 
 export class GlyphNote extends Note {
   protected options: GlyphNoteOptions;
+  static format(notes: GlyphNote[], state: ModifierContextState): boolean {
+    return false;
+  }
   constructor(glyph: Glyph | undefined, noteStruct: NoteStruct, options?: GlyphNoteOptions) {
     super(noteStruct);
     this.options = {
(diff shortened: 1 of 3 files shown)


AaronDavidNewman:
Here is the test case I added:

image

src/glyphnote.ts
```diff
@@ -1,15 +1,22 @@
...
+import { ModifierContextState } from './modifiercontext';
+
 export interface GlyphNoteOptions {
   ignoreTicks?: boolean;
   line?: number;
 }
 
 export class GlyphNote extends Note {
   protected options: GlyphNoteOptions;
+  static format(notes: GlyphNote[], state: ModifierContextState): boolean {
```
AaronDavidNewman, line 17: Do I need to do this?  I am following the pattern of StaveNote, and I needed to do this to get the format() of the modifiers called.  But maybe there is a way I can add the modifiers and not the GlyphNote itself.

AaronDavidNewman: Indeed I do not, I can just format the modifiers and not the thing itself.  Removed the empty call.

src/glyphnote.ts
```diff
@@ -1,15 +1,20 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 
+import { Vex } from './vex';
 import { BoundingBox } from './boundingbox';
 import { Glyph } from './glyph';
 import { Note, NoteStruct } from './note';
+import { ModifierContext } from './modifiercontext';
+import { ModifierContextState } from './modifiercontext';
```
rvilarl, line 8: This line can be deleted, `ModifierContextState` is not used.

src/glyphnote.ts
```diff
@@ -36,31 +41,54 @@ export class GlyphNote extends Note {
...
+    for (let i = 0; i < this.modifiers.length; ++i) {
+      this.modifierContext.addMember(this.modifiers[i]);
+    }
+    this.setPreFormatted(false);
     return this;
   }
-  */
 
   preFormat(): this {
+    if (!this.preFormatted && this.modifierContext) {
```
rvilarl, line 56: Should an error be thrown if `this.modifierContext` is `undefined`?

src/glyphnote.ts
```diff
@@ -36,31 +41,54 @@ export class GlyphNote extends Note {
...
+    ctx.closeGroup();
+  }
   draw(): void {
-    const stave = this.checkStave();
-    const ctx = stave.checkContext();
+    if (!this.stave) {
+      throw new Vex.RERR('NoStave', 'No stave attached to this note.');
+    }
+
+    const ctx = this.stave.checkContext();
```
rvilarl, line 77: `this.checkStave()` works with `this.stave` like `this.checkContext()` with `this.context`:
- `this.checkStave()` will throw an `RERR` if `this.stave` is `undefined`. 

Please revert the change.

src/modifiercontext.ts
```diff
@@ -7,6 +7,7 @@
 
 import { Vex } from './vex';
 import { StaveNote } from './stavenote';
+import { GlyphNote } from './glyphnote';
```
rvilarl, line 10: This line can be deleted. `GlyphNote` is not used.

rvilarl (changes requested):

src/glyphnote.ts
AaronDavidNewman: I don't know.  I followed the lead from stavenote.ts, it allows modifierContext to be null in preFormat.

src/glyphnote.ts
AaronDavidNewman: done.  Didn't it used to throw an error when you did that?

src/modifiercontext.ts
AaronDavidNewman: done

src/glyphnote.ts
AaronDavidNewman: Sorry, that must have snuck in from my other branch.  I reverted to what it was.

src/glyphnote.ts
```diff
@@ -1,15 +1,19 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 
+import { Vex } from './vex';
```
0xfe, line 3: This line can be removed. (unused Vex).

src/glyphnote.ts
```diff
@@ -36,21 +40,39 @@ export class GlyphNote extends Note {
     return this.glyph.getBoundingBox();
   }
 
-  /*
-  addToModifierContext() {
+  // Add self to modifier context. `mContext` is the `ModifierContext`
+  // to be added to.
+  addToModifierContext(mContext: ModifierContext): this {
```
0xfe, line 45: Is this necessary? Isn't it derived from the inheritance chain?

src/glyphnote.ts
0xfe: Right, this is okay.

src/glyphnote.ts
```diff
@@ -36,21 +40,39 @@ export class GlyphNote extends Note {
...
+  drawModifiers(): void {
+    const ctx = this.checkContext();
+    ctx.openGroup('modifiers');
+    for (let i = 0; i < this.modifiers.length; i++) {
+      const modifier = this.modifiers[i];
+      modifier.setContext(ctx);
+      modifier.drawWithStyle();
+    }
+    ctx.closeGroup();
+  }
```
0xfe, line 70: New line between method bodies please.

0xfe (commented):
Looks good. Minor comments.

src/glyphnote.ts
AaronDavidNewman: The inherited one (`tickable`) doesn't add modifiers to the context. 

We could add this loop code to `Note`, then `stavenote` and `glyphnote` could share, but we'll have to check all the other `Note` children.  We could modify the `tickable` code to check for modifiers, maybe.

AaronDavidNewman: Welp, not exactly the same.  Glyphnote doesn't need to `format`, so it doesn't add itself to the context.  So stavenote looks like this, not sure how you feel about `super`.

``` javascript
    this.modifierContext.addMember(this);
    return super.addToModifierContext(mContext);
```

AaronDavidNewman: welp, this doesn't work, I'd have to add a check for null to `this.modifierContext` which seems to defeat the purpose.  Or add an empty format method for glyphnote.  

So I think I'm back to:  it is necessary :)

```
ERROR in C:\Users\Daddy\Documents\GitHub\vexflow\src\stavenote.ts
./src/stavenote.ts 871:4-24
```

src/glyphnote.ts
AaronDavidNewman: ok

src/glyphnote.ts
rvilarl: The line is still there

src/glyphnote.ts
rvilarl: This is not reverted, did you push the change?

src/modifiercontext.ts
rvilarl: The line is still there also

src/glyphnote.ts
0xfe: kk, thanks for digging in :-)

0xfe (approved):
Looks great.

PR #997, opened 2021-05-24
Migration/pedalmarking

rvilarl:
PedalMarking

@0xfe here I had to resolve a bug mixing pedal type and element style.


src/pedalmarking.ts
```diff
@@ -9,86 +9,116 @@
 // custom text for the release/depress pedal markings.
 
 import { Vex } from './vex';
-import { Element } from './element';
+import { Element, ElementStyle } from './element';
 import { Glyph } from './glyph';
+import { FontInfo, RenderContext } from './types/common';
+import { StaveNote } from './stavenote';
+
+export interface PedalMarkingRenderOptions {
```
0xfe, line 17: This can be inline.

src/pedalmarking.ts
```diff
@@ -9,86 +9,116 @@
...
+  color: string;
+  bracket_height: number;
+  text_margin_right: number;
+  bracket_line_width: number;
+}
+
+export interface PedalMarkingGlyph {
```
0xfe, line 24: This can be inline too.

0xfe (commented):
Some minor comments.

src/pedalmarking.ts
rvilarl: done

rvilarl:
Ready to merge!

0xfe:
Looks good, and thanks for fixing the bug.

PR #999, opened 2021-05-25
Migrate EasyScore

ronyeh:
Ready for review.

I had to use `// eslint-disable-next-line` more often than I preferred.

Some of these can be cleaned up after migrating Factory, assuming we decide to add stricter typing to the `params` objects used by the Factory methods.


src/easyscore.ts
```diff
@@ -0,0 +1,492 @@
...
+  constructor(options: EasyScoreOptions = {}) {
+    this.setOptions(options);
+    this.defaults = {
+      clef: 'treble',
+      time: '4/4',
+      stem: 'auto',
+    };
+  }
+
+  set(defaults: EasyScoreDefaults): this {
```
0xfe, line 426: If `set` is public then `EasyScoreDefaults` should be exported no?

0xfe (commented):
Thanks Ron. Mostly looks good. Some comments.

src/easyscore.ts
ronyeh: Yup. Will export the EasyScoreDefaults interface.

Question!

The next line is `Object.assign(this.defaults, defaults);`

I notice elsewhere in the code base you use Vex.Merge(a, b);

They are very very similar..... the only difference would be that Vex.Merge also copies over properties inherited via the JS prototype chain (e.g., through Object.create()).

We can discuss this later on via a GitHub issue or something, but do you think Vex.Merge can be completely replaced by Object.assign everywhere?

Or are there instances where we depend on copying over properties inherited via prototype chain?? 

I believe that Vex.Merge and Object.assign are equivalent even with properties inherited via es6 `extends` since all properties inherited through `extends` are "own properties" in the instantiated object.

ronyeh:
`npm run test:reference` passes with 0 diffs.

src/easyscore.ts
0xfe: `Vex.Merge` is super old. Today I'd rather use the spread operator, something like: `this.defaults = {...this.defaults, ...defaults}`

0xfe: To answer your question specifically, we don't need the properties from the prototype chain. `Object.assign` is almost always better. (However, I feel the spread operator is more readable.)

0xfe:
Looks great, thanks!

PR #1001, opened 2021-05-25
Migrate NoteSubGroup

ronyeh:
Ready for review.


ronyeh:
While migrating this class, I noticed a bug. 

Most of the `Modifier` subclasses set `this.index` to null, even though it is typed to be `number` in the `Modifier` class.

Additionally, classes like:
notesubgroup.ts, bend.ts, gracenotegroup.ts, stringnumber.ts, strokes.js, tremolo.js

sometimes check if `this.index != null` or `this.index !== null` or `this.index == null`.... or else it throws an exception.

Currently, since Modifier sets this.index to 0 in the constructor, this means that these checks probably don't do what we want them to do.

So I'm not an expert in the code base. What does index set to 0 mean? Does that mean it is unattached? Are notes 1-indexed? Or should we have the default index be set to -1, and then change all the null checks to check against -1 instead? 

Ideally, we can have a method in the Modifier class report if it is not attached. That way we can consolidate all of these various checks into one single point.

Anyways.... Let me know what you guys think. I can make a github issue for this task, and perhaps I can look into it.

src/notesubgroup.ts
```diff
@@ -12,14 +12,18 @@ import { Flow } from './tables';
 import { Modifier } from './modifier';
 import { Formatter } from './formatter';
 import { Voice } from './voice';
+import { ModifierContextState } from './modifiercontext';
+import { Note } from './note';
+import { Tickable } from './tickable';
```
rvilarl, line 17: You can delete this. `Tickable` is not used.

src/notesubgroup.ts
```diff
@@ -55,45 +63,48 @@ export class NoteSubGroup extends Modifier {
...
   }
-  getWidth() {
+
+  getWidth(): number {
     return this.width;
   }
 
-  draw() {
+  draw(): void {
     this.checkContext();
```
rvilarl, line 95: Get the `ctx` here, and you can delete the new line 107 below:
`const ctx = this.checkContext()`

rvilarl (changes requested):

src/notesubgroup.ts
ronyeh: thanks! wasn't aware that checkContext() returned the context, haha.

src/notesubgroup.ts
ronyeh: thanks! vscode must have inserted it when I was trying to get at one of the protected variables from Note / Tickable.

src/notesubgroup.ts
```diff
@@ -33,16 +36,20 @@ export class NoteSubGroup extends Modifier {
...
+  protected formatter: Formatter;
+  protected voice: Voice;
+
+  constructor(subNotes: Note[]) {
     super();
     this.setAttribute('type', 'NoteSubGroup');
 
-    this.note = null;
-    this.index = null;
+    this.index = -1;
```
0xfe, line 48: I think you can remove `this.index` here, since it's not really used.

src/notesubgroup.ts
```diff
@@ -55,45 +62,44 @@ export class NoteSubGroup extends Modifier {
...
 
-  draw() {
-    this.checkContext();
-
+  draw(): void {
+    const ctx: RenderContext = this.checkContext();
     const note = this.getNote();
 
-    if (!(note && this.index !== null)) {
+    if (!(note && this.index !== -1)) {
```
0xfe, line 97: This can just be `if (!note)` (and fix the error message too.)

0xfe (commented):
Looks good overall, some comments.

ronyeh:
Done. `npm run test:reference` passes w/ 0 diffs from `master`.

0xfe:
Looks great!

Issue #1000, opened 2021-05-25
Should 'state' param use ModifierContextState type?


ronyeh:
https://github.com/0xfe/vexflow/blob/3e13b92cc7437d5c6ac2b23fa2d056bb1f0f1564/src/gracenotegroup.ts#L46

@rvilarl I noticed that all other format(...) functions have a `ModifierContextState` as the second param. I assume this is one of the earlier migrations. Should we now change the type of the second param to be `ModifierContextState`?

0xfe:
Yes, correct -- it should be `ModifierContextState`

PR #1002, opened 2021-05-25
Introduce Modifier.checkAttachedNote() to check that this.note and this.index are valid.

ronyeh:
Modifier declares `index` as a number. However, subclasses check `this.index` against `null` or `undefined`. Below are some of the various ways we used to validate `this.note` and `this.index`. I consolidated these checks into a single method: `Modifier.checkAttachedNote()`. ~~Now, an invalid index is -1.~~  **Edit: invalid indexes are now `undefined`.**

```

accidental.js
    if (!(note && index != null)) {
      throw new Vex.RERR('NoAttachedNote', "Can't draw accidental without a note and index.");
    }
annotation.ts
    if (!this.note) {
      throw new Vex.RERR('NoNoteForAnnotation', "Can't draw text annotation without an attached note.");
    }
articulation.ts
    if (!note || index == null) {
      throw new Vex.RERR('NoAttachedNote', "Can't draw Articulation without a note and index.");
    }
bend.ts
    if (!(this.note && this.index != null)) {
      throw new Vex.RERR('NoNoteForBend', "Can't draw bend without a note or index.");
    }
chordsymbol.js
    if (!this.note) {
      throw new Vex.RERR('NoNoteForAnnotation', "Can't draw text annotation without an attached note.");
    }
dot.ts
    if (!this.note || this.index === undefined) {
      throw new Vex.RERR('NoNoteIndex', 'Drawing a dot requires a note and an index.');
    }
frethandfinger.ts
    if (!this.note || this.index == null) {
      throw new Vex.RERR('NoAttachedNote', "Can't draw string number without a note and index.");
    }
gracenotegroup.ts
    if (!(note && this.index !== null)) {
      throw new Vex.RuntimeError(
        'NoAttachedNote',
        "Can't draw grace note without a parent note and parent note index."
      );
    }
notesubgroup.js
    if (!(note && this.index !== null)) {
      throw new Vex.RuntimeError('NoAttachedNote', "Can't draw notes without a parent note and parent note index.");
    }
ornament.js
    if (!this.note || this.index == null) {
      throw new Vex.RERR('NoAttachedNote', "Can't draw Ornament without a note and index.");
    }
stringnumber.js
    if (!(this.note && this.index != null)) {
      throw new Vex.RERR('NoAttachedNote', "Can't draw string number without a note and index.");
    }
strokes.js
    if !(this.note && this.index != null) {
      throw new Vex.RERR('NoAttachedNote', "Can't draw stroke without a note and index.");
    }
tremolo.ts
    if (!(this.note && this.index != null)) {
      throw new Vex.RERR('NoAttachedNote', "Can't draw Tremolo without a note and index.");
    }
vibrato.ts
    if (!this.note) {
      throw new Vex.RERR('NoNoteForVibrato', "Can't draw vibrato without an attached note.");
    }

```

### DIFF
--- a/src/annotation.ts
+++ b/src/annotation.ts
@@ -7,7 +7,7 @@
 //
 // See `tests/annotation_tests.js` for usage examples.
 
-import { Vex } from './vex';
+import { RuntimeError, log } from './util';
 import { Flow } from './tables';
 import { Modifier } from './modifier';
 import { TextFont } from './textfont';
@@ -18,9 +18,8 @@ import { ModifierContextState } from './modifiercontext';
 // To enable logging for this class. Set `Vex.Flow.Annotation.DEBUG` to `true`.
 function L(
   // eslint-disable-next-line
-  ...args: any[]
-) {
-  if (Annotation.DEBUG) Vex.L('Vex.Flow.Annotation', args);
+  ...args: any []) {
+  if (Annotation.DEBUG) log('Vex.Flow.Annotation', args);
 }
 
 enum Justify {
--- a/src/annotation.ts
+++ b/src/annotation.ts
@@ -8,6 +8,7 @@
 // See `tests/annotation_tests.js` for usage examples.
 
 import { Vex } from './vex';
+import { RuntimeError } from './util';
 import { Flow } from './tables';
 import { Modifier } from './modifier';
 import { TextFont } from './textfont';
@@ -155,7 +156,7 @@ export class Annotation extends Modifier {
     const ctx = this.checkContext();
 
     if (!this.note) {
-      throw new Vex.RERR('NoNoteForAnnotation', "Can't draw text annotation without an attached note.");
+      throw new RuntimeError('NoNoteForAnnotation', "Can't draw text annotation without an attached note.");
     }
 
     this.setRendered();
--- a/src/annotation.ts
+++ b/src/annotation.ts
@@ -8,7 +8,6 @@
 // See `tests/annotation_tests.js` for usage examples.
 
 import { Vex } from './vex';
-import { RuntimeError } from './util';
 import { Flow } from './tables';
 import { Modifier } from './modifier';
 import { TextFont } from './textfont';
@@ -19,7 +18,8 @@ import { ModifierContextState } from './modifiercontext';
 // To enable logging for this class. Set `Vex.Flow.Annotation.DEBUG` to `true`.
 function L(
   // eslint-disable-next-line
-  ...args: any []) {
+  ...args: any[]
+) {
   if (Annotation.DEBUG) Vex.L('Vex.Flow.Annotation', args);
 }
 
@@ -154,13 +154,10 @@ export class Annotation extends Modifier {
   // Render text beside the note.
   draw(): void {
     const ctx = this.checkContext();
-
-    if (!this.note) {
-      throw new RuntimeError('NoNoteForAnnotation', "Can't draw text annotation without an attached note.");
-    }
-
+    const note = this.checkAttachedNote();
     this.setRendered();
-    const start = this.note.getModifierStartXY(Modifier.Position.ABOVE, this.index);
+
+    const start = note.getModifierStartXY(Modifier.Position.ABOVE, this.index);
 
     // We're changing context parameters. Save current state.
     ctx.save();
@@ -184,18 +181,18 @@ export class Annotation extends Modifier {
     } else if (this.justification === Annotation.Justify.CENTER) {
       x = start.x - text_width / 2;
     } /* CENTER_STEM */ else {
-      x = this.note.getStemX() - text_width / 2;
+      x = (note as StemmableNote).getStemX() - text_width / 2;
     }
 
     let stem_ext: Record<string, number> = {};
     let spacing = 0;
-    const has_stem = this.note.hasStem();
-    const stave = this.note.checkStave();
+    const has_stem = note.hasStem();
+    const stave = note.checkStave();
 
     // The position of the text varies based on whether or not the note
     // has a stem.
     if (has_stem) {
-      stem_ext = this.note.checkStem().getExtents();
+      stem_ext = (note as StemmableNote).checkStem().getExtents();
       spacing = stave.getSpacingBetweenLines();
     }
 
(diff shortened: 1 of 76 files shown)


rvilarl:
@ronyeh I had done a change in parallel this morning. Have a look into it, index used to be null in 3.0.9 and I tried to change it consistently to undefined. I also included the #1000:

https://github.com/rvilarl/vexflow/tree/fix/modifierIndexContextState

Do you know what is the relation between `note` and `index`? Does it make sense that one is defined and the other one not?

ronyeh:
Ah I see. I think in this case we should wait for Mohit's advice.

Setting index to undefined is an interesting approach. I guess I am not
used to seeing numbers assigned to undefined, but if we plan to stick with
that convention throughout the code base, then I can live with it.

My personal sense is that a number field should just be given an invalid
number (like -1). But we should wait for a consensus with Mohit before we
proceed.

Ron

rvilarl:
I just reverted to `undefined` because it was `null` at origin. I tend to believe that I changed to `0` because I thought that marking the problem with `note` was enough. Perhaps Mohit can answer the question:

@0xfe Do you know what is the relation between `note` and `index`? Does it make sense that one is defined and the other one not?

ronyeh:
> @ronyeh I had done a change in parallel this morning. Have a look into it, index used to be null in 3.0.9 and I tried to change it consistently to undefined. I also included the #1000:
> 
> https://github.com/rvilarl/vexflow/tree/fix/modifierIndexContextState

@rvilarl **is it okay if I integrate your ModifierContextState change into my pull request**? My PR currently uses a constant UNATTACHED_INDEX = -1. If we eventually decide to go with `undefined` I think I can just replace the instances of that constant with `undefined` and it'll mostly be good (I'll have to change a few typings and type checks).

From what I see, index is only ever used as a number.

```
const start = note.getModifierStartXY(position, index);
```

I think originally the code base used index = null as a way to say it was invalid, since index = 0 is a valid pointer into the `Note.ys[]` array.

So if we are moving to `undefined` as a way to mark things invalid, I'm happy to use that instead of -1. However, I'm just used to using things like -1 because it at least keeps index the same type. There is no need to label function return values or params like `number | undefined` :-).

Anyways, either way is fine by me as long as we are consistent across the code base, haha.

So for numbers, do we use an invalid number like -1? Or do we use `undefined`?

rvilarl:
> @rvilarl **is it okay if I integrate your ModifierContextState change into my pull request**? 

No problem, go for it.

> So for numbers, do we use an invalid number like -1? Or do we use `undefined`?

Let us 0xfe decide that and answer the question:
Do you know what is the relation between `note` and `index`? Does it make sense that one is defined and the other one not?

The question is important because it might be enough to check `note` and leave `index` unchecked.

ronyeh:
OK I integrated the ModifierContextState tiny fix into this PR.

I'm going to take a break on migration for the rest of this week. I don't want to bombard Mohit with too many more PRs haha. It woould also make the merge more difficult because some of the PRs (like this one) span multiple files, so will introduce conflicts.

0xfe:
Hi Ron -- thanks for addressing this, we should definitely make this more consistent. 

> Do you know what is the relation between note and index? 

Here `note` is usually a `StaveNote` or `TabNote` or some derivative of `Tickable`, and `index` is the specific note head in the chord. E.g., you might have a stave note `C5-E5-G5`, and putting an `Accidental` modifier `#` on `index = 0`, will attach it to the `C5`.

> Does it make sense that one is defined and the other one not?

Index does not make sense without a note, however some modifiers operate on the entire note, and indexes don't make sense there. So index should be optional / nullable.

ronyeh:
@0xfe the last question we have about `index` is can we make it a number? Then invalid indexes would be set to -1:

```
const UNATTACHED_INDEX: number = -1;
...
export class Modifier extends Element {

  protected index: number;

  constructor() {

    this.index = UNATTACHED_INDEX;
```

If we make it optionally `undefined`, it'll make the code a bit messier as we will have to use `(number | undefined)` and possibly add some ? operators as needed. It's definitely more straightforward if we keep it as a number, and do one check against -1 before this.index is used.

0xfe:
Yep, gotcha, however I feel like the semantics of `number | undefined` is more correct, and more readable, otherwise you're just using `-1` as a new kind of `undefined`.

ronyeh:
> Yep, gotcha, however I feel like the semantics of `number | undefined` is more correct, and more readable, otherwise you're just using `-1` as a new kind of `undefined`.

Sounds good! I'll make the changes such that index can be a number or undefined (when invalid).

ronyeh:
OK, this is ready for code review.

The basic idea is Modifier now gets two methods:

```
  checkAttachedNote(): void {
    if (!this.note || this.index === undefined) {
      throw new RuntimeError('NoAttachedNote', `Can't draw ${this.getCategory()} without a note and index.`);
    }
  }

  checkIndex(): number {
    if (this.index === undefined) {
      throw new RuntimeError('NoIndex', 'Modifier has an invalid index.');
    }
    return this.index;
  }
```

I used the checkXXX naming to follow with Rodrigo's convention for when we verify fields are not undefined.

Now index is a number | undefined.

src/annotation.ts
```diff
@@ -154,13 +154,11 @@ export class Annotation extends Modifier {
   // Render text beside the note.
   draw(): void {
     const ctx = this.checkContext();
-
-    if (!this.note) {
-      throw new RuntimeError('NoNoteForAnnotation', "Can't draw text annotation without an attached note.");
-    }
-
+    this.checkAttachedNote();
```
rvilarl, line 157: As the other check methods, this one could return the note and avoid the need fo the `getNote` below.

rvilarl (changes requested):

src/annotation.ts
ronyeh: Sounds good. It does seem odd to have a method named "check" that is actually a getter. :-D But as long as we are consistent with it, hehe.

rvilarl: I just followed the convention that was already there with `checkContext`. I produced a `checkStave`...

ronyeh:
`npm run test:reference` passes with 0 diffs from master.

0xfe:
Looks great! Lots of other little fixes here that I really appreciate too.

If you could fix the conflict please, I can merge. :-)

ronyeh:
OK, I think this is ready to go. I ran the tests again after merging with upstream and conflict resolution.

PR #1005, opened 2021-05-25
Fix inconsistencies with the way we check against null vs. undefined.

ronyeh:
In reviewing music.ts, I noticed that sometimes we were incorrectly comparing against `null`. For example: https://github.com/0xfe/vexflow/blob/3e13b92cc7437d5c6ac2b23fa2d056bb1f0f1564/src/music.ts#L192-L197

When we expect a number (as in the example above) or an undefined parameter (see below), comparing against `null` is incorrect.
https://github.com/0xfe/vexflow/blob/3e13b92cc7437d5c6ac2b23fa2d056bb1f0f1564/src/music.ts#L296-L297

Well, this latter line is actually fine... since undefined is falsy and null is falsy so it will do the right thing. :-D

This PR improves consistency so that we:

- check against undefined / null with strict equality operators ===
- utilize optional parameter syntax when appropriate.

### DIFF
--- a/src/music.ts
+++ b/src/music.ts
@@ -190,10 +190,7 @@ export class Music {
   }
 
   isValidNoteValue(note: number): boolean {
-    if (note == null || note < 0 || note >= Music.NUM_TONES) {
-      return false;
-    }
-    return true;
+    return note >= 0 && note < Music.NUM_TONES;
   }
 
   isValidIntervalValue(interval: number): boolean {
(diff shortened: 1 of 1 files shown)


rvilarl:
Good catch, I am sure there are more inconsistencies specially in the first migrations.

0xfe:
Nice fix! Thank you. Merging.

PR #1006, opened 2021-05-25
Migrate SVGContext

ronyeh:
Ready for review.

One thing I noticed is that many `CanvasRenderingContext2D` API methods return void. However, our SVGContext methods return `this`, which allows for chaining method calls. (The code base doesn't utilize chaining though.)


src/svgcontext.ts
```diff
@@ -2,40 +2,65 @@
 // @author Gregory Ristow (2015)
 
 import { Vex } from './vex';
+import { RenderContext } from './types/common';
 
-const attrNamesToIgnoreMap = {
+const attrNamesToIgnoreMap: { [key: string]: any } = {
```
rvilarl, line 7: There are quite a few `any` without
// eslint-disable-next-line
If they cannot be resolved I suggest to include the comments

src/svgcontext.ts
```diff
@@ -487,25 +517,24 @@ export class SVGContext {
     this.path +=
       'M' + x1 + ' ' + y1 + ' A' + radius + ' ' + radius + ' 0 ' + largeArcFlag + ' ' + sweepFlag + ' ' + x2 + ' ' + y2;
     if (!isNaN(this.pen.x) && !isNaN(this.pen.y)) {
-      this.peth += 'M' + this.pen.x + ' ' + this.pen.y;
+      this.path += 'M' + this.pen.x + ' ' + this.pen.y; // BUG? this.peth => this.path
```
rvilarl, line 520: @ronyeh did you get any difference using
`npm run test:reference`

rvilarl (changes requested):
Very good!

src/svgcontext.ts
0xfe: The type here can be `Attributes` from below (and then you won't need the `eslint-disable`.)

src/svgcontext.ts
```diff
@@ -2,40 +2,65 @@
...
-  const fontAttrNamesToIgnore = {
-    'font-family': true,
-    'font-weight': true,
-    'font-style': true,
-    'font-size': true,
-  };
+// Create the SVG in the SVG namespace:
+const SVG_NS = 'http://www.w3.org/2000/svg';
+
+type Attributes = { [key: string]: any };
```
0xfe, line 33: `eslint-disable-next-line no-explicit-any`.

src/svgcontext.ts
```diff
@@ -87,13 +112,13 @@ export class SVGContext {
...
-  create(svgElementType) {
-    return document.createElementNS(this.svgNS, svgElementType);
+  create(svgElementType: string): SVGElement {
+    return document.createElementNS(SVG_NS, svgElementType);
   }
 
   // Allow grouping elements in containers for interactivity.
-  openGroup(cls, id, attrs) {
-    const group = this.create('g');
+  openGroup(cls: string, id: any, attrs: { pointerBBox: boolean }): SVGGElement {
```
0xfe, line 120: I think `id` can be `string`, no?

src/svgcontext.ts
```diff
@@ -332,8 +364,7 @@ export class SVGContext {
   }
 
   // ## Rectangles:
-
-  rect(x, y, width, height, attributes) {
+  rect(x: number, y: number, width: number, height: number, attributes?: any): this {
```
0xfe, line 367: Can't this be `attributes?: Attribute`?

src/svgcontext.ts
0xfe: Yikes! You can grep for functions that use this and see if there are any tests that would be affected.

src/svgcontext.ts
```diff
@@ -522,7 +551,7 @@ export class SVGContext {
     return this;
   }
 
-  fill(attributes) {
+  fill(attributes: any): this {
```
0xfe, line 554: Same `attributes: Attributes`

src/svgcontext.ts
```diff
@@ -601,14 +630,14 @@ export class SVGContext {
...
   }
 
-  fillText(text, x, y) {
+  fillText(text: string, x: number, y: number): this {
     if (!text || text.length <= 0) {
-      return;
+      return this;
     }
-    const attributes = {};
+    const attributes: any = {};
```
0xfe, line 640: Here too.

0xfe (changes requested):

rvilarl:
@ronyeh 
- are you familiar with `git rebase` to resolve the conflict with master?
- could you replace also the instances of `Vex.Merge` with the spread operator?

ronyeh:
Yeah I know how to resolve conflicts. I've done it with both git rebase and git merge. At this point I think I prefer the git merge workflow.

@0xfe @rvilarl Are we replacing all Vex.Merge with spread operator ... ? How about Object.assign? Should those be replaced with ... too? Should we do a separate PR where we replace them both with ... ?

There is also a eslint flag we can use to prevent Object.assign in the future: https://eslint.org/docs/rules/prefer-object-spread

rvilarl:
> Yeah I know how to resolve conflicts. I've done it with both git rebase and git merge. At this point I think I prefer the git merge workflow.

Someone was not used to using branches I was not sure if it was you :)

 
> @0xfe @rvilarl Are we replacing all Vex.Merge with spread operator ... ? How about Object.assign? Should those be replaced with ... too? Should we do a separate PR where we replace them both with ... ?

yes, have a look at this one:
https://github.com/0xfe/vexflow/pull/1009#issuecomment-851402492

and also #1011 
 
> There is also a eslint flag we can use to prevent Object.assign in the future: https://eslint.org/docs/rules/prefer-object-spread

rvilarl:
@ronyeh you did not make any push, did you?

ronyeh:
https://github.com/0xfe/vexflow/blob/faf894c4adb650b864d7a10ab02e904fde5fd60e/src/svgcontext.js#L228

There's currently a compilation type error where render_options.line_dash is a string, but ctx.setLineDash actually expects an array of integers. So either I need to change the declared types, or do some conversion. I'll investigate this later.

```
    if (render_options.line_dash) {
      ctx.setLineDash(render_options.line_dash);
    }
```

src/svgcontext.ts
ronyeh: I added console.logs around that line....

So the containing method `arcHelper(...)` is called 1560 times, but every single time this.pen.x and this.pen.y are both `NaN`.

So this buggy line of code was never used.

```
    this.pen.x = x;
    this.pen.y = y;
```

The above code only appears in methods like: moveTo, lineTo, bezierCurveTo, quadraticCurveTo.

So the buggy line of code would have only been run if arc() is called after one of the above methods. However, in the code base, arc is only called right after beginPath(), so this.pen.x and this.pen.y are both NaN.

Also, arc() is only ever used to draw circles.

I think the intention of this line was to move the SVG path's current point back to where we were before we decided to draw the dot using arc().

So if we did: moveTo, lineTo, arc, lineTo....  then the second lineTo would continue where the first lineTo left off.

ronyeh:
`npm run test:reference` results in 0 diffs from master.

There is one remaining eslint warning for `formatter.ts`.

```
formatter.ts
484:92  warning  'voiceIndex' is defined but never used  @typescript-eslint/no-unused-vars
```

I didn't migrate that file, so we can address it in a separate PR.

src/types/common.d.ts
```diff
@@ -48,51 +53,46 @@ export interface TypeProps extends KeyProps {
   position: string;
 }
 
-/** Contexts common interface */
+export interface GroupAttributes {
```
0xfe, line 56: Let's remove this interface and move it inline please.

src/types/common.d.ts
```diff
@@ -5,11 +5,16 @@ export interface FontInfo {
   style?: string;
 }
 
+export interface Point {
```
0xfe, line 8: I'm in two minds about this -- it doesn't feel useful enough for all the indirections. Better to remove this interface and inline the structure explicitly: e.g., `function drawLine(a: {x: number, y: number}, b: {x:number, y:number})`.

0xfe (commented):
Overall looks good. Some minor comments.

src/types/common.d.ts
ronyeh: Sounds good to me!

My default inclination is to remove duplication and give names to things that represent particular concepts (e.g., Points, Vectors, Rects).

I've seen `{ x: number; y: number }` used in a few places in VexFlow and thought it might be good to consolidate them under an interface. (Looks like the TypeScript compiler doesn't generate any extra code; it's just for type checking.)

In this case since it's just {x, y} tuples, I'm happy to inline them everywhere we use them.

ronyeh:
OK. Good to go. :-)

0xfe:
Looks great!

PR #1009, opened 2021-05-30
Migration/vex

rvilarl:
@0xfe @ronyeh I need help, I donot know what I am oing wrong. Everything compiles but Vex is not being exported correctly.


ronyeh:
When I check out your branch and do a npm install, I get some errors.

```
npm WARN deprecated highlight.js@9.18.5: Support has ended for 9.x series. Upgrade to @latest

> puppeteer@4.0.1 install vexflow/node_modules/puppeteer
> node install.js

Downloading Chromium r756035 - 124.1 Mb [                    ] 1% 31.1s ERROR: Failed to set up Chromium r756035! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.
Error: 4601402816:error:1408F119:SSL routines:ssl3_get_record:decryption failed or bad record mac:../deps/openssl/openssl/ssl/record/ssl3_record.c:677:

  -- ASYNC --
    at BrowserFetcher.<anonymous> (vexflow/node_modules/puppeteer/lib/helper.js:116:19)
    at fetchBinary (vexflow/node_modules/puppeteer/install.js:151:8)
    at download (vexflow/node_modules/puppeteer/install.js:55:9) {
  library: 'SSL routines',
  function: 'ssl3_get_record',
  reason: 'decryption failed or bad record mac',
  code: 'ERR_SSL_DECRYPTION_FAILED_OR_BAD_RECORD_MAC'
}
npm WARN typedoc@0.20.36 requires a peer of typescript@3.9.x || 4.0.x || 4.1.x || 4.2.x but none is installed. You must install peer dependencies yourself.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! puppeteer@4.0.1 install: `node install.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the puppeteer@4.0.1 install script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.


```

ronyeh:
I did a `git apply 1009.patch` and it seems to build fine. So I'll see if I can figure out what's going on.

https://patch-diff.githubusercontent.com/raw/0xfe/vexflow/pull/1009.patch

ronyeh:
When I load flow.html, I get this in the console:

```
Uncaught TypeError: Cannot read property 'MakeException' of undefined
    at Object../src/registry.ts (vexflow-debug.js:22437)
    at __webpack_require__ (vexflow-debug.js:30)
    at Object../src/element.ts (vexflow-debug.js:5009)
    at __webpack_require__ (vexflow-debug.js:30)
    at Object../src/tables.ts (vexflow-debug.js:28332)
    at __webpack_require__ (vexflow-debug.js:30)
    at Object../src/vex.ts (vexflow-debug.js:32730)
    at __webpack_require__ (vexflow-debug.js:30)
    at Object../src/index.ts (vexflow-debug.js:18737)
    at __webpack_require__ (vexflow-debug.js:30)
```

If you look at the stack it looks like index.ts => vex.ts => tables.ts => element.ts => registry.ts which uses `Vex.MakeException`. At this point it looks like Vex isn't done initializing yet.

Maybe it's a circular dependency via the imports?

I notice that vex.ts has a import { Flow } from ./tables, and then at the top of tables.ts, there is a import { Vex } from ./vex.

Another way to inspect this is to go to the line where there is a runtime error, and drop this line before the line with the error:

`debugger;`

So my first 4 lines of registry.ts looks like:

```
import { Vex } from './vex';

debugger;
const X = Vex.MakeException('RegistryError');

```
Then refresh `flow.html`.

Chrome will break at the debugger line, and you can inspect that Vex is currently undefined.


I see that you're so far sticking with the original method of keeping Vex defined as a JS object literal { }.

Perhaps the TypeScript way would be to declare Vex as a `class` and then have static methods and static properties? Or maybe a TypeScript namespace?

In any case, it seems like currently there is a circular chain of imports. Ideally Vex shouldn't have any dependencies, so that all of its static methods like `MakeException` will be defined before other classes need them.

0xfe:
Ron is right about the circular dependencies. Also, many of those methods are decade-old and no longer necessary. 

I'd do two things here: 1) move the helper functions (`MakeException`, `Merge`, `forEach`, etc.) to a new `util.ts`, 2) update references that use those calls to import from `util.ts` instead, 3) if everything works, you can also remove them (e.g., replace `Merge` with spread operator, `forEach` with native `Array.forEach`, etc.

rvilarl:
When you say util.ts, would you arrange the elements like in vex.ts or would you leave the helper functions stand alone? I would start with RuntimeError. Then Util.RuntimeError or RuntimeError?

0xfe:
Standalone functions not static member functions.

rvilarl:
Utils merged, ready for review!

src/tables.ts
```diff
@@ -0,0 +1,1526 @@
...
+  // This will be deprecated in the future. This is a temporary solution until
+  // we have more robust text metrics.
+  TEXT_HEIGHT_OFFSET_HACK: 1,
+
+  /* Kerning (DEPRECATED) */
+  IsKerned: true,
+  Element: Element,
+  Fraction: Fraction,
+  Renderer: Renderer,
+  Formatter: Formatter,
```
0xfe, line 425: Can we update these imported elements to `index.ts`?

src/tables.ts
```diff
@@ -0,0 +1,1526 @@
...
+      'a@a': { code: 'fermataAbove', between_lines: false }, // Fermata above staff
+      'a@u': { code: 'fermataBelow', between_lines: false }, // Fermata below staff
+      'a|': { code: 'stringsUpBow', between_lines: false }, // Bow up - up stroke
+      am: { code: 'stringsDownBow', between_lines: false }, // Bow down - down stroke
+      'a,': { code: 'pictChokeCymbal', between_lines: false }, // Choked
+    };
+    return articulations[artic];
+  },
+
+  accidentalCodesAccidentals: accidentals,
```
0xfe, line 760: Let's call this `accidentalMap`

src/tables.ts
rvilarl: Typescript does not like to include additional properties to the variable once it is declared.

src/tables.ts
rvilarl: Done

rvilarl:
Ready to merge

rvilarl:
@0xfe I just remembered that this was the right time to remove common.ts.

rvilarl:
Ready to review. I think we are both wishing to get to the end!

src/tables.ts
0xfe: Putting it in here makes it really messy. You can do this:

```typescript
// tables.ts
export const Flow: Record<any, any> = {
  DEFAULT_FONT_STACK: ...,
 // ...
};
```

```typescript
// index.ts
import {Flow} from './tables'

Flow.Element = Element;
```

Here's another option: in this file call it `Tables` instead of `Flow`.
```typescript
// tables.ts
export const Tables = {
  DEFAULT_FONT_STACK: ...,
 // ...
};
```

Then inside `index.ts` you can declare `Flow` like so:

```typescript
// index.ts
import {Tables} from './tables'
import {Element} from './element'
// ...

export const Flow = {
  Element: Element,
  // ...
  ...Tables
}
```

I like this second approach better, if it works.

rvilarl: Me too, I will give it a try. All the other PRs are also ready for review :)

rvilarl: There is a problem, `Flow` is also required by `Vex`. I could do an additional file called `flow.ts` following the approach above:
`table.ts` exports `Table`
`flow.ts` imports `table.ts` and exports `Flow`
`vex.ts` imports `flow.ts` and exports `Vex`

I could remove `index.ts` and declare `vex.ts` the entry point of the library as well.

0xfe: Thanks Rodrigo. Additional `flow.ts` sounds good. Keep entry point as `index.ts` please.

rvilarl:
@0xfe the separation of Tables now working. Ready to merge, hopefully!

src/flow.ts
```diff
@@ -25,7 +21,7 @@ import { Note } from './note';
 import { ModifierContext } from './modifiercontext';
 import { MultiMeasureRest } from './multimeasurerest';
 import { TickContext } from './tickcontext';
-import { Articulation } from './articulation';
+import { Articulation, ArticulationStruct } from './articulation';
```
0xfe, line 24: lint error: `'ArticulationStruct' is defined but never used`

0xfe (commented):
Looks great. Minor comment.

src/flow.ts
rvilarl: done

rvilarl:
Ready

0xfe:
Great! Thanks for grinding through this! :-)

PR #1011, opened 2021-06-01
migration/util

rvilarl:
log, warn, max, min

### DIFF
--- a/src/textfont.ts
+++ b/src/textfont.ts
@@ -6,16 +6,11 @@
 // VEX modules can take advantage of font metrics in a uniform way.
 //
 
-import { Vex } from './vex';
 import { RuntimeError, log } from './util';
 import { PetalumaScriptTextMetrics } from './fonts/petalumascript_textmetrics';
 import { RobotoSlabTextMetrics } from './fonts/robotoslab_textmetrics';
 import { FontGlyph } from './font';
 import { FontInfo } from './types/common';
-export interface TextFontMetrics {
-  advanceWidth: number;
-  ha: number;
-}
 
 export interface TextFontRegistry {
   [name: string]: unknown;
@@ -44,7 +39,7 @@ export class TextFont {
   protected static debug: boolean;
   protected resolution: number = 1000;
   protected name?: string;
-  protected glyphs: Record<string, TextFontMetrics> = {};
+  protected glyphs: Record<string, FontGlyph> = {};
   protected family: string = '';
   protected serifs?: boolean;
   protected monospaced?: boolean;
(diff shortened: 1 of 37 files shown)


src/textfont.ts
```diff
@@ -290,7 +300,7 @@ export class TextFont {
     if (!metric) {
       return 0.65 * this.pointsToPixels;
     }
-    return (metric.advanceWidth / this.resolution) * this.pointsToPixels;
+    return ((metric.advanceWidth ?? 1000) / this.resolution) * this.pointsToPixels;
```
0xfe, line 303: This looks like a silent failure. What's happening here?

src/tuplet.ts
```diff
@@ -98,24 +97,23 @@ export class Tuplet extends Element {
       throw new RuntimeError('BadArguments', 'No notes provided for tuplet.');
     }
 
-    this.options = Vex.Merge({}, options);
+    this.options = { ...options };
     this.notes = notes;
-    this.num_notes = 'num_notes' in this.options ? this.options.num_notes : notes.length;
+    this.num_notes = this.options.num_notes ? this.options.num_notes : notes.length;
```
0xfe, line 102: This will break for `num_notes == 0`, no?

src/textfont.ts
rvilarl: Two possible explanations:
- Not all the `FontGlyphs` define `advanceWidth` so a default value is required. See `csymDiagonalArrangementSlash` in `Bravura`
- This is not a `FontGlyphs` and is a `TextFontMetric` there was a bug in the code mixing both types. Using any of the two, the test pass fine.

src/tuplet.ts
rvilarl: `undefined` used

rvilarl:
@0xfe dependencies on `Vex` are almost done, how do you want to proceed with the dependencies on `Flow`? I would say that it would make sense to move the tables and associated function to the most appropriate class. 
We also have quite a few constants defined in `tables.ts` should I move them to `util.ts`?
@ronyeh what do you think?

src/textfont.ts
0xfe: Gotcha. I think a default is a good idea, but this is not a good place for it (it will be hard to find in the future.) Maybe put it in `FontGlyph`?

0xfe:
+1 for moving the tables and functions to appropriate classes, that's been my approach so far as well. Anything that isn't clear leave here. Let's keep `util.ts` just for common generic utility functions.

src/textfont.ts
rvilarl: @0xfe One question prior to the change, the right type for `metric` is `FontGlyphs` or `TextFontMetric`?

0xfe: @AaronDavidNewman do  you have thoughts here?

rvilarl: Fixed ?? no longer necessary

rvilarl:
Ready to merge.

0xfe:
Awesome!

Issue #1013, opened 2021-06-02
Duplicate/nested for loop


ronyeh:
https://github.com/0xfe/vexflow/blob/faf894c4adb650b864d7a10ab02e904fde5fd60e/src/stringnumber.ts#L52-L58

It's late and I'm seeing double. :-D

While working on my PR I ran across these lines which stood out to me and just look wrong.

I believe the outer for loop can just be removed, correct? Since they both iterate on index i, by the time the inner for loop finishes, the outer one will also finish.

sschmidTU:
yeah, looks like one of the loops can be removed, they do the same thing.
It definitely doesn't make sense to have nested loops on the same index variable (i).

ronyeh:
Thanks for the sanity check @sschmidTU . I added a PR with the fix.

PR #1014, opened 2021-06-02
Fix duplicated for loop: issue https://github.com/0xfe/vexflow/issues/1013

ronyeh:
I removed the duplicated for loop in issue: https://github.com/0xfe/vexflow/issues/1013

I verified that the outer loop was never run more than once even though the inner one goes up to 3 times, with some awesome println debugging.

```
console.log('outer for loop ' + i);
...
      console.log('inner for loop ' + i);
```

Tests pass with no diffs.

### DIFF
--- a/src/stringnumber.ts
+++ b/src/stringnumber.ts
@@ -45,46 +45,37 @@ export class StringNumber extends Modifier {
     let shift_left = 0;
     let shift_right = 0;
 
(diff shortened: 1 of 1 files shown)


0xfe:
This is incredibly strange. I have no idea what the original intent was here, but thanks for fixing!

PR #1015, opened 2021-06-03
Add `ver` query param to flow.html to allow switching between different vexflow builds.

ronyeh:
For example:
`flow.html?ver=(build|reference|releases)`

You can open different versions in different tabs and switch between them to see the visual differences.

### DIFF
--- a/tests/flow.html
+++ b/tests/flow.html
@@ -3,13 +3,11 @@
   <head>
     <title>VexFlow - JavaScript Music Notation and Guitar Tab</title>
     <link rel="stylesheet" href="flow.css" type="text/css" media="screen" />
-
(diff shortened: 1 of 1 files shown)


ronyeh:
By default, the flow.html test page looks the same.

If we add a `ver=reference` query param to the URL, it will load the vexflow version in the `reference/` folder. The `ver` can be any string, like `releases` or `temp_hack` and it will just look for that corresponding directory and load the correct: `vexflow-debug.js` and `vexflow-tests.js`.

This allows us to load the same tests from two different versions and then switch tabs to compare them, like in the example below. When comparing the current build against the `releases` build, I noticed that the current build doesn't seem to test Gonville & Petaluma, while the `vexflow-tests.js` in the `releases` folder **does** test those two fonts.

vex_version

0xfe:
Nice!

PR #1016, opened 2021-06-03
Replace instances of Barline.type with BarlineType.

ronyeh:
The `Barline.type` getter just returns `BarlineType`. I did not remove the getter, in case it is important for backwards compatibility.

However, I replaced instances of Barline.type in the VexFlow library with `BarlineType`, to skip the getter.

Tests remain unchanged for now. We can fix them as we migrate tests to TS.


0xfe:
Looks great!

PR #1017, opened 2021-06-05
Migration/keysignature

rvilarl:
keysignature is now migrated, however the result is different, the signatures are now more compact.
@0xfe @ronyeh please look if I introduced a bug.

The reason might also be a bug in the code `KeySigNote` was accessing `width` directly in `keysignature` and not through `getWidth`. As a consequence, `format` was not called.


rvilarl:
**current**

Clef_Keys Major_Key_Clef_Test_Current

**reference**

Clef_Keys Major_Key_Clef_Test_Reference

rvilarl:
Conflicts resolved

rvilarl:
> keysignature is now migrated, however the result is different, the signatures are now more compact.
> @0xfe @ronyeh please look if I introduced a bug.
> 
> The reason might also be a bug in the code `KeySigNote` was accessing `width` directly in `keysignature` and not through `getWidth`. As a consequence, `format` was not called.

My guess was not correct, the difference has nothing to do with `getWidth` and `format`
@ronyeh could you make a review?

rvilarl:
@0xfe @ronyeh please look into that, I really do not see the difference.

rvilarl:
@0xfe this is the last class to get the src migration complete!

src/keysignature.ts
```diff
@@ -114,11 +124,11 @@ export class KeySignature extends StaveModifier {
     }
 
     // Place the glyph on the stave
-    this.placeGlyphOnLine(glyph, this.stave, acc.line);
+    this.placeGlyphOnLine(glyph, this.checkStave(), acc.line);
     this.glyphs.push(glyph);
 
     const xPosition = this.xPositions[this.xPositions.length - 1];
-    const glyphWidth = glyph.getMetrics().width + extraWidth;
+    const glyphWidth = glyph.getMetrics().width ?? 0 + extraWidth;
```
0xfe, line 131: Can we remove the `?? 0` -- there could be a silent failure here.

src/keysignote.ts
```diff
@@ -25,7 +25,7 @@ export class KeySigNote extends Note {
     this.setPreFormatted(true);
     this.keySignature.setStave(this.checkStave());
     this.keySignature.format();
```
0xfe, line 27: You can remove this (since `getWidth` calls `format`).

src/keysignature.ts
rvilarl: done

src/keysignote.ts
rvilarl: done

rvilarl:
@0xfe ready

0xfe:
Does the same visual diff show up?

rvilarl:
@0xfe I believe so I will check

rvilarl:
@0xfe No! There is no difference, you fixed it! Welldone!

0xfe:
👍 great!

PR #1018, opened 2021-06-05
migration/keymanager

rvilarl:
KeyManager!

I have separated the mix of Note and Key under `NoteParts` that we did in music.ts.


rvilarl:
@0xfe is there any concern with this PR? It was open long ago.

src/keymanager.ts
```diff
@@ -58,7 +68,13 @@ export class KeyManager {
     return this;
   }
 
-  getAccidental(key) {
+  getAccidental(
```
0xfe, line 71: This formatting feels off. Can you put the function and params in one line?

0xfe (commented):
Sorry, I may have missed this one.

src/keymanager.ts
rvilarl: you are right, fixed

rvilarl:
@0xfe ready

0xfe:
Looks good.

PR #1020, opened 2021-06-05
fix formatter.preCalculateMinTotalWidth to return enough pixels, and related formatting changes

AaronDavidNewman:
There are a lot of changes here.  I'll post some before/after images also so you can see what the effect is.

1. There is one actual formatting change.  The formatter will always try to give you exactly the width you justify to, within a small threshold (and if there's room).  In other words, the padding on the right of the measure will always be consistent within a few pixels.  You can always increase the padding by just making the measure wider (or decrease justify width).
2. For estimating `preCalculateMinTotalWidth`, a few changes:
    1. we add additional padding if there are non-aligned beats between justified voices.  This is the main reason previous estimates were often too low.
    2. If there are flags on the stavenote, we adjust width to the width of the flag.  Previously the width used was the notehead which is just a teency bit smaller.  Very small change.
    
3. Where stave padding is used to calculate music width, I added both the padding at the beginning and end of the measure.  Previously (like in `Formatter.formatToStave`) only the beginning padding was considered.    I also added a constant to Stave that returns this value.  And since change 1 makes the right padding consistent, this value will also be consistent.  If you increase the staff width beyond that, those pixels will appear as padding at the end of the measure.
4. I added a static method to formatter called `estimateJustifiedMinWidth` that takes voice array as a parameter.  This can be used to get the value of `preCalculateMinTotalWidth`, without creating a formatter directly.  If you are using the factory, this is how you would get it.  It would be nice if the factory had a 'formatAndDrawToMinWidth' or something.
5. Change default iterations of formatter from 2 to 5 (for point 1 above).
6. Rewrote most formatter test cases to use `preCalculateMinTotalWidth` instead of using hard-coded width.


src/fonts/bravura_metrics.ts
```diff
@@ -4,6 +4,9 @@ export const BravuraMetrics = {
 
   stave: {
     padding: 12,
+    endPaddingMax: 10,
```
AaronDavidNewman, line 7: This is the staff padding.  endPaddingMin is the minimum padding on right.  endPaddingMax is the maximum.  unalignedNotePadding is the extra padding used to calculate min width for unaligned beats.  `padding` is (and always was) the startX of the first note from the left.

src/formatter.ts
```diff
@@ -691,14 +704,22 @@ export class Formatter {
...
+    const paddingMax = musicFont.lookupMetric('stave.endPaddingMax');
+    const maxX = adjustedJustifyWidth + lastContext.getMetrics().notePx - paddingMin;
 
     let iterations = this.formatterOptions.maxIterations;
-    while (actualWidth > maxX && iterations > 0) {
-      // If we couldn't fit all the notes into the jusification width, it's because the softmax-scaled
-      // widths between different durations differ across stave (e.g., 1 quarter note is not the same pixel-width
-      // as 4 16th-notes). Run another pass, now that we know how much to justify.
-      targetWidth -= actualWidth - maxX;
+    while ((actualWidth > maxX && iterations > 0) || (actualWidth + paddingMax < maxX && iterations > 1)) {
```
AaronDavidNewman, line 716: `>1` because we always want to end with < actualWidth pixels, we don't want adjusting right as our last iteration.

0xfe:
Thanks @AaronDavidNewman. There's a lot going on here. I can review this, but is there a way to break this up into smaller PRs?

AaronDavidNewman:
Fine with me.  A subset of the changes will show some problems that the other changes fix, is the only thing.  But if it makes the changes easier to digest, I can submit them one at a time.

0xfe:
Maybe start with package.json + `Stave.defaultPadding`

ronyeh:
I would love to see some before / after images (or a GIF that flips between the two, hehe).

AaronDavidNewman:
@ronyeh  there are quite a few in the parent issue 846

AaronDavidNewman:
If that's the approach, I think I will cancel this PR and start over with the smaller change.  Github doesn't make this so easy - I can't just make a branch and push from there (unless I want to pay for a premium account :)

0xfe:
Thanks Aaron!

PR #1027, opened 2021-06-09
Migrate CanvasContext

ronyeh:
Interestingly, one visual diff. This migration might have fixed a bug??

TextNote

The reference / master version does not have the subscripts, but the post migration version restores the subscripts. I have no idea why.... will have to dig into this!


ronyeh:
Summary: the CanvasContext migration 1) fixed a visual bug which 2) uncovered a bug related to measuring text height & width.

This issue is also present in 3.0.9, but is not present in the version hosted on vexflow.com.

The failing test cases are:

-   TextNote (Canvas): TextNote Superscript and Subscript
-   TextNote (SVG Bravura): TextNote Superscript and Subscript

Basically, the workaround is to estimate the height by measuring the width of the capital letter **M**, for example: 
`const height = ctx.measureText('M').width;`

The current version does ` const height = ctx.measureText(this.text).height;` which doesn't work, because the TextMetrics object returned by CanvasRenderingContext2D.measureText() does not have a height field. (so height is `undefined` and the subsequent calculations result in NaN).

https://github.com/0xfe/vexflow/blame/fe151b3722248a25a0fbe9e5a5840a3faf3ac4a1/src/textnote.ts#L210

Anyways, I propose adding new methods to the RenderContext interface called   
```
getTextHeight(text: string): number;
getTextWidth(text: string): number;
```

Then we can consolidate all the text measurement into those functions. (We currently are not consistent with the way we estimate text height. Sometimes we measure a lowercase m. Other times we measure a uppercase M.)

Alternatively, we can add a single method `getTextBounds(text: string): {width: number, height: number};` But most of the time in VexFlow, we just need the width.... so it seems like extra work to measure the letter M if we don't end up using the height.

ronyeh:
This PR is good to go. Ready for review.

The only visual difference is from the bug fix. On master, the superscripts/subscripts are hidden.

TextNote.TextNote_Superscript_and_Subscript.Petaluma.png
TextNote TextNote_Superscript_and_Subscript Bravura

TextNote.TextNote_Superscript_and_Subscript.Gonville.png
TextNote TextNote_Superscript_and_Subscript Gonville

TextNote.TextNote_Superscript_and_Subscript.Bravura.png
TextNote TextNote_Superscript_and_Subscript Petaluma

rvilarl:
@ronyeh checks are not running

ronyeh:
OK, I think this is good to go. I rebased off of revision a85a83276a3ce27adf76e2a0fc3bf5c8792fb405 (from Sat Jun 12 19:26:23 2021). 

The visual diffs are due to the `CanvasRenderingContext2D.measureText()` bug fix included with this PR. For example, TextNote.TextNote_Superscript_and_Subscript.Bravura_Current.png now shows the superscripts whereas before they were hidden.

Note: I will be away until Thursday. I think this merge is pretty safe, but if you want, feel free to wait until then to merge.

ronyeh:
Ready for code review.

0xfe:
Looks great. Thanks for catching and fixing that `measureText` bug.

PR #1028, opened 2021-06-09
Fix #1026 by using Flow.DEFAULT_FONT_STACK instead of a constant DefaultFontStack.

### LINKED ISSUE #1026: Cannot customize DEFAULT_FONT_STACK
I think we introduced a bug into element.ts:
https://github.com/0xfe/vexflow/blob/2c77ad2af0a5cc79ba5857d3c640b6dd1280741d/src/element.ts#L65-L66

I think the original code allowed developers to customize `Vex.Flow.DEFAULT_FONT_STACK`, which would affect the music notation font. DEFAULT_FONT_STACK is no longer referenced anywhere in the core library.

I noticed this while cleaning up `vexflow_test_helpers.js`, which sets VF.DEFAULT_FONT_STACK to generate the different outputs. Now all the tests are rendered in Bravura.

ronyeh:
Fixes https://github.com/0xfe/vexflow/issues/1026

I also turned on visual diffs for **all three fonts by default**.

For faster development, each individual developer can temporarily diff fewer fonts by customizing `generate_png_images.js` line 20. Before sending in a PR, the developer should make sure to run the visual diff for all three fonts.

In the future, we can add a npm script to allow testing of fewer fonts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
     "lint": "grunt eslint",
     "qunit": "grunt test",
     "generate:current": "node ./tools/generate_png_images.js ../build ./build/images/current",
-    "generate:current:all-fonts": "node ./tools/generate_png_images.js ../build ./build/images/current --fonts=all",
     "generate:reference": "node ./tools/generate_png_images.js ../reference ./build/images/reference",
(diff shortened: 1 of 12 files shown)


0xfe:
Looks good. Thanks for catching and fixing, Ron!

PR #1029, opened 2021-06-10
Improve webpack build process to include vexflow version, git hash, and banner.

ronyeh:
I found a visual bug and was trying to compare the same test between different vexflow versions, to see if I could determine when the bug was introduced. For example, there is a test case that works on https://www.vexflow.com/tests/ but fails in the current master HEAD revision. 

To make it easier to identify the build, I think it would be nice to include the version and git hash as publicly accessible variables. So I added `Vex.Flow.VERSION` and `Vex.Flow.BUILD` via webpack.DefinePlugin.

I also noticed the gruntfile has a banner which is no longer getting included, since we transitioned from concatenating JS files to using webpack. So I restored the banner, which includes VexFlow version, build timestamp, and git hash via webpack.BannerPlugin.

Hopefully this PR will be useful, for identifying which version a particular website uses! We will be able to open the dev console and just type `Vex.Flow.VERSION` and/or `Vex.Flow.BUILD` to pinpoint the exact build.


sschmidTU:
I also added `.Version` to OSMD a while ago for similar reasons, can be very useful.

0xfe:
This looks good! Can you fix the conflict please?

ronyeh:
I fixed the conflict, but now that we have introduced `flow.ts`, this PR causes errors with the typescript compiler and typedoc. However, it still builds fine when using `--force`, because webpack define does the right magic.

I might have to break out the definition of Vex.Flow.VERSION and Vex.Flow.BUILD into a separate tiny JS file that is included by webpack and ignored by eslint / ts compiler / typedoc.

Question for @0xfe : What should we do with the header.js file? https://github.com/0xfe/vexflow/blob/master/src/header.js
**Can I remove it as part of this PR?** It is no longer included anywhere, and the LICENSE file and this pull request covers what the header.js was originally intended to do. I can also move the credit to the font author into the top of each font metrics file.

rvilarl:
I had not installed the latest update with `npm install` and I was not getting the typedoc error. I will look into that.

0xfe:
Thanks @ronyeh -- removing `header.js` sounds good. Reviewing your other PR.

ronyeh:
OK, it's ready for review!

This PR requires a `npm install` to update your dependencies. I have added `webpack-inject-plugin` which allows us to directly inject code into the bundle. This avoids the typescript compiler errors, because the lines of code defining the VERSION and BUILD are injected by webpack.

This PR modifies the Gruntfile to inject the following properties:
* Vex.Flow.VERSION  (e.g., "3.0.9")
* Vex.Flow.BUILD  (e.g., "5463be0ea2570d54ba47f58c66c60b599b7f2b32")
* Vex.Flow.Test.VERSION  (we put the same properties into the `vexflow-tests.js` file, so we can verify if the tests are from the same build as the core library)
* Vex.Flow.Test.BUILD

This PR also removes `header.js` and adds a banner to the top of each file.

npm run test:reference shows no diffs from master @ 605b9a7bc15ec2eb687fd17c49ab87199b698999.

0xfe:
Great!

PR #1035, opened 2021-06-13
multimeasurerest fix

### LINKED ISSUE #1034: MultiMeasureRest.Simple_Test regression due to recent merge
MultiMeasureRest.Simple_Test.Bravura.png
MultiMeasureRest Simple_Test Bravura

MultiMeasureRest.Simple_Test.Petaluma.png
MultiMeasureRest Simple_Test Petaluma

rvilarl:
#1034 fixed

### DIFF
--- a/src/multimeasurerest.ts
+++ b/src/multimeasurerest.ts
@@ -250,7 +250,7 @@ export class MultiMeasureRest extends Element {
       timeSig.point = this.render_options.number_glyph_point;
       timeSig.setTimeSig(timeSpec);
       timeSig.setStave(stave);
-      timeSig.setX(left + (right - left) * 0.5 - timeSig.getWidth() * 0.5);
+      timeSig.setX(left + (right - left) * 0.5 - timeSig.getTimeSig().glyph.getMetrics().width * 0.5);
       timeSig.bottomLine = this.render_options.number_line;
       timeSig.setContext(ctx).draw();
     }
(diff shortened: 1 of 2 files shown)


src/multimeasurerest.ts
```diff
@@ -250,7 +250,7 @@ export class MultiMeasureRest extends Element {
       timeSig.point = this.render_options.number_glyph_point;
       timeSig.setTimeSig(timeSpec);
       timeSig.setStave(stave);
-      timeSig.setX(left + (right - left) * 0.5 - timeSig.getWidth() * 0.5);
+      timeSig.setX(left + (right - left) * 0.5 - timeSig.getTimeSig().glyph.getMetrics().width * 0.5);
```
ronyeh, line 253: I admit to knowing nothing about this section of the code.

What was the bug?

How come the fixed line looks so different from the buggy line? From a cursory glance, it looks like the buggy line actually makes sense. :-) In fact, the fixed line looks weird to me, because why would we call getTimeSig() on timeSig??? :-)

rvilarl: I thought the same. The original code was very similar to what we have now. I thought `getWidth` would return the same but obviously it does not :). The original was:
```
timeSig.timeSig.glyph.getMetrics().width
```

0xfe: The original code was using the width of just the glyph, the incorrect update used the entire width.

Suggestions below.

0xfe:
Merging this to fix the bug, but suggestions: rename `getTimeSig` to `getInfo`. Rename `TimeSignature.timeSig` to `TimeSignature.info`.

PR #1037, opened 2021-06-13
Correct staff padding to include end padding

AaronDavidNewman:
Currently Formatter.formatToStave uses the pre-first-note padding but does not consider padding at the end of the measure.  Also, the formatting doesn't really consider end padding at all, so the actual end point of the bar is pretty inconsistent.

2 new metrics are defined for stave:  endPaddingMin and endPaddingMax.  default padding for calculation purposes is `padding + endPaddingMax`.  The formatter decides the current format fits within the measure if the right-most point is > `staffWidth - endPaddingMin` and < `staffWidth - endPaddingMax`.

Note that many test cases don't actually use formatToStave (even some that think they do) due to some other bugs that will be addressed in other changesets.

Here is a test case that _does_ use formatToStave.  Three images, pre-changes, changing only the padding, and changing the formatting to consider the padding.
Prior to this cs:
pr_formatter_before

Only considering additional end padding:
pr-correctpadding

Using both end padding and formatting to the end padding:
image

### DIFF
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -17,7 +17,6 @@
 // See `tests/formatter_tests.js` for usage examples. The helper functions included
 // here (`FormatAndDraw`, `FormatAndDrawTab`) also serve as useful usage examples.
 
-import { Vex } from './vex';
 import { RuntimeError, midLine, log } from './util';
 import { Beam } from './beam';
 import { Flow } from './tables';
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -693,10 +693,13 @@ export class Formatter {
       firstContext.getMetrics().totalLeftPx;
     let targetWidth = adjustedJustifyWidth;
     let actualWidth = shiftToIdealDistances(calculateIdealDistances(targetWidth));
-    const maxX = adjustedJustifyWidth + lastContext.getMetrics().notePx;
+    const musicFont = Flow.DEFAULT_FONT_STACK[0];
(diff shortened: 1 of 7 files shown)


src/stave.ts
```diff
@@ -62,6 +62,13 @@ export class Stave extends Element {
   protected bounds: Bounds;
   protected readonly modifiers: StaveModifier[];
 
+  // This is the sum of the padding that normally goes on left + right of a stave during
+  // drawing.  Used to size staves correctly with content width
+  static get defaultPadding() {
```
0xfe, line 67: Needs a return type: `static get defaultPadding() : number`.

src/formatter.ts
```diff
@@ -693,10 +693,13 @@ export class Formatter {
       firstContext.getMetrics().totalLeftPx;
     let targetWidth = adjustedJustifyWidth;
     let actualWidth = shiftToIdealDistances(calculateIdealDistances(targetWidth));
-    const maxX = adjustedJustifyWidth + lastContext.getMetrics().notePx;
+    const musicFont = Flow.DEFAULT_FONT_STACK[0];
```
0xfe, line 696: Not related to this PR, but lint error. Can you remove the `import { Vex } ...` from the top. (`Vex` is no longer used here.)

0xfe (commented):
Thanks. Minor comments.

AaronDavidNewman:
OK!

0xfe:
LGTM, thanks

PR #1038, opened 2021-06-14
Add additional padding to left of accidentals

AaronDavidNewman:
I thought it made sense to do the actual formatting changes first, b/c it's hard to evaluate the success of an estimate if there are formatting issues.

This just adds 2 pixels of padding to an accidental block, so they don't get squished against the notes to the left.  Also moved a couple of formatting magic numbers to metrics.  

Before:
pr-accidental-padding

After, 1px padding:
image

After, 2px padding (this PR):
image

### DIFF
--- a/src/accidental.js
+++ b/src/accidental.js
@@ -34,9 +34,11 @@ export class Accidental extends Modifier {
 
   // Arrange accidentals inside a ModifierContext.
   static format(accidentals, state) {
-    const noteheadAccidentalPadding = 1;
+    const musicFont = Flow.DEFAULT_FONT_STACK[0];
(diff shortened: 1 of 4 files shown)


sschmidTU:
NIce!
Can the padding be modified though? That would be great. I can see myself prefering only 1 pixel of padding (if at all), to keep scores more compact. Every pixel counts ;)

AaronDavidNewman:
@sschmidTU it is a metric of the font, so I'd image you can override it at a score/application level.
I also included 1 px as a reference.

0xfe:
Looks good, thanks!

PR #1040, opened 2021-06-23
Correctly check for empty array

eliot-akira:
I found a very tiny bug where it's checking for empty array using `notes === []`, which is always false because `[] !== []` in JavaScript. Changed to `notes.length === 0`.

Searching the code base, the file `src/beams.ts` is the only place where this comparison was used.


rvilarl (approved):
Thanks

0xfe:
Thanks for catching.

PR #1041, opened 2021-06-27
Migration/update (updates node_modules and fixes typedoc)

rvilarl:
`grunt` cannot be updated due to the dependencies from `grunt-bump` & `grunt-git`


0xfe:
Looks good.

PR #1044, opened 2021-07-01
fix formatting merge bug in system.js

AaronDavidNewman:
stave with padding needs to consider both beginning and ending padding.


0xfe:
Thanks for the fix!

PR #1045, opened 2021-07-01
Fix wrong global definition genereated by Webpack in UMD styles

aleen42:
Requiring releases under Node.js will throw an error:

```
})(window, function() {
   ^

ReferenceError: window is not defined
```

REF: https://github.com/webpack/webpack/issues/6522

After checking the history, I can confirm that it is a problem after upgrading Webpack above 4.x: https://github.com/0xfe/vexflow/pull/1045#discussion_r674445728 and 0802a99 has set `globalObject: 'this'` to solve the problem, but it is not a perfect way.

Other:

- Support passing elements directly when constructing `Renderer`s, so that developers can:
  1. Use [`node-canvas`](https://github.com/Automattic/node-canvas) to render with this tool.
  2. Use [`jsdom`](https://github.com/jsdom/jsdom) to simulate and generate SVG headlessly.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -37,7 +37,7 @@ module.exports = (grunt) => {
         library: libraryName,
         libraryTarget: 'umd',
         libraryExport: 'default',
-        globalObject: 'this',
+        globalObject: `typeof window !== 'undefined' ? window : this`,
         publicPath: 'auto',
       },
       resolve: {
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -13,7 +13,7 @@ export type ContextBuilder = typeof Renderer.getSVGContext | typeof Renderer.get
  * Support Canvas & SVG rendering contexts.
  */
 export class Renderer {
-  protected elementId?: string | HTMLElement;
+  protected elementId?: string | HTMLCanvasElement;
   protected element: HTMLCanvasElement;
   protected backend: number;
 
@@ -43,7 +43,7 @@ export class Renderer {
   static lastContext: RenderContext | undefined = undefined;
 
   static buildContext(
-    elementId: string | HTMLElement,
+    elementId: string | HTMLCanvasElement,
     backend: number,
     width: number,
     height: number,
@@ -145,18 +145,20 @@ export class Renderer {
     context.stroke();
   }
 
-  constructor(elementId: string | HTMLElement, backend: number) {
+  constructor(elementId: string | HTMLCanvasElement, backend: number) {
     if (!elementId) {
       throw new RuntimeError('BadArgument', 'Invalid id for renderer.');
     }
 
-    this.element = document.getElementById(elementId as string) as HTMLCanvasElement;
-    if (!this.element) this.element = elementId as HTMLCanvasElement;
+    this.element =
+      typeof elementId === 'string' && typeof document !== 'undefined'
+        ? (document.getElementById(elementId as string) as HTMLCanvasElement)
+        : (elementId as HTMLCanvasElement);
 
     // Verify backend and create context
     this.backend = backend;
     if (this.backend === Renderer.Backends.CANVAS) {
-      if (!this.element.getContext) {
+      if (!this.element?.getContext) {
         throw new RuntimeError('BadElement', `Can't get canvas context from element: ${elementId}`);
       }
       this.ctx = Renderer.bolsterCanvasContext(this.element.getContext('2d'));
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -146,8 +146,7 @@ export class Renderer {
   }
 
   constructor(elementId: string | HTMLElement, backend: number) {
-    this.elementId = elementId;
-    if (this.elementId === undefined) {
+    if (!elementId) {
       throw new RuntimeError('BadArgument', 'Invalid id for renderer.');
     }
 
@@ -170,9 +169,6 @@ export class Renderer {
 
   resize(width: number, height: number): this {
     if (this.backend === Renderer.Backends.CANVAS) {
-      if (!this.element.getContext) {
-        throw new RuntimeError('BadElement', `Can't get canvas context from element: ${this.elementId}`);
-      }
       [width, height] = CanvasContext.SanitizeCanvasDims(width, height);
 
       const devicePixelRatio = window.devicePixelRatio || 1;
--- a/demos/README.md
+++ b/demos/README.md
@@ -1,0 +1,5 @@
+# Demos
+
+This folder contains examples of ways you can use VexFlow.
+
+To run the demos, first build the VexFlow library by running `npm install` and then `grunt` from the main folder.
--- a/demos/node/README.md
+++ b/demos/node/README.md
@@ -1,0 +1,7 @@
+# Using VexFlow in Node JS
+
+You can use VexFlow in Node JS by calling `const Vex = require(...)` on the JS library.
+
+`node canvas.js > output.html` creates an HTML page containing the VexFlow output.
+
+`node svg.js > output.svg` creates a SVG image file containing the VexFlow output.
--- a/demos/node/canvas.js
+++ b/demos/node/canvas.js
@@ -1,0 +1,18 @@
+// node canvas.js > output.html
+
+const { createCanvas } = require('canvas');
+const Vex = require('../../build/vexflow-debug');
+const VF = Vex.Flow;
+
+const canvas = createCanvas(500, 500);
+
+const renderer = new VF.Renderer(canvas, VF.Renderer.Backends.CANVAS);
+const context = renderer.getContext();
+context.setFont('Arial', 10, '').setBackgroundFillStyle('#eed');
+
+const stave = new VF.Stave(10, 40, 400);
+stave.addClef('treble');
+stave.addTimeSignature('4/4');
+stave.setContext(context).draw();
+
+console.log(`<!DOCTYPE html><html><body><img src="${canvas.toDataURL()}"></body></html>`);
--- a/demos/node/svg.js
+++ b/demos/node/svg.js
@@ -1,0 +1,33 @@
+// node svg.js > output.svg
+
+const Vex = require('../../build/vexflow-debug');
+const { JSDOM } = require('jsdom');
+const fs = require('fs');
+
+const VF = Vex.Flow;
+
+const dom = new JSDOM('<!DOCTYPE html><html><body><div id="vf"></div><body></html>');
+
+global.document = dom.window.document;
+
+// Create an SVG renderer and attach it to the DIV element named "vf".
+const div = document.getElementById('vf');
+const renderer = new VF.Renderer(div, VF.Renderer.Backends.SVG);
+
+// Configure the rendering context.
+renderer.resize(500, 500);
+const context = renderer.getContext();
+context.setFont('Arial', 10, '').setBackgroundFillStyle('#eed');
+
+// Create a stave of width 400 at position 10, 40 on the canvas.
+const stave = new VF.Stave(10, 40, 400);
+
+// Add a clef and time signature.
+stave.addClef('treble').addTimeSignature('4/4');
+
+// Connect it to the rendering context and draw!
+stave.setContext(context).draw();
+
+const svg = div.innerHTML.replace('<svg ', '<svg xmlns="http://www.w3.org/2000/svg" ');
+
+console.log(svg);


0xfe:
@ronyeh can you review this please? (I think the `globalObject: this` popped up in a previous PR.)

rvilarl:
@0xfe @ronyeh 
globalObject definition was required  to allow the update to webpack 5.x. I was getting the error-self-is-not-defined.

I got the solution from:
https://stackoverflow.com/questions/64639839/typescript-webpack-library-generates-referenceerror-self-is-not-defined

0xfe:
Thanks @rvilarl -- so I guess we didn't notice it for our headless tests because we're not testing with UMD.

rvilarl:
I guess you are right @0xfe.

src/renderer.js
```diff
@@ -127,19 +127,18 @@ export class Renderer {
...
-    if (!this.element) this.element = elementId;
+    this.element =
+      typeof elementId === 'string' && typeof document !== 'undefined' ? document.getElementById(elementId) : elementId;
 
     this.ctx = null;
     this.paper = null;
     this.backend = backend;
     if (this.backend === Renderer.Backends.CANVAS) {
-      if (!this.element.getContext) {
+      if (!(this.element || 0).getContext) {
```
rvilarl, line 141: why is ||0 required?

src/renderer.js
```diff
@@ -127,19 +127,18 @@ export class Renderer {
...
-    this.elementId = elementId;
-    if (!this.elementId) {
+    if (!elementId) {
       throw new RuntimeError('BadArgument', 'Invalid id for renderer.');
     }
 
-    this.element = document.getElementById(elementId);
-    if (!this.element) this.element = elementId;
+    this.element =
+      typeof elementId === 'string' && typeof document !== 'undefined' ? document.getElementById(elementId) : elementId;
```
rvilarl, line 135: Can document be undefined? When?

ronyeh: We can add a console.log of document to see if it ever prints out undefined. When we do the visual diffs via node JS on the command line, is document defined? Are there any other instances where we would be outside of a browser context?

ronyeh:
Sorry I am out of the loop on this. @aleen42 can you send us an example page where this PR fix helps? That would help me understand which file you are requiring, and the use case.

rvilarl:
> Sorry I am out of the loop on this. @aleen42 can you send us an example page where this PR fix helps? That would help me understand which file you are requiring, and the use case.

@ronyeh this seems to be related to headless machine testing in https://github.com/0xfe/vexflow/wiki/Visual-Regression-Tests the command mentioned is:
`$ xvfb-run -a npm test`

src/renderer.js
aleen42: Apparently, to avoid NPE of `this.element` after remove the following code, which has the same meaning:

```js
if (!this.element) this.element = elementId;
```

src/renderer.js
aleen42: Exactly, you can see the commit message, where I have noticed that developers may hope to use this library outside a browser context, like Node, where we can simulate Canvas, and where without document definition.

aleen42:
A simple example with `node-canvas` running under Node context:

```js
const Vex = require('vexflow/releases/vexflow-debug');
const VF = Vex.Flow;

const { createCanvas, loadImage } = require('canvas')
const canvas = createCanvas(200, 200);

// Pass canvas element directly
const renderer = new VF.Renderer(canvas, Renderer.Backends.CANVAS);
// Configure the rendering context.
renderer.resize(500, 500);
const context = renderer.getContext();
context.setFont("Arial", 10, "").setBackgroundFillStyle("#eed");

// Create a stave of width 400 at position 10, 40 on the canvas.
const stave = new VF.Stave(10, 40, 400);

// Add a clef and time signature.
stave.addClef("treble").addTimeSignature("4/4");

// Connect it to the rendering context and draw!
stave.setContext(context).draw();
```

And then we can get the image data:

```js
console.log(canvas.toDataURL());
```

If hoping to define Vexflow with UMD styles, please avoid using `this` directly as `globalThis` under the Webpack configuration, and avoid using `document` at the same time.

ronyeh:
Thanks for your example. Is the use case to generate JPG or PNG files outside of the browser context?

It definitely makes sense to support such use cases.

VexFlow does something similar (search for "canvas.toDataURL" in the codebase) but I think we currently use JSDOM so we never ran into the missing `document` and `this` issues.

aleen42:
@ronyeh I know it, and hope it can support especially when we can simulate DOM under the Node context.

aleen42:
Another example of `SvgContext` under Node environment:

```js
const Vex = require('vexflow/releases/vexflow-debug');
const VF = Vex.Flow;

const {JSDOM} = require('jsdom');
// Register simulated document
Object.assign(self, {document : new JSDOM('<!DOCTYPE html>')});


// Pass canvas element directly
const div = document.createElement('div');
const renderer = new VF.Renderer(, Renderer.Backends.SVG);
const context = renderer.getContext();
context.setFont('Arial', 10, '').setBackgroundFillStyle('#eed');

// Create a stave of width 400 at position 10, 40 on the canvas.
const stave = new VF.Stave(10, 40, 400);

// Add a clef and time signature.
stave.addClef('treble').addTimeSignature('4/4');

// Connect it to the rendering context and draw!
stave.setContext(context).draw();
```

And then we easily generate a SVG file under Node with:

```js
console.log(div.innerHTML);
```

src/renderer.js
rvilarl: I suggest to leave it unchanged.

src/renderer.js
rvilarl: This is fine then.

rvilarl:
@0xfe I see nothing against merging this PR. Only the minor comment to revert the '|| 0'

ronyeh:
Can we add a small test example that uses this PR? Like maybe under
tests/other/no_document.js or something? It can be one of the examples
provided where we just import the library in the node context and console
log the SVG string of a simple staff with one whole note.

src/renderer.ts
```diff
@@ -137,19 +137,20 @@ export class Renderer {
...
+    this.element =
+      typeof elementId === 'string' && typeof document !== 'undefined'
+        ? (document.getElementById(elementId as string) as HTMLCanvasElement)
+        : (elementId as HTMLCanvasElement);
 
     // Verify backend and create context
     this.backend = backend;
     if (this.backend === Renderer.Backends.CANVAS) {
-      if (!this.element.getContext) {
+      if (!(this.element || 0).getContext) {
```
rvilarl, line 153: @aleen42 please revert this change, this.element is not optional.

aleen42: There is an exception when getElementabyId rerurns null.

rvilarl: ok

ronyeh:
Sorry for my ignorance on this line of code. If this.element is null or
undefined, won't that expression return 0.getContext ? Is the intention to
skip this if clause?

aleen42:
Yup, exactly avoid exceptions when passed element is not existed. @ronyeh

src/renderer.ts
aleen42: @0xfe Please confirm whether need to revert this change.

0xfe: It's cryptic :-) Since this is typescript, you can do: `if (!this.element?.getContext) { .. }` which might be clearer, no?

aleen42: You're right

aleen42: I have amended and pushed it already.

Gruntfile.js
```diff
@@ -38,7 +38,7 @@ module.exports = (grunt) => {
         library: libraryName,
         libraryTarget: 'umd',
         libraryExport: 'default',
-        globalObject: 'this',
+        globalObject: `typeof window !== 'undefined' ? window : this`,
```
ronyeh, line 41: Why is this line necessary? In my PR #1072 I got the two examples to work without having to modify the gruntfile. 

From the webpack docs: "To make UMD build available on both browsers and Node.js, set output.globalObject option to 'this'."

https://webpack.js.org/configuration/output/#outputglobalobject

aleen42: It is wired, I have checked the history of `vexflow-debug.js` and found that 23c6ac0bbba1f907f0f7f3377a21147da1a0017f release as `typeof self !== 'undefined' ? self : this` while 0793189194ca216207b4ac1d2503156da24ebde7 revert to `window`.

It seems that a bug arisen between upgrading Webpack, you can see https://github.com/webpack/webpack/issues/6522.

aleen42: So currently I can revert this commit and re-release the folder to solve the problem.

0xfe:
Thanks @aleen42! @ronyeh I'll wait for your lgtm before I start my review.

Gruntfile.js
ronyeh: Hello! Do you have a small example which shows that this gruntfile change is necessary? An example which fails with the current gruntfile, and works better when we use your solution:

```
globalObject: `typeof window !== 'undefined' ? window : this`,
```

I confess that I am not a webpack and UMD expert. I am just going by the current webpack documentation: https://webpack.js.org/configuration/output/#outputglobalobject

The comment you reference is 3 years old. Could this have been solved by webpack in the meantime?

The two Node JS demos we currently have work fine without the gruntfile change. I'd like to see an example where it fails.

ronyeh (changes requested):

ronyeh:
@0xfe I looked over the PR diff and it seems fine. My only outstanding question is whether or not `globalObject` needs to be modified as suggested. I think we need an example which uses one of the JS import/require/script-tag methods that demonstrate that the current `globalObject: this` is insufficient.

Gruntfile.js
aleen42: You can reproduce like this:

```html
<script type="module" src="releases/vexflow-debug.js"></script>
```

```js
// releases/vexflow-debug.js
console.log(this); // => undefined
console.log(typeof window !== 'undefined' ? window : this); // => Window
```

aleen42: @ronyeh If you dig into the implementation of jQuery, you can find that they have adopted the style when exporting UMD style modules.

- REF: https://github.com/jquery/jquery/blob/a684e6ba836f7c553968d7d026ed7941e1a612d8/src/wrapper.js#L37

0xfe: Hi @aleen42 -- this change looks good overall. This is the only part I'm still confused about. I guess the question is: what breaks when you leave it as `this`? It seems like the demos work fine that way.

My suggestion would be to revert this line to keep it simple (unless, ofcourse, we can find a good reason not to.) :-)

ronyeh: I think I now understand. Let me repro the issue and put together an example.

ronyeh:
Here are my findings....

Normally we just include `vexflow-debug.js` in a classic script tag. 

However, if you set the `type=module` on that script tag, it fails with error:

```
vexflow-debug.js:14 Uncaught TypeError: Cannot set property 'Vex' of undefined
    at webpackUniversalModuleDefinition (vexflow-debug.js:14)
    at vexflow-debug.js:15
```

That's because our webpack header is trying to use `this` as the global object, but `this` is undefined with script `type=module`.

If we open the vexflow-debug.js file, we see this at the top:

```
(function webpackUniversalModuleDefinition(root, factory) {
    if(typeof exports === 'object' && typeof module === 'object')
        module.exports = factory();
    else if(typeof define === 'function' && define.amd)
        define([], factory);
    else if(typeof exports === 'object')
        exports["Vex"] = factory();
    else
        root["Vex"] = factory();
})(this, function() {
```

The **last line** passes in `this` into the `root` param. `this` is undefined, so `root["Vex"] = factory();` is an error.

If we change that last line:

```
})(this, function() {

==>

})(typeof window !== 'undefined' ? window : this, function() {
```

... then script type module will work fine.


This PR's change to the `Gruntfile.js` addresses this problem, because it will pass in `window` if it is available.

However, I think we could do a **tiny** bit better.

I think that Gruntfile line should look like:

```
globalObject: `typeof window !== 'undefined' ? window : typeof globalThis !== 'undefined' ? globalThis : this`,
```

In my node example at: https://github.com/ronyeh/vexflow/blob/examples/demos/modules/index.node.js

... `window` and `this` are both not available. However, `globalThis` works.

Thanks @aleen42 for pointing out this issue.

Can you change the Gruntfile's webpack option to:

```
globalObject: `typeof window !== 'undefined' ? window : typeof globalThis !== 'undefined' ? globalThis : this`,
```

Otherwise, this PR looks good to go.

Gruntfile.js
ronyeh: I've tested the above on my branch. It does indeed fix a error with script type=module. Can we actually modify this line to:

```
globalObject: `typeof window !== 'undefined' ? window : typeof globalThis !== 'undefined' ? globalThis : this`,
```

This would also fix situations where both window and this are unavailable. Thanks!

aleen42:
@ronyeh If using `globalThis`, it means that Vexflow has no compatibility on older browsers, are you ensure that?

ronyeh:
Ok, don't worry about globalThis for your PR. If you can fix the conflict,
Mohit can review and merge your PR.

We will do more research in the next PR to see if globalThis can be used.

Thanks.

rvilarl:
@ronyeh is there anything pending here, if not I would suggest @0xfe to merge it.The PR is open now for a month.

ronyeh:
If no conflicts, it should be safe to merge. It comes with Node JS demos to show how it works.

0xfe:
Okay, merging. Thanks everyone.

PR #1046, opened 2021-07-02
Flow.ts: Use shorthand property names. Sort imports and keys.

ronyeh:
While working on migrating test pages, I cleaned up flow.ts a bit. I extracted that change into this PR so it's easier to review.

In addition to sorting the imports and keys, I removed two entries that don't seem to be used anywhere on the test pages (and they are also not in version 3.0.9).

`PetalumaScriptTextMetrics`
`RobotoSlabTextMetrics`

I believe those two are used exclusively by `textfont.ts`, so they should be local to that file.

### DIFF
--- a/src/flow.ts
+++ b/src/flow.ts
@@ -1,162 +1,158 @@
-import { Element } from './element';
-import { Fraction } from './fraction';
-import { Renderer } from './renderer';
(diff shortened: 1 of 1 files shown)


0xfe:
Thanks for cleaning this up.

Issue #1004, opened 2021-05-25
Comment Style /** ... */ vs //


ronyeh:
This issue can be for future discussion, post TS migration.

I've noticed that we are not consistent with the way methods are commented. Many times the /** */ vs // styles are intermixed within the same file. :-) Should we standardize on the way we document methods? 

For example in music.ts:

```
  /* Returns the interval of a note, given a diatonic scale.
   *
   * E.g., Given the scale C, and the note E, returns M3
   */
  getIntervalBetween(note1: number, note2: number, direction: number = 1): number {
    if (direction !== 1 && direction !== -1) {
      throw new Vex.RERR('BadArguments', `Invalid direction: ${direction}`);
    }
    ...
    return difference;
  }

  // Create a scale map that represents the pitch state for a
  // `keySignature`. For example, passing a `G` to `keySignature` would
  // return a scale map with every note naturalized except for `F` which
  // has an `F#` state.
  createScaleMap(keySignature: string): Record<string, string> {
    ...

```

In note.ts:

```
  // Get and set spacing to the left and right of the notes.
  getLeftDisplacedHeadPx(): number {
    return this.leftDisplacedHeadPx;
  }
  ...

  /** Returns true if this note has no duration (e.g., bar notes, spacers, etc.) */
  shouldIgnoreTicks(): boolean {
    return this.ignore_ticks;
  }

```

ronyeh:
Even though I personally prefer the look of multiline comments starting with // ...

It looks like Microsoft is suggesting this standard:

https://tsdoc.org/play
https://github.com/microsoft/tsdoc/
https://github.com/microsoft/tsdoc/tree/master/tsdoc

**TypeDoc** is a popular tool that seems to mostly follow the TSDoc standard proposal:
https://typedoc.org/guides/doccomments/

So I guess it's a good idea to move toward the /** */ style of multi-line comments.

At some point we could use TypeDoc to generate documentation for the VexFlow API. We could even keep docco around to preserve the literate style documentation as an alternate way for developers to learn VexFlow.

mscuthbert:
I don't think it's a matter of multiline vs single line.  Use two stars if the lines are intended to appear in the documentation, and one star or `//` if it is for developer reference only.  Sometimes there are multiline comments intended for devs only, like explaining how a complex algorithm works.  Maybe they should all be `//` lines, maybe one star, but not two-star.

ronyeh:
[...]

PR #1047, opened 2021-07-03
doc/voice (and VoiceGroup removed)

rvilarl:
- VoiceGroup Removed 
- Voice Documented

I have taken the following approach (to be agreed):
- Exclusion of protected and private members in the ouput from typedoc.
- Usage of imperative rather that third person in comments.
- Protection of methods only used within the class.
- Simple Header: product, copyright and license (the license statement on each file help the automatic sw clearance tools) 

@0xfe @AaronDavidNewman @mscuthbert @ronyeh 
I have tried to address the concerns in #1004. Once we agree with this PR we can address the next ones in a similar well.


src/voice.ts
```diff
@@ -28,12 +22,15 @@ export enum VoiceMode {
   FULL = 3,
 }
 
+/**
+ * Voice class is mainly a container
+ * object to group `Tickables` for formatting.
+ */
```
ronyeh, line 28: How about:

\`Voice\` is mainly a container object to group \`Tickables\` for formatting.

src/voice.ts
```diff
@@ -44,12 +41,14 @@ export class Voice extends Element {
...
-  // Modes allow the addition of ticks in three different ways:
-  //
-  // STRICT: This is the default. Ticks must fill the voice.
-  // SOFT:   Ticks can be added without restrictions.
-  // FULL:   Ticks do not need to fill the voice, but can't exceed the maximum
-  //         tick length.
+  /**
+   * Modes allow the addition of ticks in three different ways:
+   *
+   * STRICT: This is the default. Ticks must fill the voice.<br>
```
ronyeh, line 47: Is the \<br\> tag the only way to force a line break? Is the alternative to have an extra empty line between the STRICT line and SOFT line?

Personally I would prefer having as few HTML tags in the comments as possible, but I could live with the br tags if there is no good alternative.

What about putting the three different ways in a bulleted list?

```
Modes allow the addition of ticks in three different ways:

- STRICT: This is the default. Ticks must fill the voice.
- SOFT:   Ticks can be added without restrictions.
- FULL:   Ticks do not need to fill the voice, but can't exceed the maximum tick length.
```

src/voice.ts
```diff
@@ -99,59 +98,63 @@ export class Voice extends Element {
...
   }
 
-  // Get the total ticks used in the voice by all the tickables
+  /** Get the total ticks used in the voice by all the tickables */
   getTicksUsed(): Fraction {
     return this.ticksUsed;
   }
 
-  // Get the largest width of all the tickables
+  /** Get the largest width of all the tickables */
```
ronyeh, line 111: Super duper minor, but I would prefer that sentence comments end with a period.

src/voice.ts
```diff
@@ -205,26 +192,30 @@ export class Voice extends Element {
...
     this.options.softmaxFactor = factor;
     return this;
   }
 
-  // Calculate the sum of the exponents of all the ticks in this voice to use as the denominator
-  // of softmax.
-  reCalculateExpTicksUsed(): number {
+  /**
+   * Calculate the sum of the exponents of all the ticks in this voice to use as the denominator
+   * of softmax.
```
ronyeh, line 209: Can this all fit on one line? :-)

Overall, this style of commenting looks good!

Over time, I imagine we will add the `@param ` as needed, to explain any params that might not be obvious.

rvilarl: @param is already used with `mode`

src/voice.ts
rvilarl: done

src/voice.ts
rvilarl: Done

src/voice.ts
rvilarl: Done

rvilarl:
@ronyeh Ready

rvilarl:
@0xfe I changed all the comments from @ronyeh, anything else from yourself?

0xfe:
This looks good to me.

PR #1048, opened 2021-07-03
Replace VF.Test.TIME4_4 with VF.TIME4_4.

ronyeh:
VF.Test.TIME4_4 is now undefined due to our TypeScript migration.

`npm run test:reference` results in 0 diffs from current master.


rvilarl:
Looks good

0xfe:
Looks good.

PR #1052, opened 2021-07-04
Fix/ledger lines (double width by default)

rvilarl:


src/stavenote.ts
```diff
@@ -1119,7 +1119,12 @@ export class StaveNote extends StemmableNote {
       ctx.stroke();
     };
 
-    const style = { ...stave.getStyle(), ...(this.getLedgerLineStyle() || {}) };
+    const style = {
+      ...stave.getStyle(),
+      ...(this.getLedgerLineStyle() || {
```
0xfe, line 1124: Can we update `getLedgerLineStyle` instead?

rvilarl: i@0xfe If we do it inside `getLedgerLine` we will have to move exactly the same code in the function to make sure that it defaults to double the "current" `stave.getStyle().lineWidth`, otherwise it will be calculated once and that's it. 
Do you want me to do that?
```
getLedgerLineStyle(): ElementStyle {
    const stave = this.checkStave();
    const style = {
      ...stave.getStyle(),
      ...(this.getLedgerLineStyle() || {
        lineWidth: stave.getStyle().lineWidth * Flow.LEDGER_LINE_THICKNESS_MULTIPLIER,
      }),
    };
    return style;
  }
```

0xfe: Thanks. Actually, I think it's better to initialize `this.ledgerLineStyle` in the constructor to `{ lineWidth: stave.getStyle().lineWidth * 2.0 }` and remove `Flow.LEDGER_LINE_THICKNESS_MULTIPLIER`.

Then this part does not have to change.

rvilarl: @0xfe stave is not defined in the constructor. I suggest to initialise simply to 2 as default.

rvilarl: Done with simple initialisation to 2.0. Now the changes are really minor.

rvilarl:
@0xfe ready!

0xfe:
Thanks!

mscuthbert:
Thanks for taking this over @rvilarl !

tommadams:
It's funny how tastes differ! I honestly though the double width ledger lines were a bug and went digging around until I found this PR.

They look a strange to my eye primarily because the ledger & stave lines are grey, not black. I wonder if the intent of the "ledger lines should be twice as thick" advice is that they should have twice as much _emphasis_ as stave lines. In a black & white printed medium, the only way to achieve this is to double the width.

Given that we're rendering to true colour displays, I was wondering if we could make the ledger lines darker instead of wider:

So instead of:
wide ledger lines

We would have:
darker ledger lines

Another way to think about it is that the staff lines could actually be fully black but less than a pixel wide. Then anti-aliasing would make them appear grey. For example: if the staff line width was 0.5px, they'd render 1px wide on screen but roughly 50% black. Doubling that width for ledger lines would make them still 1px wide but fully black.

I'm happy if folks disagree, I can just change the line styling in my app. I just wanted to raise the suggestion.

ronyeh:
Yeah I agree that **emphasis** is important so that the eye can sight read faster (but I'd have to ask a pro sight reader for their feedback. I'm terrible.).

I also think the thicker but gray lines look a little out of place, especially when zoomed in. I like the idea of the ledger lines being darker but more similar in thickness. How does it look if it's 1.5x thickness and a bit darker than the stave lines?

It also depends on your target audience. Lemme tell ya that as we age, we will all enjoy bigger fonts and higher contrast. Thin gray lines and tiny staves aren't fun.

tommadams:
Here are ledger lines with widths 1.0px, 1.2px 1.5px and colour `#444444` (compared to the stave line colour `#999999`).

1.0px
1.2px
1.5px

With non-integer line widths, the ledger lines start to look a little blurry. That said, I quite like 1.2px.

Another issue is that all the devices I own have high density displays, so I don't know how these look on a regular display :(

If anyone has a regular display and can test the change out, I've pushed it to [https://github.com/tommadams/vexflow/tree/ledgerLine](https://github.com/tommadams/vexflow/tree/ledgerLine)

ronyeh:
I simulate poor eyesight / bad display by hitting CMD/CTRL minus to shrink the image.

I'd probably vote for something in between 1.2 to 1.5... but luckily this field is easily customizable by developers.

Would be interested to hear others' opinions.

mscuthbert:
I think 1.5 looks fine if people think 2.0 is too much -- a nice way of getting the effect of 2x when the normal staff is gray.  But 1.0 just looks totally wrong to me from reading other scores.

Never change the normal gray staff -- it's so nice.

tommadams:
I'm a big fan of the grey staff lines too, I just think that having double-width ledger lines in the same light shade looks a little odd, especially considering they're four pixels wide on high density displays. At that width, the fact that the ledger lines are around 50% grey really sticks out to me.

I took a look at some manuscripts from my collection, published across multiple genres, countries and decades and the emphasis used for ledger lines can vary rather noticeably.

The ledgers lines from this manuscript are 1.5x thicker or less than the staff lines.
thin

While these from a different manuscript are monsters!
thick

PR #1055, opened 2021-07-04
Calculate preCalculateMinWidth with enough (but not too much) padding (Issue 826)

AaronDavidNewman:
The method is to calculate the min width padding using 3 methods, and take the largest value.

- For each unaligned beat, add 10 px of padding
- Multiple normalized (0-1) stddev of width * 10px, per tick context.
- Multiple normalized (0-1) stddev of duration * 10px, per tick cpntext.

If measures are all quarter notes of equal width, there are no unaligned beats, stddev is 0 for width and duration, and no extra padding is needed, since softmax will calculate same width for all.

If a tickable is not aligned with another tickable in a different voice, the formatter will need to place the later tickable to the right of the earlier tickable by as much as one note head, hence it may need extra space.

If a tickable is wider than notes of smaller duration (due to accidentals, for instance), the formatter will need to adjust the notes to get the ideal distance since softmax function will short-change the wider note.  The same thing can happen when notes are the same width, and the durations are very different.  This is more likely when there are many different durations/widths, hence the stddev as an approximation to note-width entropy.

This PR only adjusts the width calculation and I will paste a few examples of before/after.  Sometimes the 'before' is a little tighter, but the 'after' never overflows.

The next PR will fix factory and system to allow formatter's pre-calculated width to be used with factory/easyScore, and rewrite a whole bunch of test cases that currently use hard-coded widths.


AaronDavidNewman:
`Formatter.accidentalJustification`
Before
image
After
image

`Formatter.longMeasureProblems`
Before
image
after 
image


 `Formatter.unalignedNoteDurations`
Before
image
after:
image

`Vex.Flow.Test.Accidental.formatAccidentalSpaces`
Before
image
After
image
`Formatter.multiStaves`
Before:
image
After:
image

src/stave.ts
```diff
@@ -68,6 +68,10 @@ export class Stave extends Element {
     const musicFont = Flow.DEFAULT_FONT_STACK[0];
     return musicFont.lookupMetric('stave.padding') + musicFont.lookupMetric('stave.endPaddingMax');
   }
+  static get leftPadding(): number {
```
0xfe, line 71: This is unused, can we remove?

src/formatter.ts
```diff
@@ -427,20 +436,50 @@ export class Formatter {
...
+
+    const davg = dsum / durations.length;
+    const dvar = durations.map((ll) => Math.pow(ll - davg, 2)).reduce((a, b) => a + b);
+    const dpads = Math.pow(dvar / durations.length, 0.5) / davg;
+
+    // Find max of 3 methods and use that
+    // const padmax = Math.max(dpads, wpads) * contextList.length * unalignedPadding;
+    const padmax = Math.max(dpads, wpads) * contextList.length * unalignedPadding;
+    const unalignedPad = unalignedPadding * unalignedCtxCount;
+    L('wpad/dpad/upad:', wpads, dpads, unalignedPad);
```
0xfe, line 473: Can we remove these log lines?

src/formatter.ts
```diff
@@ -427,20 +436,50 @@ export class Formatter {
...
-      })
-      .reduce((a: number, b: number) => a + b, 0);
+    contextList.forEach((tick) => {
+      const context = contextMap[tick];
+      context.preFormat();
+      if (context.getTickables().length < voices.length) {
+        unalignedCtxCount += 1;
+      }
+      const width = context.getWidth();
+      const duration = context.getMaxTicks().value();
```
0xfe, line 449: Unused variable.

src/formatter.ts
```diff
@@ -413,6 +413,15 @@ export class Formatter {
 
   // Calculate the minimum width required to align and format `voices`.
   preCalculateMinTotalWidth(voices: Voice[]): number {
+    const unalignedPadding = Flow.DEFAULT_FONT_STACK[0].lookupMetric('stave.unalignedNotePadding');
```
0xfe, line 416: I think we should use the current font stack, no?

0xfe (commented):
Thanks Aaron -- added some early thoughts. Can you add some comments in the sections around the calculations to say what you're doing at a high level?

Also, how does this affect the other tests? Any corner cases to be aware of?

src/formatter.ts
AaronDavidNewman: um..yeah how do you do that?  I thought DEFAULT_FONT_STACK was the current one.

AaronDavidNewman:
>> Also, how does this affect the other tests? Any corner cases to be aware of?

Only the tests with screenshots attached are affected

AaronDavidNewman:
@0xfe I think this is ready for merge, unless you have more comments on current font stack.  I thought DEFAULT_FONT_STACK is the current font stack.  That's the way it's treated in the tests, and that's how I treated it in chordSymbol module.

src/formatter.ts
0xfe: Okay, yes, this is fine. I was thinking about element overrides with `setMusicFont`. This is good.

0xfe:
Great -- thanks for adding the comments, looks great!

Issue #221, opened 2014-05-09
Improve articulations


Silverwolf90:
While looking at the [lilypond glyphs](http://lilypond.org/doc/v2.17/Documentation/notation/the-feta-font#script-glyphs) I noticed a few things.
1. We're using the downwards facing pedal toe glyph for marcato instead of the prettier available glyph
   - Additionally, marcato should be flipped when positioned on the bottom.
   - There is a downwards facing glyph for marcato
2. Staccatissimo has a downwards glyph
3. We're missing the portato articulation

So it becomes clear that having an up and down glyph is common for many articulations.
- Fermata
  - alternate fermatas also available in font
- Marcato
- Staccatissimo
- Portato (in font, not VF)
- Pedal heel (in font, not VF)
- Pedal toe (in font, not VF)

`Articulation` should automatically figure out which glyph to use, as opposed to manually having to decide the direction like we do with fermatas.

PR #1056, opened 2021-07-05
Add missing setters to maintain compatibility with CanvasRenderingContext2D API

ronyeh:
Many test cases set the `fillStyle`, `strokeStyle`, and `font` directly via setters. However, `SVGContext` and `CanvasContext` did not actually implement those setters.

For example, if we write code like this:

```
      ctx.fillStyle = '#F21';
      ctx.strokeStyle = '#F21';
      ctx.font = '10pt Arial';
```

It will only affect the CanvasContext, since the web browser's native `CanvasRenderingContext2D` implements those setters.

I added the setters so that now it also works with `SVGContext`. See below for before/after:

CanvasAPI

One question: We never actually create any CanvasContext objects. There are no test cases that set `Renderer.USE_CANVAS_PROXY` to true. What types of environments require us to use the CanvasContext() object?

```
    if (Renderer.USE_CANVAS_PROXY) {
      return new CanvasContext(ctx);
    }
```

Anyways, this PR is ready for code review!

Surprisingly there were no visual diffs with the test cases (or perhaps the `#221` color was close enough to the previous color that it was not detectable). However, I did verify that if I set the colors to #F00, that there is the expected visual diff on the test cases that use SVGContext (e.g., the black/red GIF above).

### DIFF
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -98,7 +98,7 @@ export class CanvasContext implements RenderContext {
     this.vexFlowCanvasContext.fillRect(0, 0, this.canvas.width, this.canvas.height);
     this.vexFlowCanvasContext.fillStyle = oldFillStyle;
     */
-    this.background_fillStyle = style; // TODO: Is this a BUG? The field is never referenced anywhere else.
+    this.background_fillStyle = style;
     return this;
   }
 
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -1,13 +1,11 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
-// Mohit Muthanna <mohit@muthanna.com>
-//
-// A rendering context for the Canvas backend (CanvasRenderingContext2D).
-//
-// Copyright Mohit Cheppudira 2010
-
+// MIT License
 import { RenderContext } from './types/common';
 import { warn } from './util';
 
+/**
+ * A rendering context for the Canvas backend (CanvasRenderingContext2D).
+ */
 export class CanvasContext implements RenderContext {
   vexFlowCanvasContext: CanvasRenderingContext2D;
   canvas: HTMLCanvasElement | { width: number; height: number };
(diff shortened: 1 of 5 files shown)


src/renderer.ts
```diff
@@ -82,7 +86,6 @@ export class Renderer {
       'setShadowBlur',
       'setLineWidth',
       'setLineCap',
-      'setLineDash',
```
ronyeh, line 85: `setLineDash` already exists on `CanvasRenderingContext2D`, so this code does nothing. Removing it doesn't affect the test output.

https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setLineDash

src/canvascontext.ts
```diff
@@ -94,14 +92,13 @@ export class CanvasContext implements RenderContext {
...
     /*
-    // The CanvasContext version only sets a field which is never referenced anywhere else.
     // Should it fill the entire canvas rect? For example:
     const oldFillStyle = this.vexFlowCanvasContext.fillStyle;
     this.vexFlowCanvasContext.fillStyle = style;
     this.vexFlowCanvasContext.fillRect(0, 0, this.canvas.width, this.canvas.height);
     this.vexFlowCanvasContext.fillStyle = oldFillStyle;
     */
-    this.background_fillStyle = style; // BUG? It is never referenced anywhere else.
+    this.background_fillStyle = style; // TODO: Is this a BUG? The field is never referenced anywhere else.
```
0xfe, line 101: This comment can be removed.

src/renderer.ts
0xfe: Thanks, sounds good.

0xfe (commented):
Thanks Ron. Just a minor comment.

ronyeh:
So is the this.background_fillStyle field intentional? It isn't referenced
anywhere else in the code base.

0xfe:
Thanks for the change. Let's deal with `this.background_fillStyle` separately -- I'm not sure if it's needed anymore.

Issue #1003, opened 2021-05-25
undefined vs null: last_bend should be compared against undefined


ronyeh:
https://github.com/0xfe/vexflow/blob/3e13b92cc7437d5c6ac2b23fa2d056bb1f0f1564/src/bend.ts#L249-L292

In the first line we set last_bend to `undefined`, but later on we compare it against `null`:

https://github.com/0xfe/vexflow/blob/3e13b92cc7437d5c6ac2b23fa2d056bb1f0f1564/src/bend.ts#L276
```
        if (last_bend === null) {    // <== this is always false
```

Also, we check 
https://github.com/0xfe/vexflow/blob/3e13b92cc7437d5c6ac2b23fa2d056bb1f0f1564/src/bend.ts#L290
```
    if (!last_bend || last_bend.x == undefined) {  // use strict equality, or else null will also pass this test
```

I think for checks against `undefined` we should generally use triple equal === for more strict comparisons (especially since we are moving away from usage of `null`).

rvilarl:
Agree with your comments

0xfe:
Thanks for catching this Ron, do the bend tests catch this?

0xfe:
i.e., do the visual tests show a difference? (if not, we prob should add a test too.)

rvilarl:
@ronyeh will you do a PR on that?

ronyeh:
Sure, I can do a PR.

rvilarl:
@ronyeh I believe it is better to use !last_bend so as to have backwards compatibility with `null`. I have included the change in #1058

ronyeh:
Hi Rodrigo,

The benefit of !x is that it is very concise. However, !x is true if x is
0, null, false, or undefined. This kind of check for a falsy value can be
buggy if we are not careful.

I think we converted the entire codebase to use undefined if a property was
optional, correct? In the past, vexflow often used null instead.

Here is one case where the not operator would cause a bug.


*export class Modifier extends Element {*
*  protected index?: number;*

So by default, index is undefined. However, the number 0 is a valid value.
So* !mod.index* will return true if the index is undefined, and also true
if the index is 0.

Since we use undefined widely in the vexflow codebase, I think we should
explicitly check for undefined.

This might be overkill, but we could add a utility function
isUndefined()... 😄

...it turns out there is a NPM package for this (I am not recommending that
we npm install it haha). That package has three contributors!? The current
definition is:

[...]

PR #1058, opened 2021-07-06
documentation update

rvilarl:
Accidental, Annotation, Articulation, util, ...

### DIFF
--- a/src/bend.ts
+++ b/src/bend.ts
@@ -1,8 +1,5 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
-//
-// ## Description
-//
-// This file implements tablature bends.
+// MIT License
 
 import { RuntimeError } from './util';
 import { Flow } from './flow';
@@ -24,40 +21,7 @@ export interface BendRenderOptions {
   line_style: string;
 }
 
-/**
-   @param text Text for bend ("Full", "Half", etc.) (DEPRECATED)
-   @param release If true, render a release. (DEPRECATED)
-   @param phrase If set, ignore "text" and "release", and use the more
-                 sophisticated phrase specified.
-
-   Example of a phrase:
-
-     [{
-       type: UP,
-       text: "whole"
-       width: 8;
-     },
-     {
-       type: DOWN,
-       text: "whole"
-       width: 8;
-     },
-     {
-       type: UP,
-       text: "half"
-       width: 8;
-     },
-     {
-       type: UP,
-       text: "whole"
-       width: 8;
-     },
-     {
-       type: DOWN,
-       text: "1 1/2"
-       width: 8;
-     }]
- */
+/** Bend implements tablature bends. */
 export class Bend extends Modifier {
   protected text: string;
   protected release: boolean;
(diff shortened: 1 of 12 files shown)


src/bend.ts
```diff
@@ -270,7 +271,7 @@ export class Bend extends Modifier {
           renderRelease(start.x, start.y, last_drawn_width, bend_height);
         }
 
-        if (last_bend === null) {
+        if (!last_bend) {
```
ronyeh, line 274: @rvilarl @0xfe I think this is fine for now, since it fixes the bug at least.

However, this line and lines above that start with

`if (last_bend`

almost make me think last_bend is a boolean type. Maybe we should actually have isUndefined() and isNotUndefined() utility functions??? 😬

rvilarl: @ronyeh I answered in 
https://github.com/0xfe/vexflow/issues/1003#issuecomment-876529023
what do you think?

ronyeh:
!last_bend is fine with me. Do we have an explicit typing for the local
variable declaration, like let last_bend: Bend;

I'm on mobile email so I can't really inspect the code.

We just have to be careful if we ever encounter a optional number (like in
modifier) or optional boolean that we need to compare against undefined
explicitly. (Or use a utility function.)

rvilarl:
> !last_bend is fine with me. Do we have an explicit typing for the local variable declaration, like let last_bend: Bend; I'm on mobile email so I can't really inspect the code. 

No we are not explicit typing local variables. It is getting the type `BendPhrase` implicitly from `this.phrase[i]`.

>We just have to be careful if we ever encounter a optional number (like in modifier) or optional boolean that we need to compare against undefined explicitly. (Or use a utility function.)

yes, you are right.

src/bend.ts
```diff
@@ -1,15 +1,12 @@
...
-// This file implements tablature bends.
+// MIT License
 
 import { RuntimeError } from './util';
 import { Flow } from './flow';
 import { Modifier } from './modifier';
 import { ModifierContextState } from './modifiercontext';
 
-export interface BendPhrase {
+interface BendPhrase {
```
0xfe, line 9: This should be exported, it's part of the API (in constructor.)

src/tables.ts
```diff
@@ -434,7 +434,7 @@ export const Tables = {
       'B##': { index: 6, int_val: 13, accidental: '##' },
       BB: { index: 6, int_val: 10, accidental: 'b' },
       BBB: { index: 6, int_val: 9, accidental: 'bb' },
-      R: { index: 6, int_val: 9, rest: true }, // Rest
+      R: { index: 6, rest: true }, // Rest
```
0xfe, line 437: Are you sure this doesn't need an `Int_val`?

0xfe (commented):
This has both doc and code updates, and not clear to me if this affects output. Any regressions in the visual tests?

src/tables.ts
rvilarl: A rest I believe has no pitch. This is also the way to identify that it is not a note like the X.

rvilarl:
> This has both doc and code updates, and not clear to me if this affects output. Any regressions in the visual tests?

No changes in visual tests.

rvilarl:
I will rebase the changes in master and re-check that there is no change. :)

src/bend.ts
rvilarl: done

rvilarl:
@0xfe review comments done, rebased to latest changes in master and no visual changes.

src/tables.ts
rvilarl: `Music.getNoteValue` must throw an error if it is closed with `R`. There is even a test for that. This made me realised that rests have no value.

0xfe:
Great, thanks!

PR #1063, opened 2021-07-11
use formatter.preCalculateMinWidth in system/factory

AaronDavidNewman:
Possibly final issue with bug 846. 
If system is created with autoWidth set to true, it will calculate use formatter.preCalculateMinWidth and justify all voices.
autoWIdth flag is set by default if no width is provided, and justification is not specifically disabled with a flag.  I'll paste before/after.

### DIFF
--- a/src/stave.ts
+++ b/src/stave.ts
@@ -68,6 +68,11 @@ export class Stave extends Element {
     const musicFont = Flow.DEFAULT_FONT_STACK[0];
     return musicFont.lookupMetric('stave.padding') + musicFont.lookupMetric('stave.endPaddingMax');
   }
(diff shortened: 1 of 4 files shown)


AaronDavidNewman:
2 test cases in formatter tests and a few in easy_score are affected by this change.  Other cases either create draw staves directly or provide explicit width.

Before (specific width):
image
AFter:
image

Before:
image
After:
image

Easy score:
Before:
image

After:
image

Before:
image

After.  It seems like there might be an issue here mixing the western and micro-tonal accidentals, not quite sure:
image

0xfe:
This is awesome! Thanks! :-)

PR #1064, opened 2021-07-11
Fix/OpenSheetMusicDisplay

rvilarl:
@0xfe I have built the version 1.0.0 of OpenSheetMusicDisplay using `npm link ../vexflow/` and as I said @types/vexflow is no longer necessary. However OSMD was complaining about two topics:
- MakeException returning a private type. Here I decided to simply use `RuntimeError` and `log` if extra data is required.
- One unused variable (`num_shitL` already identified in comments). Here I decided simply to delete it.
- The set accessors in an interface in common are also not accepted.

No changes on visual regressions, all tests passed. :)

### DIFF
--- a/src/easyscore.ts
+++ b/src/easyscore.ts
@@ -5,7 +5,7 @@
 
 /* eslint max-classes-per-file: "off" */
 
-import { RuntimeError, MakeException, log } from './util';
+import { RuntimeError, log } from './util';
 import { StaveNote } from './stavenote';
 import { Match, Parser, Result, Rule, RuleFunction } from './parser';
 import { Articulation } from './articulation';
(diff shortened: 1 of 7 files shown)


rvilarl:
@sschmidTU with this PR you should have no problem to give it a try on OSMD. I think that after the changes from @AaronDavidNewman it is now possible to replace VexFlow 1.2.93.

0xfe:
Not sure I understand why you're removing `MakeException` -- those exceptions are part of the interface, so users that are expecting to catch it will break. I think we should keep it (and export the type.)

rvilarl:
@0xfe I did look into that:
https://stackoverflow.com/questions/35314114/how-can-i-return-a-class-from-a-typescript-function

I did not like the idea of returning `any` and I thought that the `interface` solution was just too complicated for the additional benefit provided by `MakeException` with the change to `RuntimeError` we a throwing and Error anyway.

**RuntimeError** has a `code`
**MakeException** has a `name` which is the class name, a `message` which is equivalent to the `code` and a `data` which I do not like, it is a `any` and I cannot imagine that it is used by the users.

I also wanted to standardize: **MakeExecption are thrown 5 times** in the library and **RuntimeError 185**.

0xfe:
Okay, those are good points. Let's just remove it then.

PR #1065, opened 2021-07-11
Migrate Rests Tests

ronyeh:
Ready for code review. 

For now, we keep the Vex.Flow.* prefix in front of all class names, since removing the prefix breaks some test cases due to VexFlow being included twice in the flow.html.

During the migration, I did remove the prefix to allow the TS compiler to do its magic. As a result, this PR includes a few small fixes, for example:

stave.ts: the fourth parameter is now optional, since most test cases only provide three params.
`constructor(x: number, y: number, width: number, options?: Partial<StaveOptions>) {`

stavenote.ts: In the StaveNoteStruct, `auto_stem` and `clef` are now optional. Test casses frequently omit these options.

Tests show no visual diffs relative to master @ revision 73b01c61ea67c3a8539ec77eec424d2bc6ff5aba.


src/element.ts
```diff
@@ -44,8 +44,8 @@ export abstract class Element {
   protected style?: ElementStyle;
   private attrs: ElementAttributes;
   protected boundingBox?: BoundingBox;
-  protected fontStack: Font[];
-  protected musicFont: Font;
+  protected fontStack!: Font[];
+  protected musicFont!: Font;
```
rvilarl, line 48: Please add a comment saying that they are initialized by SetFonts called by `constructor.

src/note.ts
```diff
@@ -387,10 +387,8 @@ export abstract class Note extends Tickable {
   }
 
   /** Gets the stave line number for the note. */
-  getLineNumber(
-    // eslint-disable-next-line
-    isTopNote?: boolean
-  ): number {
+  // eslint-disable-next-line
+  getLineNumber(isTopNote?: boolean): number {
```
rvilarl, line 391: Have you checked if typedoc copes with the comment? Is the function still documented? (use `grunt typedoc` and have a look into build/docs)

src/note.ts
```diff
@@ -400,8 +398,8 @@ export abstract class Note extends Tickable {
   }
 
   /** Get the glyph associated with this note. */
-  getGlyph(): // eslint-disable-next-line
-  any {
+  // eslint-disable-next-line
+  getGlyph(): any {
```
rvilarl, line 402: Same here regarding typedoc.

src/note.ts
```diff
@@ -545,10 +543,8 @@ export abstract class Note extends Tickable {
     return this;
   }
   /** Get the coordinates for where modifiers begin. */
-  getModifierStartXY(
-    // eslint-disable-next-line
-    position?: number, index?: number, options?: any
-  ): { x: number; y: number } {
+  // eslint-disable-next-line
+  getModifierStartXY(position?: number, index?: number, options?: any): { x: number; y: number } {
```
rvilarl, line 547: Same here regarding typedoc.

src/stavenote.ts
```diff
@@ -370,10 +368,10 @@ export class StaveNote extends StemmableNote {
   constructor(noteStruct: StaveNoteStruct) {
     super(noteStruct);
     this.setAttribute('type', 'StaveNote');
-    // Ledger Lines default width 2.0
-    this.ledgerLineStyle = { lineWidth: 2.0 };
-    this.clef = noteStruct.clef;
-    this.octave_shift = noteStruct.octave_shift;
+
+    this.ledgerLineStyle = { lineWidth: 2.0 }; // Set default width of ledger lines to 2.0.
```
rvilarl, line 372: Comments are on the line above in the `coding template`: accidental.ts according to the wiki.

src/tables.ts
```diff
@@ -374,12 +374,9 @@ export const Tables = {
...
   octave_shift for clef ottavation (0 = default; 1 = 8va; -1 = 8vb, etc.).
-*/
-  keyProperties(
-    key: string,
-    clef?: string,
-    // eslint-disable-next-line
-  params?: any): any {
+  */
+  // eslint-disable-next-line
+  keyProperties(key: string, clef?: string, params?: any): any {
```
rvilarl, line 379: Same here regarding typedoc.

tests/rests_tests.ts
```diff
@@ -0,0 +1,351 @@
...
+import { QUnit, ok, Assert } from './declarations';
+import { VexFlowTests } from './vexflow_test_helpers';
+import { RenderContext } from '../src/types/common';
+import { Flow } from '../src/flow';
+import { StaveNoteStruct } from '../src/stavenote';
+import { Stave } from '../src/stave';
+import { ContextBuilder } from '../src/renderer';
+
+// TODO: Fix later! Currently we need to prepend all classes with VF.
+const VF: any = Vex.Flow;
```
rvilarl, line 14: Add:
`//eslint-disable-next-line`

rvilarl (changes requested):

rvilarl:
> For now, we keep the Vex.Flow.* prefix in front of all class names, since removing the prefix breaks some test cases due to VexFlow being included twice in the flow.html.

At the begining of the migration I suggested to have two html files `flow.html` including js test and `flow_ts.html` including the ts file. I thought it was better to remove the prefix in order to make usage of the type testing. I also realised the problem of double inclusion, I thought that it did not matter anyway as we would have a single html file once the migration was finised, but my proposal was not accepted :(

src/note.ts
ronyeh: Thanks for your review. I haven't checked typedoc yet, and I will do so to make sure that the eslint-disable-next-line comments don't break the typedoc.

The only reason I've been moving the eslint-disable-next-line up to the line before the method is due to aesthetics. I definitely prefer if we don't have a comment in the middle of the method signature.

However, if it breaks typedoc, then I'll have to live with it! (So far I have been depending on the mouseover feature in VS code to check that the documentation looks OK. I just move my mouse cursor over the method name and a tooltip shows the comment.)

src/stavenote.ts
ronyeh: Sounds good! I'll make sure that implementation comments precede the line of code.

ronyeh:
Thanks for the review @rvilarl ! I think @0xfe suggested that we stick with a single flow.html, but only make sure to bundle VexFlow once.

My idea for that is to:

1. add a new test file that verifies all the existing Vex.Flow.* exists. For example, we can do a qunit sanity check on the Vex.Flow.StaveNote.CATEGORY and other properties to make sure they aren't undefined.
2. then rename all the existing tests to *.ts, and remove all the VF. prefixes. This will require the tests to directly import the classes when needed.
3. in flow.html, we remove vexflow-debug.js completely. Since vexflow-tests.js will be generated by webpack, it should include all the VexFlow.* classes as needed.

Does this general approach seem sound?

For now, I will migrate a few test pages (while leaving the VF.* prefixes in place). During migration I actually remove the VF.* prefixes to let the TypeScript compiler catch bugs (and it has caught several already). Then at the end, I put the VF.* prefixes back in place.

src/note.ts
ronyeh: It looks like typedoc generates the correct documentation even if the code has the eslint-disable-next-line right before the method:

```
  /** Gets the stave line number for the note. */
  // eslint-disable-next-line
  getLineNumber(isTopNote?: boolean): number {
    return 0;
  }
```

@rvilarl is the above OK with you? Otherwise, the eslint directive breaks up the method signature and makes the code a bit harder to read, in my opinion.

rvilarl: This is fine, I made it because I thought that it would break typedoc. It looks much better this way.

rvilarl:
> My idea for that is to:
> 
> 1. add a new test file that verifies all the existing Vex.Flow.* exists. For example, we can do a qunit sanity check on the Vex.Flow.StaveNote.CATEGORY and other properties to make sure they aren't undefined.
> 2. then rename all the existing tests to *.ts, and remove all the VF. prefixes. This will require the tests to directly import the classes when needed.
> 3. in flow.html, we remove vexflow-debug.js completely. Since vexflow-tests.js will be generated by webpack, it should include all the VexFlow classes as needed.
> 
> Does this general approach seem sound?

Sounds very good.

ronyeh:
DotFix

One of the fixes in this PR is the above GIF. We had never tested the 128th rest, even though we include the glyph in VexFlow. When I added the 128th rest to the rest_tests, I noticed the dot was too low. I moved the dot up to the top, and it now looks like 128th rests generated by other scoring apps.

ronyeh:
`npm run test:reference` shows the expected 3 diffs relative to master, as shown in the GIF in a previous comment.

```
Rests.Rests___Dotted.Gonville 34.7312
Rests.Rests___Dotted.Petaluma 31.2511
Rests.Rests___Dotted.Bravura 22.9915
```

This diff is the result of adding a test for 128th rests (and fixing a bug with dot placement for 128th rests).

ronyeh:
Note: the visual diffs now show a warning because I changed the title of the tests to make it look visually cleaner in flow.html.

```
Warning: Rests.Auto_Align___Beamed_Notes_Stems_Down.Bravura.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Beamed_Notes_Stems_Down.Gonville.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Beamed_Notes_Stems_Down.Petaluma.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Beamed_Notes_Stems_Up.Bravura.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Beamed_Notes_Stems_Up.Gonville.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Beamed_Notes_Stems_Up.Petaluma.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Multi_Voice.Bravura.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Multi_Voice.Gonville.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Multi_Voice.Petaluma.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Single_Voice__Align_All_.Bravura.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Single_Voice__Align_All_.Gonville.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Single_Voice__Align_All_.Petaluma.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Single_Voice__Default_.Bravura.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Single_Voice__Default_.Gonville.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Single_Voice__Default_.Petaluma.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Tuplets_Stems_Down.Bravura.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Tuplets_Stems_Down.Gonville.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Tuplets_Stems_Down.Petaluma.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Tuplets_Stems_Up.Bravura.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Tuplets_Stems_Up.Gonville.png missing in ./build/images/reference.
Warning: Rests.Auto_Align___Tuplets_Stems_Up.Petaluma.png missing in ./build/images/reference.
Warning: Rests.Dotted.Bravura.png missing in ./build/images/reference.
Warning: Rests.Dotted.Gonville.png missing in ./build/images/reference.
Warning: Rests.Dotted.Petaluma.png missing in ./build/images/reference.
```

I have manually verified that there are no unexpected diffs.

0xfe:
Can I review this, Ron?

ronyeh:
Yeah, ready for review.

ronyeh:
Actually, I would recommend skipping this PR.

PR #1074 includes the changes from this PR, and adds other small improvements to the Rests Tests.

I will close this PR and recommend #1074 instead.

PR #1066, opened 2021-07-12
Documentation update

rvilarl:
@0xfe Only documentation and removal of one interface which was only used once.


0xfe:
Thanks for all this cleanup! It's great!

Issue #729, opened 2019-11-14
Is it possible to set a default ledger line style?


newlandsvalley:
(I haven't found a way to do it so far)

rvilarl:
@0xfe @newlandsvalley is right, there is no way to set a default. `protected ledgerLineStyle: ElementStyle;` is an attribute for each `StaveNote`. Is it expected to have different styles on each `StaveNote`? There indeed examples with different colours.

0xfe:
It's part of the note head. Open to an API to set a default -- perhaps at the `Stave`.

rvilarl:
@0xfe Stave does not hold a list of the notes so I decided to place it in system. I hope this is fine :)

0xfe:
`System` is not that widely used, but notes are always rendered onto a `Stave`. So you can get the style from the stave during render time. (Alternatively, we can make it a global default setting in `Flow`.)

rvilarl:
@0xfe now in Stave as suggested, we can close this once we merge #1069

PR #1069, opened 2021-07-16
setLedgerLineStyle in Stave

rvilarl:
Addresses #729.
setLedgerLineStyle sets the ledger line style of all the stave notes in the system.


rvilarl:
@0xfe now in Stave as suggested.

0xfe:
Great!

PR #1074, opened 2021-07-25
Migrate all tests

ronyeh:
I have transitioned `flow.html` to only include `vexflow-tests.js`, which is bundled by webpack (and includes any VexFlow library classes that are imported by the tests).

My plan is to drop a `/* eslint-disable */` at the top of each test file for now. Once every test is renamed to *.ts, then we can go back and actually add types to enforce type checking and discover bugs.

This PR subsumes https://github.com/0xfe/vexflow/pull/1065 so we can ignore https://github.com/0xfe/vexflow/pull/1065.


ronyeh:
A big milestone!

All the tests have been migrated to *.ts files. Everything is webpack bundled into a single `vexflow-tests.js`, and flow.html now only includes one copy of VexFlow.

All tests under `flow.html` pass.

For now, I put this at the top of each tests file to hide any eslint / ts issues:

```
/* eslint-disable */
// @ts-nocheck
```

After we eventually merge this PR, we will need to revisit each tests file and remove those two lines to see what issues crop up.

The most common issue I see is that tests will omit parameters, but our current TypeScript-ified API declares those parameters are required. We will need to revisit those cases to see which params can be made optional via param? or via  param=default_value.

There are still some open issues with this PR, because the command line scripts need to be updated.

src/easyscore.ts
```diff
@@ -15,6 +15,7 @@ import { RenderContext } from './types/common';
 import { Accidental } from './accidental';
 import { Modifier } from './modifier';
 import { Voice } from './voice';
+import { Note } from 'note';
```
rvilarl, line 18: change to `from './note'`

src/staveline.ts
```diff
@@ -154,7 +154,9 @@ export class StaveLine extends Element {
   protected font: FontInfo;
   protected first_indices!: number[];
   protected last_indices!: number[];
-  protected notes: StaveLineNotes;
+
+  // Initialized by this.setNotes(notes) in the constructor.
+  protected notes!: StaveLineNotes;
```
rvilarl, line 159: The linter does not complain?

src/tables.ts
```diff
@@ -1308,7 +1305,7 @@ export const Tables = {
             flag: false,
             rest: true,
             position: 'B/4',
-            dot_shiftY: 1.5,
+            dot_shiftY: -2.5,
```
rvilarl, line 1308: This change does not affect the output of the tests?

src/tabslide.ts
```diff
@@ -26,18 +26,19 @@ export class TabSlide extends TabTie {
     return new TabSlide(notes, TabSlide.SLIDE_DOWN);
   }
 
+  /**
+   * @param notes is a struct that has:
```
rvilarl, line 30: you could comment each attribute
@param notes.first_note 
...
if you wanted

src/tabtie.ts
```diff
@@ -16,18 +16,18 @@ export class TabTie extends StaveTie {
     return new TabTie(notes, 'P');
   }
 
+  /**
+   * @param notes is a struct that has:
```
rvilarl, line 20: as above, you could comment each attribute if you wanted.

src/tuplet.ts
```diff
@@ -73,7 +73,7 @@ export class Tuplet extends Element {
   protected y_pos: number;
   protected x_pos: number;
   protected width: number;
-  protected location: number;
+  protected location!: number;
```
rvilarl, line 76: The linter does not comply?

rvilarl (changes requested):

rvilarl:
@0xfe I have review the code change, I suggest to accept the test changes without review on the basis of identical output, otherwise the revisison would take too much effort.
@ronyeh this is great.

ronyeh:
Ready for comments and code review!

ronyeh:
To do a sanity check on the old API (e.g., `new VF.Formatter()`), I added some new example test cases. As we remove more VF.* prefixes, we can add more of these checks along the way.

```


// Use QUnit to verify that the classes are identical.
VF_Prefix(): void {
  equal(Voice, VF.Voice);
  equal(Registry, VF.Registry);
  propEqual(new Formatter(), new VF.Formatter(), 'new Formatter()');
  equal(StaveNote, VF.StaveNote);
  notEqual(Stave, VF.StaveNote);
  equal(Stave, VF.Stave);
  equal(Beam, VF.Beam);
  equal(Note, VF.Note);
  equal(Bend, VF.Bend);
  equal(Annotation, VF.Annotation);
  equal(StaveConnector, VF.StaveConnector);
},
```

ronyeh:
There are 4 tests with visual diffs. 3 of the diffs are due to my small changes and bug fixes. 1 of the diffs was unexpected, but I believe the output is now correct, whereas the output was previously incorrect. (So maybe I fixed a small bug somewhere?)

I will put together some before and after GIFs.

Visual Diffs: 
```
Rests.Rests___Dotted.*.png
This was expected because I added a 128th rest to the test case, and also fixed the dot's vertical location.

KeySignature.Key_Signature_Change_test.*.png
This was expected. The width of the stave changed, because I fixed a bug in the test.
The previous test code was buggy. It called `factory.Stave(10, 10, 800)` even though `factory.Stave(options)` only accepts 1 param.

StaveModifier.Vertical_Bar_Test.*.png
The previous test output was incorrect. The current test output is correct. However, this diff was not expected. I may have fixed a tiny bug somewhere as I ported the tests. I might have to investigate this one.

StringNumber.Multi_Voice_With_Strokes__String___Finger_Numbers.*.png
I changed the order in which the notes/voices were drawn, so that the ledger lines from the second voice do not display on top of the note heads from voice 1. While fixing this test case, I discovered that VexFlow draws TWO ledger lines for middle C when two separate voices both require a middle C ledger line.



```

src/tables.ts
ronyeh: We never tested 128th rests before. When I added a 128th rest to the test cases, I noticed the dot was in the wrong place.

src/staveline.ts
ronyeh: Surprisingly, I am not getting any eslint complaints when I add a ! 
I need to investigate...

src/tuplet.ts
ronyeh: It seems like adding a ! is equivalent to asking the TS compiler to avoid flagging this line. It's basically like saying "trust me, I know what I am doing here because I actually do initialize this property later".

eslint does not complain. Perhaps there is a setting we can turn on if we care to.

ronyeh: More info here:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#definite-assignment-assertions

We definitely need to be careful when using the **definite assignment** operator. (It's similar to the non-null assertion, which the eslint will definitely complain about.)

For now, we can just require a comment above the instance variable that indicates where it is assigned. In this case, I will need to add a comment to tuplet.ts to show where `location` is assigned.

ronyeh: It's actually easy to search for where we use this feature.

Doing a project wide search for the string `!:` I see that I have 40 instances of it. The master version has 34. So I will need to check where I added them and provide comments to verify that those instance variables are definitely assigned via a helper function.

ronyeh:
I understand this is a pretty big PR :-) and I would like to make sure we are extremely confident that it doesn't introduce bugs before we merge. (In fact I believe it fixes a few old bugs, and exposes TS migration issues that we will need to address.)

I think a good first step will be for Mohit and Rodrigo to directly git clone my branch:

https://github.com/ronyeh/vexflow/tree/migration/tests

Then npm install and run grunt and poke around through all the test files.

You will see that the work is not complete. There are still many instances of VF.* that we need to remove. There are also eslint-disable and @ts-nocheck at the top of each file. If you temporarily remove those, you will see the TypeScript issues that pop up. The most common one is where the test case omits a param, but the TS method signature declares that the param is required. In many cases, the simplest fix is to add a ? to the method param (after verifying that the method does provide a default value if the param is undefined).

My next step will be to investigate the one unexpected visual diff (I believe the master revision had buggy output).

I will also try to put together a list of any small bugs I squashed, and add comments to any place I added the ! definite assignment assertion operator.

ronyeh:
Here are the six ! operators that I added. I made sure that all of them have comments. There are already other uses of this operator in the code base. We can add more comments next time.

Several of the ! operators were added to the `notes` instance variable. I discovered an issue where `TabSlide` extends `TabTie` extends `StaveTie`, but all three constructors called `this.setNotes()`. So creating a new `TabSlide` would call setNotes() three times (via super()).

```
element.ts
  // fontStack and musicFont are both initialized by the constructor via this.setFontStack(...).
  protected fontStack!: Font[];
  protected musicFont!: Font;

tuplet.ts
  // location is initialized by the constructor via setTupletLocation(...).
  protected location!: number;

stavetie.ts
  // notes is initialized by the constructor via this.setNotes(notes).
  protected notes!: TieNotes;

stavehairpin.ts
  // notes is initialized by the constructor via this.setNotes(notes).
  protected notes!: Record<string, Note>;

staveline.ts
  // These five instance variables are all initialized by the constructor via this.setNotes(notes).
  protected notes!: StaveLineNotes; // <= I added this !
  protected first_note!: StaveNote;
  protected first_indices!: number[];
  protected last_note!: StaveNote;
  protected last_indices!: number[];
```

src/easyscore.ts
ronyeh: While migrating tests, I noticed that I could omit paths for any class under src/.

For example:
```
import { EasyScore } from '../src/easyscore';
import { StaveNote } from '../src/stavenote';
import { System } from '../src/system';
```
can be rewritten as
```
import { EasyScore } from 'easyscore';
import { StaveNote } from 'stavenote';
import { System } from 'system';
```

I think this is due to the baseURL field in the tsconfig.

If VS Code and TS compiler and webpack don't complain, and everything builds fine, does that mean we can omit all the ./ in front of all the imports?

We can do a separate PR for this. I think it would make everything look slightly cleaner if we can omit all the starting ./

rvilarl:
Fixes #1079

ronyeh:
A

This was one of the four visual diffs: `StaveModifier.Vertical_Bar_Test.*.png`

The reference test output was incorrect. The current test output (with two visible lines) is correct. 

This diff was NOT expected. I will try to investigate what small change I made that resulted in this bug fix.

rvilarl:
> I think a good first step will be for Mohit and Rodrigo to directly git clone my branch:
> 
> https://github.com/ronyeh/vexflow/tree/migration/tests
> 
> Then npm install and run grunt and poke around through all the test files.

@ronyeh I cloned your branch and used `npm run test:reference`to get the differences on the tests output compared to master. Here they are in pairs (first this branch, second master), only 4 differences:

1) KeySignature Key_Signature_Change_test Bravura
KeySignature Key_Signature_Change_test Bravura_Current

KeySignature Key_Signature_Change_test Bravura_Reference

2) Rests Rests___Dotted Bravura **(128r was added, this is fine)**
Rests Rests___Dotted Bravura_Current

Rests Rests___Dotted Bravura_Reference

3) StaveModifier Vertical_Bar_Test Bravura 
StaveModifier Vertical_Bar_Test Bravura_Current

StaveModifier Vertical_Bar_Test Bravura_Reference

4) StringNumber Multi_Voice_With_Strokes__String___Finger_Numbers 
StringNumber Multi_Voice_With_Strokes__String___Finger_Numbers Bravura_Current

StringNumber Multi_Voice_With_Strokes__String___Finger_Numbers Bravura_Reference

rvilarl:
> 
> This diff was NOT expected. I will try to investigate what small change I made that resulted in this bug fix.

I also realised that, we are both looking at the same thing :)

rvilarl:
> > This diff was NOT expected. I will try to investigate what small change I made that resulted in this bug fix.
> 
> I also realised that, we are both looking at the same thing :)

@ronyeh looking at the code I think that the current representation is the right one, you also commented on the ouput

```js
/**
   * Draw three vertical lines. The last line is not visible, because it overlaps
   * the right border of the stave at x = 300.
   */
  drawVerticalBar(options: TestOptions, contextBuilder: ContextBuilder): void {
    const ctx = contextBuilder(options.elementId, 400, 120);
    const stave = new Stave(10, 10, 300);
    stave.setContext(ctx);
    stave.draw();
    stave.drawVerticalBar(100);
    stave.drawVerticalBar(150);
    stave.drawVerticalBar(300);

    ok(true, 'all pass');
  },
```
StaveModifier Vertical_Bar_Test Bravura_Current

ronyeh:
@rvilarl I figured it out!

The master branch is running the wrong test code:

https://github.com/0xfe/vexflow/blob/2315e7aba40a9bd0328ccb0b6c155eeff4489417/tests/stavemodifier_tests.js#L8-L12

ronyeh:
As you can see above, the master branch was calling .draw and .drawVerticalBar in the Stave tests, not the StaveModifier tests.

So when I converted it to TypeScript, it revealed the error and I pointed it to the draw and drawVertical bar defined in StaveModifierTests. I think it's as simple as that.

rvilarl:
@0xfe checked the test results are all fine! From my point of view, ready to merge.

src/easyscore.ts
0xfe: Yep, agreed.

src/tables.ts
0xfe: Good catch!

.vscode/settings.json
```diff
@@ -1,6 +1,14 @@
 // Place your settings in this file to overwrite default and user settings.
 {
   "eslint.enable": true,
+  "files.exclude": {
```
0xfe, line 4: Thanks for this. vscode-eslint starts to get laggy after a while.

0xfe:
Ron, this is fantastic work! Great job! 

Rodrigo, thanks for reviewing this. I agree that this is ready to merge. Absolutely thrilled at how far we've come, thanks to you two.

PR #1077, opened 2021-07-27
Fix Glyph.getOutlineBoundingBox.

tommadams:
getOutlineBoundingBox was not passing the first control point for either
qudratic or cubic Bezier curves and left the last (x, y) parameters to
addQuadraticCurve and addBezierCurve as undefined. This caused the
computed bounding boxes for individual curve pieces to be significantly
too small.

Fortunately, this bug only caused the bounding boxes for the entire
glyph to be incorrect by less than one or two pixels in each dimension.

This pull request causes minor visual diffs of around one or two pixels in
note spacing, as one might expect.

I've also added some unit tests.


0xfe:
Great catch! Thanks for fixing. Overall looks good, however we just switched the tests to TS -- can you fix the conflicts please?

tommadams:
I've updated the tests to TS. Please take a look.

0xfe:
Thanks Tom. This looks great! Merging!

PR #1085, opened 2021-07-31
restore `music.noteValues`

### LINKED ISSUE #1084: What happened to VF.Music.noteValues?
It seems some of the theory routines that used to be in vex are gone, or maybe just changed and I don't understand their new form.  I used to use to get Object.keys(VF.Music.noteValues) to cycle through enharmonics, for instance.  I can create a new class that contains what it used to contain, but I'm wondering, is it there and I'm just not seeing it?

rvilarl:
`music.noteValues` is used by smoosic. Assumption of private element was wrong.
fixes #1084

### DIFF
--- a/src/music.ts
+++ b/src/music.ts
@@ -1,5 +1,5 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
-// MIT Licnse
+// MIT License
 
 import { RuntimeError } from './util';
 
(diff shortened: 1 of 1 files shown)


src/music.ts
```diff
@@ -1,7 +1,6 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 // MIT Licnse
```
ronyeh, line 2: typo. :-)

rvilarl: resolved :)

0xfe (commented):
Looks good.

PR #1090, opened 2021-08-01
Restore the ability to run 'npm run test:reference' 

### LINKED ISSUE #1088: generate_png_images complains about VFT
@ronyeh  I have realised that after the migration of tests to TS, I get the following:

vexflow/tools/generate_png_images.js:50
VFT.RUN_CANVAS_TESTS = false;
                     ^

TypeError: Cannot set property 'RUN_CANVAS_TESTS' of undefined
    at Object.<anonymous> (/home/rodrigo/git/vexflow/tools/generate_png_images.js:50:22)

ronyeh:
Now that PR #1074 has been merged, it has become the new 'reference'. Thus generate_png_images.js needs to load the 'reference' library the new way (one JS file). 

Fixes #1088.

### DIFF
--- a/tools/generate_png_images.js
+++ b/tools/generate_png_images.js
@@ -28,19 +28,16 @@ if (process.argv.length >= 5) {
   }
 }
 
-if (scriptDir.includes('reference') || scriptDir.includes('releases')) {
+if (scriptDir.includes('releases')) {
   // THE OLD WAY loads two JS files.
-  // TODO: Remove line 32 "scriptDir.includes('reference') ||"
-  //       after PR #1074 has been merged, becoming the new 'reference/'.
-  // TODO: Remove this entire block lines 32-40, after the new version has been moved to 'releases/'
+  // TODO: Remove this block lines 31-37, after the new version has been moved to 'releases/'
   global.Vex = require(`${scriptDir}/vexflow-debug.js`);
   require(`${scriptDir}/vexflow-tests.js`);
   global.Vex.Flow.shims = { fs, process };
 } else {
   // THE NEW WAY loads a single JS file.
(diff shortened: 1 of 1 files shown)


rvilarl:
> Fixes Issue #1088.

@ronyeh I learnt today that if you write `fix`or `fixes`rather than `Fixes issues`you get the issue and the PR automatically link. You will see a merge icon on the issue and an issue icon in the PR. I think it shoud work if you edit it.

0xfe:
Looks good. Are you tracking those TODOs somewhere? (We can remove those lines after the next release is cut, correct?)

ronyeh:
@0xfe is there a place you'd prefer for tracking todos for the next official release? Is a markdown task list **in the wiki** best? That way any of us can edit it.

Yes, this check for `if (scriptDir.includes('releases')) {` can be removed after we move the final vexflow-min.js into the releases/ folder.

I'd recommend making subdirectories inside releases/ for organizing previous releases. Something like:
```
releases/1.2.90/*
releases/3.0.9/*
releases/4.0.0/*
```

That way our test scripts can be smarter about doing visual diffs. When loading the library from 3.0.9 and before, it needs to load both vexflow-debug.js and vexflow-tests.js. When loading the library from 4.0.0 and later, it only needs vexflow-tests.js (which is bundled by webpack).

ronyeh:
@0xfe @rvilarl I added https://github.com/0xfe/vexflow/wiki/TODOs for a lightweight way to track TODOs. If we need something more permanent, we could always commit a TODOS.md into the repository. But for now hopefully a wiki page is enough for now.

One benefit of a TODOS.md in the repository is that as we clean up items, we can move them over to a CHANGELOG.md.

0xfe:
That sounds good, Ron.

PR #1091, opened 2021-08-02
Note.keys keeps accidental

### LINKED ISSUE #1075: Accidentals: hash and sharp right location in Note.
Notes can have the accidentals hash and sharp in two different attributes:
- `keys` i.e.: `c#/5`
- `modifiers` (defined in tickable) as instance of `Accidental`

The function `Accidental.applyAccidentals` expects it to be in `keys`.

@0xfe what is the right location?

rvilarl:
fix #1075


rvilarl:
int_val have to be updated for the microtones.

src/tables.ts
```diff
@@ -390,47 +390,138 @@ export const Tables = {
       }
     > = {
       C: { index: 0, int_val: 0 },
+      'C+': { index: 0, int_val: 0, accidental: '+' },
```
0xfe, line 393: Thanks Rodrigo. Can you move all the microtonals to a separate section to reduce clutter?

```typescript
{
  'C': ...,
  'C#': ...,
  ...
  // Microtonals
  'C+': ...
  ...
}
```

rvilarl: Done

rvilarl:
@0xfe I forgot to commit the fix to `applyAccidentals`

mscuthbert:
I have concerns about putting microtones into the keys as opposed to other ways of setting microtones' display and effect (`.setMicrotonalAccidental()` and `.setMicrotonalTuning()` or something).  They tend to multiply out of control, different composers have used the same symbol to mean different things (or different symbols to mean the same thing), and can often be combined with each other (C# + 7th-harmonic-microtone, etc.).

For the next version of music21, I plan to remove shortcut setters even for 1/2-sharp and 1/2-flat because of problems with display and no-one really knowing what they mean (I can never remember if C' is half-sharp and C~ is half-flat or vice versa).

Once you support 20 microtone types, it's very easy to find a project quickly needing to support 100, and some with unusual placement or other specialty notational requirements.

rvilarl:
I noticed that there are more. I included the ones defined in the Grammar for Easyscore:

https://github.com/0xfe/vexflow/blob/37e70ad84f26ebce2546c52ac0c9c3e7c4a127c9/src/easyscore.ts#L169-L171

I find that the Musicxml approach with alter which is only a number indicating the required pitch is a nice one

https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/alter/

Then the accidental includes the accidental-value used for the representation

https://www.w3.org/2021/06/musicxml40/musicxml-reference/data-types/accidental-value/

0xfe:
Thanks Michael -- that context super helpful. So maybe we should remove microtones entirely out of `ACCIDENTALS`, and add a specific grammar for microtones.

```typescript
  ACCIDENTALS(): Rule {
    return { token: 'bb|b|##|#|n' };
  }

  MICROTONES(): Rule {
    return { token: 'bbs|bss|bs|db|d|\\+\\+-|\\+-|\\+\\+|\\+|k|o' };
  }

  ACCIDENTAL(): Rule {
    return {
      expect: [this.ACCIDENTALS, this.MICROTONES],
      maybe: true,
      or: true,
    };
  }
```

And then make sure that accidentals are part of the `key` and microtones are not.

rvilarl:
Just to check I understood the proposal:
1) `cbbs4/q` will produce a note with `keys 'C/4` with an added `Accidental('bbs')`
2) `c#4/q` will produce a note with `keys 'C#/4` with an added `Accidental('#')`
3) `applyAccidentals` will ignore microtones and leave them untouched
@0xfe is that right?

@mscuthbert does `makeAccidentals` work like 3) ?

0xfe:
That's correct @rvilarl

rvilarl:
@0xfe done, ready to merge

src/easyscore.ts
```diff
@@ -350,7 +354,15 @@ export class Builder {
 
     // Build StaveNotes.
     const { chord, duration, dots, type } = this.piece;
-    const keys: string[] = chord.map((notePiece) => notePiece.key + '/' + notePiece.octave);
+    const keys: string[] = chord.map(
+      (notePiece) =>
+        notePiece.key +
+        (notePiece.accid && ['bb', 'b', '##', '#', 'n'].includes(notePiece.accid.toLowerCase())
```
0xfe, line 360: Can you make this:

```typescript
const keys: string[] = chord.map((notePiece) => notePiece.key + ['bb', 'b', '##', '#', 'n'].includes(notePiece.accid) ? notePiece.accid : '' + '/' + notePiece.octave);
```

No need for `toLowerCase` because the parser already restricts case, and no need for the extra `notePiece.accid &&`.

0xfe (commented):
Apologies for the delay here -- I dropped this.

src/easyscore.ts
rvilarl: `notePiece.accid` can be `undefined` and `notePiece.accid &&` was required in order to avoid the complain in `.includes(notePiece.accid)`

I changed to `.includes(notePiece.accid ?? '')` which is more compact.

src/accidental.ts
```diff
@@ -427,6 +427,17 @@ export class Accidental extends Modifier {
           // modified the accidental state
           const previouslyModified = modifiedPitches.indexOf(pitch) > -1;
 
+          // Remove accidentals
+          note.getModifiers().forEach(function (modifier, index) {
```
0xfe, line 431: Not sure what's going on here. Why do we need this block?

0xfe:
Can you add some tests to validate the new behavior?

src/accidental.ts
rvilarl: This is required when the notes already include the accidentals before calling `applyAccidentals` (ie when they are created with `EasyScore`. I have included an additional test with EasyScore in accidentals. If this block is remved, the result is:

image

0xfe:
Okay, makes sense.

Issue #1093, opened 2021-08-02
TabNoteStruct's str & fret should probably be numbers instead of strings.


ronyeh:
https://github.com/0xfe/vexflow/blob/7a682f7075017ccdcebd30d06dcc0f9661126dd2/src/tabnote.ts#L21-L23

In tabslide_tests.ts, str and fret are consistently provided as numbers (which makes sense to me).

I did a quick search in the tests for "new TabNote(" and it looks like other tests also assume `str` and `fret` are numbers.

ronyeh:
https://github.com/0xfe/vexflow/blob/7a682f7075017ccdcebd30d06dcc0f9661126dd2/tests/tabslide_tests.ts#L90-L101

https://github.com/0xfe/vexflow/blob/7a682f7075017ccdcebd30d06dcc0f9661126dd2/tests/vibrato_tests.ts#L70-L89

0xfe:
Actually, `fret` takes `X` too (muted notes), which is rendered into a glyph via `Flow.tabToGlyph` (which prob should be moved to `tabnote.ts`.)

ronyeh:
I changed `str` to `number`.
I changed `fret` to `number | string`.

...I don't believe this changes any of the public APIs (at least the tests aren't affected). These type labels were introduced during the TS migration.

Issue #1087, opened 2021-08-01
Use unicode characters directly?


ronyeh:
https://github.com/0xfe/vexflow/blob/2315e7aba40a9bd0328ccb0b6c155eeff4489417/src/tables.ts#L735-L747

Minor thing I noticed. Can we replace the unicode table with the actual unicode characters, since source files support utf-8?

The following are all equivalent:
```
String.fromCharCode(parseInt('00F8', 16))
String.fromCharCode(0x00F8)
'ø'
```

Ideally we use the bottom one. We can keep the hex number code point in a comment above each entry.

ronyeh:
So the new unicode table would look like

```
unicode: { 
  // 0x266F
  sharp: "♯",
  // 0x266D 
  flat: "♭", 
  // 0x266E
  natural: "♮", 
```

mscuthbert:
I think this is a good idea until you get to Astral plane ( > 0xFFFF) characters which some Javascript engines have a hard time with.  So sharp, flat, natural are easy, but double-flat and double-sharp are astral characters which need to be constructed more gently.

0xfe:
I would prefer it the other way round -- put the actual characters in the comments. Unicode breaks in too many ways particularly on older tooling or misconfigured apps.

ronyeh:
Done. I put the unicode characters in comments. But I removed the `parseInt()` in favor of using hex numbers directly. Hopefully JS engines don't have difficulty with 0xFFF style numbers.

Also, my text labels are inspired by the SMuFL naming convention. I understand the triangle is originally just a major triad, but SMuFL calls it major seventh so I went with that.

https://github.com/0xfe/vexflow/blob/a4a528d8991fbaf6899decd3208b4453ebbb5d5c/src/tables.ts#L731-L746

PR #1095, opened 2021-08-02
Small fixes to TabNoteStruct and Unicode

ronyeh:
TabNoteStruct: Use `number | string` for fret positions. Handle edge cases. Fixes: https://github.com/0xfe/vexflow/issues/1093
Unicode: Remove `parseInt()` because I think JS engines should be able to deal with numbers starting with `0x`. Add comments showing the unicode output. Fixes: https://github.com/0xfe/vexflow/issues/1087

**No visual diffs from master.**

I have been using a VS Code extension called "TypeScript Import Sorter" so the imports are getting alphabetized by module name when I save the file. At some point we can add a [eslint rule](https://eslint.org/docs/rules/sort-imports) so we can have consistent imports. :-)

### DIFF
--- a/src/tables.ts
+++ b/src/tables.ts
@@ -2,12 +2,11 @@
 
 /* eslint-disable key-spacing */
 
-import { RuntimeError } from './util';
+import { ArticulationStruct } from './articulation';
+import { Fonts } from './font';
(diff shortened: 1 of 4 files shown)


0xfe:
Looks great.

Issue #1096, opened 2021-08-02
Consider whether instanceof checks will create problems for users


mscuthbert:
One thing I've noticed is appearing more in the Vexflow code after moving to TypeScript are more instanceof tests, which Typescript really likes (and helps with typing).  There's a significant downside of them when embedding Vexflow in large projects, which often appears with JQuery projects.

Say I'm using a program that uses Vexflow (OSMD, or music21j, or something) but I also use Vexflow directly in my project.  My code may look something like this:

```
import * as music21 from 'music21j';
import * as Vex from 'vexflow';

const vfn = <Vex.Flow.StaveNote> new music21.note.Note('C#').vexflowNote;
const staff = new Vex.Flow.Stave();
...  // etc. place vfn into staff, and finally something like:
staff.draw()
```

The problem is that as far as `staff` is concerned, `vfn instanceof Vex.Flow.StaveNote` is false!  Why?  Because most packaging systems will load `music21j` from its already compiled/packed form which includes its own copy of Vexflow.  Even if they are identical versions, etc., `instanceof` checks that `vfn.constructor === Vex.Flow.StaveNote` using triple-equal memory-location equality.   This can lead to some very subtle bugs that are hard to diagnose, even though there's no reason why a Vex.Flow.StaveNote loaded from one copy of Vexflow can't be processed by objects loaded from another copy of Vexflow (assuming they're the same or similar versions).

This happens quite often with projects involving JQuery (which somehow always gets loaded at least twice into every webpack...) where some routine will read something like:

```
getText(el: HTMLElement|JQuery<HTMLElement>) {
   let html_el: HTMLElement;
   if (el instanceof JQuery) { html_el = el[0]; } 
   else { html_el = el; } 
   return html_el.innerText;   // possible error:  JQuery element has no attribute innerText
}
```
...where el is a JQuery element that is not instanceof (someone else's) JQuery constructor.  And I'm afraid this will start happening in Vexflow as well.

ronyeh:
This is precisely what happened with the flow.html test page while we were migrating tests to TS.

flow.html included vexflow-debug.js and also vexflow-tests.js. Since both were bundled separately by webpack, there ended up being two copies of VexFlow on the flow.html webpage.

And instanceof and strict equality checks started failing because Vex.Flow was not the same object as Flow, etc.


It's a good discussion to have. The big question is, should we support this case, where a webpage may have multiple (slightly incompatible) versions of VexFlow?

[...]

PR #1098, opened 2021-08-03
Remove instanceof in favor of a more flexible alternative.

ronyeh:
Implement `isKindOf<T>(obj, T)` as a flexible type guard alternative to instanceof.

Fixes https://github.com/0xfe/vexflow/issues/1096

@mscuthbert  can you clone my branch, build, and see if it solves your issue?

https://github.com/ronyeh/vexflow/tree/instanceof

This PR definitely needs some discussing and extensive testing. :-)


ronyeh:
@0xfe I see some existing string properties that can help us implement `isCategory`.

`Element` subclasses all have `this.setAttribute('type', 'Tickable')`, `this.setAttribute('type', 'Accidental')`, etc. 

63 classes have a setAttribute.

33 classes have a
```
    return 'repetitions';
  }
```

Which one should we use in isCategory? Should we use both? We can check .CATEGORY first, and then fall back to getAttribute()?

We would probably need to implement more `.CATEGORY` properties for every class.

ronyeh:
Merge https://github.com/0xfe/vexflow/pull/1099 first. This branch will be rebased off of that work.

mscuthbert:
This looks good to me.  If you want a pure drop in replacement for instanceof, then you can walk the Constructor prototype chain and check categories there.

ronyeh:
> ... you can walk the Constructor prototype chain and check categories there.

Done! isCategory() now walks up the prototype chain by default. I added an optional flag to disable this, in case there is a performance issue or some other reason.

ronyeh:
No visual diffs from master.

ronyeh:
@mscuthbert can you build this branch and see if VexFlow StaveNote objects generated from your project pass the isCategory() tests inside this version of VexFlow?

https://github.com/ronyeh/vexflow/tree/instanceof

I should probably add some isCategory unit tests...

mscuthbert:
This looks great to me!  THANK YOU!  This will save a lot of people problems!  (I just ran into the JQuery instanceof problem again yesterday!  Glad not to have that problem here in VexFlow)

src/typeguard.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Use this instead of `instanceof` as a more flexible type guard.
+ * @param cls a JavaScript class, such as `StaveNote`.
+ * @param obj check if this object is an instance of the provided `cls`.
+ * @param checkAncestors defaults to `true`, so we walk up the prototype chain to look for a matching `.getCategory()`.
+ *        If `false`, we do not check the superclass or other ancestors.
+ * @returns true if `obj` is an instance of `ClassName`, or has a `.getCategory()` that matches `ClassName.CATEGORY`.
+ */
+export function isCategory<Class>(cls: Constructor<Class>, obj: any, checkAncestors: boolean = true): obj is Class {
```
0xfe, line 15: Thanks Ron -- this change is great overall. QQ: why do wee need `cls` here? Can we not derive it from `Class`?

(Also, I prefer to call the template type `T` instead of `Class`.)

src/typeguard.ts
```diff
@@ -0,0 +1,55 @@
...
+        return true;
+      }
+      obj = Object.getPrototypeOf(obj);
+    }
+    return false;
+  } else {
+    // Do not walk up the prototype chain. Just check this object's .getCategory().
+    return 'getCategory' in obj && obj.getCategory() === categoryToMatch;
+  }
+}
```
0xfe, line 45: Can you add some tests for this class?

src/typeguard.ts
mscuthbert: Class is a typescript annotation, so it's removed in compilation.  It might be possible to do the opposite (get Class from cls) but I couldn't figure out a way to do it in current typescript.

ronyeh:
I will rename to T when I'm back at my computer.

I put together this solution by combining various stack overflow answers
and blog posts. 😳 But I don't think I can avoid the duplication of T at
the moment. That's why I suggest using utility typeguard methods like
isStaveNote(obj) so at least we keep all the ugly duplication inside
typeguard.ts.

Then the utility methods can each call isCategory<T>(T, obj)....

0xfe:
Thanks for the changes, Ron. Is this good to go? (I see it's still WIP.)

ronyeh:
Let me do a little bit more testing and then I'll change it to "Ready for review".

While working on this PR I discovered some inconsistencies with the CATEGORY property that I'd like to clean up in a future PR. I wrote up my proposal in https://github.com/0xfe/vexflow/issues/1107

ronyeh:
OK I think this one is good to go. 

npm run test:reference produces no visual diffs.

flow.html tests seem to run fine.

mscuthbert:
Great new tests.

I'm happy to merge -- but given the edge cases that we're having to deliberately fail tests on, I wonder if it's a good case to move towards targeting ES6 sooner rather than later.  My company switched its target to ES6 about 12 months ago (and even to native ShadowDOM support 4 months ago) and we've seen no complaints and very very few issues.  Microsoft ended its support for IE11 in their own webapps today (well, yesterday in your time zones: https://docs.microsoft.com/en-us/lifecycle/announcements/m365-ie11-microsoft-edge-legacy) and Edge pre-Chromium is already deprecated.  

I think the solution is -- merge this, and then quickly target ES6 while these tests are still in the memory.  But I'm not confident enough that that's the right action to merge without a nod from @0xfe or some time for people to object/support.

ronyeh:
https://github.com/0xfe/vexflow/pull/1098/commits/77d3d2a93bd8fc8c5a993eea605c915a921c6e82

I updated the test case now that we are targeting ES6. `isCategory()` now works as expected in all cases!

No visual diffs. Ready to go. :-)

0xfe:
Thanks Ron. Looks great!

PR #1103, opened 2021-08-13
Optimizations to the glyph code

tommadams:
These optimisations speed up the time spent during parsing and rendering the "Bach demo" test by a factor of approximately 1.6 in both Chrome and Firefox on my machine. The total time spent running all tests is reduced from approximately 16 seconds to 12 seconds.

Some more details:
 - Convert the cached outline from an array of strings to an array of integers. This avoids repeatedly calling `parseInt` whenever the outline is processed.
 - Cache the glyph metrics and bounding box, which avoids repeatedly loading the metrics and parsing the outline data.
 - Add a static `Glyph.getWidth` method that calculates the width from the cached bounding box rather than creating a complete `Glyph` object when only the width is needed.
 - Replace all the implementation of the `getWidth` calls in `tables.ts` to use the new `Glyph.getWidth` method.
 - I also made `duration_codes` in `tables.ts` a global variable that gets lazily initialised on the first call to `getGlyphProps`. Even though `duration_codes` is marked `const`, the compiled JS code was creating new instances of the object literal every single time the function was called.

### DIFF
--- a/src/tables.ts
+++ b/src/tables.ts
@@ -3,6 +3,7 @@
 /* eslint-disable key-spacing */
 
 import { ArticulationStruct } from './articulation';
+import { Flow } from './flow';
 import { Fonts } from './font';
 import { Fraction } from './fraction';
 import { Glyph } from './glyph';
@@ -317,6 +318,9 @@ const accidentals: Record<string, { code: string; parenRightPaddingAdjustment: n
   accidentalWilsonPlus: { code: 'accidentalWilsonPlus', parenRightPaddingAdjustment: -1 },
   accidentalWilsonMinus: { code: 'accidentalWilsonMinus', parenRightPaddingAdjustment: -1 },
 };
+
+// eslint-disable-next-line
+let duration_codes: Record<string, any>;
 
 export const Tables = {
   STEM_WIDTH: 1.5,
@@ -801,8 +805,7 @@ export const Tables = {
     type?: string
   ): // eslint-disable-next-line
   any {
-    // eslint-disable-next-line
-    const duration_codes: Record<string, any> = {
+    duration_codes = duration_codes || {
       '1/2': {
         common: {
           stem: false,
@@ -821,14 +824,14 @@ export const Tables = {
             // Breve note
             code_head: 'noteheadDoubleWhole',
             getWidth(scale = Tables.DEFAULT_NOTATION_FONT_SCALE): number | undefined {
-              return new Glyph('noteheadDoubleWhole', scale).getMetrics().width;
+              return Glyph.getWidth(Flow.DEFAULT_FONT_STACK, 'noteheadDoubleWhole', scale);
             },
           },
           h: {
             // Breve note harmonic
             code_head: 'unpitchedPercussionClef1',
             getWidth(scale = Tables.DEFAULT_NOTATION_FONT_SCALE): number | undefined {
-              return new Glyph('unpitchedPercussionClef1', scale).getMetrics().width;
+              return Glyph.getWidth(Flow.DEFAULT_FONT_STACK, 'unpitchedPercussionClef1', scale);
             },
           },
           m: {
@@ -836,7 +839,7 @@ export const Tables = {
             code_head: 'vexNoteHeadMutedBreve',
             stem_offset: 0,
             getWidth(scale = Tables.DEFAULT_NOTATION_FONT_SCALE): number | undefined {
-              return new Glyph('vexNoteHeadMutedBreve', scale).getMetrics().width;
+              return Glyph.getWidth(Flow.DEFAULT_FONT_STACK, 'vexNoteHeadMutedBreve', scale);
             },
           },
           r: {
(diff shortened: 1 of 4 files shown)


src/tables.ts
```diff
@@ -3,6 +3,7 @@
 /* eslint-disable key-spacing */
 
 import { ArticulationStruct } from './articulation';
+import { Flow } from './flow';
```
rvilarl, line 6: It is not required to import `Flow`. This creates also a circular dependency. Please remove it.

@0xfe @ronyeh I just realised that this is a general issue in the code, I will make a PR to resolve the other ones which should use `Tables` rather than `Flow`.

src/tables.ts
```diff
@@ -821,22 +824,22 @@ export const Tables = {
             // Breve note
             code_head: 'noteheadDoubleWhole',
             getWidth(scale = Tables.DEFAULT_NOTATION_FONT_SCALE): number | undefined {
-              return new Glyph('noteheadDoubleWhole', scale).getMetrics().width;
+              return Glyph.getWidth(Flow.DEFAULT_FONT_STACK, 'noteheadDoubleWhole', scale);
```
rvilarl, line 827: Here and in the instances below use:
`Tables.DEFAULT_FONT_STACK`
rather than
`Flow.DEFAULT_FONT_STACK`

rvilarl (changes requested):
It is great to improve the perfomance, thank you!

rvilarl:
@tommadams have you used:
- `npm run reference` on master, and then,
- `npm run test:reference` on your branch
to check that no visual change was introduced?

src/tables.ts
tommadams: Unfortunately, using `Tables.DEFAULT_FONT_STACK` here doesn't work here because it doesn't get updated if a user assigns a new value to `Flow.DEFAULT_FONT_STACK`. I don't like the circular reference either but I saw that pattern was used elsewhere in the code without apparent problems.

tommadams:
I can confirm that this PR doesn't introduce any visual diffs.

src/tables.ts
tommadams: As I mentioned in my other comment below, I don't think the circular references can be avoided without some larger restructuring of the code :(

I hope I'm wrong though.

tommadams: Looking a little more, circular dependencies are pervasive throughout the codebase. For example [`types/common.d.ts`](https://github.com/0xfe/vexflow/blob/24f91ee58fa2de3536b56ee90604d2dab36e498c/src/types/common.d.ts#L1) transitively imports `Flow` via [`note.ts`](https://github.com/0xfe/vexflow/blob/24f91ee58fa2de3536b56ee90604d2dab36e498c/src/note.ts#L5).

src/tables.ts
rvilarl: I think that it should work as arrays are copied by reference. See https://jsfiddle.net/27s0L1xh/

src/tables.ts
rvilarl: Agree I am working on a PR removing those.

src/tables.ts
tommadams: Sorry, I didn't explain myself clearly.

The problem is that in the compiled JS code, `Flow` and `Tables` are different objects (because of how `Flow` is [initialised here](https://github.com/0xfe/vexflow/blob/24f91ee58fa2de3536b56ee90604d2dab36e498c/src/flow.ts#L80-L81)). Assigning a new array to `Flow.DEFAULT_FONT_STACK` cannot possibly affect the value of `Tables.DEFAULT_FONT_STACK`. Consequently, any code that refers to `DEFAULT_FONT_STACK` _must_ use `Flow.DEFAULT_FONT_STACK` rather than `Tables.DEFAULT_FONT_STACK` to function correctly.

The compiled JS code is effectively doing the following:

```javascript
let Tables = { DEFAULT_FONT_STACK: [1] };
let Flow = { DEFAULT_FONT_STACK: Tables.DEFAULT_FONT_STACK };
```

Clearly, assigning a new array to `Flow.DEFAULT_FONT_STACK` cannot affect the value of`Tables.DEFAULT_FONT_STACK`.

tommadams: One more thing I was perhaps not clear about:

The issue I'm referring to is when the a user assigns an entirely new array to `Flow.DEFAULT_FONT_STACK` rather than modifying the existing one in-place. This is an important use case because it's the [documented method](https://github.com/0xfe/vexflow/wiki/VexFlow-Font-Rendering#the-vexflow-font-stack) of modifying the font stack.

0xfe:
Thanks for the change @tommadams, and for the review @rvilarl! I'll accept this PR, and note the two issues for separate PRs.

1) Figure out how to remove circular dependencies from the code base.
2) Create a new explicit API call for changing the font stack. E.g., `Flow.setFontStack(...)`.

0xfe:
Optimizations look great -- thanks for doing them @tommadams! Merging.

Issue #1106, opened 2021-08-16
Target es6?


ronyeh:
https://github.com/0xfe/vexflow/blob/5480e70c4cfbe47df692d873a59e9441d3b90736/tsconfig.json#L3

When would we be comfortable changing the typescript compilation output to target es6? Are there lots of projects out there that use VexFlow and also require es5 compatibility?

One reason I ask is because I ran into an interesting es5 edge case regarding `extends` and `instanceof` in my branch exploring the implementation of `isCategory()`

If we do decide to target es6 (which I personally lean toward), and a developer discovers that they really need es5, they could probably just fork this project and change the one line in the tsconfig.json and then build their own es5-compatible version.

0xfe:
I think it's reasonable to target es6 for the next version bump. What's the browser adoption rate there? (i.e., what percent of browsers will break?)

ronyeh:
Here are some quick search results:

https://caniuse.com/?search=es6
- 95.42% supported worldwide
- 2.7% with partial support

https://caniuse.com/?search=es5
- 98.08% supported worldwide
- 1.63% partial support

https://www.w3schools.com/js/js_es6.asp
- Chrome 58 from Jan 2017
- Edge 14 from Aug 2016
- Firefox 54 from Mar 2017
- Safari 10 from Jul 2016
- Opera 55 from Aug 2018

ronyeh:
And here's a rather overwhelming compatibility table:
https://kangax.github.io/compat-table/es6/

But it seems like the biggest browser to break would be Internet Explorer which could be ~1.62% of the browser market:
https://www.stetic.com/market-share/browser/

0xfe:
Okay, great, seems like pretty low risk to switch, and wait and see if anyone complains.

mscuthbert:
> Okay, great, seems like pretty low risk to switch, and wait and see if anyone complains.

Nice to see this happening on the same day the Microsoft Office Online themselves dropped support for IE11.  Non-updating browsers are dead.  The only potential complete loss are people stuck on old versions of iOS Safari whose phones are too old to run Opera or Chrome iOS.  But people embedding VexFlow in projects can always compile from source to target a lower browser version.

PR #1108, opened 2021-08-18
Change target to es6.

ronyeh:
Fix https://github.com/0xfe/vexflow/issues/1106

I checked that `npm run test:reference` passed with zero visual diffs.


mscuthbert:
OH Hey I was just talking about this in the `instanceOf` thread.  If even Microsoft is no longer supporting IE11 in their own apps, I think we can take that route also.

The only possible concern is less tech-savvy profs who never update their macs and thus still have Safari 9 or 8 or 7 or (as my next-office-over-neighbor-has) 6.   But consuming clients who absolutely need ES5 (or even ES3!!!) support, can load the Vexflow sources themselves into their applications and define their own built target.

0xfe:
Great!

Issue #1109, opened 2021-08-18
Tests - run.ts produces different visual output if the imports and test cases are sorted alphabetically


ronyeh:
I tried cleaning up the run.ts tests by sorting everything alphabetically.

However, this changes the color of the stave lines in some of the tests. Perhaps the stave lines' color is initialized improperly either before or after the sort, due to the ordering of dependencies/imports?

Need to investigate this... as ideally the test suite should be more robust to sorting of imports.

0xfe:
Good catch! The test suite should be robust to this. Part of the problem here is that we have some global configuration state.

I think we can do two things here:

- Add a `VF.reset()` that restores all globals to it's original state. (Ideally, we'll have a class to capture all the globals to simplify this.
- Update our test harness (`VexFlowTests.runTest(...)`) code to call `VF.reset()` before each test begins.

Does that make sense?

PR #1110, opened 2021-08-18
Fix a test case that permanently modified Vex.Flow.STAVE_LINE_THICKNESS

ronyeh:
Summary: doing so allows us to alphabetically sort the tests and test imports in `run.ts` without breaking things.

---

In an older version of VexFlow, ( circa Oct 2015 / commit b8d27229d2ac69d732d7b8df0b4b75c1b6c8a609 ) `STAVE_LINE_THICKNESS` was 2 by default. Currently it is 1.

`staveconnector_tests` had a test case that temporarily set the thickness to 1, and then changed it back to 2. This worked back in the day, but now that the default is 1, it has the problem of permanently changing the thickness away from the default for all test cases that come after the StaveConnectorTests.

This PR results in 363 visual diffs, all related to the stave lines in test cases that were run after `StaveConnectorTests.Start()`;
The *reference images were wrong* because they do not render the stave lines correctly.

This fixes https://github.com/0xfe/vexflow/issues/1109

### DIFF
--- a/tests/staveconnector_tests.ts
+++ b/tests/staveconnector_tests.ts
@@ -19,7 +19,7 @@ const StaveConnectorTests = {
     const run = VexFlowTests.runTests;
     test('VF.* API', this.VF_Prefix);
     run('Single Draw Test', this.drawSingle);
-    run('Single Draw Test, 1px Stave Line Thickness', this.drawSingle1pxBarlines);
+    run('Single Draw Test, 4px Stave Line Thickness', this.drawSingle4pxStaveLines);
     run('Single Both Sides Test', this.drawSingleBoth);
     run('Double Draw Test', this.drawDouble);
     run('Bold Double Line Left Draw Test', this.drawRepeatBegin);
@@ -54,8 +54,9 @@ const StaveConnectorTests = {
     ok(true, 'all pass');
   },
 
-  drawSingle1pxBarlines(options: TestOptions, contextBuilder: ContextBuilder): void {
-    VF.STAVE_LINE_THICKNESS = 1;
+  drawSingle4pxStaveLines(options: TestOptions, contextBuilder: ContextBuilder): void {
+    const oldThickness = VF.STAVE_LINE_THICKNESS;
+    VF.STAVE_LINE_THICKNESS = 4;
     const ctx = contextBuilder(options.elementId, 400, 300);
     const stave = new Stave(25, 10, 300);
     const stave2 = new Stave(25, 120, 300);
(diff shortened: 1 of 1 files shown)


tests/staveconnector_tests.ts
```diff
@@ -19,7 +19,6 @@ const StaveConnectorTests = {
     const run = VexFlowTests.runTests;
     test('VF.* API', this.VF_Prefix);
     run('Single Draw Test', this.drawSingle);
-    run('Single Draw Test, 1px Stave Line Thickness', this.drawSingle1pxBarlines);
```
rvilarl, line 22: It seems to be the only test changing the Stave Line Thickness. So I suggest leaving it changing `1px` to `2px`.

tests/staveconnector_tests.ts
```diff
@@ -54,24 +53,6 @@ const StaveConnectorTests = {
     ok(true, 'all pass');
   },
 
-  drawSingle1pxBarlines(options: TestOptions, contextBuilder: ContextBuilder): void {
-    VF.STAVE_LINE_THICKNESS = 1;
```
rvilarl, line 58: Change `1` to `2`.

tests/staveconnector_tests.ts
```diff
@@ -54,24 +53,6 @@ const StaveConnectorTests = {
...
-    const stave2 = new Stave(25, 120, 300);
-    stave.setContext(ctx);
-    stave2.setContext(ctx);
-    const connector = new StaveConnector(stave, stave2);
-    connector.setType(VF.StaveConnector.type.SINGLE);
-    connector.setContext(ctx);
-    stave.draw();
-    stave2.draw();
-    connector.draw();
-    VF.STAVE_LINE_THICKNESS = 2;
```
rvilarl, line 70: Change `2` to `1`

rvilarl (changes requested):
@ronyeh good catch!

tests/staveconnector_tests.ts
ronyeh: Sure, I could save the default value, change it to something different, then restore the old default value at the end of the test case. Having a hard-coded 2 vs 1 will break again if we ever decide to change the default in the future.

rvilarl: Savig the default value would be better, you are right.

0xfe:
So why don't we do the following instead?

```
drawSingle1pxBarlines(options: TestOptions, contextBuilder: ContextBuilder): void {
    const oldThickness = VF.STAVE_LINE_TICKNESS;
    VF.STAVE_LINE_TICKNESS = 1;

    ... test code ...

    VF.STAVE_LINE_TICKNESS = oldThickness;
```

ronyeh:
Done! I used **4px stave lines**, so that it looks more visually obvious on the `flow.html` page and also in visual regression tests.

363 visual diffs, as expected.

0xfe:
Great!

PR #1094, opened 2021-08-02
Migration/tests

ronyeh:
This PR improves the tests while revealing ways we can improve (or fix) the TypeScript typings in VexFlow.

Jump down to the most recent discussion: https://github.com/0xfe/vexflow/pull/1094#issuecomment-906897447

---

Here is the entire list of 67 modified files:
```
tests/accidental_tests.ts
tests/annotation_tests.ts
tests/articulation_tests.ts
tests/auto_beam_formatting_tests.ts
tests/bach_tests.ts
tests/barline_tests.ts
tests/beam_tests.ts
tests/bend_tests.ts
tests/boundingbox_tests.ts
tests/boundingboxcomputation_tests.ts
tests/chordsymbol_tests.ts
tests/clef_tests.ts
tests/curve_tests.ts
tests/dot_tests.ts
tests/easyscore_tests.ts
tests/factory_tests.ts
tests/formatter_tests.ts
tests/fraction_tests.ts
tests/ghostnote_tests.ts
tests/glyphnote_tests.ts
tests/gracenote_tests.ts
tests/gracetabnote_tests.ts
tests/key_clef_tests.ts
tests/keymanager_tests.ts
tests/keysignature_tests.ts
tests/mocks.ts
tests/modifier_tests.ts
tests/multimeasurerest_tests.ts
tests/music_tests.ts
tests/notehead_tests.ts
tests/notesubgroup_tests.ts
tests/ornament_tests.ts
tests/parser_tests.ts
tests/pedalmarking_tests.ts
tests/percussion_tests.ts
tests/registry_tests.ts
tests/rests_tests.ts
tests/rhythm_tests.ts
tests/stave_tests.ts
tests/staveconnector_tests.ts
tests/stavehairpin_tests.ts
tests/staveline_tests.ts
tests/stavemodifier_tests.ts
tests/stavenote_tests.ts
tests/stavetie_tests.ts
tests/stringnumber_tests.ts
tests/strokes_tests.ts
tests/style_tests.ts
tests/tabnote_tests.ts
tests/tabslide_tests.ts
tests/tabstave_tests.ts
tests/tabtie_tests.ts
tests/textbracket_tests.ts
tests/textnote_tests.ts
tests/threevoice_tests.ts
tests/tickcontext_tests.ts
tests/timesignature_tests.ts
tests/tuning_tests.ts
tests/tuplet_tests.ts
tests/typeguard_tests.ts
tests/types/qunit.d.ts
tests/vexflow_test_helpers.ts
[...]

PR #1112, opened 2021-08-18
Sort Imports & Test Cases

ronyeh:
This PR is part of a larger effort to improve all the tests. It was broken off from a larger effort as documented in https://github.com/0xfe/vexflow/pull/1094

This PR produces no visual diffs from master.


tests/staveconnector_tests.ts
```diff
@@ -19,7 +19,6 @@ const StaveConnectorTests = {
     const run = VexFlowTests.runTests;
     test('VF.* API', this.VF_Prefix);
     run('Single Draw Test', this.drawSingle);
-    run('Single Draw Test, 1px Stave Line Thickness', this.drawSingle1pxBarlines);
```
rvilarl, line 22: As discussed jn #1110

tests/vf_prefix_tests.ts
```diff
@@ -0,0 +1,178 @@
...
+    equal(NoteSubGroup, VF.NoteSubGroup);
+    equal(Ornament, VF.Ornament);
+    equal(Parser, VF.Parser);
+    equal(PedalMarking, VF.PedalMarking);
+    equal(Registry, VF.Registry);
+    equal(Renderer, VF.Renderer);
+    equal(RepeatNote, VF.RepeatNote);
+    equal(Repetition, VF.Repetition);
+    equal(Stave, VF.Stave);
+    notEqual(Stave, VF.StaveNote);
```
rvilarl, line 144: Is this intended to be `notEqual`?

tests/support/qunit_api.ts
```diff
@@ -5,12 +5,21 @@
...
+//
+// According to https://qunitjs.com/upgrade-guide-2.x/
+// QUnit 2 no longer uses global functions. The assertion methods
+// are exposed through an assert object that is bound to each test.
+//
+// However, we stick with the global methods for readability.
+declare const global: any;
 export const QUnit = global.QUnit;
 export const test = global.test;
 export const expect = global.expect;
```
rvilarl, line 25: Is this still required? These are no longer imported from `declarations`?

rvilarl (changes requested):

tests/vf_prefix_tests.ts
ronyeh: Yeah, I was just asserting that Stave and **StaveNote** are not the same. I can remove that line if it looks ugly. :-)

tests/support/qunit_api.ts
ronyeh: I actually just renamed `declarations.ts` and moved it into the `support/` folder. I realized that declarations was all about QUnit types. It's good to keep this around because otherwise TypeScript complains about the QUnit globals in all the test files.

ronyeh:
The main contribution of this PR is `vf_prefix_tests.ts` which makes sure the classes are still available under the global `Vex.Flow.*`. The test cases has lines that look like: `equal(Note, VF.Note);`

There are no visual diffs relative to https://github.com/0xfe/vexflow/pull/1110

rvilarl (approved):

tests/support/qunit_api.ts
rvilarl: your choice :)

0xfe:
This looks great, Ron. I think we should update the wiki with some info around how these tests are laid out.

Also, the instructions on how to add a test.

Issue #1097, opened 2021-08-03
Voice contains Tickables or Notes?


ronyeh:
https://github.com/0xfe/vexflow/blob/37e70ad84f26ebce2546c52ac0c9c3e7c4a127c9/src/voice.ts#L40

I've made some more progress cleaning up the tests, and ran into this question. In the tests, we create a `MockTickable` object that gets added to a `Voice` or a `TickContext`.

`Voice` and `TickContext` both assume that tickables are all `Note` objects. **Is this always true**?

If so, that's fine and I was just confused on terminology. I can put an eslint-disable in `MockTickable` to hide the errors.

But if there is any other type of Tickable, then we will need to change references to `Note` to `Tickable` instead, and then add instanceof type checks into `Voice` where necessary.

rvilarl:
I believe that `Note` is the right answer. `Tickable` does not have `setStave` and `Note` does. `MockTickable` implements `setStave`.

rvilarl:
`Voice` contains Notes I believe. It does not contain TickContexts which is the only other class taht extends `Tickable`

ronyeh:
That's fine if Note[] is the intended type. I was just confused by naming. Since it's called tickables, setTickables, etc.... I assumed the original intention was to use the more generic Tickable class.

rvilarl:
I assume the same.

0xfe:
It was supposed to be a derivative of `Tickable`.

`Tickable` is supposed to be an abstract class (hence `MockTickable`), but the lack of type enforcement caused some drift over time.

With Typescript now, I think we should specify `abstract class Tickable` and add `abstract` methods `setStave`, `getStave`, etc. Can we try that?

rvilarl:
@ronyeh This will also work but then we should remove them from `Note`.

ronyeh:
OK I will explore making Tickable abstract and add appropriate abstract methods. 

I can do this with my continued cleanup of the tests.

The reason I got stuck here was because TS was complaining about MockTickable not being compatible with methods that expect a Note or a Tickable.

ronyeh:
I had to add some abstract methods to Tickable, to avoid adding too many type guards.

Additionally, there was a method signature incompatibility which requires us to break the existing API (which might be OK if we log it under **Breaking Changes** in a CHANGELOG.md).

Tickable had:
`addModifier(mod: Modifier): this {`

StaveNote (which extends StemmableNote ❮ Note ❮ Tickable) had:
[...]

PR #1116, opened 2021-08-22
Tickable

ronyeh:
This PR fixes https://github.com/0xfe/vexflow/issues/1097

While migrating tests (the one that used a MockTickable), I discovered that the the separation between Note, StaveNote, and Tickable was not very clear. Multiple methods declared that they accepted Tickable, but assumed that the tickables were all Notes or StaveNotes.

This PR makes the separation a bit more clear, and uses `Tickable` instead of `Note`/`StaveNote` where appropriate. I'm sure it can be improved more in the future as we discover more bugs/API-leaks/etc, but I think it's in a good enough state at the moment to be merged into master.

No visual diffs from current master.


src/tickable.ts
```diff
@@ -324,4 +351,38 @@ export abstract class Tickable extends Element {
...
+
+  /** Get `ModifierContext`. */
+  getModifierContext(): ModifierContext | undefined {
+    return this.modifierContext;
+  }
+
+  /**
+   * This is also used by things like beams and dots for positioning.
+   */
+  isRest(): boolean {
```
0xfe, line 376: What happens if we remove this? It feels semantically out of place.

tests/support/qunit_api.ts
```diff
@@ -5,12 +5,21 @@
 /* eslint-disable */
 
 // global.VF == Vex.Flow and is available in all of the tests.
+// TODO: Remove this after we have cleaned up the VF.* prefixes in the tests.
```
0xfe, line 8: I know I merged this in the other PR, however I just realized that the `support` directory is for drop-in third party code.

I'd prefer this to remain in the `tests/` directory -- we can call it `tests/qunit.ts`.

0xfe (commented):
Good change overall. Some quick comments here.

(And we should have a separate discussion about a `Tickable` interface, just concerned about the time, not positioning/rendering, which was the original intent.)

tests/support/qunit_api.ts
ronyeh: OK, will do.

src/tickable.ts
ronyeh: Originally it was in note.ts. I think I pulled it up into Tickable due to some TS warnings, but now that isCategory() is implemented, I think I can move it back into note.ts where it probably belongs!

However, I think the original comment is incorrect. Can you suggest a better comment?

note.ts:
```
  /**
   * Don't play notes by default, call them rests. This is also used by things like
   * beams and dots for positioning.
   */
  isRest(): boolean {
    return false;
  }
```

It says "call them rests" but the actual method returns false, so we aren't actually calling them rests. :-)

Or we can just use a simpler comment like in stavenote.ts:
```
  // Determine if current note is a rest.
  isRest(): boolean {
    return this.glyph.rest;
  }
```

There is another small "bug" that I have noted. this.glyph.rest is actually `true` or `undefined`... JS treats undefined as false in all the places we currently use isRest(), but at some point we should make sure isRest() actually returns a boolean.

ronyeh:
I think I found a good solution for QUnit:

https://github.com/0xfe/vexflow/blob/98d90995d4e5d44ece8fb2ecea2f84b0a04f1f84/tests/types/qunit.d.ts

I moved it into `tests/types/qunit.d.ts` since we also have a `src/types/common.d.ts` with type declarations.

Since `flow.html` provides all the QUnit assertion functions as global functions, I declared them inside `declare global { ... }` which makes the functions available in all test files without having to import the QUnit module. 

This makes the test code slightly closer to what "real code" might look like, since there are now no references to `assert.equal()`, or QUnit imports. The test harness blends into the background and makes the code easier to skim for someone who is learning VexFlow.

No visual diffs from master @ 366057d9d21c710de49c9553e9c4025d885a8615

src/tickable.ts
0xfe: Okay, let's make this work, and deal with the abstractions after. Re: comment "Returns true if this note is a type of rest. Rests don't have pitches, but take up space in the score."

0xfe:
Overall, looks good. If you can fix the conflict, I can merge. Nice job.

ronyeh:
Fixed conflict. No visual diffs from master @ e7f0426a662cc9fb28760b3c547cd3a2b357e0da.

0xfe:
Nice job!

PR #1117, opened 2021-08-23
Clean up EasyScore & Parser.

ronyeh:
Use TypeScript's Partial<Type> for EasyScoreOptions.
Use spread operator instead of Object.assign, as discussed in https://github.com/0xfe/vexflow/pull/999#discussion_r641946148
Parser should not know about EasyScore. Grammar is now a simple interface with one required begin() function.

### DIFF
--- a/src/easyscore.ts
+++ b/src/easyscore.ts
@@ -1,13 +1,11 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 // MIT License
 
-/* eslint max-classes-per-file: "off" */
-
 import { Accidental } from './accidental';
 import { Articulation } from './articulation';
 import { Factory } from './factory';
 import { FretHandFinger } from './frethandfinger';
-import { Parser, Match, Result, Rule, RuleFunction } from './parser';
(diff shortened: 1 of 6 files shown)


src/easyscore.ts
```diff
@@ -379,30 +372,19 @@ export class Builder {
...
-    this.commitHooks.forEach((fn) => fn(options, note, this));
+    this.commitHooks.forEach((commitHook) => commitHook(options, note, this));
 
     this.elements.notes.push(note);
     this.elements.accidentals.push(accidentals);
     this.resetPiece();
   }
 }
 
-function setId(options: IDUpdate, note: StaveNote) {
```
ronyeh, line 390: I moved these two helper functions to the bottom of the file.

src/parser.ts
```diff
@@ -3,7 +3,6 @@
 // A generic text parsing class for VexFlow.
 
 import { RuntimeError, log } from './util';
-import { Grammar } from './easyscore';
```
ronyeh, line 6: Parser is a generic class and should not know about EasyScore.

0xfe:
Nice job, Ron!

Issue #1115, opened 2021-08-20
Change: check<Stave>(this.stave) => defined(this.stave)


ronyeh:
I just recently learned that the TypeScript compiler does "type argument inference" on generics, so that you don't have to explicitly pass in the `<T>` in functions like `check<T>(obj)` and `isCategory<T>(obj, ...)`

Search for "type argument inference" here:
https://www.typescriptlang.org/docs/handbook/2/generics.html#:~:text=type%20argument%20inference

For example:
`const stave = check<Stave>(this.stave);`

Can be written as:
`const stave = check(this.stave);`

The compiler knows that the first argument is `Stave | undefined`, so the return type will be `Stave`.

Going forward, should we slowly remove the `<T>` from functions like `check<T>(obj)` when possible? Or should we keep the `<T>` so that it is more explicit?

ronyeh:
@rvilarl would love to hear your opinion on this, since you implemented the check() function.

ronyeh:
Advantages of each approach:

`const stave = check<Stave>(this.stave);`
* explicitly states the expected type
* TS compiler will show an error if `this.stave` does not match the type `<T>`

`const stave = check(this.stave);`
* shorter and possibly easier to read, because it makes use of TS type inference.

0xfe:
What happens if the variable has a union type? (e.g., `number | string | undefined`)?

0xfe:
Thinking about this more, the goal of those check statements is to assert that they have values set. The type system does the rest. So, yeah `check(this.stave)` should be sufficient I think.

ronyeh:
The implementation of check() verifies that it is NOT undefined. So the compiler is actually pretty smart about this.

```
      const x: number | string | undefined = Math.random() < 0.33 ? 1 : Math.random() < 0.5 ? 'hello' : undefined;
      const y = check(x);
```

If you hover the mouse pointer over the const y, the VS code tooltip says "const y: string | number". So the compiler decided that it can no longer be undefined, but it can still be either of the other two types.

0xfe:
That's great, this is way better. While we're at it should we rename it to something clearer, like `defined(x)` or `assertDefined(x)`? The latter sounds more like an assertion, but assertions don't typically return values, so I like the former.

ronyeh:
`const stave = defined(this.stave)` looks fine to me (or `checkDefined(this.stave)` if we want to be more verbose, haha).

`isDefined()` would imply returning a boolean.
[...]

PR #1123, opened 2021-08-30
Rename check<T>() to defined<T>().

ronyeh:
Also omit type `<T>` from calls to `defined(...)` when possible.

Some classes already have `checkStave()`. Implement `checkStave()` as a convenience method in other classes.


ronyeh:
Fixes https://github.com/0xfe/vexflow/issues/1115

0xfe:
Looks great. Thanks for the PR.

PR #1128, opened 2021-09-03
System.format not calculating this.boundingBox

rvilarl:
`System.format` is misiing the update of `this.boundingBox`


0xfe:
Thanks! Good find.

Issue #1130, opened 2021-09-04
Failure to build latest


AaronDavidNewman:
There are a couple of places like this in the test code.  Am I doing something wrong?

``` javascript
 catch (e) {
    equal(e.code, 'BadArguments', 'Invalid note: r');
  ...
```
```
ERROR in C:\Users\Daddy\Documents\GitHub\vexflow\tests\music_tests.ts
49:12-13
[tsl] ERROR in C:\Users\Daddy\Documents\GitHub\vexflow\tests\music_tests.ts(49,13)
      TS2571: Object is of type 'unknown'.
```

Easy to fix by making `catch (e: any)`, but I'd imagine the CI tool does not see this, so what's different with me?   
 
```
+ typedoc@0.21.9
+ @typescript-eslint/eslint-plugin@4.30.0
+ npm@7.22.0
+ webpack@5.52.0
+ @typescript-eslint/parser@4.30.0
```

0xfe:
All the tests are now Typescript. Thanks Aaron, did you try `npm i` again?

@ronyeh FYI

ronyeh:
I just pulled down master @ c712ef340fbc7de8f34a951561c47ad0584abe0e to a new directory and it builds:

```
git clone git@github.com:0xfe/vexflow.git .

npm install

grunt

>> webpack 5.48.0 compiled with 2 warnings in 10719 ms

grunt watch
```

Then I opened `flow.html` in a browser:
```
>> Tests completed in 16571 milliseconds. 6864 assertions of 6864 passed, 0 failed.
```

What version of TypeScript are you running?

**WAIT!!**: Now that I open the fresh git clone in my vs code, I do see the red squiggly underlines with `catch(e) {` and `Object is of type 'unknown'.ts(2571)`

Sorry we missed this. I can put a PR together with typings for the try / catch statements (unless there is a better solution).

ronyeh:
These were the library versions in my fresh git clone / npm install:
```
typescript@4.3.5
typedoc@0.21.5
@typescript-eslint/eslint-plugin@4.29.0
@typescript-eslint/parser@4.29.0
webpack@5.48.0
npm@7.20.3
```

I'm not a eslint / webpack guru. I wonder if there's a good way to throw errors or warnings for this issue.

ronyeh:
I'll submit a PR to fix this later this evening.

[...]

PR #1131, opened 2021-09-05
Replace try / catch blocks in tests with the QUnit throws() assertion.

ronyeh:
Fixes https://github.com/0xfe/vexflow/issues/1130

### DIFF
--- a/tests/keysignature_tests.ts
+++ b/tests/keysignature_tests.ts
@@ -37,11 +37,7 @@ function parser(): void {
   expect(11);
 
   function catchError(spec: string): void {
-    try {
-      Flow.keySignature(spec);
-    } catch (e) {
-      equal(e.code, 'BadKeySignature', e.message);
-    }
+    throws(() => Flow.keySignature(spec), /BadKeySignature/);
   }
 
   catchError('asdf');
@@ -52,7 +48,6 @@ function parser(): void {
   catchError('b');
   catchError('Kb');
   catchError('Fb');
-  catchError('Ab');
   catchError('Dbm');
   catchError('B#m');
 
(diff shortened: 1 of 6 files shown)


ronyeh:
OK, hopefully this fixes the build issue! I added a single space in the RuntimeError message string to make it easier to read.

I converted all the try / catch blocks to QUnit's throws().

For future reference, the official recommendation for try / catch blocks is to set the type of `e` to `unknown` and then do a type check before using it:

https://www.typescriptlang.org/play?ts=4.0.2#example/unknown-in-catch

```
try {
  // ..
} catch (e: unknown) {
  // You cannot use `e` at all until the type system learns what it is.
  e.stack; // <= compiler will complain!

  if (e instanceof SyntaxError) {
    e.stack;
  }
}
```

tests/keysignature_tests.ts
```diff
@@ -52,7 +48,6 @@ function parser(): void {
   catchError('b');
   catchError('Kb');
   catchError('Fb');
-  catchError('Ab');
```
rvilarl, line 55: Is this removal intentional?

rvilarl (changes requested):

tests/keysignature_tests.ts
ronyeh: Thanks for the review! Yes, it is intentional.

The previous catchError() method would do nothing if there was no exception thrown. When I changed it to QUnit's `throws()`, it is now asserting that an exception WILL be thrown. If the line of code succeeds, the assertion fails.

`Flow.keySignature('Ab')` is a valid key signature. It also appears 7 lines down, and runs without throwing any exception.

0xfe:
Thanks for the quick fix. The code is also easier to read.

PR #1132, opened 2021-09-06
Fix exception when calling preformat twice on TextDynamic

AaronDavidNewman:
See issue 1122.  If you call preCalculateMinWidth and then format, preformat is called twice.  This was resulting in the glyphs getting set twice, which caused an index out of range.  I just added a check to see if the glyphs had been created.


0xfe:
Thanks @AaronDavidNewman -- looks like tests are failing in the CI. @ronyeh FYI...

AaronDavidNewman:
Annoying :)  This should be fixed now.

tests/textnote_tests.ts
```diff
@@ -270,10 +270,16 @@ function textDynamics(options: TestOptions): void {
     { time: '7/4' }
   );
 
-  f.Formatter().joinVoices([voice]).formatToStave([voice], stave);
-
-  f.draw();
-
+  // This is refactored to use preCalculateMinWidth... to excersize
```
ronyeh, line 273: excersize typo

AaronDavidNewman: bless you sir :)

src/textdynamics.ts
```diff
@@ -91,6 +91,10 @@ export class TextDynamics extends Note {
   /** Preformat the dynamics text. */
   preFormat(): this {
     let total_width = 0;
+    // Issue 1122: length of this.glyphs must be <=
```
0xfe, line 94: You can remove this issue number.

Note that the standard pattern we use in other classes is to define a `this.preFormatted` boolean in the constructor.

Then here you just `if (preFormatted) return;`

I'd prefer we do that to stay consistent unless there's a good reason to do something different.

AaronDavidNewman: @0xfe  that is what I thought at first.  But looking into it, this is actually the same behavior as `StaveNote`.  When `formatter.format` is called, it calls `createTickContext` which sets preformat back to false.  So it is consistent, but not optimal.

The larger can of worms is that we should keep formatter from creating the tick contexts a second time.  But I think we might also need to add some checks to make sure that no modifiers etc. were added between the calls to `preCalculateMinTotalWidth` and `format`' that might affect the width calculation.  And I wasn't sure if there were other ways the list of tick contexts could be non-empty.  This would result in some tickables/modifiers being preformatted, and others not.

I'm happy to take this on also, but I don't know if it should be for this PR.  I know that the change for TextDynamics won't be a breaking change, it's harder to be sure with Formatter since it is used in different ways.

ronyeh:
> Thanks @AaronDavidNewman -- looks like tests are failing in the CI. @ronyeh FYI...

The tests were all migrated to typescript, and VF is no longer available as a magic global in the test files. From now, we import the desired class directly. 

If you still wish to use VF.StaveNote, etc, you'd need to assign the VF constant manually in your module. For example:

https://github.com/0xfe/vexflow/blob/e17079653cd5f8a5103fd7d09fb11187c5068f17/tests/vf_prefix_tests.ts#L94-L100

0xfe:
Thanks!

PR #1133, opened 2021-09-07
Minor gruntfile fix

tommadams:
Fix a minor regex bug in `Gruntfile.js`. The regex was previously matching files that end in `.t`, .`ts`, `.j` or `.js`. Fix it to only match `.ts` and `.js`.


0xfe:
Good catch, that was weird! :-)

PR #1135, opened 2021-09-07
`TickContext` not extending `Tickable`

rvilarl:
This PR implements the discussion below with @ronyeh. `TickContext`does not extend `Tickable` and it follows the same approach as `ModifierContext`.

**The discussion for reference:**

Let me try my best to analyze it objectively, but I am not familiar with the original intent of `Tickable` and `TickContext`.

---

### There are only two classes that `extends Tickable` (i.e., Note and TickContext). 
If we could make `TickContext` an independent type, then `Tickable` and `Note` could be merged.

So I investigated whether we can make `TickContext` not extend Tickable, by commenting it out `/*extends Tickable*/`.

This produces errors with these missing methods and properties:
```
this.setAttribute('type', 'TickContext'); 
this.width 
this.ignore_ticks = false;
this.setPreFormatted(false);
this.preFormatted
this.postFormatted
```

TickContext has similarities to ModifierContext. In `Formatter`, they are used in preFormat() and postFormat() to lay out tickables on a score. The ModifierContext is actually used inside Note's getMetrics(), which Formatter relies on.

**ModifierContext does NOT extend Modifier.** So if the classes are truly mirrors of each other, maybe TickContext should also not extend Tickable.

It's worth noting that ModifierContext implements its own fields for things like:
```
protected preFormatted: boolean = false;
protected postFormatted: boolean = false;
protected width: number;
```

So maybe TickContext can just implement the fields it needs, without extending Tickable. It does seem weird that a TickContext is a Tickable..... isn't it just a helper class for the Formatter to use?

Looking at the class comment for `Tickable`, we see "Tickable represents a element that sit on a score and has a duration, i.e., Tickables occupy space in the musical rendering dimension."

This suggests that indeed TickContext shouldn't really be a Tickable, since I don't think the TickContext sits on the score.

So what else could extend Tickable?

We could argue that some classes that extend Note should actually be extending Tickable instead. For example, is a Crescendo a Note? **Shouldn't Crescendo extend Tickable?** The class comment for Note includes "Notes have some common properties: All of them have a value (e.g., pitch, fret, etc.)" So while a crescendo has a duration like Notes and Tickables, it doesn't really have a pitch.

--- 

So in my professional, extremely biased opinion, I think we can merge the current Tickables PR, and then continue to think about what things should `extends Tickable` vs `extends Note`.

My hunch is that Tickable and Note should still be separate entities, but that TickContext should not extend Tickable, and implement the fields it needs to help the Formatter do its magic.

_Originally posted by @ronyeh in https://github.com/0xfe/vexflow/issues/1097#issuecomment-908556689_

### DIFF
--- a/src/tickcontext.ts
+++ b/src/tickcontext.ts
@@ -5,7 +5,7 @@
 // tabs, etc.
 
 import { RuntimeError } from './util';
-import { Tickable } from './tickable';
+import { FormatterMetrics, Tickable } from './tickable';
 import { Fraction } from './fraction';
 import { NoteMetrics } from './note';
 import { Stave } from './stave';
@@ -22,7 +22,7 @@ export interface TickContextOptions {
 /**
  * TickContext formats abstract tickable objects, such as notes, chords, tabs, etc.
  */
-export class TickContext extends Tickable {
+export class TickContext {
   protected readonly tickID: number;
   protected readonly tickables: Tickable[];
   protected readonly tickablesByVoice: Record<string, Tickable>;
@@ -45,10 +45,9 @@ export class TickContext extends Tickable {
   protected minTickable?: Tickable;
   tContexts: TickContext[];
 
-  // eslint-disable-next-line
-  draw(...args: any[]): void {
-    // DO NOTHING.
-  }
+  protected preFormatted: boolean = false;
+  protected postFormatted: boolean = false;
+  protected width: number;
 
   static getNextContext(tContext: TickContext): TickContext | undefined {
     const contexts = tContext.tContexts;
@@ -58,9 +57,7 @@ export class TickContext extends Tickable {
   }
 
   constructor(options?: TickContextOptions) {
-    super();
     this.tickID = options && options.tickID ? options.tickID : 0;
-    this.setAttribute('type', 'TickContext');
     this.currentTick = new Fraction(0, 1);
 
     this.maxTicks = new Fraction(0, 1);
@@ -85,6 +82,8 @@ export class TickContext extends Tickable {
     this.totalLeftPx = 0; // Total left pixels
     this.totalRightPx = 0; // Total right pixels
     this.tContexts = []; // Parent array of tick contexts
+
+    this.width = 0;
   }
 
   getTickID(): number {
@@ -198,7 +197,7 @@ export class TickContext extends Tickable {
 
   setCurrentTick(tick: Fraction): void {
     this.currentTick = tick;
-    this.setPreFormatted(false);
+    this.preFormatted = false;
   }
 
   addTickable(tickable: Tickable, voiceIndex?: number): this {
@@ -207,8 +206,6 @@ export class TickContext extends Tickable {
     }
 
     if (!tickable.shouldIgnoreTicks()) {
-      this.ignore_ticks = false;
-
       const ticks = tickable.getTicks();
 
       if (ticks.greaterThan(this.maxTicks)) {
@@ -228,7 +225,7 @@ export class TickContext extends Tickable {
     tickable.setTickContext(this);
     this.tickables.push(tickable);
     this.tickablesByVoice[voiceIndex || 0] = tickable;
-    this.setPreFormatted(false);
+    this.preFormatted = false;
     return this;
   }
 
@@ -271,12 +268,27 @@ export class TickContext extends Tickable {
     return this;
   }
 
-  getStave(): Stave | undefined {
-    throw new RuntimeError('NotImplemented', 'getStave() not implemented.');
-  }
+  getFormatterMetrics(): FormatterMetrics {
+    return {
+      // The freedom of a tickable is the distance it can move without colliding
+      // with neighboring elements. A formatter can set these values during its
+      // formatting pass, which a different formatter can then use to fine tune.
+      freedom: { left: 0, right: 0 },
 
-  // eslint-disable-next-line
-  setStave(stave: Stave): this {
-    throw new RuntimeError('NotImplemented', 'setStave() not implemented.');
+      // The simplified rational duration of this tick as a string. It can be
+      // used as an index to a map or hashtable.
+      duration: '',
+
+      // The number of formatting iterations undergone.
+      iterations: 0,
+
+      // The space in pixels allocated by this formatter, along with the mean space
+      // for tickables of this duration, and the deviation from the mean.
+      space: {
+        used: 0,
+        mean: 0,
+        deviation: 0,
+      },
+    };
   }
 }
--- a/src/tickcontext.ts
+++ b/src/tickcontext.ts
@@ -47,6 +47,7 @@ export class TickContext {
   protected preFormatted: boolean = false;
   protected postFormatted: boolean = false;
   protected width: number;
+  protected formatterMetrics: FormatterMetrics;
 
   static getNextContext(tContext: TickContext): TickContext | undefined {
     const contexts = tContext.tContexts;
@@ -83,6 +84,27 @@ export class TickContext {
     this.tContexts = []; // Parent array of tick contexts
 
     this.width = 0;
+    this.formatterMetrics = {
+      // The freedom of a tickable is the distance it can move without colliding
+      // with neighboring elements. A formatter can set these values during its
+      // formatting pass, which a different formatter can then use to fine tune.
+      freedom: { left: 0, right: 0 },
+
+      // The simplified rational duration of this tick as a string. It can be
+      // used as an index to a map or hashtable.
+      duration: '',
+
+      // The number of formatting iterations undergone.
+      iterations: 0,
+
+      // The space in pixels allocated by this formatter, along with the mean space
+      // for tickables of this duration, and the deviation from the mean.
+      space: {
+        used: 0,
+        mean: 0,
+        deviation: 0,
+      },
+    };
   }
 
   getTickID(): number {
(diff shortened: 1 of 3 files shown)


rvilarl:
There is one visual difference:

Current after the change:

Formatter Proportional_Formatting__20_iterations_ Bravura_Current

Reference:

Formatter Proportional_Formatting__20_iterations_ Bravura_Reference

sschmidTU:
the visual difference is a clear improvement!

ronyeh:
Wow the tuplets are spaced much more evenly now. Do you know which line or lines fixed the bug?

rvilarl:
> Wow the tuplets are spaced much more evenly now. Do you know which line or lines fixed the bug?

No, I could not figure it out. I just made `TickContext` independent from `Tickable`.

ronyeh:
:-D I'm not a fan of bug fixes we don't understand, so I pulled down your branch to dig around.

The test case that was fixed was:
```
Formatter.Proportional_Formatting__20_iterations_.Bravura_Current.png
Formatter.Proportional_Formatting__20_iterations_.Bravura_Reference.png
```

I opened flow.html with your build and compared it against a recent reference:
```
/tests/flow.html?module=Formatter&filter=20%20iter
/tests/flow.html?module=Formatter&filter=20%20iter&ver=reference
```

By adding query parameter `ver=reference` to the end of the flow.html, we can pull the vexflow-tests.js from the reference/ directory.

I uncommented the block of debug code at the bottom of the `proportional()` test case:
```
  const typeMap = Registry.getDefaultRegistry().index.type;
  const table = Object.keys(typeMap).map((typeName) => typeName + ': ' + Object.keys(typeMap[typeName]).length);
  console.log(table);
```

Your PR branch prints:
```
[
    "Base: 0",
    "System: 1",
    "Tickable: 0",
    "Note: 0",
    "StemmableNote: 0",
    "StaveNote: 21",
    "Stem: 21",
    "NoteHead: 42",
    "Glyph: 68",
    "Tuplet: 3",
    "Voice: 5",
    "Stave: 5",
    "StaveModifier: 0",
    "Barline: 10",
    "Clef: 5",
    "TimeSignature: 5",
    "StaveConnector: 1"
]
```

Reference prints:
```
[
    "Base: 0",
    "System: 1",
    "Tickable: 0",
    "Note: 0",
    "StemmableNote: 0",
    "StaveNote: 21",
    "Stem: 21",
    "NoteHead: 42",
    "Glyph: 68",
    "Tuplet: 3",
    "Voice: 5",
    "Stave: 5",
    "StaveModifier: 0",
    "Barline: 10",
    "Clef: 5",
    "TimeSignature: 5",
    "StaveConnector: 1",
    "TickContext: 32"
]
```

❗❗❗ Note the 32 TickContext objects in the Registry in the reference version.

The class comment for Tickable says "Tickables occupy space in the musical rendering dimension." 

So my current guess is that TickContexts were added into the voice and the Formatter placed it onto the score. They were basically ghost ticks that took up space.

I'll post any further findings here!

src/tickcontext.ts
```diff
@@ -271,12 +267,27 @@ export class TickContext extends Tickable {
...
+      iterations: 0,
+
+      // The space in pixels allocated by this formatter, along with the mean space
+      // for tickables of this duration, and the deviation from the mean.
+      space: {
+        used: 0,
+        mean: 0,
+        deviation: 0,
+      },
+    };
```
ronyeh, line 291: This method always returns a dummy object with everything set to 0.

In formatter.ts, Formatter.evaluate() modifies the dummy object, so it's basically a no-op.

Formatter.tune() uses information from getFormatterMetrics(). However, your PR returns a dummy object so tune() gets all zeroes.

I confirmed that if I comment out this line, the buggy test case will look like your PR:
https://github.com/0xfe/vexflow/blob/490970b96aace865669d5d2d44acfeb349bc112f/src/system.ts#L229

So maybe tune() is buggy?

rvilarl:
> :-D I'm not a fan of bug fixes we don't understand, so I pulled down your branch to dig around.

I have the same oppinion, thanks for looking into it :)

> Your PR branch prints:
> 
> ```
> [
>     "Base: 0",
>     "System: 1",
>     "Tickable: 0",
>     "Note: 0",
>     "StemmableNote: 0",
>     "StaveNote: 21",
>     "Stem: 21",
>     "NoteHead: 42",
>     "Glyph: 68",
>     "Tuplet: 3",
>     "Voice: 5",
>     "Stave: 5",
>     "StaveModifier: 0",
>     "Barline: 10",
>     "Clef: 5",
>     "TimeSignature: 5",
>     "StaveConnector: 1"
> ]
> ```
> 
> Reference prints:
> 
> ```
> [
>     "Base: 0",
>     "System: 1",
>     "Tickable: 0",
>     "Note: 0",
>     "StemmableNote: 0",
>     "StaveNote: 21",
>     "Stem: 21",
>     "NoteHead: 42",
>     "Glyph: 68",
>     "Tuplet: 3",
>     "Voice: 5",
>     "Stave: 5",
>     "StaveModifier: 0",
>     "Barline: 10",
>     "Clef: 5",
>     "TimeSignature: 5",
>     "StaveConnector: 1",
>     "TickContext: 32"
> ]
> ```
> 
>  Note the 32 TickContext objects in the Registry in the reference version.

This is expected, `TickContext` does no longer extend `Element`. I took this out to follow the `ModifierContext` approach.

ronyeh:
Yeah, I'm still not sure if we "fixed" a bug or introduced a bug.

First, I think the `getFormatterMetrics()` should probably follow the approach taken in `Tickable`. It shouldn't return a new object every time its called, otherwise there is no way for `tune()` to do what it wants to do.

Now the question is... what is `tune()` supposed to do? I looked at the proportional formatting test case with 20 iterations and I tweaked the number of iterations multiple times. In the current master, the layout seems to look best at 0 iterations, and it gets progressively more distorted as we increase it to 20 or even 25 iterations.

However, if we pull up an old test page: https://vexflow.com/tests/index.html?module=Formatter

Scroll to the bottom. You'll see that with 2 iterations, the formatting looks distorted, but it gets more and more evened out at 14 iterations. The tune() algorithm seems to be trying to even out the 7-tuplet so that its rhythm visually align with the other tuplets above it.

@0xfe is tune() no longer necessary? What's the purpose of tune()?

ronyeh:
The comment for `tune()` is actually very clear.

```
  /**
   * Run a single iteration of rejustification. At a high level, this method calculates
   * the overall "loss" (or cost) of this layout, and repositions tickcontexts in an
   * attempt to reduce the cost. You can call this method multiple times until it finds
   * and oscillates around a global minimum.
   * @param alpha the "learning rate" for the formatter. It determines how much of a shift
   * the formatter should make based on its cost function.
   */
  tune(options?: { alpha: number }): number {
```

Clearly the intention is that the more you call tune(), the "better" it should look.

So maybe there is a bug in that we are calculating the cost incorrectly, or the adjustment methods are doing the wrong thing?

rvilarl:
> The comment for `tune()` is actually very clear.
> 
> ```
>   /**
>    * Run a single iteration of rejustification. At a high level, this method calculates
>    * the overall "loss" (or cost) of this layout, and repositions tickcontexts in an
>    * attempt to reduce the cost. You can call this method multiple times until it finds
>    * and oscillates around a global minimum.
>    * @param alpha the "learning rate" for the formatter. It determines how much of a shift
>    * the formatter should make based on its cost function.
>    */
>   tune(options?: { alpha: number }): number {
> ```
> 
> Clearly the intention is that the more you call tune(), the "better" it should look.
> 
> So maybe there is a bug in that we are calculating the cost incorrectly, or the adjustment methods are doing the wrong thing?

The bug must be in tune() and this is now a different issue, thank you for your help. I have made formatterMetrics a member varaible and there is now no visual difference.

rvilarl:
Ready to review and merge.

0xfe:
Thanks for this folks, the render looks better, but something is off here.

1) The duplicated `formatterMetrics` is a smell that suggests that `TickContext` must not be separated from `Tickable`
2) Rodrigo, which test exactly are those images from?
3) Ron, `tune` is a red herring, none of our code paths should be calling it -- it was for the old formatter. (I'd like to remove it eventually, but keeping it to support some old code.)

sschmidTU:
Maybe mark `tune` deprecated then?

rvilarl:
> Thanks for this folks, the render looks better, but something is off here.
> 
> 1. The duplicated `formatterMetrics` is a smell that suggests that `TickContext` must not be separated from `Tickable`

On the other side the lack of need for `draw`, `setStave`, `getStave` suggests the opposite. `ModifierContext` is also not a `Modifier`.

> 2. Rodrigo, which test exactly are those images from?

Formatter.Proportional_Formatting__20_iterations_.Bravura_Current.png
Formatter.Proportional_Formatting__20_iterations_.Bravura_Reference.png

With the last commit, there is no longer  a visual differences.

> 3. Ron, `tune` is a red herring, none of our code paths should be calling it -- it was for the old formatter. (I'd like to remove it eventually, but keeping it to support some old code.)

ronyeh:
I believe `TickContext` should not extend `Tickable`.

From the class comments: 
```
Tickable represents a element that sit on a score and has a duration, i.e., Tickables occupy space in the musical rendering dimension.

TickContext formats abstract tickable objects, such as notes, chords, tabs, etc.
```

If those comments are roughly correct, it implies that a TickContext is a helper object that places Tickables onto the score. 

TickContext has a `addTickable()` which pushes Tickables onto its internal `this.tickables` array. During `preFormat()`, the TickContext loops through all of its Tickables and calls `tickable.preFormat()` on each Tickable.

Will a TickContext ever contain child TickContexts in its `this.tickables` array? If not, then I think it's safe to have TickContext no longer extend Tickable.




I think these are all candidates for removal:
* `Formatter.tune()` – Currently, System.format() is the only place that calls tune(), inside a for loop over `this.options.formatIterations`. @0xfe said that this is for legacy code, so I assume `formatIterations` is 0 in modern VexFlow apps. Question for @0xfe: if tune() is buggy when used with our current formatter, can we just remove the tune() for loop in `System.format()`?
* `Formatter.evaluate()` – The number it calculates is never used anywhere. It's extra work every time we call format().
* If we get rid of `tune()` and `evaluate()`, then we can get rid of `formatterMetrics` too, I believe. The only other place that references it is `Note.plotMetrics()`, which is a debug visualization for the numbers that evaluate() calculated.


In summary I believe that:
* TickContext is a helper for Tickable, and TickContext is **not** a Tickable.
* We should remove the two instances of `return this.evaluate()` from `Formatter.preFormat()`, and return `this` instead. Separate PR?
* Deprecate `tune()`, `evaluate()`, and `getFormatterMetrics()` related methods & options. Put in the CHANGELOG that these will be candidates for complete removal in a future version bump. Separate PR? (Or if we are extra bold & spicy, simply remove them all from this upcoming 4.0.0 version, and refer folks to 3.0.9 if they want to use `tune()`).

0xfe:
Alright, thanks for the thorough investigation -- I'm not 100% convinced of this, but I trust your judgement and it's better than the status quo, so merging!

Thanks @rvilarl and @ronyeh!

Issue #1127, opened 2021-09-02
Improvements to Glyph Rendering in SVGContext / CanvasContext


ronyeh:
This thread has evolved into discussing ways to improve Glyph rendering. I'll keep this issue open for now, even though there is no bug to fix. If we prefer to move it to the GitHub discussions tab, I can close off this issue.

---

Original title:
## Should OutlineCode enum (in glyph.ts) use ints other than 0, 1, 2, 3?

This isn't a _bug_, but I was reading glyph.ts and saw that `GlyphOutline.parse()` returns a `number[]` where the **_position_** of the number changes its meaning.

For example, Bach Demo produces an array like:
`[0,0,60,3,140,180,0,135,62,180,3,425,60,268,180,425,62,3,285,180,425,134,367,180,3,0,60,127,180,0,63]`

The first 0 is actually a "move to" instruction, and the second 0 is an x coordinate with value 0, and then comes a y coordinate with value 60. The 3 after that starts a bezier curve....

Then, `Glyph.renderOutline()` and `Glyph.getOutlineBoundingBox()` loop over these outline arrays to draw the glyphs to the context.

```
    while (!go.done()) {
      switch (go.next()) {
        case OutlineCode.MOVE: // => 0
          ctx.moveTo(go.nextX(), go.nextY());
          break;
        case OutlineCode.LINE: // => 1
          ctx.lineTo(go.nextX(), go.nextY());
          break;
          ...
```

So the switch statement is looking for numbers 0, 1, 2, 3.

Anyways, like I said, this isn't a bug because the three methods all follow the same convention (a number 0 implies an x, y and a number 3 is followed by 6 ints specifying the points of a cubic bezier curve).

I know it was part of @tommadams massive improvements on glyph performance & correctness in https://github.com/0xfe/vexflow/pull/1103 and  https://github.com/0xfe/vexflow/pull/1077.

Obviously we don't want to go back to the old / slow way. 

But do we have any ideas on if / how we could improve this, or make it potentially less brittle / future proof?

I can't think of anything off the top of my head at the moment, other than starting the enum at a really big number (that is still well under `Number.MAX_SAFE_INTEGER`), like:

```
export const enum OutlineCode {
  MOVE = 0xfffffffffff0,
  LINE = 0xfffffffffff1,
  QUADRATIC = 0xfffffffffff2,
  BEZIER = 0xfffffffffff3,
}
```

Anyways, can we improve this in any way? If not, I'll close as "not a bug."

[...]

PR #1137, opened 2021-09-08
Clean up text measuring & add a text measurement cache to SVGContext

tommadams:
While investigating #1127, I noticed that measuring text in the `SVGContext` is quite slow because it forces a DOM reflow, so I added a cache. This reduces the time taken to run all tests from _12.5 seconds_ to _10.5 seconds_ on my machine.

In real world use cases that lay out systems one time only, this change will probably only have a marginal effect. In interactive cases, where the same page is repeatedly laid out with only minor changes each time (like my personal use case), this will greatly speed up all text measurement.

Changes in this PR:
 - Return a `TextMeasure` object from `RenderContext.measureText` that has both `width` & `height`. Previously, either a `SVGRect` or `TextMetrics` object would be returned depending on whether the context was canvas or SVG. In particular, `height` is always set (it was missing from the `TextMetrics` returned by `CanvasContext.measureText`.
 - `CanvasContext.measureText` calculates the text height directly from the font size.

I also fixed a bug in `TextNote` where the tick context width was incorrectly being used to position superscript and subscript text instead of the main text's width.

To make this PR simpler, I removed a couple of features:
 - Removed `Renderer.bolsterCanvasContext` and always return a `CanvasContext`. I measured and this does not make any measurable difference to the performance: actually very little time is spent in the calls to the render contexts themselves. This avoids having to do an ugly hack to make `measureText` work because that method already exists on `CanvasRenderingContext2D`.
 - Removed the IE polyfill for italic text: IE is no longer supported since #1108.

I also removed the `bounds` attribute from `Stave` because it isn't actually used for anything.

### DIFF
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -1,6 +1,6 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 // MIT License
-import { RenderContext } from './types/common';
+import { RenderContext, TextMeasure } from './types/common';
 import { warn } from './util';
 
 /**
@@ -10,6 +10,7 @@ export class CanvasContext implements RenderContext {
   vexFlowCanvasContext: CanvasRenderingContext2D;
   canvas: HTMLCanvasElement | { width: number; height: number };
   background_fillStyle?: string;
+  textHeight: number = 0;
 
   static get WIDTH(): number {
     return 600;
@@ -38,9 +39,6 @@ export class CanvasContext implements RenderContext {
   }
 
   /**
-   * This constructor is only called if Renderer.USE_CANVAS_PROXY is true.
-   * In most instances, we do not need to create a CanvasContext object.
-   * See Renderer.bolsterCanvasContext().
    * @param context
    */
   constructor(context: CanvasRenderingContext2D) {
@@ -76,11 +74,17 @@ export class CanvasContext implements RenderContext {
 
   setFont(family: string, size: number, weight: string): this {
     this.vexFlowCanvasContext.font = (weight || '') + ' ' + size + 'pt ' + family;
+    this.textHeight = (size * 4) / 3;
     return this;
   }
 
   setRawFont(font: string): this {
     this.vexFlowCanvasContext.font = font;
+
+    const fontArray = font.split(' ');
+    const size = Number(fontArray[0].match(/\d+/));
+    this.textHeight = (size * 4) / 3;
+
     return this;
   }
 
@@ -129,15 +133,8 @@ export class CanvasContext implements RenderContext {
     return this;
   }
 
-  // setLineDash is the one native method in a canvas context
-  // that begins with set, therefore we don't bolster the method
-  // if it already exists (see Renderer.bolsterCanvasContext).
-  // If it doesn't exist, we bolster it and assume it's looking for
-  // a ctx.lineDash method, as previous versions of VexFlow
-  // expected.
   setLineDash(dash: number[]): this {
-    // eslint-disable-next-line
-    (this.vexFlowCanvasContext as any).lineDash = dash;
+    this.vexFlowCanvasContext.setLineDash(dash);
     return this;
   }
 
@@ -222,8 +219,12 @@ export class CanvasContext implements RenderContext {
     return this;
   }
 
-  measureText(text: string): TextMetrics {
-    return this.vexFlowCanvasContext.measureText(text);
+  measureText(text: string): TextMeasure {
+    const metrics = this.vexFlowCanvasContext.measureText(text);
+    return {
+      width: metrics.width,
+      height: this.textHeight,
+    };
   }
 
   fillText(text: string, x: number, y: number): this {
@@ -242,7 +243,7 @@ export class CanvasContext implements RenderContext {
   }
 
   set font(value: string) {
-    this.vexFlowCanvasContext.font = value;
+    this.setRawFont(value);
   }
 
   get font(): string {
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -83,6 +83,8 @@ export class CanvasContext implements RenderContext {
 
     const fontArray = font.split(' ');
     const size = Number(fontArray[0].match(/\d+/));
+    // The font size is specified in points, scale it to canvas units.
+    // CSS specifies dpi to be 96 and there are 72 points to an inch: 96/72 == 4/3.
     this.textHeight = (size * 4) / 3;
 
     return this;
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -33,13 +33,6 @@ export class Renderer {
     DOWN: 3, // Downward leg
   };
 
-  /**
-   * Set this to true if you're using VexFlow inside a runtime
-   * that does not allow modifying canvas objects. There is a small
-   * performance degradation due to the extra indirection.
-   */
-  static readonly USE_CANVAS_PROXY = false;
-
   static lastContext?: RenderContext = undefined;
 
   static buildContext(
@@ -67,43 +60,6 @@ export class Renderer {
 
   static getSVGContext(elementId: string, width: number, height: number, background?: string): RenderContext {
     return Renderer.buildContext(elementId, Renderer.Backends.SVG, width, height, background);
-  }
-
-  // eslint-disable-next-line
-  static bolsterCanvasContext(ctx: any): RenderContext {
-    if (Renderer.USE_CANVAS_PROXY) {
-      return new CanvasContext(ctx);
-    }
-
-    // Modify the CanvasRenderingContext2D to include the following methods, if they do not already exist.
-    // TODO: Is a Proxy object appropriate here?
-    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy
-    const methodNames = [
-      'clear',
-      'setFont',
-      'setRawFont',
-      'setFillStyle',
-      'setBackgroundFillStyle',
-      'setStrokeStyle',
-      'setShadowColor',
-      'setShadowBlur',
-      'setLineWidth',
-      'setLineCap',
-      'openGroup',
-      'closeGroup',
-      'getGroup',
-    ];
-
-    ctx.vexFlowCanvasContext = ctx;
-
-    methodNames.forEach((methodName) => {
-      if (!(methodName in ctx)) {
-        // eslint-disable-next-line
-        ctx[methodName] = (CanvasContext.prototype as any)[methodName];
-      }
-    });
-
-    return ctx;
   }
 
   // Draw a dashed line (horizontal, vertical or diagonal
@@ -172,7 +128,7 @@ export class Renderer {
       if (!canvasElement.getContext) {
         throw new RuntimeError('BadElement', `Can't get canvas context from element: ${canvasId}`);
       }
-      this.ctx = Renderer.bolsterCanvasContext(canvasElement.getContext('2d'));
+      this.ctx = new CanvasContext(canvasElement.getContext('2d')!);
     } else if (this.backend === Renderer.Backends.SVG) {
       this.ctx = new SVGContext(this.element);
     } else {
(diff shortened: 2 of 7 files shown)


tommadams:
Oh, it looks like I merged the most recent changes into my branch incorrectly. Let me take a look.

sschmidTU:
Stave.bounds may not be used in Vexflow, but it could be used by existing applications. Even though you can get the same values with stave.x, stave.y and stave.width, we probably shouldn't risk breaking applications that use it. It's basically a public field.
(i checked and our application doesn't, but you know)

src/svgcontext.ts
```diff
@@ -113,18 +188,20 @@ export class SVGContext implements RenderContext {
...
   }
 
+  create(svgElementType: 'g'): SVGGElement;
+  create(svgElementType: 'path'): SVGPathElement;
+  create(svgElementType: 'rect'): SVGRectElement;
+  create(svgElementType: 'svg'): SVGSVGElement;
+  create(svgElementType: 'text'): SVGTextElement;
   create(svgElementType: string): SVGElement {
     return document.createElementNS(SVG_NS, svgElementType);
   }
```
ronyeh, line 200: Hey, I learned something new (overload signatures)!

https://www.typescriptlang.org/docs/handbook/2/functions.html#overload-signatures-and-the-implementation-signature

Do you think others might be confused that the last signature `create(svgElementType:string)` cannot be used directly?

Maybe we need a comment that create() can only be called with 5 different parameters: svgElementType = 'g' | 'path' | ....

tommadams:
> `Stave.bounds` may not be used in Vexflow, but it could be used by existing applications. Even though you can get the same values with stave.x, stave.y and stave.width, we probably shouldn't risk breaking applications that use it. It's basically a public field.
> (i checked and our application doesn't, but you know)

`Stave.bounds` was a protected member, so if anyone is using it they should be ashamed of themselves.

tommadams:
I did some more research into the text height issue. It turns out that the height of the text bounding box returned by `SVGTextElement.getBBox()` is a fixed value specified by the font. The height returned is greater or equal to the tallest glyph in the font.

On the other hand, the `actualBoundingBoxAscent` value returned by `CanvasRenderingContext2D.measureText()` is the height of the bounding box that tightly fits _only_ the glyphs being rendered. `measureText()` does also return `fontBoundingBoxAscent` on a subset of modern browsers, which contains what we'd like but it's currently only supported by about [56% of browsers](https://caniuse.com/?search=fontBoundingBoxAscent).

However, it turns out that you can trivially convert from a font's point size to the context coordinates by scaling by a factor of 96 / 72 (or more simply 4 / 3). The 96 is the DPI [specified in CSS](https://www.w3.org/TR/css3-values/#resolution) and 1 point is [1/72th of an inch](https://www.w3.org/TR/css3-values/#absolute-lengths).

Since that's by far the easiest solution, that's what I chose to do and now both `CanvasContext` and `SVGContext` measure text height the same way, without any hacks.

Current head (bad alignment on the superscripts; the spacing of superscripts and subscripts is different between canvas and svg, most noticeably on the `ii^6_4`):
old

With this PR (all fixed!):
new

This does introduce a few minor visual diffs due to text height values now being slightly different:
TextBracket Simple_TextBracket Bravura
TextNote TextNote_Superscript_and_Subscript Petaluma

src/renderer.ts
```diff
@@ -69,43 +62,6 @@ export class Renderer {
...
-
-    methodNames.forEach((methodName) => {
-      if (!(methodName in ctx)) {
-        // eslint-disable-next-line
-        ctx[methodName] = (CanvasContext.prototype as any)[methodName];
-      }
-    });
-
-    return ctx;
-  }
```
ronyeh, line 107: So we are effectively setting USE_CANVAS_PROXY to true from now on? If the performance impact is minimal, then I'm totally on board with the simpler approach. I don't know the history behind USE_CANVAS_PROXY and it is not currently referenced anywhere in the tests.

And in fact, we are already proxying almost all of the method calls, since we modify the CanvasRenderingContext2D to have a new instance var `vexFlowCanvasContext` and then add on new methods which close but not exactly the same as the CanvasRenderingContext2D API's :-D. I guess the main difference is that the VexFlow methods are chainable.

src/svgcontext.ts
```diff
@@ -60,13 +116,11 @@ export class SVGContext implements RenderContext {
...
   groups: SVGGElement[];
   fontString: string = '';
-  fontSize: number = 0;
-  ie: boolean = false; // true if the browser is Internet Explorer.
 
   constructor(element: HTMLElement) {
     this.element = element;
 
-    const svg = this.create('svg') as SVGSVGElement;
+    const svg = this.create('svg');
```
ronyeh, line 123: Nice work with the overload signatures. Makes this a bit cleaner.

src/svgcontext.ts
tommadams: Oops, I forgot to add the generic overload. Passing an arbitrary string to `create` will compile now.

src/renderer.ts
tommadams: I was unable to measure any performance impact. Modern JS engines already inline on hot functions, so if there was any overhead at all, it would come from the code optimizer having to work a tiny bit harder.

The primary reason I tried deleting it to check what the performance impact might be is that `CanvasRenderingContext2D` already has a `measureText` method. So adding a `measureText` method to the `CanvasContext` that called into the `CanvasRenderingContext2D` would require stuffing in a `realMeasureText` attribute too, which just smelt bad.

Following on from the discussion in #1127, I'd like to add a `fillGlyph` method to the context ([I have it written for the `SVGContext` already](https://github.com/tommadams/vexflow/blob/1ee462ee01f1f47314d9ad80e7ebce0cad81c4fa/src/svgcontext.ts#L656)). Once this is added, the canvas context's `fillGlyph` would call the `moveTo`, `lineTo`, `bezierCurveTo`, etc directly on the `CanvasRenderingContext2D` context, which would elide a lot of the wrapped calls.

src/svgcontext.ts
tommadams: Yeah, TypeScript's type system is pretty nice on the whole.

src/renderer.ts
ronyeh: CanvasContext `fillGlyph()` sounds like a good idea! The current loop where we process path instructions and then call moveTo, lineTo, etc isn't ideal. Even if that loop were just moved into CanvasContext, it would remove the proxy overhead, as you noted. But if the SVG paths could somehow be shoved straight into a Canvas Path2D, that'd be extra cool. For example, could we use a single Path2D instance for all of the black noteheads? We might have to make use of a lot of: 
```
    ctx.translate(x,y);
    ctx.stroke(noteheadPath);
    ctx.translate(x,y);
    ctx.stroke(noteheadPath);
```
And probably also `ctx.save()` and `ctx.restore()`.

ronyeh:
> `Stave.bounds` was a protected member, so if anyone is using it they should be ashamed of themselves.

The `private`, `protected`, `(empty/public)` access modifiers didn't exist earlier this year. We migrated VexFlow to TypeScript starting around 5 months ago (thanks in large part to @rvilarl with guidance from @0xfe and feedback from community members).

So > 5 months ago, every field was public.... because JavaScript. :-)

So we have no idea if someone out there is relying on a once public, recently-protected field that we decide to remove, unless they file a bug and ask us to reimplement.

If we do decide that a field is _useless_, then I'm in favor of removing it. It's not fun to keep old crusty code forever. We can just put it in the CHANGELOG as a possible issue while upgrading from VexFlow 3.0.9 to 4.0.0.

---

We might need to do a review of the access modifiers and adjust (or add accessors) if needed.

I feel like for a library like VexFlow, we should prefer `protected` over private. This allows a developer to subclass / customize one of our classes, without having to fork VexFlow and modify the code directly.

If something truly needs to be **private**, maybe it can be moved outside of the class body. Every one of our files is a JS module, so any let or const outside of the class declaration is private to that module anyways.

Some of our classes still have public instance variables (e.g., Beam, Note, StaveNote). Which one of those fields should actually be protected? For example, Beam is the only one with a public `postFormatted: boolean;`

src/svgcontext.ts
```diff
@@ -113,18 +167,21 @@ export class SVGContext implements RenderContext {
...
+  create(svgElementType: 'text'): SVGTextElement;
+  create(svgElementType: string): SVGElement;
   create(svgElementType: string): SVGElement {
     return document.createElementNS(SVG_NS, svgElementType);
   }
 
   // Allow grouping elements in containers for interactivity.
   openGroup(cls: string, id?: string, attrs?: { pointerBBox: boolean }): SVGGElement {
-    const group: SVGGElement = this.create('g') as SVGGElement;
+    const group: SVGGElement = this.create('g');
```
ronyeh, line 184: Should we remove `: SVGGElement`? TypeScript can infer the type, and the variable name is pretty descriptive.

We can keep explicit local variable types when there is potential for confusion.

Your call. Either way works for me!

src/svgcontext.ts
```diff
@@ -113,18 +167,21 @@ export class SVGContext implements RenderContext {
     };
 
     this.state_stack = [];
-
-    // Test for Internet Explorer
-    this.iePolyfill();
   }
 
+  create(svgElementType: 'g'): SVGGElement;
```
ronyeh, line 172: Can we add a comment along the lines of:
```
  /**
   * Use one of the overload signatures to create an SVG element of a specific type.
   * The last overload accepts an arbitrary string, and is identical to the implementation signature.
   */
```
It's probably redundant for TS gurus, but this syntax was new to me and a comment would have helped me for sure.

src/renderer.ts
tommadams: That's a neat idea. You'd need to handle scaling and make sure the context's transform matrix was reset at the end of glyph rendering but it could definitely work.

src/svgcontext.ts
tommadams: Done

src/svgcontext.ts
tommadams: Done

tommadams:
Thanks for explaining the history behind the TypeScript conversion, I only discovered this project a couple of months ago and I was wondering why so many fields were protected rather than private :)

It sounds like the question of what kind of possibly breaking changes the VexFlow developers are willing to introduce between major versions is a bit of a philosophical one. For example, there's a chance that #1138 could break clients because of something like the type name change of `accidentals` to `Accidental`. I'm personally all in favour of adding breaking changes with a major semantic version change, _if_ they make the code cleaner and are easy enough for for clients to fix.

I've added `Stave.bounds` back in for now, since it's not relevant to the rest of the PR.

> Sometimes I wish math / measurement oriented classes (e.g., `BoundingBox`) had **public** fields. :-D This usually happens when I need to read or write code that has lots of `getWidth()`, `getX()`, `getY()`....

I absolutely agree on that point.

ronyeh:
I agree re: breaking changes. Since we are in the progress of bumping the major revision, it's a good time to clean up the crusty old stuff. As long as we note changes in our CHANGELOG, it's probably fine. 

We discussed the #1138 change to the CATEGORY strings, and reasoned that it's unlikely projects out there are using those strings directly. It was mostly a way for VexFlow to internally group/distinguish different types of modifiers and other elements. But yes, it's a breaking change and we'll need to mention it. Thanks for reminding me to update the CHANGELOG as well. 👍

0xfe:
Thanks for the PR and the discussion, folks. Is this ready for review?

tommadams:
Yep, I think this is ready for review now. Thanks!

src/canvascontext.ts
```diff
@@ -76,11 +74,19 @@ export class CanvasContext implements RenderContext {
 
   setFont(family: string, size: number, weight: string): this {
     this.vexFlowCanvasContext.font = (weight || '') + ' ' + size + 'pt ' + family;
+    this.textHeight = (size * 4) / 3;
```
0xfe, line 77: Can you add the comment from `setRawFont` here too please?

src/renderer.ts
0xfe: I added `USE_CANVAS_PROXY` when building the VexTab Google Docs plugin -- at the time they had a special canvas implementation which I had to hack around with this to make work. Now that they use a fully-featured runtime, we don't need this anymore (I think.)

src/svgcontext.ts
0xfe: +1

0xfe (approved):
Thanks Tom, looks great, merging!

Issue #1107, opened 2021-08-17
getCategory() and this.setAttribute('type', ...) should use the same strings


ronyeh:
This is a proposal for a PR to standardize `CATEGORY` and `getCategory()` and `this.setAttribute('type', elementType)` 

### PROPOSAL
- Standardize the static CATEGORY property to always be lowercase and singular. (If you prefer UpperCamelCase, that's fine too. As long as we are consistent!)
- Make sure every class that has a CATEGORY property also has a getCategory() method (and vice versa).
- this.setAttribute('type', ...) should use the CATEGORY property, instead of using a separate hard-coded string.

### BACKGROUND INFO

I discovered this issue while working on https://github.com/0xfe/vexflow/pull/1098

There are 33 classes that implement the static `CATEGORY` property:

https://github.com/0xfe/vexflow/search?q=%22static+get+CATEGORY%28%29%22

There are 33 **slightly different** classes that implement the `getCategory()` method:

https://github.com/0xfe/vexflow/search?q=%22getCategory%28%29%22

For example: 
- stavemodifier.ts has a getCategory() (which returns an empty string) but it does NOT have the static CATEGORY property. 
- textfont.ts has the static CATEGORY property, but it does not implement getCategory(). TextFont.CATEGORY is currently unused.

Additionally, the capitalization & number (singular/plural) of the category string is not consistent.

For example: 
- Bend.CATEGORY === 'bends'
- Stem.CATEGORY === 'stem'
- ChordSymbol.CATEGORY === 'chordSymbol'
- TimeSignature.CATEGORY === 'timesignatures'

I also noticed that 63 subclasses of `Element` call `this.setAttribute('type', elementTypeString)` in the constructor. The string here is always UpperCamelCase, and always singular. 

### RISKS
If there are projects out there that rely on the CATEGORY and 'type' attribute being a certain way, this PR could break those projects. 

`.getAttribute('type')` is used only once in the core library, and once in the tests (accidental_tests.ts). That test case can be updated to use isCategory() instead. 

`.getCategory()` is used more often (13 core library files and 2 test files). Those can be updated to use `isCategory()`.

mscuthbert:
My thought is:

1.  Standardize the case to be the same as the class name -- thus UpperCamelCase -- risk the compatibility failure
2. Make a static property CATEGORY that appears in every class that `.getCategory()` was in
3. Mark getCategory() as deprecated but for now just return `.CATEGORY`.

ronyeh:
[...]

PR #1138, opened 2021-09-09
Standardize `CATEGORY` property.

ronyeh:
Standardize `CATEGORY` / `getCategory()` / `this.setAttribute('type', ...)`.
The string is now always UpperCamelCaseSingular.
Improve the typeguard functions to check for the `CATEGORY` property directly.
Element's `this.attrs.type` is set to `this.getCategory()` in the `Element` constructor.

Fixes https://github.com/0xfe/vexflow/issues/1107

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,14 @@ Please help test this beta release and [report issues on GitHub](https://github.
 
 ## Breaking
 
-- The tsconfig.json `compilerOptions.target` has been updated to ES6 / ES2015. If you are targeting an older environment such as Internet Explorer, you will need to build directly from source code (and change the target back to ES5).
+- The tsconfig.json `compilerOptions.target` has been updated to ES6 / ES2015. If you are targeting an older environment, you will need to build directly from source code (and change the target back to ES5).
 - `Stave.setNumLines(n: number)` requires a number. Previously, a string would also work. See: [stave.ts](https://github.com/0xfe/vexflow/blob/master/src/stave.ts) and [#1083](https://github.com/0xfe/vexflow/issues/1083).
 - `Note.addModifier(modifier: Modifier, index?: number): this` now throws a RuntimeError if the parameters are reversed.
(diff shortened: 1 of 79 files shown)


ronyeh:
Now there is a single `getCategory()` function (in the `Element`) class. Any descendant of Element inherits this method, and only needs to provide its own static `CATEGORY` property. The isCategory() typeguard utility method has been refactored to look for `CATEGORY`.

```
  getCategory(): string {
    return (<typeof Element>this.constructor).CATEGORY;
  }

```

ronyeh:
@mscuthbert if you have time next week, can you help review the new implementation of `isCategory()` in typeguard.ts?

https://github.com/0xfe/vexflow/pull/1138/files#diff-b4e62682a03cf67278ca1b02054f986cd7eccdf0b25ee5bf1f5bda7bbeea1dcaR36-R74

The overall idea is from now on, any object that wants to masquerade as a VexFlow class will need to declare a matching `static CATEGORY` property. For example:

```
  class FakeStaveNote {
    static CATEGORY = 'StaveNote';
    ...
  }
```

Additionally, I was able to remove all of the individual `getCategory()` methods, and declare it only in the `Element` class. The implementation is to return the static CATEGORY property, which is now always UpperCamelCaseSingular:
```
  /** Get element category string. */
  getCategory(): string {
    return (<typeof Element>this.constructor).CATEGORY;
  }
```

ronyeh:
**This one's good to go.** 

No visual diffs relative to the current master @ 46e4cac2f4bc92a576f38b58f660a4e7d9ebaa98

There is now only one getCategory() function, in Element:
https://github.com/ronyeh/vexflow/blob/e2e7c5d76f954e4dd28881f24cd3493e4597ad5c/src/element.ts#L72-L75

Every descendant of Element that had a getCategory() now only has a static CATEGORY property. The property has been standardized to always be singular, and UpperCamelCase.

The `isCategory()` type check method has been updated to look for the static `CATEGORY` property directly. Previously it called the `getCategory()` method. Since this feature has been in VexFlow for only about a month, it seems low risk to change it. :-} Hopefully we won't need to change this any more.

See: https://github.com/ronyeh/vexflow/blob/category/src/typeguard.ts

0xfe:
Looks good overall. Would love to rethink this in the new typescript world. Having a CATEGORY is important for inventory management, but we could potentially also add interfaces for where they're used for their behaviours, so we get compile-time checking.

PR #1139, opened 2021-09-09
Fix/factory grace note

rvilarl:
TODO: Factory.GraceNote  resolved.
TODO: Stroke constructor 2nd parameter resolved.

### DIFF
--- a/src/gracenote.ts
+++ b/src/gracenote.ts
@@ -24,17 +24,17 @@ export class GraceNote extends StaveNote {
     return 0.66;
   }
 
-  constructor(note_struct: Partial<GraceNoteStruct>) {
+  constructor(noteStruct: Partial<GraceNoteStruct>) {
     super({
       ...{
         glyph_font_scale: Flow.DEFAULT_NOTATION_FONT_SCALE * GraceNote.SCALE,
         stroke_px: GraceNote.LEDGER_LINE_OFFSET,
       },
-      ...note_struct,
+      ...noteStruct,
     });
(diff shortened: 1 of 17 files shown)


src/gracenote.ts
```diff
@@ -24,7 +24,7 @@ export class GraceNote extends StaveNote {
     return 0.66;
   }
 
-  constructor(note_struct: GraceNoteStruct) {
+  constructor(note_struct: Partial<GraceNoteStruct>) {
```
ronyeh, line 27: Sometimes I see note_struct and sometimes it is noteStruct. Should we standardize to `noteStruct` where possible? Maybe the underscores should be reserved for keys of parameter objects, or legacy public fields.

src/stemmablenote.ts
```diff
@@ -22,7 +22,7 @@ export abstract class StemmableNote extends Note {
   protected flag?: Glyph;
   protected stem_extension_override?: number;
 
-  constructor(note_struct: NoteStruct) {
+  constructor(note_struct: Partial<NoteStruct>) {
```
ronyeh, line 25: note_struct => noteStruct :-)

I figure we might as well rename it now while we are working on this PR. (I see stavenote.ts uses noteStruct).

src/gracenote.ts
rvilarl: Agree, it will add a few files though

0xfe:
Thanks for this @rvilarl. Also, thanks for cleaning up the `noteStruct` inconsistencies.

PR #1140, opened 2021-09-10
Fix shadow/glow rendering

tommadams:
This PR makes the shadow/glow appearance consistent across the SVG context and canvas contexts at different device pixel ratios. The old SVG glow code is deleted and replaced with a `drop-shadow` CSS filter, a feature that now has more than [95% browser support](https://caniuse.com/?search=drop-shadow).

It also fixes a bug with the canvas shadow where the shadow blur radius wasn't being scaled by the canvas transform. It turns out that `CanvasRenderingContext2D` ignores scaling when applying a shadow. This meant that glow/shadow had a different appearance when depending on whether or not the user had a high density display.

I had to update the shadow blur sizes in the `StaveNote` tests to account for the fact that the blur radius now uses the correct units.

All visual diffs from this PR are caused by differences in the glow rendering.

I also fixed a bug the canvas size sanitisation: the pixel ratio scaling should be applied before sanitisation.

## Appearance before this PR

Low resolution canvas (simulated by forcing `devicePixelRatio` to 1):
LR canvas

High resolution canvas (note the glow is much tighter):
HR canvas

SVG (the glow has a different look):
svg

## Appearance after this PR

Low resolution canvas
LR canvas

High resolution canvas
HR canvas

SVG
svg

### DIFF
--- a/tests/stavenote_tests.ts
+++ b/tests/stavenote_tests.ts
@@ -670,7 +670,7 @@ function drawKeyStyles(options: TestOptions, contextBuilder: ContextBuilder): vo
   const note = new StaveNote({ keys: ['g/4', 'bb/4', 'd/5'], duration: 'q' })
     .setStave(stave)
     .addAccidental(1, new Accidental('b'))
-    .setKeyStyle(1, { shadowBlur: 15, shadowColor: 'blue', fillStyle: 'blue' });
+    .setKeyStyle(1, { shadowBlur: 2, shadowColor: 'blue', fillStyle: 'blue' });
 
   new TickContext().addTickable(note).preFormat().setX(25);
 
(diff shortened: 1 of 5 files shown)


tests/stavenote_tests.ts
```diff
@@ -845,7 +845,7 @@ function drawBeamStyles(options: TestOptions, contextBuilder: ContextBuilder): v
   staveNotes[1].setKeyStyle(0, { fillStyle: 'darkturquoise' });
 
   staveNotes[5].setStyle({ fillStyle: 'tomato', strokeStyle: 'tomato' });
-  beam3.setStyle({ shadowBlur: 20, shadowColor: 'blue' });
+  beam3.setStyle({ shadowBlur: 4, shadowColor: 'blue' });
```
0xfe, line 848: This is possibly going to break existing users that use `shadowBlur`. @ronyeh to capture this in the changelog.

0xfe:
Thanks Tom. Modulo my comment about capturing the new blur scale in the changelogs, LGTM!

PR #1141, opened 2021-09-12
Notes constructor should take a Partial<...>

rvilarl:
Further consistency on noteStruct approach using Partial

### DIFF
--- a/src/chordsymbol.ts
+++ b/src/chordsymbol.ts
@@ -572,14 +572,14 @@ export class ChordSymbol extends Modifier {
 
   // ### addGlyphSuperscript
   // add a glyph block with superscript modifier
-  addGlyphSuperscript(glyph: Glyph): this {
+  addGlyphSuperscript(glyph: string): this {
     const symbolType = ChordSymbol.symbolTypes.GLYPH;
     const symbolModifier = ChordSymbol.symbolModifiers.SUPERSCRIPT;
     return this.addSymbolBlock({ glyph, symbolType, symbolModifier });
   }
 
   // eslint-disable-next-line
-  addGlyph(glyph: string, parameters: any): this {
+  addGlyph(glyph: string, parameters?: any): this {
     parameters = parameters == null ? {} : parameters;
     parameters.glyph = glyph;
     parameters.symbolType = ChordSymbol.symbolTypes.GLYPH;
@@ -592,7 +592,7 @@ export class ChordSymbol extends Modifier {
   // `addGlyphOrText("(+5#11)")`
   // will use text for the '5' and '11', and glyphs for everything else.
   // eslint-disable-next-line
-  addGlyphOrText(text: string, parameters: any): this {
+  addGlyphOrText(text: string, parameters?: any): this {
     parameters = parameters == null ? {} : parameters;
     let str = '';
     for (let i = 0; i < text.length; ++i) {
@@ -615,7 +615,7 @@ export class ChordSymbol extends Modifier {
   // ### Add a line of the given width, used as a continuation of the previous
   // symbol in analysis, or lyrics, etc.
   // eslint-disable-next-line
-  addLine(width: number, parameters: any): this {
+  addLine(width: number, parameters?: any): this {
     parameters = parameters == null ? {} : parameters;
     parameters.symbolType = ChordSymbol.symbolTypes.LINE;
     parameters.width = width;
--- a/src/chordsymbol.ts
+++ b/src/chordsymbol.ts
@@ -490,8 +490,8 @@ export class ChordSymbol extends Modifier {
   // ChordSymbol allows multiple blocks so we can mix glyphs and font text.
   // Each block can have its own vertical orientation
   // eslint-disable-next-line
-  getSymbolBlock(parameters: any): any {
-    parameters = parameters == undefined ? {} : parameters;
+  getSymbolBlock(parameters?: any): any {
+    parameters = parameters ?? {};
     const symbolType = parameters.symbolType ? parameters.symbolType : ChordSymbol.symbolTypes.TEXT;
     const text = parameters.text ? parameters.text : '';
     const symbolModifier = parameters.symbolModifier ? parameters.symbolModifier : ChordSymbol.symbolModifiers.NONE;
@@ -548,7 +548,7 @@ export class ChordSymbol extends Modifier {
   // Add a text block
   // eslint-disable-next-line
   addText(text: string, parameters?: any): this {
-    parameters = parameters == null || parameters == undefined ? {} : parameters;
+    parameters = parameters ?? {};
     parameters.text = text;
     parameters.symbolType = ChordSymbol.symbolTypes.TEXT;
     return this.addSymbolBlock(parameters);
@@ -580,7 +580,7 @@ export class ChordSymbol extends Modifier {
 
   // eslint-disable-next-line
   addGlyph(glyph: string, parameters?: any): this {
-    parameters = parameters == null ? {} : parameters;
+    parameters = parameters ?? {};
     parameters.glyph = glyph;
     parameters.symbolType = ChordSymbol.symbolTypes.GLYPH;
     return this.addSymbolBlock(parameters);
@@ -593,7 +593,7 @@ export class ChordSymbol extends Modifier {
   // will use text for the '5' and '11', and glyphs for everything else.
   // eslint-disable-next-line
   addGlyphOrText(text: string, parameters?: any): this {
-    parameters = parameters == null ? {} : parameters;
+    parameters = parameters ?? {};
     let str = '';
     for (let i = 0; i < text.length; ++i) {
       if (ChordSymbol.glyphs[text[i]]) {
@@ -616,7 +616,7 @@ export class ChordSymbol extends Modifier {
   // symbol in analysis, or lyrics, etc.
   // eslint-disable-next-line
   addLine(width: number, parameters?: any): this {
-    parameters = parameters == null ? {} : parameters;
+    parameters = parameters ?? {};
     parameters.symbolType = ChordSymbol.symbolTypes.LINE;
     parameters.width = width;
     return this.addSymbolBlock(parameters);
--- a/src/notehead.ts
+++ b/src/notehead.ts
@@ -17,21 +17,20 @@ function L(...args: any[]) {
 }
 
 export interface NoteHeadStruct extends NoteStruct {
-  glyph_font_scale?: number;
-  slashed?: boolean;
-  style?: ElementStyle;
-  stem_down_x_offset?: number;
-  stem_up_x_offset?: number;
-  custom_glyph_code?: string;
-  x_shift?: number;
-  line?: number;
-  stem_direction?: number;
-  displaced?: boolean;
-  //  duration: string;
+  glyph_font_scale: number;
+  slashed: boolean;
+  style: ElementStyle;
+  stem_down_x_offset: number;
+  stem_up_x_offset: number;
+  custom_glyph_code: string;
+  x_shift: number;
+  line: number;
+  stem_direction: number;
+  displaced: boolean;
   note_type: string;
-  y?: number;
-  x?: number;
-  index?: number;
+  y: number;
+  x: number;
+  index: number;
 }
 
 /**
@@ -108,7 +107,7 @@ export class NoteHead extends Note {
   protected custom_glyph: boolean = false;
   protected stem_up_x_offset: number = 0;
   protected stem_down_x_offset: number = 0;
-  protected note_type: string;
+  protected note_type?: string;
   protected displaced: boolean;
   protected stem_direction: number;
 
@@ -118,17 +117,16 @@ export class NoteHead extends Note {
   protected index?: number;
   protected slashed: boolean;
 
-  constructor(head_options: NoteHeadStruct) {
-    super(head_options);
+  constructor(noteStruct: Partial<NoteHeadStruct>) {
+    super(noteStruct);
 
-    this.index = head_options.index;
-    this.x = head_options.x || 0;
-    this.y = head_options.y || 0;
-    this.note_type = head_options.note_type;
-    this.duration = head_options.duration;
-    this.displaced = head_options.displaced || false;
-    this.stem_direction = head_options.stem_direction || Stem.UP;
-    this.line = head_options.line || 0;
+    this.index = noteStruct.index;
+    this.x = noteStruct.x || 0;
+    this.y = noteStruct.y || 0;
+    this.note_type = noteStruct.note_type;
+    this.displaced = noteStruct.displaced || false;
+    this.stem_direction = noteStruct.stem_direction || Stem.UP;
+    this.line = noteStruct.line || 0;
 
     // Get glyph code based on duration and note type. This could be
     // regular notes, rests, or other custom codes.
@@ -136,22 +134,22 @@ export class NoteHead extends Note {
     defined(this.glyph, 'BadArguments', `No glyph found for duration '${this.duration}' and type '${this.note_type}'`);
 
     this.glyph_code = this.glyph.code_head;
-    this.x_shift = head_options.x_shift || 0;
-    if (head_options.custom_glyph_code) {
+    this.x_shift = noteStruct.x_shift || 0;
+    if (noteStruct.custom_glyph_code) {
       this.custom_glyph = true;
-      this.glyph_code = head_options.custom_glyph_code;
-      this.stem_up_x_offset = head_options.stem_up_x_offset || 0;
-      this.stem_down_x_offset = head_options.stem_down_x_offset || 0;
+      this.glyph_code = noteStruct.custom_glyph_code;
+      this.stem_up_x_offset = noteStruct.stem_up_x_offset || 0;
+      this.stem_down_x_offset = noteStruct.stem_down_x_offset || 0;
     }
 
-    this.style = head_options.style;
-    this.slashed = head_options.slashed || false;
+    this.style = noteStruct.style;
+    this.slashed = noteStruct.slashed || false;
 
     this.render_options = {
       ...this.render_options,
       ...{
         // font size for note heads
-        glyph_font_scale: head_options.glyph_font_scale || Flow.DEFAULT_NOTATION_FONT_SCALE,
+        glyph_font_scale: noteStruct.glyph_font_scale || Flow.DEFAULT_NOTATION_FONT_SCALE,
         // number of stroke px to the left and right of head
         stroke_px: 3,
       },
--- a/src/notehead.ts
+++ b/src/notehead.ts
@@ -107,7 +107,6 @@ export class NoteHead extends Note {
   protected custom_glyph: boolean = false;
   protected stem_up_x_offset: number = 0;
   protected stem_down_x_offset: number = 0;
-  protected note_type?: string;
   protected displaced: boolean;
   protected stem_direction: number;
 
@@ -123,15 +122,15 @@ export class NoteHead extends Note {
     this.index = noteStruct.index;
     this.x = noteStruct.x || 0;
     this.y = noteStruct.y || 0;
-    this.note_type = noteStruct.note_type;
+    if (noteStruct.note_type) this.noteType = noteStruct.note_type;
     this.displaced = noteStruct.displaced || false;
     this.stem_direction = noteStruct.stem_direction || Stem.UP;
     this.line = noteStruct.line || 0;
 
     // Get glyph code based on duration and note type. This could be
     // regular notes, rests, or other custom codes.
-    this.glyph = Flow.getGlyphProps(this.duration, this.note_type);
-    defined(this.glyph, 'BadArguments', `No glyph found for duration '${this.duration}' and type '${this.note_type}'`);
+    this.glyph = Flow.getGlyphProps(this.duration, this.noteType);
+    defined(this.glyph, 'BadArguments', `No glyph found for duration '${this.duration}' and type '${this.noteType}'`);
 
     this.glyph_code = this.glyph.code_head;
     this.x_shift = noteStruct.x_shift || 0;
@@ -269,7 +268,7 @@ export class NoteHead extends Note {
 
     const y = this.y;
 
-    L("Drawing note head '", this.note_type, this.duration, "' at", head_x, y);
+    L("Drawing note head '", this.noteType, this.duration, "' at", head_x, y);
 
     // Begin and end positions for head.
     const stem_direction = this.stem_direction;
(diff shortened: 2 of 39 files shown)


0xfe:
Thanks Rodrigo -- looks good at a first pass. @ronyeh for comments.

rvilarl:
@0xfe @ronyeh I have decide to include some simple corrections on parameter issues (optional, double types ...).

ronyeh:
Sorry folks. I've been a bit busy. I'll give this a review in a bit....

I have been thinking about our approach to options objects / parameter structs. It seems like VexFlow prefers to have as many optional fields as possible for a provided options object. It aims to provide decent defaults for missing options. 

Is this our main philosophy? If so, I think converting all of our options arguments to accept `Partial<OptionsType>` is okay. However, there are some small considerations.

In the old approach, we would have some fields marked required, and others marked optional (with a ?):

```
export interface NoteStruct {
  duration: string;
  line?: number;
  dots?: number;
  keys?: string[];
  type?: string;
  align_center?: boolean;
  duration_override?: Fraction;
}
```

As a developer, it is immediately clear to me that the most important field to provide is `duration`. If we move to the new way, toward `Partial<NoteStruct>`, we will remove all the ? from the interface declaration.

```
export interface NoteStruct {
  duration: string;
  line: number;
  dots: number;
  keys: string[];
  type: string;
  align_center: boolean;
  duration_override: Fraction;
}
```

This looks a bit cleaner to me, but then our VexFlow methods that accept a `NoteStruct` will now have to instead accept `Partial<NoteStruct>` and provide a default duration if it doesn't already exist. Or if there are enough of these methods, they could all call a shared `validateNoteStruct(noteStruct)` function that provides defaults for missing fields.

The one nice thing about the old way is that the compiler does some checks for us. If duration is required, and the developer forgets to provide it, TS will complain:

```
Property 'duration' is missing in type '{}' but required in type 'NoteStruct'
```

In the new way (where everything is optional due to `Partial`), the code will work even if important fields are omitted, so maybe the developer will need to investigate if the output is buggy.

In the below example, there are zero TS compiler errors, even though 'glyph' is misspelled:

```
interface Options {
    text: string;
    glyph: string;
    smooth: boolean;
}

function doSomethingUseful(options: Partial<Options>) {
    // ... provide sane defaults for missing fields ...
    console.log(options.glyph);
}

const opts = {
    text: 'Hello',
    glyhp: 'gClef',
    smooth: true,
};

doSomethingUseful(opts);

```

Or should we just ditch usage of Partial and instead mark all fields as optional with question marks (and provide sane defaults where needed)?

I don't really know the answer!

Regardless, I think we need to add more comments to individual fields in the various options object types. If one or more fields are very important, we should note them as such, and possibly move them to the top of the interface declaration.

## Advantages of different approaches to handling options objects
### A - Mark some fields as required, and the rest optional.  
Advantages: a developer can see which fields are important. TS compiler will complain if a field is accidentally omitted. 

### B - Mark all fields as optional, with the ? symbol
Advantages: maximum flexibility for the developer. If a developer writes a helper method that takes in the options object as an argument, they can simply type `NoteStruct` instead of `Partial<NoteStruct>`.

### C - Mark all fields as required (omit the ? symbol), but use `Partial<T>` where needed.
Advantages: The interface declaration looks clean. It includes the advantages of option B, like maximum flexibility for the developer. Disadvantages: In certain cases, the developer will have to type out `Partial<NoteStruct>` instead of just `NoteStruct`. This might be a stumbling block or a tiny bit of mental overhead for a developer who is using VexFlow.

tests/tabtie_tests.ts
```diff
@@ -33,7 +33,7 @@ const TabTieTests = {
 /**
  * Helper function to create TabNote objects.
  */
-const tabNote = (struct: TabNoteStruct) => new TabNote(struct);
+const tabNote = (noteStruct: Partial<TabNoteStruct>) => new TabNote(noteStruct);
```
ronyeh, line 36: This is what I mean about "slightly more keystrokes or mental overhead" for the developer.

If someone wants to create a helper method, they might have to type out `Partial<TabNoteStruct>` because using just `TabNoteStruct` here will trigger a TS compiler error.

tests/tabnote_tests.ts
```diff
@@ -178,7 +178,7 @@ function drawStemsUp(options: TestOptions, contextBuilder: ContextBuilder): void
   stave.setContext(ctx);
   stave.draw();
 
-  const specs: TabNoteStruct[] = [
+  const specs: Partial<TabNoteStruct>[] = [
```
ronyeh, line 181: If we consider that the `xxxx_tests.ts` files are examples of code that a developer using the VexFlow library would write, I would say that seeing all these instances of `Partial<TabNoteStruct>` are evidence that maybe we should avoid Partial, and instead just mark all fields as optional with ?.

@0xfe what do you think? Is it annoying for a dev to have to type out `Partial<TabNoteStruct>`? It would look slightly cleaner if the developer could just use `TabNoteStruct` directly.

tests/notehead_tests.ts
```diff
@@ -72,15 +67,15 @@ function basic(options: TestOptions, contextBuilder: ContextBuilder): void {
...
+function showNote(noteStruct: Partial<StaveNoteStruct>, stave: Stave, ctx: RenderContext, x: number) {
+  const note = new StaveNote(noteStruct).setStave(stave);
   new TickContext().addTickable(note).preFormat().setX(x);
   note.setContext(ctx).draw();
   return note;
 }
 
 function variousHeads(options: TestOptions, contextBuilder: ContextBuilder): void {
-  const notes: StaveNoteStruct[] = [
+  const notes: Partial<StaveNoteStruct>[] = [
```
ronyeh, line 78: Hmm... I'm not sure I'm a fan of seeing the `Partial<...>` leak into the tests files. That's a sign that the developer who runs `npm install vexflow` will have to utilize `Partial<...>` in his/her code, simply because we choose to accept Partials.

tests/barline_tests.ts
```diff
@@ -33,7 +28,7 @@ function simple(options: TestOptions): void {
 
   const notes = [
     f.StaveNote({ keys: ['d/4', 'e/4', 'f/4'], stem_direction: -1, duration: '2' }),
-    f.BarNote({ type: 'single' }), // => f.BarNote({ type: BarlineType.SINGLE })
+    f.BarNote({ type: 'single' }),
```
ronyeh, line 31: Oh, my comment here was a question: should we change this line of code to use the BarlineType enum, for enhanced compiler checks? Or do we prefer developers to stick with 'single' (it's shorter to type, for sure!).

src/textnote.ts
```diff
@@ -13,12 +13,12 @@ export enum Justification {
...
-  font?: FontInfo;
-  subscript?: string;
-  superscript?: string;
+  ignore_ticks: boolean;
+  smooth: boolean;
+  glyph: string;
+  font: FontInfo;
+  subscript: string;
+  superscript: string;
   text: string;
```
ronyeh, line 22: Should we move `text` to the top of the TextNoteStruct declaration? It seems like the most important field here.

ronyeh (approved):
Looks good! I think it's safe to merge.

My main question from reviewing this is.... do we want to continue moving everything toward Partials? Or should we instead move toward making all the options objects declare every field to be optional (with ?). 

It's nice to not have to mark everything with ?, but as we saw from the xxxx_tests.ts files, using Partials might mean that our client developers will also have to explicitly declare some parameters with `Partial<...>` as well. Basically, we are adding slightly more work to developers who are using VexFlow.

src/textnote.ts
rvilarl: I think that in many cases it is very difficult to figure out what is most important parameter. For exampla should we also put the scripts on the second position? In what order sub, super or super, sub?
I believe that the best approach would be to document them.

tests/barline_tests.ts
rvilarl: I saw that `BarNote` supported internally in `SetType` both possibilities and I adapted the parameter of the constructor.

tests/notehead_tests.ts
rvilarl: I do not quite agree. We have to use Partial due to the fact that we decided to use Partial definitions in the tests. If we defined the complete structure in the test Partial would not be required.

tests/tabnote_tests.ts
rvilarl: We have to use Partial due to the fact that we decided to use Partial definitions in the tests. If we defined the complete structure in the test Partial would not be required.

rvilarl:
> I have been thinking about our approach to options objects / parameter structs. It seems like VexFlow prefers to have as many optional fields as possible for a provided options object. It aims to provide decent defaults for missing options.

VexFlow did not have optional parameters as JavaScript does not care about completion. They were introduced in the migration.

> 
> Is this our main philosophy? If so, I think converting all of our options arguments to accept `Partial<OptionsType>` is okay. However, there are some small considerations.
> 
> In the old approach, we would have some fields marked required, and others marked optional (with a ?):
> 
> ```
> export interface NoteStruct {
>   duration: string;
>   line?: number;
>   dots?: number;
>   keys?: string[];
>   type?: string;
>   align_center?: boolean;
>   duration_override?: Fraction;
> }
> ```
> 
> As a developer, it is immediately clear to me that the most important field to provide is `duration`. If we move to the new way, toward `Partial<NoteStruct>`, we can choose to remove all the ? from the interface declaration.

I would not say that it is the most important one. Many times it only means that it has no default value. I believe, as you stated below, that the importance and meaning should be documented.

> 
> ```
> export interface NoteStruct {
>   duration: string;
>   line: number;
>   dots: number;
>   keys: string[];
>   type: string;
>   align_center: boolean;
>   duration_override: Fraction;
> }
> ```
> 
> This looks a bit cleaner to me, but then our VexFlow methods that accept a `NoteStruct` will now have to instead accept `Partial<NoteStruct>` and provide a default duration if it doesn't already exist. Or if there are enough of these methods, they could all call a shared `validateNoteStruct(noteStruct)` function that provides defaults for missing fields.
> 
> The one nice thing about the old way is that the compiler does some checks for us. If duration is required, and the developer forgets to provide it, TS will complain:
> 
> ```
> Property 'duration' is missing in type '{}' but required in type 'NoteStruct'
> ```
> 
> In the new way (where everything is optional due to `Partial`), the code will work even if important fields are omitted, so maybe the developer will need to investigate if the output is buggy.
> 
> In the below example, there are zero TS compiler errors, even though 'glyph' is misspelled in the `opts` object:
> 
> ```
> interface Options {
>     text: string;
>     glyph: string;
>     smooth: boolean;
> }
> 
> function doSomethingUseful(options: Partial<Options>) {
>     // ... provide sane defaults for missing fields ...
>     console.log(options.glyph);
> }
> 
> const opts = {
>     text: 'Hello',
>     glyhp: 'gClef',
>     smooth: true,
> };
> 
> doSomethingUseful(opts);
> ```
> 
> Or should we just not use `Partial<>` and instead mark all fields as optional with question marks (providing sane defaults where needed)?
> 
Partial is just a helper to make all fields optional.

> I don't really know the answer!
> 
> Regardless, I think we need to add more comments to individual fields in the various options object types. If one or more fields are very important, we should note them as such, and possibly move them to the top of the interface declaration.

Agree, with typedoc they will end up in the documentation.

> 
> ## Advantages of different approaches to handling options objects
> ### A - Mark some fields as required, and the rest optional.
> Advantages: a developer can see which fields are important. TS compiler will complain if a field is accidentally omitted.
> 
> ### B - Mark all fields as optional, with the ? symbol
> Advantages: maximum flexibility for the developer. If a developer writes a helper method that takes in the options object as an argument, they can simply type `NoteStruct` instead of `Partial<NoteStruct>`.
> 
> ### C - Mark all fields as required (omit the ? symbol), but use `Partial<T>` where needed.
> Advantages: The interface declaration looks clean. It includes the advantages of option B, like maximum flexibility for the developer. Disadvantages: In certain cases, the developer will have to type out `Partial<NoteStruct>` instead of just `NoteStruct`. This might be a stumbling block or a tiny bit of mental overhead for a developer who is using VexFlow.

Personally I prefer Partial, I stopped using ? as I learnt that this was possible during the migration.

src/chordsymbol.ts
```diff
@@ -572,14 +572,14 @@ export class ChordSymbol extends Modifier {
...
+  addGlyphSuperscript(glyph: string): this {
     const symbolType = ChordSymbol.symbolTypes.GLYPH;
     const symbolModifier = ChordSymbol.symbolModifiers.SUPERSCRIPT;
     return this.addSymbolBlock({ glyph, symbolType, symbolModifier });
   }
 
   // eslint-disable-next-line
-  addGlyph(glyph: string, parameters: any): this {
+  addGlyph(glyph: string, parameters?: any): this {
     parameters = parameters == null ? {} : parameters;
```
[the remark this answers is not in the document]
rvilarl: done

src/chordsymbol.ts
```diff
@@ -592,7 +592,7 @@ export class ChordSymbol extends Modifier {
   // `addGlyphOrText("(+5#11)")`
   // will use text for the '5' and '11', and glyphs for everything else.
   // eslint-disable-next-line
-  addGlyphOrText(text: string, parameters: any): this {
+  addGlyphOrText(text: string, parameters?: any): this {
     parameters = parameters == null ? {} : parameters;
```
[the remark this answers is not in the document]
rvilarl: done

src/notehead.ts
```diff
@@ -105,7 +104,7 @@ export class NoteHead extends Note {
   protected custom_glyph: boolean = false;
   protected stem_up_x_offset: number = 0;
   protected stem_down_x_offset: number = 0;
-  protected note_type: string;
+  protected note_type?: string;
```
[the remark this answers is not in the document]
rvilarl: Duplication resolved

src/stave.ts
```diff
@@ -317,7 +317,7 @@ export class Stave extends Element {
   setText(
     text: string,
     position: number,
-    options: {
+    options?: {
       shift_x: number;
       shift_y: number;
       justification: number;
```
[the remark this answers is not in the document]
rvilarl: You are right this has to be Partial as well(resolved). I did not remove the `ts-nocheck` from stave_tests.ts as there were a lot of changes. 
In general @0xfe always asked to declare inline when the structure was not too complex.

tests/tabtie_tests.ts
rvilarl: We have to use Partial due to the fact that we decided to use Partial definitions in the tests. If we defined the complete structure in the test Partial would not be required.

src/stave.ts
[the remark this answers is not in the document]
rvilarl: I included the changes to resolve stave_test.ts sorry. It would be more work now to remove it because I rebased.
@ronyeh please look into them. There is a visual difference on the bars because the double ones are now correctly displayed.

0xfe:
Thanks for the discussion folks -- I think Ron's point about making it simpler for users (and tooling) to know what's required and what's optional is spot on. We can really piggyback off of the type system to enforce these requirements.

My suggestion is to use `?` as much as possible, and `Partial<..>` only in cases where 1) it doesn't matter from a readability or correctness perspective, or 2) it gets messy to manage (e.g., if there are too many parameters.)

@rvilarl I know you put a bunch of work into this, and I really appreciate it, however can we refactor this with that lens? For example, if all the parameters are optional (as we see some instances), then `Partial` is okay.

@ronyeh @rvilarl thoughts?

ronyeh:
Rodrigo, we all owe you for the ton of work you've put into VexFlow. We wouldn't even be having this conversation if not for you porting VexFlow to TS, and submitting the recent PRs to clean up all the TODOs I left in the tests :-). Thanks!

I've been doing more thinking about Partials... Can we make _any_ strong case to use `Partial<...>` over just setting all fields to be optional? More and more, I am leaning toward not using `Partial` at all. 

This stackoverflow question/answer provides added insight:
https://stackoverflow.com/questions/54192544/difference-between-explicitly-setting-optional-properties-vs-using-the-partial

"Later, if you want to add a required parameter to the interface, you need to remove the Partial and go back to..." [explicitly marking fields as optional].

I think Partial can be useful if you are a developer who npm installed an existing library, and there is an existing interface declaration that you can't modify. Then you can just declare a local variable as `Partial<...>`, and when you finally have supplied all the required fields, you can cast it back to the original interface with all required properties.

I like the direction we are going, where we provide sane defaults for optional fields.

Even with all optional fields, the TS compiler can do basic checks. For example:
S 2021-09-18 at 11 43 53

```
error TS2322: Type '{ text: string; glhyp: string; smooth: true; }' is not assignable to type 'Options'.
  Object literal may only specify known properties, but 'glhyp' does not exist in type 'Options'. Did you mean to write 'glyph'?

glhyp: 'gClef',
```

However, my current view is that we should not use `Partial` anywhere, and mark as many fields optional as we can. I really would like developers who npm install vexflow to encounter fewer stumbling blocks while learning vexflow, and have as little mental overhead as possible while using vexflow. I believe Partial types add to those stumbling blocks / mental overhead.

Should we ask our core contributors for their feedback? Maybe folks like Simon, Myke, Tom, have insight on this matter.

Anyways, whatever we ultimately decide, I will totally support it with code review, actual coding, etc. For example, I'm happy to go through and remove Partials and add `?` to every field, along with comments, sane defaults, etc.

src/textnote.ts
ronyeh: I agree that the best approach will be to document each field. But we should leave this effort to a future PR. You and I can take on the various options object interfaces and add comments, especially where it is not clear what the field accomplishes.

rvilarl:
I think that Partial is the most convenient approach when a structure is used at the same time for two purposes:
1) a member variable in the class where options is a pain because you have to check all over the place if the fields are defined or not, and
2) you use it as parameter where not all the fields are required.

This fits very well for all the noteStructs. Otherwise we will probably need to define to structs, one for the member variable without options and one for the paramater with options.

Of course all the approaches are valid. :)

ronyeh:
Thanks. This is a great point. 

However, I would like developers to not have to worry about `Partial`. It seems like a potential stumbling block to learning VexFlow. 

I am totally fine if internally, we VexFlow maintainers have to deal with the hassle of validating incoming options objects, etc. That's our job.

What about this idea:

NoteStruct has all optional fields. No Partials are used anywhere. We add documentation to each field (in a future PR).

Internally, we have a validation method that takes in a NoteStruct and then provides default values for any missing options. This validation method returns `Required<NoteStruct>`, which means that all fields are now defined. So internally, the member variable is a `Required<NoteStruct>` because we have validated it in the constructor or factory method. Using the member variable should be easier because we no longer need to check each field every time we use it. The options object was funneled through the validation method, and became a `Required<NoteStruct>`.

I think this makes the external API cleaner. Developers only need to know about NoteStruct. Internally, maintainers need to know that NoteStruct has optionals, but a validated `Required<NoteStruct>` has no optionals. 

Does this solution work?

https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype

rvilarl:
Go for it! Required should work.

src/notehead.ts
```diff
@@ -18,21 +18,20 @@ function L(...args: any[]) {
...
-  style?: ElementStyle;
-  stem_down_x_offset?: number;
-  stem_up_x_offset?: number;
-  custom_glyph_code?: string;
-  x_shift?: number;
-  line?: number;
-  stem_direction?: number;
-  displaced?: boolean;
-  //  duration: string;
+  glyph_font_scale: number;
```
ronyeh, line 21: @rvilarl Are you going to try the all optional ? plus the `Required<NoteStruct>` with this PR? If you need help with testing or implementation, let me know and I can work off your branch.

src/notehead.ts
```diff
@@ -18,21 +18,20 @@ function L(...args: any[]) {
...
+  slashed: boolean;
+  style: ElementStyle;
+  stem_down_x_offset: number;
+  stem_up_x_offset: number;
+  custom_glyph_code: string;
+  x_shift: number;
+  line: number;
+  stem_direction: number;
+  displaced: boolean;
   note_type: string;
```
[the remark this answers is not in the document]
ronyeh: @0xfe @rvilarl is it ok if we delete this note_type field from the NoteHeadStruct? It extends the NoteStruct interface which seems to have the same field (.type).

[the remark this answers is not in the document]
ronyeh: We can add it to the CHANGELOG as part of this PR. "Deleted note_type field from NoteHeadStruct. Use `type` instead."

rvilarl:
> Anyways, whatever we ultimately decide, I will totally support it with code review, actual coding, etc. For example, I'm happy to go through and remove Partials and add `?` to every field, along with comments, sane defaults, etc.

@ronyeh I am not going to change this PR to go back to ? I let you go through all the Partials. What do you prefer to merge this one and modify them or to close this one and resolve the `ts-check` from current master?

ronyeh:
There's good stuff in this PR. If there aren't any unexpected visual diffs, I'm happy to have this merged. I can revisit the Required type approach in the next PR to see if it makes things cleaner.

But going forward, are we all agreeing to mark options object fields as optional with the question mark operator?
I can slowly remove Partial over the next few PRs....

ronyeh:
Or an alternative solution, I can submit a PR to your branch, haha. Or is that too complicated?

rvilarl:
I am happy with both solutions (merging and rework / PRs to my branch) @0xfe your choice

ronyeh:
Hi! I just started looking into this, and plan to send a PR to your branch. :-)

Then you can feel free to squash / rebase any commits to your liking, and update this PR with our combined commits.

(I'm treating this as a learning experience, haha.)

0xfe:
+1 for working on Rodrigo's branch :-)

mscuthbert:
Sorry to be so absent on things lately (even when @ tagged) -- beginning of semester is heaviest workload for me as a prof. and Artusi music as well.

In Artusi and music21j we've just gotten used to writing our enums with lots of ? for everything that's optional.  (And always put the non-optional ones at the top, usually with a blank line between those and the optional ones if the Enum is long).  As people on the thread have figured out, we end up losing the compiler errors if we mark things as optional that are actually required.  The one keystroke ? eventually just becomes second-nature.

ronyeh:
> Sorry to be so absent on things lately (even when @ tagged) -- beginning of semester is heaviest workload for me as a prof. and Artusi music as well.

Thanks for the insight! I'm currently working on Rodrigo's branch, and am transitioning Partial to explicit ?. It's a little more work for the VexFlow maintainers, but I believe it presents the easiest external API for developers to learn.

Good luck with your teaching/research/business!

rvilarl:
Included in #1148

PR #1142, opened 2021-09-13
Fix a couple of defects in SVGContext.arc

tommadams:
This PR updates the `SVGContext.arc` behaviour to match that of `CanvasRenderingContext2D`:
 - The pen position should end up at the end of the arc: the code that was moving the pen back at the end of `arcHelper` was incorrect.
 - If the length of the arc specified by `angleStart`, `angleEnd` and `radius` is longer than the circle's circumference, a completely closed circle should be drawn.

None of this really matters because this change doesn't cause any image diffs, but it's nice to have consistency across the contexts.

While I was here, I simplified the implementation a bit.

### DIFF
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -2,7 +2,7 @@
 // MIT License
 // @author Gregory Ristow (2015)
 
-import { RuntimeError, prefix } from './util';
+import { RuntimeError, normalizeAngle, prefix } from './util';
 import { RenderContext, TextMeasure } from './types/common';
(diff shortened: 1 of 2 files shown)


0xfe:
Thanks Tom. Can you fix the conflict please?

tommadams:
> Thanks Tom. Can you fix the conflict please?

Done. Thanks for the reveiws!

0xfe:
Thanks for the PR, Tom.

PR #1144, opened 2021-09-18
A few ledger line tweaks

tommadams:
- Move the default ledger line style from `StaveNote` to `Stave`. Prior to this change, a user could only change the default ledger line width by explicitly setting it on every single `StaveNote`. Now, they only have to change the default ledger line style on each `Stave`.
 - Following on from the recent discussion on #1052, change the ledger lines to be thinner but darker. In the end, I chose a line width of 1.4 so that the anti-aliasing on the darker line was close to the old ledger line shade.
 - Early out of `StaveNote.drawLedgerLines` if there are no ledger lines to be drawn.

I reviewed every one of the 576 image diffs that result from this PR and they were all caused by the change in ledger line thickness and stroke style, i.e. moving the default style from `StaveNote` to `Stave` had no effect on the tests.

Before:
old ledger lines

After:
old ledger lines

### DIFF
--- a/src/stave.ts
+++ b/src/stave.ts
@@ -116,7 +116,7 @@ export class Stave extends Element {
     };
     this.bounds = { x: this.x, y: this.y, w: this.width, h: 0 };
     this.options = { ...this.options, ...options };
-    this.defaultLedgerLineStyle = {};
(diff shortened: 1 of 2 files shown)


0xfe:
Thanks Tom. Looks great. Also, the darker lines do improve legibility.

PR #1145, opened 2021-09-19
Don't stroke rectangles drawn with fillRect

tommadams:
This PR fixes a bug in `SVGContext.fillRect`.

Prior to this fix the `SVGContext` was incorrectly applying the current stroke style to the rectangle; the rectangle should be drawn without a stroke to match the behaviour of the `CanvasContext`. The most noticeable effect of this bug was that the barlines & stave connectors were drawn incorrectly.

All visual diffs caused by this change are, rather surprisingly, under the threshold.

SVG before this PR:
svg old

SVG after this PR:
svg new

Canvas:
canvas

### DIFF
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -458,12 +458,8 @@ export class SVGContext implements RenderContext {
   }
 
   fillRect(x: number, y: number, width: number, height: number): this {
-    if (height < 0) {
-      y += height;
-      height *= -1;
-    }
-
-    this.rect(x, y, width, height, this.attributes);
+    const attributes = { fill: this.attributes.fill };
+    this.rect(x, y, width, height, attributes);
     return this;
   }
 


AaronDavidNewman:
Is the (now invisible) stroke included in the bounding box of the rectangle?  I notice the barline doesn't take up less space, there is just more whitespace.  It looks better in this case, but I'm not sure where else rectangles are used in vex.

tommadams:
> Is the (now invisible) stroke included in the bounding box of the rectangle? 

No, the stroke has no effect on the rectangle's bounding box for layout purposes. This is purely a graphical change.

> I'm not sure where else rectangles are used in vex.

`fillRect` is used as part of rendering the following elements:
 - tuplet brackets
 - barlines & stave connectors
 - multi-measure rests
 - voltas
 - brush up & down strokes
 - the note stem in the tempo (which is odd because normal stave notes use a line to render the stem)

src/svgcontext.ts
```diff
@@ -458,12 +458,8 @@ export class SVGContext implements RenderContext {
   }
 
   fillRect(x: number, y: number, width: number, height: number): this {
-    if (height < 0) {
-      y += height;
-      height *= -1;
-    }
```
ronyeh, line 464: Do we know what this was supposed to accomplish? Is there a test case for negative heights?

Supposedly the Canvas API supports negative widths and heights. Not sure how closely we are trying to replicate the Canvas API:

https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillRect

tommadams: SVG doesn't support negative widths or heights, so this code was correcting for that.

However, `fillRect` calls through to `rect`, which also performs the same correction so this code is entirely redundant.

ronyeh: Ah makes sense. The default github diff only showed this snippet, and I neglected to expand the view to see the implementation of rect(). I suppose we don't need to handle negative widths yet. :-)

sschmidTU:
> All visual diffs caused by this change are, rather surprisingly, under the threshold.

Try changing the threshold in `visual_regression.sh`:
`THRESHOLD=0.0001`
The default `0.01` is too low in my opinion, it misses a lot of thin line changes. It's a good first pass.
We use `0.0001` by default in OSMD.

ronyeh:
Does changing the threshold slow down the tests any? If not, I vote for
matching what OSMD uses.

tommadams:
So I think the reason that no visual diffs were detected is that there aren't any!

Even though master branch definitely renders SVG barlines incorrectly when viewed in the browser, the exact same code renders the barlines correctly in the visual regression test.

I don't really have a good explanation for this other than a possible bug in jsdom.

ronyeh:
Yeah it's definitely possible that node canvas paints things differently
(or it could be the underlying graphics package that node canvas uses).

I recently tried to draw a 1px thin rectangle and it kept drawing it with
translucency (even though alpha was set to 1). When I changed the stroke
thickness to 2px, everything looked as I expected (2px thick black outlined
rectangle).

ronyeh:
OH wait. We are talking about changes to SVGContext, right?

The command line visual regression tests only use `Renderer.Backends.CANVAS`.

I guess we have always assumed that the SVG and CanvasContexts render things identically (which obviously they do not). Recently I discovered that the SVGContext was not stacking multiple calls to `.scale()` incorrectly.

Maybe we eventually need a sanity test suite that just verifies that our SVGContext and CanvasContext do the same thing with basic drawing and transformation calls.

tommadams:
Oh, well that would certainly explain it!

ronyeh:
I think Rodrigo had this idea a while back.... For command line diffing of SVG output, we could render out the path string. If the path string doesn't change at all, then it's a zero diff. If the path changes, we can save out two SVG files using those paths (assuming we set the viewbox properly). On the file system, we can manually visually compare the two SVG files to see what changed.

sschmidTU:
Or you convert the SVG to PNG and do the usual visual regression with it. The diff in the SVG text would also be interesting, but I think often it's not easy to imagine the visual difference from the text.

tommadams:
Unfortunately, it looks like #1150 will need fixing before we can make any progress on diffing the SVG output.

mscuthbert:
#1150 is now in.

0xfe:
Hi folks -- trying to follow the thread and understand why we're trying to diff SVG output? (SVG representations are implementation detail, which can change across SVG libraries and browser versions.)

ronyeh:
Ah, there are two different things:
A) visual diff the SVG vs the Canvas output.
B) visual diff the SVG between reference and HEAD

Tom and I have recently encountered bugs in VexFlow where our Canvas and SVG contexts do different things. It'd be nice if our SVG and Canvas output the exact same PNG (or as close as possible). **Method A** is for detecting these inconsistencies.

Tom fixed the bugs in SVGContext, but our visual diffs showed no changes. That's because the command line tests **only test the CanvasContext**. The command line tests completely miss all changes to SVG-only code paths. **Method B** would be helpful here. We can run B only if we think we changed SVG-related code.

**Method B** might also be useful _if it's faster_ than our current visual diff via PNG comparisons. If the test suite could save out the SVG paths, then maybe there's a chance that doing string comparisons is faster than doing visual diffs? This can be a useful first pass. If there is any string diff, then the visual PNG diff can be run for only those test cases.

0xfe:
Ah, okay, gotcha -- good idea, thanks! LMK when this PR is ready for review.

tommadams:
Can we proceed with reviewing this PR please?

Originally, I'd hoped to add SVG support to the visual diff before submitting this PR, but jsdom doesn't support `getBBox` on SVG elements: https://github.com/jsdom/jsdom/issues/918

So text measuring doesn't work.

I've been thinking and once #1161 is in, it will be easier to configure the render context. We could then abstract texture measurement out into a `TextMeasurement` class that gets passed to the render context constructor (defaulting to the existing behaviour if unspecified). Then, when running under Node, we could pass in a `TextMeasurement` implementation that has its own canvas and 2d context that it uses solely for measuring text (it looks like node-canvas supports the `measureText` method).

0xfe:
Looks good, Tom. Thanks.

PR #1148, opened 2021-09-22
Fix Tests, Remove Partial<...>, Prefer Optional Fields

ronyeh:
This builds directly on top of Rodrigo's https://github.com/0xfe/vexflow/pull/1141

It includes Rodrigo's 5 commits, so I believe if we merge this one, we can close https://github.com/0xfe/vexflow/pull/1141

This is part 1 of migrating away from using `Partial<...>`. The main goal is to prevent Partial types from leaking out into the public API.

Part 2 will come in a second PR, which will remove the rest of the Partial, and include a couple `Required<...>`. Part 2 also comes with improvements to how we handle fonts (for text on the score, not for the music engraving).

There are 2 visual diffs. Both are due to bug fixes.

### Stave.Vertical_Bar_Test.Bravura
BEFORE:
Stave Vertical_Bar_Test Bravura_Reference

AFTER:
Stave Vertical_Bar_Test Bravura_Current

DIFF:
Stave Vertical_Bar_Test Bravura

The test now correctly shows a double vertical bar. This is due to Rodrigo's bug fix. Not sure why we have two methods that were almost identical.... (now they're actually identical!). We can address that in a future PR:
```
  stave.drawVerticalBar(50, true);
  stave.drawVerticalBar(150, false);
  stave.drawVertical(250, true);
  stave.drawVertical(300);
```

### Stave.Volta___Modifier_Measure_Test.Bravura
DIFF:
Stave Volta___Modifier_Measure_Test Bravura

The test now correctly shows the volta number.
It was a minor bug fix. The test previously passed in a constant that didn't exist. Thanks to TypeScript, we now realized it doesn't exist, and changed it to the correct constant.

Ready for review!

### DIFF
--- a/src/chordsymbol.ts
+++ b/src/chordsymbol.ts
@@ -492,9 +492,10 @@ export class ChordSymbol extends Modifier {
   // eslint-disable-next-line
   getSymbolBlock(parameters?: any): any {
     parameters = parameters ?? {};
-    const symbolType = parameters.symbolType ? parameters.symbolType : ChordSymbol.symbolTypes.TEXT;
-    const text = parameters.text ? parameters.text : '';
-    const symbolModifier = parameters.symbolModifier ? parameters.symbolModifier : ChordSymbol.symbolModifiers.NONE;
+    const symbolType = parameters.symbolType ?? ChordSymbol.symbolTypes.TEXT;
+    const text = parameters.text ?? '';
+    const symbolModifier = parameters.symbolModifier ?? ChordSymbol.symbolModifiers.NONE;
+
     const xShift = 0;
     const yShift = 0;
     const vAlign = 0;
--- a/src/chordsymbol.ts
+++ b/src/chordsymbol.ts
@@ -1,9 +1,9 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 //
-// ## Description
+// Author: @AaronDavidNewman
 //
-// This file implements chord symbols as modifiers that can be attached to
-// notes.  Chord symbols can contain multiple 'blocks' which can contain
+// This implements chord symbols as modifiers that can be attached to notes.
+// Chord symbols can contain multiple 'blocks' which can contain
 // text or glyphs with various positioning options.
 //
 // See `tests/chordsymbol_tests.ts` for usage examples.
@@ -15,17 +15,7 @@ import { TextFont } from './textfont';
 import { Modifier } from './modifier';
 import { FontInfo } from './types/common';
 import { StemmableNote } from './stemmablenote';
-
-export interface ChordSymbolBlock {
-  vAlign: boolean;
-  symbolModifier: number;
-  text: string;
-  yShift: number;
-  xShift: number;
-  width: number;
-  glyph: Glyph;
-  symbolType: number;
-}
+import { ModifierContextState } from 'modifiercontext';
 
 // To enable logging for this class. Set `Vex.Flow.ChordSymbol.DEBUG` to `true`.
 // eslint-disable-next-line
@@ -33,6 +23,45 @@ function L(...args: any[]): void {
   if (ChordSymbol.DEBUG) log('Vex.Flow.ChordSymbol', args);
 }
 
+export interface ChordSymbolBlock {
+  text: string;
+  symbolType: SymbolTypes;
+  symbolModifier: SymbolModifiers;
+  xShift: number;
+  yShift: number;
+  vAlign: boolean;
+  width: number;
+  glyph?: Glyph;
+}
+
+export enum HorizontalJustify {
+  LEFT = 1,
+  CENTER = 2,
+  RIGHT = 3,
+  CENTER_STEM = 4,
+}
+
+export enum VerticalJustify {
+  TOP = 1,
+  BOTTOM = 2,
+}
+
+export enum SymbolTypes {
+  GLYPH = 1,
+  TEXT = 2,
+  LINE = 3,
+}
+
+export enum SymbolModifiers {
+  NONE = 1,
+  SUBSCRIPT = 2,
+  SUPERSCRIPT = 3,
+}
+
+/**
+ * ChordSymbol is a modifier that creates a chord symbol above/below a chord.
+ * As a modifier, it is attached to an existing note.
+ */
 export class ChordSymbol extends Modifier {
   static DEBUG: boolean = false;
 
@@ -40,58 +69,47 @@ export class ChordSymbol extends Modifier {
     return 'ChordSymbol';
   }
 
-  protected static noFormat: boolean;
-
   protected symbolBlocks: ChordSymbolBlock[];
-
   protected horizontal: number;
   protected vertical: number;
   protected useKerning: boolean;
   protected reportWidth: boolean;
   protected font: FontInfo;
   protected textFont: TextFont;
+  /** Currently unused. */
+  protected static noFormat: boolean = false;
 
   // Chord symbols can be positioned and justified relative to the note.
-  static readonly horizontalJustify = {
-    LEFT: 1,
-    CENTER: 2,
-    RIGHT: 3,
-    CENTER_STEM: 4,
+  static readonly horizontalJustify = HorizontalJustify;
+
+  static readonly horizontalJustifyString: Record<string, HorizontalJustify> = {
+    left: HorizontalJustify.LEFT,
+    right: HorizontalJustify.RIGHT,
+    center: HorizontalJustify.CENTER,
+    centerStem: HorizontalJustify.CENTER_STEM,
   };
 
-  static readonly horizontalJustifyString: Record<string, number> = {
-    left: ChordSymbol.horizontalJustify.LEFT,
-    right: ChordSymbol.horizontalJustify.RIGHT,
-    center: ChordSymbol.horizontalJustify.CENTER,
-    centerStem: ChordSymbol.horizontalJustify.CENTER_STEM,
-  };
+  static readonly verticalJustify = VerticalJustify;
 
-  static readonly verticalJustify = {
-    TOP: 1,
-    BOTTOM: 2,
+  static readonly verticalJustifyString: Record<string, VerticalJustify> = {
+    top: VerticalJustify.TOP,
+    above: VerticalJustify.TOP,
+    below: VerticalJustify.BOTTOM,
+    bottom: VerticalJustify.BOTTOM,
   };
 
   static get superSubRatio(): number {
     return ChordSymbol.chordSymbolMetrics.global.superSubRatio;
   }
 
-  // ### NOTEXTFORMAT
-  // used to globally turn off text formatting, if the built-in formatting does not
-  // work for your font..
-  static get NOTEXTFORMAT(): boolean {
-    return typeof ChordSymbol.noFormat === 'undefined' ? false : ChordSymbol.noFormat;
-  }
-
-  static set NOTEXTFORMAT(val: boolean) {
+  /** Currently unused: Globally turn off text formatting, if the built-in formatting does not work for your font. */
+  static set NO_TEXT_FORMAT(val: boolean) {
     ChordSymbol.noFormat = val;
   }
 
-  static readonly verticalJustifyString: Record<string, number> = {
-    top: ChordSymbol.verticalJustify.TOP,
-    above: ChordSymbol.verticalJustify.TOP,
-    below: ChordSymbol.verticalJustify.BOTTOM,
-    bottom: ChordSymbol.verticalJustify.BOTTOM,
-  };
+  static get NO_TEXT_FORMAT(): boolean {
+    return ChordSymbol.noFormat;
+  }
 
   // eslint-disable-next-line
   static getMetricForGlyph(glyphCode: string): any {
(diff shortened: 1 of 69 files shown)


tests/bach_tests.ts
```diff
@@ -24,7 +25,7 @@ function minuet1(options: TestOptions): void {
   const id = (id: string) => registry.getElementById(id) as StaveNote;
 
   const f: Factory = VexFlowTests.makeFactory(options, 1100, 900);
-  const score = f.EasyScore({ throwOnError: true });
+  const score = f.EasyScore({ throwOnError: true } as EasyScoreOptions);
```
rvilarl, line 28: EasyScoreOptions should allow optional fields to avoid this casting.

rvilarl (changes requested):

ronyeh:
Hi Rodrigo (@rvilarl),

All the excessive casts in this PR are a temporary bridge to the final product, which will come in the next PR.

When I sent in the very big PR to your repo, we decided that it was too big to review properly and so I broke it up into 2 (or maybe 3) separate PRs.

The end goal for removing Partial is here: https://github.com/ronyeh/vexflow/tree/rvilarl-fix-notestruct

You will see that there are very minimal number of casts, due to extra use of Required and ?. 

Do you want me to include those changes into this PR? It will make it bigger, for sure.

tests/bach_tests.ts
ronyeh: Yup. The final product looks like: https://github.com/ronyeh/vexflow/blob/3118234a440b0549e6d89fea09a145861a70a621/src/easyscore.ts#L389-L394

If you'd like me to include all those changes in this PR, let me know. It will make this PR bigger, and it might have to include some of the font fixes.

src/voice.ts
```diff
@@ -85,7 +86,7 @@ export class Voice extends Element {
     };
 
     // Recalculate total ticks.
-    this.totalTicks = new Fraction(this.time.num_beats * (this.time.resolution / this.time.beat_value), 1);
+    this.totalTicks = new Fraction(this.time.num_beats * ((this.time.resolution as number) / this.time.beat_value), 1);
```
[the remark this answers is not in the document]
ronyeh: Yup! The original PR I sent to your repo included all the `Required<...>`.

See: https://github.com/ronyeh/vexflow/blob/3118234a440b0549e6d89fea09a145861a70a621/src/voice.ts#L55-L59

Shall I include all of those changes into this PR?

As long as we are OK making this PR bigger, I'm cool with it. I only ported over the changes for the files you originally modified, and I left "as number" casts where needed to make it compile.

I can include 100% of the removal of Partial, if we are OK reviewing a larger PR.

src/multimeasurerest.ts
```diff
@@ -125,14 +119,14 @@ export class MultiMeasureRest extends Element {
...
-    const y = this.checkStave().getYForLine(this.render_options.line);
+    const y = this.checkStave().getYForLine(this.render_options.line as number);
     const padding = (right - left) * 0.1;
 
     left += padding;
     right -= padding;
 
     const serif = {
-      thickness: this.render_options.serif_thickness,
+      thickness: this.render_options.serif_thickness as number,
```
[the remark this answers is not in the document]
ronyeh: Ahh, this one is tough! Even in the final product, I did not use Required.

That's because MultiMeasureRest supports two optional fields (padding_left and padding_right). These fields are calculated on the fly in the draw() method. So if I want to use Required, I might have to create a whole separate interface for this. 

See: 
https://github.com/ronyeh/vexflow/blob/3118234a440b0549e6d89fea09a145861a70a621/src/multimeasurerest.ts#L91-L102

Let me know if you have a better solution for this one! It stumped me, for sure.

src/curve.ts
```diff
@@ -139,7 +141,7 @@ export class Curve extends Element {
     let metric = 'baseY';
     let end_metric = 'baseY';
 
-    function getPosition(position: string | number) {
+    function getPosition(position?: string | number) {
```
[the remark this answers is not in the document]
ronyeh: Ah, I will just copy over the changes to `curve.ts` from the final product. In that one, I use Required so this ? can be removed.

ronyeh:
So it looks like I should probably include all changes from the final version into this PR. I will try to separate out the improvements to the Font handling into its own PR. (It's a bit more work, but probably cleaner and easier to review.)

rvilarl:
> So it looks like I should probably include all changes from the final version into this PR. I will try to separate out the improvements to the Font handling into its own PR. (It's a bit more work, but probably cleaner and easier to review.)

@ronyeh leave as it is,46 files is enough and one cast does not matter

tests/bach_tests.ts
rvilarl: we can leave this one cast and resolve it in the next PR

ronyeh:
OK, I included all my `Partial/Required` changes from the other branch. The other branch still has some Font improvements for text elements that I'll send in a separate PR.

This PR now has zero instances of `Partial<...>` and 11 instances of `Required<...>`.
I was also able to remove the temporary "as number" / "as XXX" casts.

I reran the visual diffs. Same two visual diffs from the bug fixes.

I think this is ready for review :-).

ronyeh:
As a summary:
* Use optional fields in interfaces instead of `Partial<...>` because it's one less stumbling block for a developer using VexFlow. For example, when declaring typed local options objects, a dev doesn't need to write `Partial<NoteStruct>`.
* Internally, we do the validation of all the fields, and provide sane defaults where possible. Then we assign the validated options object to a `Required<...>`. This allows internal methods to proceed without having to always check whether a field is undefined or not.
* Any accessors to the validated options object will return a `Required<...>` object, which will provide guarantees that every field is set.

rvilarl (approved):

src/chordsymbol.ts
```diff
@@ -490,11 +490,12 @@ export class ChordSymbol extends Modifier {
   // ChordSymbol allows multiple blocks so we can mix glyphs and font text.
   // Each block can have its own vertical orientation
   // eslint-disable-next-line
-  getSymbolBlock(parameters: any): any {
-    parameters = parameters == undefined ? {} : parameters;
-    const symbolType = parameters.symbolType ? parameters.symbolType : ChordSymbol.symbolTypes.TEXT;
-    const text = parameters.text ? parameters.text : '';
-    const symbolModifier = parameters.symbolModifier ? parameters.symbolModifier : ChordSymbol.symbolModifiers.NONE;
+  getSymbolBlock(parameters?: any): any {
```
0xfe, line 493: Why not just `getSymbolBlock(parameters = {})` and remove the next line?

src/chordsymbol.ts
```diff
@@ -548,7 +549,7 @@ export class ChordSymbol extends Modifier {
   // Add a text block
   // eslint-disable-next-line
   addText(text: string, parameters?: any): this {
-    parameters = parameters == null || parameters == undefined ? {} : parameters;
+    parameters = parameters ?? {};
```
0xfe, line 552: Same comment as above.

src/factory.ts
```diff
@@ -83,7 +83,20 @@ export class Factory {
...
+   *
+   * Create an SVG renderer and attach it to the DIV element named "boo" to render using <page-width> 1200 and <page-height> 600
+   *
+   * `const vf: Factory = Vex.Flow.Factory.newFromElementId('boo', 1200, 600 );`
+   */
+  static newFromElementId(elementId: string | null, width = 500, height = 200): Factory {
+    return new Factory({ renderer: { elementId, width, height, backend: Renderer.Backends.SVG } });
+  }
+
+  protected options: Required<FactoryOptions>;
```
0xfe, line 99: While we're at it, maybe we shouldn't be calling required parameters "options". The constructor should have explicit parameters for required parameters, and an options struct for optional stuff. Maybe here, we'll call it `params`?

src/factory.ts
```diff
@@ -200,44 +197,37 @@ export class Factory {
     return this.options.stave.space * spacing;
   }
 
-  Stave(paramsP: { x?: number; y?: number; width?: number; options?: Partial<StaveOptions> } = {}): Stave {
-    if (!this.options.renderer) throw new RuntimeError('NoRenderer');
+  Stave(params?: { x?: number; y?: number; width?: number; options?: StaveOptions }): Stave {
```
0xfe, line 200: Can we use default parameters here (in the destructured `params`)? E.g.,

```typescript
Stave(params: {x = 0; y = 0; width?: number; options?: StaveOptions}): Stave { ... }
```

0xfe (commented):
Looks great. Thanks for doing this! Minor comments:

src/chordsymbol.ts
rvilarl: In order to still allow the usage of `null`.

src/chordsymbol.ts
rvilarl: Same as bove

src/factory.ts
ronyeh: `Required<...>` is an annotation that tells the compiler (and someone reading the code) that we know all the fields are set to something (i.e., not undefined).

Should we rename all fields that are called `options` that we know have no undefined fields? I'm happy to call them all `params` as long as we are consistent.

ronyeh: But what about accessors? Should they be getOptions() / setOptions(), or getParams() / setParams(), or getParams() / setOptions() ??? :-)

src/chordsymbol.ts
ronyeh: I'm happy with either solution. The inline default argument `parameters={}` looks clean. This approach is also used in other places in VexFlow.

However, it is true that the nullish coalescing approach is more robust. If there is older code out there that passes in a null, it will still work as expected.

If we don't care about the extra robustness / supporting `null` (since this is a 4.0.0 release with breaking changes anyways), we could just stick with the inline default argument approach.

0xfe: Yeah, I think not supporting `null` is fine here.

src/factory.ts
0xfe: I'd argue that, in the general case, we shouldn't have accessors for `options` or `params`.

Anyhow, this isn't a strong opinion -- if this were new code, I'd be more firm about separating required parameters from options, but since we're porting JS code, it's okay.

src/factory.ts
ronyeh: Apparently no.  I tried it and got.
```
A type literal property cannot have an initializer.ts(1247)
```
I think TS treats that right side of the colon as an interface / type. But instead of us giving it an interface name, we used a literal.

If we do the same thing with an interface (e.g., StaveOptions) we get a similar error:
```
An interface property cannot have an initializer.ts(1246)
```

That's probably because TS erases all types from the JS output (after it passes all its error checks).

src/factory.ts
ronyeh: I did some thinking and also read some (only slightly) related posts:
- https://stackoverflow.com/questions/36495669/difference-between-terms-option-argument-and-parameter/36495940
- https://stackoverflow.com/questions/12826977/multiple-arguments-vs-options-object
- https://www.reddit.com/r/typescript/comments/7t8jpw/creating_a_class_with_options_object_instead_of/

It seems like "**Options Object**" is a commonly understood name for the _incoming thing_ that is provided to the constructor. 

However, **after we validate these options**, should we keep them inside a `this.options` field that is typed as `Required<XXXOptions>`?

I think the most "correct" way would be to not keep an internal options object at all. We would validate all the incoming options, and then assign them directly to instance variables which determine how the element will be rendered.

After the constructor has done its work, all the instance variables will have valid values, and we can use them directly:

- `this.options.fill_style` => `this.fillStyle`
- `this.options.stave.space` => `this.staveSpace`
- ...

There are 8 instances where we call `stave.getOptions()`. Other elements rely on the stave's internal metrics. In the world where we abandon internal options objects, all of those would instead call the individual readonly accessor for each field:
- `stave.getOptions().vertical_bar_width` => `stave.getVerticalBarWidth()`
- `stave.getOptions().spacing_between_lines_px` => `stave.getSpacingBetweenLines()`
- `stave.getOptions().num_lines` => `stave.getNumLines()`


I kind of like that, to be honest!

Do you think we should do that for the 4.0 release? Or should we just try it on `stave.ts` only and see what we think?

src/chordsymbol.ts
ronyeh: OK, I'll go for the inline (parameters = {}) approach. It's also used elsewhere in VexFlow.

src/factory.ts
ronyeh: OK, I made a slight improvement here.

There were only two declarations of `getOptions()`. Stave and Factory. The Factory one was only used by a test case (sanity check). So I removed it and accessed the protected member directly. The Stave.getOptions() was also unnecessary. Most of the options already had accessors. I implemented one additional accessor: stave.getVerticalBarWidth().

So now there are no more accessors for options.

We can discuss the other ideas (e.g., whether there should even be an internal `this.options` object) post 4.0 release, since it should only affect the internals of each class.

ronyeh:
### I addressed the review comments. Ready for a re-review. :-)
- There are no more `Partial<...>`. There are 9 instances of `Required<...>`. 
- Same two visual diffs as before (from the fixed bugs).


### Future Work
In a future PR (post 4.0.0 release) we might discuss whether classes should hold onto internal `options` objects. The alternative would be to assign the validated options directly to private fields. So instead of `this.options.spacing_between_lines_px`, it would be `this.spacingBetweenLinesPx`

Here are the classes that have internal options objects:
* `render_options` declared in Beam, Note, StaveTie, TextBracket.
* `protected render_options` declared in Accidental, Articulation, Bend, Curve, GraceNoteGroup, MultiMeasureRest, Ornament, PedalMarking, StaveHairpin, **Strokes**, Vibrato, VibratoBracket.
* `protected options` declared in Factory, Glyph, GlyphNote, Stave, StaveText, **Strokes**, Tuplet, Voice.
* `protected formatterOptions` declared in Formatter.
* ... possibly more

In many cases, the internal options object only has a couple of fields. For example, in Articulation:
```
  protected render_options: { font_scale: number };
```

Unless we really love the pseudo-namespacing that these options objects provide, perhaps it's better to assign the validated options to private or protected member variables. In the above example, Articulation objects would have an internal this.fontScale.

Anyways, this discussion can be revisited after 4.0.0! :-)

0xfe:
Thanks for your work here, Ron. Looks great.

PR #1153, opened 2021-09-25
Remove redundant moveTo in Glyph.renderOutline

### LINKED ISSUE #1150: SVGs don't render in ImageMagick correctly
ronyeh brought up in #1145 the fact that the visual regression currently doesn't test canvas rendering, which has recently been found to have a few bugs (#1126, #1145, #1142, #1140). I did a quick test and ran ImageMagick on the command line to try and diff some SVGs generated by `SVGContext` and none of them rendered correctly. It's possible that the SVGs are malformed somehow and browsers are just more forgiving than ImageMagick but we'll need to fix this before if we want to run visual regression tests on SVG output.

Here's the output from converting an SVG generated by `SVGContext` to PNG (`convert test.svg test.png`):
imagemagick output

At first glance it looks like all the glyphs might be missing.

tommadams:
All glyphs already start with an `M` command, so the `moveTo` at the start of `Glyph.renderOutline` doesn't do anything and seems to break ImageMagick's SVG rendering for some reason.

This fixes #1150.

Output from from ImageMagick's converter: `convert bach.svg bach.png`

imagemagick bach demo

### DIFF
--- a/src/glyph.ts
+++ b/src/glyph.ts
@@ -297,7 +297,6 @@ export class Glyph extends Element {
     const go = new GlyphOutline(outline, x_pos, y_pos, scale);
 
     ctx.beginPath();
-    ctx.moveTo(x_pos, y_pos);
     let x, y: number;
     while (!go.done()) {
       switch (go.next()) {


mscuthbert (approved):
Looks great!

Issue #1152, opened 2021-09-24
SVG paths have 15 decimal points of precision


ronyeh:
While looking at https://github.com/0xfe/vexflow/issues/1150, I noticed that our SVG outputs 15 decimal points of precision!

For a simple VexFlow demo like this:

![output](https://user-images.githubusercontent.com/239113/134722659-364e600a-c172-425d-82b4-3219ca8c60d3.jpg)

The output SVG looks like this:
<img width="1493" alt="SVG" src="https://user-images.githubusercontent.com/239113/134722446-63bb9746-cdcb-4476-a783-18844a5dbfc6.png">

The PNG is 3KB and the SVG is 8KB of text. And indeed, when modifying our visual diffs to output SVG paths & files instead of PNG files, I noticed the tests ran slower and the output folder took up much more disk space. That's probably mostly due to PNG compression vs uncompressed SVG text.... but it would certainly make for smaller SVG files if we rounded.

For SVG path diffing, we might consider rounding down to 3 or 4 decimal points of precision.

I found this educational thread over at the d3 project:
https://github.com/d3/d3-path/issues/10#issuecomment-275373270

https://github.com/d3/d3-path/issues/10#issuecomment-613939201

https://github.com/d3/d3-path/issues/10

The discussion (from a while ago) suggests that maybe the Chrome debugging experience suffers with excessively long SVG path strings.

In any case, SVG file sizes would be a bit smaller :-). Maybe we can round for the purposes of doing SVG diffing, since some of the comments suggest that different architectures might have tiny floating point differences that affect string diffing:

"The snapshot was generated on Mac but the unit tests were ran on Linux systems, and the floating point error makes the entire snapshot testing fail."

_Originally posted in https://github.com/0xfe/vexflow/issues/1150#issuecomment-926839055_

tommadams:
Wow, that d3 issue has been open for nearly five years.

Truncating to 3 decimal points of precision should be more than sufficient. As noted in the d3 issue, we should go with a custom rounding function rather than `.toFixed` to avoid padding with zeros.

Something like:
```typescript
// Round to a maximum of 3 decimal places
function round3(x: number): string {
  return (Math.round(x*1000)/1000).toString();
}
for (const x of [Math.PI, 1.0, 0.5555, 1.5555]) {
  console.log(round3(-x), round3(x));
}

// Output:
-3.142 3.142
-1 1
-0.555 0.556
-1.555 1.556
```

tommadams:
[...]

PR #1154, opened 2021-09-25
Round SVG coordinates to 3 decimal places

tommadams:
See discussion on #1152

### DIFF
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -34,6 +34,10 @@ const attrNamesToIgnoreMap: { [nodeName: string]: Attributes } = {
 /** Create the SVG in the SVG namespace. */
 const SVG_NS = 'http://www.w3.org/2000/svg';
 
+function round3(x: number): string {
+  return (Math.round(x * 1000) / 1000).toString();
+}
+
 interface State {
   state: Attributes;
   attributes: Attributes;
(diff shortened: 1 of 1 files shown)


src/svgcontext.ts
```diff
@@ -34,6 +34,10 @@ const attrNamesToIgnoreMap: { [nodeName: string]: Attributes } = {
 /** Create the SVG in the SVG namespace. */
 const SVG_NS = 'http://www.w3.org/2000/svg';
 
+function round3(x: number): string {
+  return (Math.round(x * 1000) / 1000).toString();
```
mscuthbert, line 38: All comments online about incorrect rounding in this case are irrelevant because we don't even really care about the third decimal place (but that's why it's better to keep it than try for 2 places).

src/svgcontext.ts
```diff
@@ -561,7 +566,9 @@ export class SVGContext implements RenderContext {
 
       const sweep = !antiClockwise;
 
-      this.path += `M${x0} ${y0} A${radius} ${radius} 0 ${+large} ${+sweep} ${x1} ${y1}`;
+      this.path += `M${round3(x0)} ${round3(y0)} A${round3(radius)} ${round3(radius)} 0 ${+large} ${+sweep} ${round3(
+        x1
+      )} ${round3(y1)}`;
       this.pen.x = x1;
```
mscuthbert, line 572: better to alias `const r3 = round3`; and fit on one line so as not to introduce newlines into the path variable.

(I wonder if just want to do that in the function name so people have to look up what it is instead of assuming it'll return a number not a string?)

src/svgcontext.ts
```diff
@@ -496,28 +500,29 @@ export class SVGContext implements RenderContext {
...
+    this.path += 'L' + round3(x) + ' ' + round3(y);
     this.pen.x = x;
     this.pen.y = y;
     return this;
   }
 
   bezierCurveTo(x1: number, y1: number, x2: number, y2: number, x: number, y: number): this {
-    this.path += 'C' + x1 + ' ' + y1 + ',' + x2 + ' ' + y2 + ',' + x + ' ' + y;
+    this.path +=
+      'C' + round3(x1) + ' ' + round3(y1) + ',' + round3(x2) + ' ' + round3(y2) + ',' + round3(x) + ' ' + round3(y);
```
mscuthbert, line 518: maybe clearer as a backtick string?  With the r3 alias mentioned below?

src/svgcontext.ts
```diff
@@ -538,8 +543,8 @@ export class SVGContext implements RenderContext {
       const y1 = y + radius * Math.sin(startAngle + Math.PI);
       // There's no way to specify a completely circular arc in SVG so we have to
       // use two semi-circular arcs.
-      this.path += `M${x0} ${y0} A${radius} ${radius} 0 0 0 ${x1} ${y1} `;
-      this.path += `A${radius} ${radius} 0 0 0 ${x0} ${y0}`;
+      this.path += `M${round3(x0)} ${round3(y0)} A${radius} ${radius} 0 0 0 ${round3(x1)} ${round3(y1)} `;
+      this.path += `A${round3(radius)} ${round3(radius)} 0 0 0 ${round3(x0)} ${round3(y0)}`;
```
mscuthbert, line 547: since we're doing the same rounding multiple times, should we store `const rx0 = round3(x0)` -- it's pretty fast though.

Should `A${radius} ${radius}` be `A${round3(radius)} ${round3(radius)}` in the first line also?

mscuthbert (commented):
Some potential confusions and speedups noted; looks like a great idea!

src/svgcontext.ts
0xfe: +1 to alias

src/svgcontext.ts
0xfe: +1 for template strings

tommadams:
Closing this in favour of the far superior #1158.

PR #1158, opened 2021-09-26
Use SVG Paths Directly

ronyeh:
Very early work in progress! I will update this thread as developments arise.

**This PR will be targeted for AFTER the 4.0.0 release.**

The purpose of this PR is to simplify how we extract and render the path strings in CanvasContext and SVGContext.

So far:
* Update Bravura and Petaluma fonts to the newest releases.
* The path strings are extracted from the OTF files and directly sent to the final SVGContext and CanvasContext (via Path2D). No parsing and modification of path strings.
* Since the path strings are extracted at 1000pt, all the numbers are integers. No need to round to the 3rd decimal point.

For example:
```
<path transform="translate(17 85) scale(0.02808 0.02808)" d="M97 125C186 125 295 43 295-42C295-93 255-125 198-125C88-125 0-44 0 42C0 94 43 125 97 125Z"></path>
```

Lots of stuff to do:
* So far, only note heads are affected. Need to make this work for all other glyphs.
* Adjust the scaling and translation to get all the glyphs rendered correctly.
* Gonville.... :-( I am working on a way to either 1) convert the old gonville_glyphs path strings to the new way, or 2) extract the glyphs from the newest July 2020 release of Gonville.
* Custom Font??? Will need to discuss with @0xfe about where these glyphs came from.

S 2021-09-26 at 16 17 26


ronyeh:
It also fixes the ordering of the parameters for the bezier and quadratic curve commands, as pointed out by: https://github.com/0xfe/vexflow/issues/1127#issuecomment-912250692

0xfe:
Really getting into the gnarley stuff here :-) It'd be great if @tommadams looked at this too.

src/glyph.ts
```diff
@@ -344,16 +355,22 @@ export class Glyph extends Element {
           bboxComp.addPoint(penX, penY);
           break;
         case OutlineCode.QUADRATIC:
+          x1 = go.nextX();
+          y1 = go.nextY();
           x = go.nextX();
           y = go.nextY();
-          bboxComp.addQuadraticCurve(penX, penY, go.nextX(), go.nextY(), x, y);
+          bboxComp.addQuadraticCurve(penX, penY, x1, y1, x, y);
```
tommadams, line 362: I have a branch that changes `BoundingBoxComputation`'s API to match the drawing API, i.e. `moveTo`, `bezierCurveTo`, etc. I can send a pull request with just that change once this PR is submitted if you like: https://github.com/tommadams/vexflow/blob/svg/src/boundingboxcomputation.ts

src/svgcontext.ts
```diff
@@ -573,6 +573,14 @@ export class SVGContext implements RenderContext {
     return this;
   }
 
+  fillPath(path: string, scale: number, x: number, y: number): this {
+    this.path = path;
+    this.pen.x = NaN;
```
tommadams, line 578: FYI: I think we can actually remove the `pen` member as it's not used any more and `CanvasRenderingContext2D` doesn't expose the pen position as far as I know.

tools/smufl/smufl_fontgen.js
```diff
@@ -11,40 +8,54 @@ const opentype = require('opentype.js');
...
 function LogError(...args) {
   // eslint-disable-next-line
-  console.error(...args)
+  console.error(...args);
 }
 
-// Converte OTF glyph path to Vexflow glyph path
+// Convert OTF glyph path to Vexflow glyph path.
 function toVFPath(glyph) {
+  // TODO: What if we convert it to 1000 pt???
```
tommadams, line 16: I think the 72 here comes from the fact that there are 72 points to an inch.

Ideally, the size passed here should be as close as possible to the actual size that the glyphs will be rendered. This is because the font rendering engine will fit the curve data to the pixel grid to prevent strokes being too thin to display. There's more about it here: https://docs.microsoft.com/en-us/typography/opentype/spec/ttch01#grid-fitting-a-glyph-outline

tommadams (commented):
This is great! Thanks for digging in to the gritty details.

src/canvascontext.ts
```diff
@@ -221,6 +221,26 @@ export class CanvasContext implements RenderContext {
     return this;
   }
 
+  /**
+   * @param path a SVG path string
+   */
+  fillPath(path: string, scale: number, x: number, y: number): this {
+    const ctx = this.vexFlowCanvasContext;
+    ctx.save();
```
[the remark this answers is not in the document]
ronyeh: Great suggestion! We definitely should opt for lighter weight alternatives when possible. I was just going by the VexFlow code convention (since I've only been a contributor for about 6 months).

I see a bunch of places (30+) where we use ctx.save() and ctx.restore(). Perhaps we can optimize all of these for ver 4.1.

https://github.com/0xfe/vexflow/blob/660f488b33e71e8a3ff1484c448e2efeaddfc4fa/src/stavevolta.ts#L74-L77

https://github.com/0xfe/vexflow/blob/660f488b33e71e8a3ff1484c448e2efeaddfc4fa/src/strokes.ts#L247-L250

src/svgcontext.ts
ronyeh: Yup, it looks like it is no longer used. I wonder when we stopped using it? I will remove it from this PR.

tools/smufl/smufl_fontgen.js
ronyeh: I think this is the most important part of the discussion.

What font size should we extract the glyphs at? The reason I chose 1000 is because the OTFs are designed on a 1000 x 1000 grid. As a result, all the paths come out as integers and look super clean and nice.

However, the final paths have transforms that look like `translate(17 85) scale(0.02808 0.02808)`. So we are scaling noteheads to 2.8% of its 1000pt size (so maybe around 28 pt).

Is there a good way to measure A) the performance impact and B) the rendering quality of scaling paths down to 2.8%?

Or should we pre-scale and pre-round the paths (to 3 decimal points) so that our final transforms look more like `translate(17 85) scale(0.5 0.5)`?

But what if a developer wants to draw at 4x or 8x or 16x size (for folks with poor visibility)?

Maybe we need to run a test. Ideally there is no performance difference, because we are drawing the same number of points, and just multiplying the path by a transform matrix that includes a translate and a scale.

Any ideas on this front? I do personally really really like the look of the all-integer paths (extracted at 1000 pt). :-)

ronyeh: I think I'm beginning to understand what you're saying. So we are using the opentype.js library to extract these paths. If we extract the paths at say, 36 pt, would the opentype.js library use font hinting or some other process to make the paths look best at 36 pt?

Hopefully the Bravura & Petaluma & Gonville designers didn't make any super thin lines at 1000 x 1000 that would disappear when I do a naive scale down to 2-5%.

I think we'd need to extract sets of paths at different pt sizes to see if there are any troublesome glyphs.

ronyeh:
Check out these beautiful gifs:

Bravura

Petaluma

Black is the current HEAD revision. Red/Blue is the new paths, extracted at 1000pt, and directly placed into the `<path d="..."></path>`

The original smufl_fontgen was extracting paths at 72pt and then scaling the coordinates by 20. It turns out all I had to do was multiply my scale factor by 1.44 to get an identical result. 72 * 20 / 1000.

Currently, the notehead transforms look somewhat like:
`transform="translate(258.5144654098348 150) scale(0.0404352 0.0404352)"`

So we are scaling glyphs down to about 4% of their original size. The entire notehead path group looks like:

```
<g class="vf-notehead" pointer-events="bounding-box">
  <path transform="translate(258.5144654098348 150) scale(0.0404352 0.0404352)" fill="blue" d="M97 125C186 125 295 43 295-42C295-93 255-125 198-125C88-125 0-44 0 42C0 94 43 125 97 125Z"></path>
</g>
```

src/glyph.ts
ronyeh: Can bounding boxes be precomputed? When I extract the paths via opentype.js, I can also access `glyph.getBoundingBox()`, which provides `xMin`, `yMin`, `xMax`, `yMax`.

Assuming their bounding box calculation is accurate, can we just scale and translate the bounding box for our purposes?

ronyeh: The BoundingBoxComputation class comment includes:

> This class is merely a crutch to get bounding box information without explicit metadata. This is likely to get deprecated following SMuFL support.

So as long as we can properly extract the bounding boxes from Bravura and Petaluma (and any other SMuFL font), then we should be good to go?

Maybe we need to eventually drop Gonville and replace it with Leland: https://github.com/MuseScoreFonts/Leland

It'd be nice to be 100% SMuFL.

tommadams: It should be possible to precompute the bounding boxes; the glyph cache I added makes a copy of the bounding box so that it only has to be computed once. So the trade off would be making the glyph data a little larger but removing the need to calculate the bounding box the first time a glyph is used.

tools/smufl/smufl_fontgen.js
tommadams: Yep, I think the only way to really know for sure is to try it and see what they look like extracted at 1000 x 1000.

tommadams:
Wow, that's awesome work. Congrats, this is really exciting.

ronyeh:
Will revisit after 4.0 is released.

PR #1168, opened 2021-10-01
Remove drawVertical from Stave

### LINKED ISSUE #1156: Draw vertical bar
In Stave the following methods:
- `drawVertical` & `drawVerticalFixed` are identical
- `drawVerticalBar` & `drawVerticalBarFixed` are also identical

There is a comment in the source code:
```
// Draw Simple barlines for backward compatability
// Do not delete - draws the beginning bar of the stave
```

However the Bars of a stave are drawn in reality by the method `drawVerticalBar` from `BarLine` and the four methods from `Stave` are used only in specific tests which aim to test them.

I suggest to delete the four methods.

rvilarl:
Fix #1156

### DIFF
--- a/src/stave.ts
+++ b/src/stave.ts
@@ -773,38 +773,6 @@ export class Stave extends Element {
     return this;
   }
 
-  // Draw Simple barlines for backward compatability
-  // Do not delete - draws the beginning bar of the stave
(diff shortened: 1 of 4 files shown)


0xfe:
Looks good, thanks. @ronyeh for the changelog.

PR #1170, opened 2021-10-03
fix #1169 applyAccidentals

### LINKED ISSUE #1169: Missing accidentals when using "VF.Accidental.applyAccidentals"
There seems to be an issue with `VF.Accidental.applyAccidentals` where accidentals are not following the "octave rule".

Example code below, which adds the following notes:

```
c/4, c#/4, c#/5, c/5
```

and then calls `applyAccidentals` with a keySignature value of `C`.

Here is the VF output:

Screen Shot 2021-10-01 at 6 03 38 PM

Here are the same notes in the same key signature in Sibelius:

Screen Shot 2021-10-01 at 6 00 15 PM

As you can read [here](https://music.stackexchange.com/a/80512/65668) (quoted from "Behind Bars" by Elaine Gould), "[...] each additional octave requires a further accidental". The same rule is explained elsewhere, see [here](https://courses.lumenlearning.com/musicappreciation_with_theory/chapter/theory-lesson-music-basics/) -- "Accidentals apply within the measure and octave in which they appear [...]"

---

I believe that the issue may be with [this line](https://github.com/0xfe/vexflow/blob/961f26aca6e1ad5244382925d2ece1d8d1837a09/src/accidental.ts#L431) which checks identical *pitches*, not taking octaves into account.

But I may be wrong.

---

Repro source code:

```
import Vex from 'vexflow';

const VF = Vex.Flow;

const div = document.getElementById("vf")
const renderer = new VF.Renderer(div, VF.Renderer.Backends.SVG);

renderer.resize(1000, 1000);
const context = renderer.getContext();
context.setFont("Arial", 10, "").setBackgroundFillStyle("#eed");
const stave = new VF.Stave(0, 0, 300);
stave.addClef("treble").addTimeSignature("4/4");
stave.setContext(context).draw();

var notes = [
  new VF.StaveNote({clef: "treble", keys: ["c/4"], duration: "q" }),
  new VF.StaveNote({clef: "treble", keys: ["c#/4"], duration: "q" }),
  new VF.StaveNote({clef: "treble", keys: ["c#/5"], duration: "q" }),
  new VF.StaveNote({clef: "treble", keys: ["c/5"], duration: "q" })
];

var voice = new VF.Voice({num_beats: 4,  beat_value: 4});
voice.addTickables(notes);

// apply accidentals
VF.Accidental.applyAccidentals([voice], 'C');

var formatter = new VF.Formatter().joinVoices([voice]).format([voice], 200);

voice.draw(context, stave);
```

rvilarl:
Fixes #1169

### DIFF
--- a/src/accidental.ts
+++ b/src/accidental.ts
@@ -397,7 +397,8 @@ export class Accidental extends Modifier {
     if (!keySignature) keySignature = 'C';
 
     // Get the scale map, which represents the current state of each pitch.
-    const scaleMap = music.createScaleMap(keySignature);
+    const scaleMapKey = music.createScaleMap(keySignature);
+    const scaleMap: Record<string, string> = {};
 
     tickPositions.forEach((tickPos: number) => {
       const tickables = tickNoteMap[tickPos];
@@ -416,6 +417,7 @@ export class Accidental extends Modifier {
         const staveNote = t;
         staveNote.keys.forEach((keyString: string, keyIndex: number) => {
           const key = music.getNoteParts(keyString.split('/')[0]);
+          const octave = keyString.split('/')[1];
 
           // Force a natural for every key without an accidental
           const accidentalString = key.accidental || 'n';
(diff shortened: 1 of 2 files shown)


0xfe:
Thanks Rodrigo. I think this breaks the "remembered" aspect of the accidentals, no? If a note is flattened, the flat does not need to be rendered a second time -- it's implied. 

@mscuthbert

rvilarl:
@0xfe have you seen the images in #1169?

0xfe:
Yes, you're right -- I missed that this was across octaves. Can you add a few more tests for the corner cases please?

rvilarl:
What do you mean with corner cases?

0xfe:
The cases where the accidentals are switched or remembered.

Examples:

1) In a single octave:

- Key of C: C/4, C#/4, C#/4, C/4, C/4, Cb/4, Cb/4, C/4, C/4
- Expected Output: C, C#, C, Cn, C, Cb, C, Cn, C (no acc, sharp, no acc, natural, no acc, flat, no acc, natural, no acc)

2) Interleave the same notes as (1) but one octave higher in between these notes (e.g., C/4, C/5, C#/4, C#/5, ...), and the accidentals should switch the same way within the octaves.

3) Same as (1) in the key of C#

4) Same as (2) in the key of C#

rvilarl:
@0xfe tests included, ready to review and merge

0xfe:
Great! Thank you!

PR #1171, opened 2021-10-03
Remove "as StaveNote" from src

rvilarl:
Minor chnage to remove unnecessary casting to `StaveNote`


0xfe (commented):
Thanks Rodrigo. Looks good.

PR #1174, opened 2021-10-03
fix #1124 !this.x

### LINKED ISSUE #1124: Possible small issue with draw() checking if (!this.x)
In the draw() method of clef.ts, keysignature.ts, and timesignature.ts:

We have code that looks like `if (!this.x) throw new RuntimeError()`.

However, StaveModifier.x is a number that defaults to 0. That means !this.x will throw a RuntimeError if x is zero.

Is that the intention of this check? Can this.x ever be set to undefined?

If the code is correct as is, we should probably add a comment that says if x is 0 or undefined, we'll throw an error.

In any case, even if it's a bug, it's probably an edge case that no one ever encounters (since they would immediately get a RuntimeError() if they did encounter it).

rvilarl:
Fixes #1124

### DIFF
--- a/src/accidental.ts
+++ b/src/accidental.ts
@@ -12,7 +12,6 @@ import { Glyph } from './glyph';
 import { ModifierContextState } from './modifiercontext';
 import { Voice } from './voice';
 import { Note } from './note';
-import { StaveNote } from './stavenote';
 import { Tickable } from './tickable';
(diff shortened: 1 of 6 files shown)


rvilarl:
@ronyeh this PR is minor, could you review?

ronyeh (approved):

mscuthbert:
Looked great to me.  Thanks for the cleanup.

Issue #1177, opened 2021-10-05
Dots are gone in svg


AaronDavidNewman:
Top is canvas.
Bottom is svg :(

![image](https://user-images.githubusercontent.com/5438280/136119539-55b5e343-e21d-4ff3-a057-4b6ee09001ff.png)

ronyeh:
I see this on my head revision too, and apparently it's all of the tests, since I'm looking at Bach Demo. 

Our visual regression tests don't check SVG. :-) :-(

AaronDavidNewman:
It seems like the issue is in svg.arc.  I'm not sure what other modifiers use arc...

ronyeh:
Possible related PR https://github.com/0xfe/vexflow/pull/1142

This was before @tommadams and I discussed the fact that visual diffs don't cover SVG modifications. We should address that at some point. 😉

ronyeh:
This discussion reminds me that I think SVGContext and CanvasContext could extend an abstract RenderContext which defines a high level API like `fillCircle(x, y, radius)` or other useful methods VexFlow needs. 

Then the individual subclasses can implement those high level methods whichever way they please. CanvasContext could use arc() and SVGContext could call circle().

Currently we have a common.d.ts which defines the RenderContext interface. It could be easily moved out to its own abstract class file when we fix this bug. 

Thoughts? @0xfe @tommadams @rvilarl

tommadams:
Sorry I haven't been communicative the past few days, work's been taking up all my time. I'll take a look at this first thing tomorrow morning.

ronyeh:
@tommadams no worries! Work is important.

You can help review https://github.com/0xfe/vexflow/pull/1178

The main change is adding two equal signs: >= instead of >

"No visual diffs!" The bigger problem here is that our visual regression tests don't catch SVG regressions or bug fixes.

PR #1178, opened 2021-10-06
Restore SVG dots

ronyeh:
Add equal signs to handle the case where the arc is exactly a circle.
`>= TWO_PI`vs `> TWO_PI`

Fixes https://github.com/0xfe/vexflow/issues/1177

Rename antiClockwise to match the HTML specification naming convention:
https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-arc

### DIFF
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -34,6 +34,8 @@ const attrNamesToIgnoreMap: { [nodeName: string]: Attributes } = {
 /** Create the SVG in the SVG namespace. */
 const SVG_NS = 'http://www.w3.org/2000/svg';
 
+const TWO_PI = 2 * Math.PI;
+
 interface State {
   state: Attributes;
   attributes: Attributes;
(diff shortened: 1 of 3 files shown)


src/svgcontext.ts
```diff
@@ -521,16 +523,16 @@ export class SVGContext implements RenderContext {
...
 
-    // Handle the edge case from the Canvas spec where arc length is greater than
+    // Handle the edge case where arc length is greater than or equal to
     // the circle's circumference:
     //   https://html.spec.whatwg.org/multipage/canvas.html#ellipse-method-steps
     if (
-      (!antiClockwise && endAngle - startAngle > 2 * Math.PI) ||
-      (antiClockwise && startAngle - endAngle > 2 * Math.PI)
+      (!counterclockwise && endAngle - startAngle >= TWO_PI) ||
+      (counterclockwise && startAngle - endAngle >= TWO_PI)
```
ronyeh, line 535: The main change is using >= instead of > on these two lines.

0xfe:
Thanks for the fix. LGTM.

tommadams:
Thanks for the fix!

PR #1181, opened 2021-10-07
legacy fonts migrated

rvilarl:
No functional impact but it should improve the % of TS in the project which is a good psychological result :)
@0xfe I asked you once and you did not want to delete them. Hopefully the migration is fine (only rename was required)


ronyeh:
> improve the % of TS in the project which is a good psychological result :)

I'm ok with the fact that we currently check in `releases/`, but we should definitely consider grouping them by version number. We could have sub-folders inside the releases folder (e.g., 3.0.9, 4.0). 

Having pre-built releases makes for convenient regression testing. However, if we are against increasing the JS % in our project, we can move the releases out to a different repository, and check it out as a git submodule.

Also, as Simon says, we should continue to make use of the releases page on GitHub: https://github.com/0xfe/vexflow/releases

@rvilarl if you like to see the TS% go up, one next step could be to remove the jQuery js file :-) I already removed the main dependency from the tests in a previous PR. I think only the legacy font tests page relies on it (`glyphs.html`). Perhaps you could just point it to the correct jQuery version hosted on a CDN, or eliminate it alltogether. I think we mostly only use the $ function to run code on document.ready.

PR #1182, opened 2021-10-08
casting to StemmableNote removed

rvilarl:
Adding a few methods to `Note` it is possible to remove all the casting to `StemmableNote` in the source base. I find this generalisation very useful.

The method in `Note` are then overloaded in `StemmableNote`. The changes are not complex.

Several `ts-nocheck` resolved.

No Visual differences.

### DIFF
--- a/src/note.ts
+++ b/src/note.ts
@@ -484,13 +484,14 @@ export abstract class Note extends Tickable {
   }
 
   getStem(): Stem | undefined {
-    return undefined;
+    throw new RuntimeError('NoStem', 'No stem attached to instance');
   }
 
   getBeamCount(): number {
-    return 0;
+    throw new RuntimeError('NoStem', 'No stem attached to instance');
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   setStemDirection(direction?: number): this {
     throw new RuntimeError('NoStem', 'No stem attached to instance');
   }


0xfe:
Thanks for the change, Rodrigo. Can you fix the conflicts from the last merge please?

rvilarl:
@0xfe done, ready to merge.

@0xfe @ronyeh I had to include the path in the import from `util` from `RuntimeError` in renderer_text. VSCode was considering a Node internal module. As a consequence I would prefer to add all the paths. Of course in another PR.

ronyeh:
I'm OK with adding all the relative paths (e.g., ../src/). That will avoid
any ambiguities like Element or Util. Feel free to do a search and replace
and submit a PR on this.

The only consideration is that VS code auto imports using the short form.
We might end up with a mix of styles (which is what we had before I made
all the tests consistent). I wonder if there is a setting that I'm missing
that forced vs code to auto import with relative paths. I can do some
research on this when I'm back with my computer.

src/beam.ts
```diff
@@ -29,7 +28,7 @@ function calculateStemDirection(notes: StemmableNote[]) {
   return Stem.UP;
 }
 
-function getStemSlope(firstNote: StemmableNote, lastNote: StemmableNote) {
+function getStemSlope(firstNote: Note, lastNote: Note) {
```
0xfe, line 31: I'm trying to understand the rationale here -- it feels like we're replacing compile-time checks with runtime checks. It's good to have `StemmableNote` a parameter here (you can't beam notes are not stemmable.)

src/note.ts
```diff
@@ -482,6 +483,26 @@ export abstract class Note extends Tickable {
     return false;
   }
 
+  getStem(): Stem | undefined {
+    return undefined;
+  }
+
+  getBeamCount(): number {
+    return 0;
```
0xfe, line 491: This feels like a potential silent failure. For notes that don't have stems, this should fail loudly, no?

sschmidTU: I mean, it's true though that e.g. a whole note has 0 beams. It might depend on why and how the getter is used. If i iterate over an array of notes that might contain whole notes, should it throw on the whole note?

PR #1183, opened 2021-10-08
Make RenderContext an abstract base class

### LINKED ISSUE #1179: Make `RenderContext` an abstract base class
I'm opening this issue to see how folks feel about making `RenderContext` an abstract base class instead of an interface.

When submitted, #1161 will allow users to pass their own custom `RenderContext` to the `Renderer` constructor.

Because `RenderContext` is currently an interface we have to use a fairly obscure type guard to figure out if an argument passed to the `Renderer` constructor is a `RenderContext`:
 ```typescript
    function isRenderContext(obj: any): obj is RenderContext {
      return obj.setShadowBlur !== undefined;
    }
```

It would be more readable if we could write `obj instanceof RenderContext`.

Work required:
 - Remove the `RenderContext` interface from `types/common.d.ts` and create a new `src/rendercontext.ts` file containing the abstract class definition.
 - Change `SVGContext` and `CanvasContext` to extend `RenderContext` instead of implement it.
 - Update `RenderContext` imports to point to the new location.

The last step will require touching a lot of files but the changes are pretty straight forward.

### PR DESCRIPTION
**Pull request by tommadams**
fixes #1179

I also removed the `draw` method from `BoundingBox` because it was only used by a couple of unit tests, which didn't seem worth taking a dependency on `RenderContext` for.

### DIFF
--- a/tests/stavenote_tests.ts
+++ b/tests/stavenote_tests.ts
@@ -108,7 +108,9 @@ function draw(
   note.setContext(context).draw();
 
   if (drawBoundingBox) {
-    note.getBoundingBox().draw(context);
+    const bb = note.getBoundingBox();
+    context.rect(bb.getX(), bb.getY(), bb.getW(), bb.getH());
+    context.stroke();
   }
   return note;
 }

### CODE REVIEW

File: tests/accidental_tests.ts [hunk 1]
```diff
@@ -10,12 +10,12 @@ import { Beam } from 'beam';
 import { Formatter } from 'formatter';
 import { Modifier } from 'modifier';
 import { ModifierContext } from 'modifiercontext';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 13: `from 'rendercontext'` should include path. The compiler is not complaining due to `ts-nocheck`. I have resolved a few in #1182 

It should be `from '../src/rendercontext'`

tests/boundingboxcomputation_tests.ts
```diff
@@ -6,7 +6,7 @@
 import { VexFlowTests, TestOptions } from './vexflow_test_helpers';
 import { BoundingBoxComputation } from 'boundingboxcomputation';
 import { Glyph, OutlineCode } from 'glyph';
-import { RenderContext } from 'types/common';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 9: as above

tests/dot_tests.ts
```diff
@@ -6,11 +6,11 @@
 import { VexFlowTests, TestOptions } from './vexflow_test_helpers';
 import { Beam } from 'beam';
 import { ModifierContext } from 'modifiercontext';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 9: as above

tests/gracetabnote_tests.ts
```diff
@@ -9,9 +9,9 @@ import { Flow } from 'flow';
 import { Formatter } from 'formatter';
 import { GraceNoteGroup } from 'gracenotegroup';
 import { GraceTabNote } from 'gracetabnote';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 12: as above

tests/notehead_tests.ts
```diff
@@ -14,12 +14,12 @@ import { VexFlowTests, TestOptions } from './vexflow_test_helpers';
 import { Flow } from 'flow';
 import { Formatter } from 'formatter';
 import { NoteHead } from 'notehead';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 17: as above

tests/percussion_tests.ts
```diff
@@ -12,11 +12,11 @@
 import { VexFlowTests, TestOptions } from './vexflow_test_helpers';
 import { ContextBuilder } from 'renderer';
 import { Factory } from 'factory';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 15: as above

tests/renderer_tests.ts
```diff
@@ -10,11 +10,11 @@
 import { CanvasContext } from 'canvascontext';
 import { Factory, FactoryOptions } from 'factory';
 import { Formatter } from 'formatter';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 13: as above

tests/rests_tests.ts
```diff
@@ -4,10 +4,10 @@
 // Rests Tests
 
 import { TestOptions, VexFlowTests } from './vexflow_test_helpers';
-import { RenderContext } from 'types/common';
 import { Beam } from 'beam';
 import { Flow } from 'flow';
 import { Formatter } from 'formatter';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 10: as above

tests/stavehairpin_tests.ts
```diff
@@ -10,7 +10,7 @@
 // TODO: Incorrect property names in the options object: vo, left_ho, right_ho.
 
 import { VexFlowTests, TestOptions } from './vexflow_test_helpers';
-import { RenderContext } from 'types/common';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 13: as above

tests/stavenote_tests.ts
```diff
@@ -21,14 +21,14 @@ import { Formatter } from 'formatter';
 import { Fraction } from 'fraction';
 import { FretHandFinger } from 'frethandfinger';
 import { Modifier, ModifierPosition } from 'modifier';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 24: as above

tests/tabnote_tests.ts
```diff
@@ -6,12 +6,12 @@
 import { VexFlowTests, TestOptions } from './vexflow_test_helpers';
 import { Flow } from 'flow';
 import { Formatter } from 'formatter';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 9: as above

tests/tabslide_tests.ts
```diff
@@ -7,10 +7,11 @@ import { VexFlowTests, TestOptions } from './vexflow_test_helpers';
 import { ContextBuilder } from 'renderer';
 import { Flow } from 'flow';
 import { Formatter } from 'formatter';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 10: as above

tests/tabtie_tests.ts
```diff
@@ -8,10 +8,11 @@ import { ContextBuilder } from 'renderer';
 import { Annotation } from 'annotation';
 import { Flow } from 'flow';
 import { Formatter } from 'formatter';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 11: as above

tests/vexflow_test_helpers.ts
```diff
@@ -5,7 +5,7 @@
 
 import { Flow } from 'flow';
 import { Assert } from './types/qunit';
-import { RenderContext } from 'types/common';
+import { RenderContext } from 'rendercontext';
```
rvilarl, line 8: as above

rvilarl (changes requested):

tests/accidental_tests.ts
0xfe: Thanks @rvilarl. Why is the above like okay (`ModiferContext`) and this one not? I'm trying to understand the difference.

rvilarl: While removing the `ts-nocheck` in #1182 the editor was complaining. I assumed that it was the removal of `ts-nocheck` but now I see that it does compile. It is perhaps a configuration problem in Visual Studio Code.

Captura de pantalla de 2021-10-10 23-22-15
`

rvilarl: It is anyway strage to use the full path on the source files and not in the tests, is not it?

ronyeh: Hi all, I won't have much time for VexFlow until middle of this week, but wanted to chime in on this.

The reason we can omit the `../src/` from the import is because in the `tsconfig.json` we have:

```
"baseUrl": "./src",
```

See: https://www.typescriptlang.org/docs/handbook/module-resolution.html#relative-vs-non-relative-module-imports

This is a non-relative import that is resolved against the baseUrl. Rodrigo, this is similar to how you used path mapping to handle importing of the Bravura / Petaluma / Goneville files.

https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping

```
    "paths": {
      "@bravura": ["fonts/loadStatic"],
      "@gonville": ["fonts/loadStatic"],
      "@petaluma": ["fonts/loadStatic"],
      "@custom": ["fonts/loadStatic"]
    }
```

So like you are able to do the following without specifying `./fonts/loadStatic`...
```
import { loadBravura } from '@bravura';
import { loadGonville } from '@gonville';
import { loadPetaluma } from '@petaluma';
```

We are able to do the following:
```
import { ModifierContext } from 'modifiercontext';
```

without having to specify ../src/modifiercontext.


Now this only works if you open VSCode to the root folder that contains the tsconfig.json. If you open the editor to the tests file directly without the larger context of the project, it will not be able to resolve the module against the `baseUrl`.

The question remains.... "why did we do this?" :-)

I think I noticed the inconsistency when I was porting all of the test files to TS. Sometimes we would have 
`import {XXX} '../src/xxx' and other times I noticed import { YYY } from 'yyy'`

**I believe the short version was introduced by VSCode's auto imports.** 

For example, when I start typing `new StaveNo`... a VSCode auto complete menu pops up and suggests StaveNote. If I hit enter, it will add the import as follows:

`import { StaveNote } from 'stavenote';`

VSCode defaults to resolving against the baseUrl.

In the end, I decided to make all the tests consistent, and eliminated use of the '../src/XXX' style of imports from the tests only.

The benefits:
* I think it looks cleaner.
* I think it is slightly more representative of what a developer's code will look like. I'm not yet sure what the import style will look like when a developer npm installs vexflow, but I am pretty sure it will _not_ look like: `import { RenderContext } from '../src/rendercontext'`. I would prefer if the test suite can serve as a good set of examples to learn from.

For reference, from some of my other TS projects, imports can look like:
```
import { useRouter } from "next/router";
import React, { useEffect, useState } from "react";
import theme from "styles/theme";
import styled from "@emotion/styled";
```

We will have to figure out how developers will be able to import individual classes from within the vexflow npm package. I know in the past, we recommended that devs just import the top level VF object, and then instantiate classes like new VF.XXX(), but I think this doesn't feel like a proper TypeScript-y way to do things.

We should definitely figure this out before npm pushing vexflow 4.0. :-)


So what do you think?

If you want, you can feel free to add ../src/XXX to all the imports. But at the moment, all tests use the short form (non-relative import paths) while all the src/ files use relative import paths (e.g., `./XXX`).

I generally prefer the cleaner look.

ronyeh: > While removing the `ts-nocheck` in #1182 the editor was complaining. I assumed that it was the removal of `ts-nocheck` but now I see that it does compile. It is perhaps a configuration problem in Visual Studio Code.

Did you open the `accidental_tests.ts` file directly? I see it says "No se encuentra el módulo". If you always open VSCode at the root folder of the project (i.e., `vexflow/`) it will see the `tsconfig.json`, and all the modules should resolve correctly.

rvilarl: > > While removing the `ts-nocheck` in #1182 the editor was complaining. I assumed that it was the removal of `ts-nocheck` but now I see that it does compile. It is perhaps a configuration problem in Visual Studio Code.
> 
> Did you open the `accidental_tests.ts` file directly? I see it says "No se encuentra el módulo". If you always open VSCode at the root folder of the project (i.e., `vexflow/`) it will see the `tsconfig.json`, and all the modules should resolve correctly.

Thanks for the tip, I was opening VSCode from the launch menu. You are right if I open it from the command line at vexflow/ , VSCode resolves it correctly. There must be a way to make it work without using the command line, hopefully. I will look into that.

rvilarl: I did not find a way to make it work consistently with VSCode but knowing the tip from @ronyeh for me both approaches are fine (with paths or without path)

@0xfe what is your preference? I will update #1182 if we go without path.

ronyeh: Honestly I am fine with either style of import. The only issue is that vscode (on macOS at least) seems to automatically insert the short form when I use the drop down to auto import. 

Can you confirm if that is the case too on your OS? If you start typing code like `new StaveNo` a autocomplete context menu should show up suggesting an import for StaveNote. If you select it and hit enter, what is inserted at the top of the file?

Mostly I'm just being lazy. If vscode chooses the non-relative import by default, I would prefer not having to go up and change it to be a relative path like ../src/stavenote

0xfe: I'm okay with either too, so long as we're consistent. For now, in the interest of not holding this PR up any longer, let's treat that as a separate issue that we may (or may not) want to fix.

tommadams: Somewhat related: I tried removing the `@ts-nocheck` from `renderer_tests.ts` and found that the ['util' import](https://github.com/0xfe/vexflow/blob/9f84f36fdd537750024615f57b6cbd94777895ba/tests/renderer_tests.ts#L18) is broken because it seems to resolve to `node_modules/@types/node/util.d.ts` instead of `src/util.ts`. The build fails with the following error:

```
ERROR in /Users/tommadams/code/web/vexflow/tests/renderer_tests.ts
15:9-21
[tsl] ERROR in /Users/tommadams/code/web/vexflow/tests/renderer_tests.ts(15,10)
      TS2305: Module '"util"' has no exported member 'RuntimeError'.
```

_However_, when I set `traceResolution = true` and rebuild, the compiler reports that it has resolved `util` to the correct location, so I don't really know what's going on here:

```
======== Resolving module 'util' from '/Users/tommadams/code/web/vexflow/tests/renderer_tests.ts'. ========
Explicitly specified module resolution kind: 'NodeJs'.
'baseUrl' option is set to '/Users/tommadams/code/web/vexflow/src', using this value to resolve non-relative module name 'util'.
'paths' option is specified, looking for a pattern to match module name 'util'.
'baseUrl' option is set to '/Users/tommadams/code/web/vexflow/src', using this value to resolve non-relative module name 'util'.
Resolving module name 'util' relative to base url '/Users/tommadams/code/web/vexflow/src' - '/Users/tommadams/code/web/vexflow/src/util'.
Loading module as file / folder, candidate module location '/Users/tommadams/code/web/vexflow/src/util', target file type 'TypeScript'.
File '/Users/tommadams/code/web/vexflow/src/util.ts' exist - use it as a name resolution result.
======== Module name 'util' was successfully resolved to '/Users/tommadams/code/web/vexflow/src/util.ts'. ========
```

ronyeh: I'm sure there will be problems importing "util" and "element" and other really common names that might conflict with node packages or HTML types.

I've been reading up on this, and many folks seem to be against using relative imports (like ../src/). One commonly suggested solution is to use path aliases for webpack & vscode to resolve imports. We already do this for dynamically loading fonts (thanks to Rodrigo's PR this summer). For example we already do this in our tsconfig.json:

```
    "paths": {
      "@bravura": ["fonts/loadStatic"],
      "@gonville": ["fonts/loadStatic"],
      "@petaluma": ["fonts/loadStatic"],
      "@custom": ["fonts/loadStatic"]
    },
```

```
import { loadBravura } from '@bravura';
```

So we could modify the tsconfig.json to look like this:

```
...
    "baseUrl": "./",
    "paths": {
      "@vex/*": ["src/*"],
      "@tests/*": ["tests/*"],
      "@bravura": ["src/fonts/loadStatic"],
      "@gonville": ["src/fonts/loadStatic"],
      "@petaluma": ["src/fonts/loadStatic"],
      "@custom": ["src/fonts/loadStatic"]
    }
...
```

This would allow us to do this in our test files:
```
import { VexFlowTests, TestOptions } from '@tests/vexflow_test_helpers';
import { Flow } from '@vex/flow';
```

The old way:
```
import { VexFlowTests, TestOptions } from './vexflow_test_helpers';
import { Flow } from 'flow';
```

We could do this for just our test files for now. This would probably resolve issues with importing modules like "util".

tommadams: I agree that using a `@vex/` prefix for imports in test code sounds like a good idea, but I still don't understand why the module under node_modules is preferred over the one under src. The typescript docs say this about non-relative imports with using node module resolution:

> All module imports with non-relative names are assumed to be relative to the baseUrl.

So it seems like the imports should be resolved correctly.

PR #1184, opened 2021-10-12
Address issue #1070

### LINKED ISSUE #1070: render_options (protected?)
@0xfe @ronyeh Considering #348 I thought that `render_options` in many classes was a good candidate to start with. However I have realised that the tests do access this attribute although we declared it protected. This is one thing that we will realise anyway during the migration of the test to TypeScript.

 @0xfe do you want to leave `render_options` public? Is this part of the public API? Or would you prefer an access function?

rvilarl:
fixes #1070

Fix done in two steps:
- all `render_options` are now protected, required acessors implemented.
- `render_options` renamed to `renderOptions` to make sure that incorrect access on applications is detected.

I would still recommend:
1) inline the interfaces associated to render options that are only used once in the code: `BendRenderOptions` & `NoteRenderOptions`
2) make the helper functions member functions and inline the interface `RenderOptions` in `StaveLine`.

Notable:
1) Two classes accept the render options in the constructor: `Curve`, `MultiMeasureRest`. The others do not.
2) `StaveHairpin` has one member to update all the options `setRenderOptions`. The others do not.
3) The helper static function `FormatByTicksAndDraw` from `StaveHairpin` is not used at all.
4) The helper static function `renderVibrato` from `Vibrato` is used by `Vibrato` but also `VibratoBracket`.


rvilarl:
Changes as discussed in #1070, ready to review and merge

PR #1185, opened 2021-10-15
Add Groups to Key Signatures, Time Signatures, Clef & Stave (Addresses #1172)

### LINKED ISSUE #1172: Add Group Tags around elements
Add group tags with class and pointer ids for most musical elements:

Clefs
Time Signature (within which should be groups for numerator and denominator)
Staff lines (all in one group)
Key Signature (within which, group for each accidental)
Others...

(Not sure if this is enough of a breaking change to need to go to 5.0)

rvilarl:
Fixes #1172 

@mscuthbert is this what you meant? I have done an example to group keysignature and to check if this is the right concept.

### DIFF
--- a/src/stave.ts
+++ b/src/stave.ts
@@ -728,6 +728,7 @@ export class Stave extends Element {
     const ctx = this.checkContext();
     this.setRendered();
 
+    ctx.openGroup('stave', this.getAttribute('id'));
     if (!this.formatted) this.format();
 
     const num_lines = this.options.num_lines;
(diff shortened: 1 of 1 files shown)


rvilarl:
@ronyeh @sschmidTU feedback from your side?

sschmidTU:
Looks good, but i'd edit the first post to not say `Fixes #1172` because it's not a complete fix, just part of it.
You could write `Part of #1172` or `Partly fix for #1172` (the syntax `Fixes #[issuenumber]` usually means it closes the issue).

Issue #555, opened 2017-06-22
Tremolo not positioning on note stem UP


gciluffo:
Was wondering if anyone has had issues with tremolos not positioning correctly on a note stem? Seems like setPosition() and setYShift() methods on Modifier class which Tremolo extends, is not moving the tremolo at all before it gets drawn.

Im doing it like so, where staveNote is of type vexflow.Flow.StaveNote

```
addTremoloToNote(staveNote) {
        let tremolo = new vexflow.Flow.Tremolo(1);
        staveNote.addArticulation(0, tremolo);
    }
```

Renders:

<img width="108" alt="screen shot 2017-06-22 at 12 21 12 pm" src="https://user-images.githubusercontent.com/14182616/27449345-529c9370-5745-11e7-9c1b-f65f6fec5474.png">

When inspecting the object in chrome after using setYShift() on the tremolo, I see that its y_shift value has been modified. But the draw method does not seem to account for it. Also I noticed that the tremolo positions correctly if the stems are facing downward on the stavenote (stem_direction: -1).

gristow:
@gciluffo, could you post a fiddle with more of your code? (You can start from the [VexFlow Sandbox](https://jsfiddle.net/gs4v6k6d/2/).) My guess is it has something to do with how you're constructing the beams but I can't see that in the code you've posted. (But I'm also realizing we don't seem to have any Tremolo tests in the test suite.)

gciluffo:
@gristow

Okay I made a fiddle <a href="https://jsfiddle.net/gciluffo/xbvrsmy0/3/">here</a>. Just added a couple notes, one with stem up and stem down with tremolos.

gristow:
Yep, that's a bug for sure. I'm trying to knock out a bunch of little bugs this afternoon, so I'll add this to the list.

jjyyxx:
the bug remains

rvilarl:
It seems that there is no test for `Tremolo`, right?

rvilarl:
The tests included in #1186 show that Bravura is fine. However there are still problems with the other fonts.

![image](https://user-images.githubusercontent.com/22865285/137597824-a8739d7d-a311-426e-9f49-2f6736abef95.png)

rvilarl:
> Probably you already found this, but there are some tremolo tests in percussion tests.

@AaronDavidNewman I did not notice that. I would live the merged explicit tremolo test, they are easier to find. Could you fix the alingment metrics?

[...]

PR #1186, opened 2021-10-16
tremolo tests added

rvilarl:
Test added to check if #555 was still an issue.

It is fine with Bravura but it is not right with the other fonts.
@AaronDavidNewman any idea?

image


rvilarl:
@0xfe please review this PR, I think it is ready to merge. I will update the issue in #555

0xfe:
Thanks for the tests, I think the issue fix is pretty straightforward too.

PR #1197, opened 2021-10-24
fix issues 555, 792.  Fix metrics with tremolo Gonville and Petaluma, and Petaluma grace notes

### LINKED ISSUE #555: Tremolo not positioning on note stem UP

### LINKED ISSUE #792: Petaluma font grace notes look goofy
Even in the unit tests, a grace note has a very large note head compared to the stem.  I played around a bit and it appears the note head for that font is not responding to the glyph scale.

AaronDavidNewman:
resolves #555, resolves #792 

Add/repair formatting for tremolo and grace notes for some fonts.  I am no designer and I don't have such a good eye for small visual details, but these look pretty good to me.   Gonville tremolo is hard to judge b/c it's so slanty.

before:
image
after:
image

before:
image
after:
image

### DIFF
--- a/src/fonts/gonville_metrics.ts
+++ b/src/fonts/gonville_metrics.ts
@@ -92,20 +92,20 @@ export const GonvilleMetrics = {
 
   tremolo: {
     default: {
-      point: 40,
+      point: 25,
       spacing: 4,
-      offsetYStemUp: -9,
-      offsetYStemDown: -21,
-      offsetXStemUp: 6,
-      offsetXStemDown: -2,
+      offsetYStemUp: -17,
+      offsetYStemDown: -2,
+      offsetXStemUp: 9,
+      offsetXStemDown: -0.5,
     },
     grace: {
-      point: 30,
+      point: 15,
       spacing: 4,
-      offsetYStemUp: -9,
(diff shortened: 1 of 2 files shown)


AaronDavidNewman:
This is ready for merge :)

0xfe:
Aaran -- Thanks for that pic. The random lengths (and positions) are a fantastic idea! (and possibly straightforward to implement too.)

PR #1198, opened 2021-10-24
fixes `getTieEndX` in `Stave` (StaveModifiers placed at the end of the Stave collide with StaveTies without end notes.)

### LINKED ISSUE #489: StaveModifiers placed at the end of the Stave collide with StaveTies without end notes.

rvilarl:
Fixes #489

**StaveTie › No End Note** 

before:
StaveTie No_End_Note Petaluma_Reference

after:
StaveTie No_End_Note Petaluma_Current

No other visual difference

### DIFF
--- a/src/stave.ts
+++ b/src/stave.ts
@@ -195,7 +195,7 @@ export class Stave extends Element {
   }
 
   getTieEndX(): number {
-    return this.x + this.width;
+    return this.end_x;
   }
 
   getX(): number {
--- a/tests/stavetie_tests.ts
+++ b/tests/stavetie_tests.ts
@@ -23,8 +23,10 @@ const StaveTieTests = {
     run('Simple StaveTie', simple);
     run('Chord StaveTie', chord);
     run('Stem Up StaveTie', stemUp);
-    run('No End Note', noEndNote);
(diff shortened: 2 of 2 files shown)


0xfe:
Thanks Rodrigo. Does this change anything if there's no stave end modifier? Does the tie still reach the end of the stave?

rvilarl:
It does not produce any other visual change. Is there already a test for what you refer to? I am not sure that I really understand the case.

0xfe:
Is there a test where the Tie begins on the last note and there's nothing on the right? (That's what the 'No End Note' test should have been, but it seems like it now ends with a stave modifier.) In that case, the tie should hit the right stave barline.

rvilarl:
@0xfe I have added two tests related to cases without note and without Clef both at start and at end.

I believe that the output is correct:

image

image

image

image

rvilarl:
@0xfe hopefully ready to merge

0xfe:
Looks great, thanks for adding those tests!

PR #1201, opened 2021-10-28
Rename vexFlowCanvasContext to context2D.

ronyeh:
Rename vexFlowCanvasContext to context2D and other smaller fixes. For example, I lowercased the first letter of `sanitizeCanvasDims()`. This should be safe, since it's only used internally by this class. (It's basically a private helper function.)

Following Rodrigo's advice, I'm extracting some of the smaller changes out of my [Font mega-PR](https://github.com/0xfe/vexflow/pull/1163).

Apologies if the diff looks big. I moved the font accessors to the bottom of the file so that it is closer to the version in the Font PR.

Ran the tests. No visual diffs.

### DIFF
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -1,5 +1,6 @@
 // [VexFlow](http://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 // MIT License
+
 import { GroupAttributes, RenderContext, TextMeasure } from './rendercontext';
 import { warn } from './util';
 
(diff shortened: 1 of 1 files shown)


0xfe:
This LGTM. I found `vexFlowCanvasContext` a real eyesore. :-)

sschmidTU:
ronyeh, thanks for the explanation! I was also confused why it was named `context2D` and not `canvasContext`, but your explanation above cleared it up. Maybe the `context2D` declaration could use a comment that it's not Vexflow's CanvasContext (but part of a browser API). But i'll remember from this explanation.

PR #1202, opened 2021-10-28
Font Tools

ronyeh:
I did run the tests, which resulted in no visual diffs. This PR doesn't modify core `src/` or `tests/` files.

## What this PR contributes:

Fix the `.gitattributes` file so that git no longer corrupts our font files.

Add `tools/fonts/@/` which is a separate NPM package called `vexflow-fonts`.
This is used for hosting web font files that can be used with VexFlow 4.0.
The font files are served via CDNs:
- https://unpkg.com/browse/vexflow-fonts@1.0.3/
- https://cdn.jsdelivr.net/npm/vexflow-fonts@1.0.3/

It is a separate package so that we can update/fix font files without having
to publish a new `vexflow` version to NPM.

Improved font generation scripts. Now they use the same .prettierrc.js
configuration to produce JSON files that our eslint/prettier are happy with.
The old font files (gonville_all.ts, gonville.ts, vexflow_font.ts, etc)
have been updated to match the new format. Small fixes were applied, including
renaming the key `familyName` to `fontFamily` so that our new scripts can
process these files again.

Improve old glyph viewer to support more fonts.

Add a new glyph inspector that loads OTF files directly.

Moved the legacy/ folder out of src/ and into tools/fonts/. It's not part of our
main VexFlow source, and things like transform.html was actually a tool for
combining multiple fonts. Use the anchor download attribute for downloading
files. Removed jQuery from these old pages. :-)

_FYI: in the Font mega PR branch, jQuery no longer exists._


ronyeh:
> @ronyeh would it make sense to use ts-node and port the .js scripts to .ts?
> 
> https://github.com/TypeStrong/ts-node

Earlier this year I used ts-node for some of my node scripts and I ran into an issue where it worked fine in regular node (after translating to JS), but it was failing in ts-node. I abandoned using ts-node for my own project.

I do **love** TypeScript, but I'm not sure we should require folks to install ts-node just to run some of our font generation scripts. :-)

tools/fonts/README.md
```diff
@@ -0,0 +1,63 @@
+## Font Generation Scripts
```
0xfe, line 1: This is going to diverge very quickly from the Wiki. I don't have a strong opinion for where this goes, but for consistency, the wiki would be it.

Whatever we decide, can we have one redirect to the other?

tools/fonts/@/robotoslab/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
+
```
0xfe, line 1: Thanks for adding these license files.

0xfe:
Thanks Ron! Lots going on here (and lots of good stuff here) -- I'm trying to understand the extent of the changes.

Few questions:

1) We're taking a dependency on the `vexflow-fonts.` I can see the `package.json`, however who owns the existing NPM?
2) What happens to users' existing workflows? (i.e., those that don't use `vexflow-fonts`)
3) How are our developer workflows affected? 
  + Does it change how we add new fonts, change them, etc.?
  + What's the delta from the wiki docs.
4) How is our overall build impacted?

ronyeh:
This one needs some explanation. :-)

To clarify, this PR has nothing to do with our music engraving fonts. Bravura, Gonville, Petaluma, and Custom are still loaded the same way, via precompiled `xxx_glyphs.ts` files.

# ...but why web fonts?

In the current master revision, the ChordSymbol test case is buggy when using CanvasContext. It uses `@font-face` CSS to load PetalumaScript (via relative URL `../releases/fonts/petalumascript-webfont.woff`) and Roboto Slab (via Google Fonts). See:

https://github.com/0xfe/vexflow/blob/7f5004a673248e15cd670f55d92948f810f66af0/tests/flow.css#L4-L12

https://github.com/0xfe/vexflow/blob/7f5004a673248e15cd670f55d92948f810f66af0/tests/flow.html#L6

This approach works okay in SVG, since once the fonts are asynchronously loaded, the SVG text elements will pick up that font, assuming the `font-family` attribute was assigned properly.

This approach does NOT work in Canvas. Instead, you need to use the FontFace API to asynchronously load web fonts before rendering.

https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/font#loading_fonts_with_the_css_font_loading_api

The current master revision hardcodes 'Roboto Slab' and 'PetalumaScript' [with fallbacks](https://github.com/0xfe/vexflow/blob/7f5004a673248e15cd670f55d92948f810f66af0/src/chordsymbol.ts#L377-L380), but the developer is responsible for ensuring those fonts are loaded somehow.

It would be nice to have a simple API to ensure that these two fonts are loaded, otherwise ChordSymbol elements (and any other elements that depend on web fonts) will render improperly in Canvas. The FontFace API is nice, and is equivalent to the @font-face CSS approach (the older approach with more browser support).

Google Fonts are hosted on a CDN already, but the SMuFL fonts are not. To use the FontFace API, we'll need to bundle the PetalumaScript woff file with our distribution, or make it available via CDN.

This PR proposes serving the web font files from a CDN, so that developers using VexFlow can conveniently load those web fonts if they desire. Doing so is completely optional.

This PR allows the developer to choose either approach (`@font-face` OR `FontFace`) for loading web fonts.

# How to use it?

See these 50 lines of code in font.ts:

https://github.com/ronyeh/vexflow/blob/8fe00b4b1e239b6cc714a15209413ee4d261baf6/src/font.ts#L301-L349

You can **optionally** load the included web fonts with a single line:

```
await Vex.Flow.Font.loadWebFonts();
```

In flow.html, we do it like this:

https://github.com/ronyeh/vexflow/blob/8fe00b4b1e239b6cc714a15209413ee4d261baf6/tests/flow.html#L87-L93

Furthermore, developers can customize the font CDN and set of fonts they want to load, by modifying `Font.WEB_FONT_HOST` and `Font.WEB_FONT_FILES`.

# Why another npm package?

-   The SMuFL and Gonville font files are not available from any CDN. Uploading the fonts to npm allows them to be served via unpkg or jsDelivr. This allows developers to load them via the FontFace API if so they desire.
-   As a separate npm package, the fonts can be added to or updated separately from the VexFlow release cycle.
-   You own the npm package. Well, I'm also a maintainer and have been doing the pushes so far. See: https://www.npmjs.com/package/vexflow-fonts

# Build Impact

-   There is no impact to the way we build or release VexFlow.
-   Using web fonts is optional, via `Font.loadWebFonts()`.
-   This is not a breaking change, since it only adds an optional API.
-   Version 3.0.9 does not have `ChordSymbol` or any support for web fonts. So developers upgrading from 3.0.9 shouldn't have any issues. It also shouldn't break @aarondavidnewman 's code, since the API is optional and he can continue to use @font-face in smoosic.

ronyeh:
setFont_2_ChordSymbol

This shows the bug fix for the Canvas test case. It now correctly picks up Roboto Slab.

0xfe:
Okay, yeah -- thanks for that explanation Ron. (I really like your very thorough explanations, btw.) Reviewing.

0xfe:
Looks great. Merging. Can you please add/update docs with instructions on this?

(BTW, there's a lot of different changes bundled here, which makes the reviewing harder.)

ronyeh:
> Looks great. Merging. Can you please add/update docs with instructions on this?

Will do. 

> (BTW, there's a lot of different changes bundled here, which makes the reviewing harder.)

I apologize! At first, all I wanted to do was to consolidate the setFont() API across classes. Then you looped in Aaron as a reviewer and we had a looong discussion about what his TextFont class did. This lead to the breakup of TextFont into Font/TextFormatter, which lead to music fonts, font tools, dynamic font loading, web fonts, etc etc. 

What I really should have done was stop after setFont() and have that merged as its own PR before delving into Aaron's TextFont and ChordSymbol. Small PRs are always easier to merge. 

That said, I'd say I am now a world expert on fonts haha. 😂

0xfe:
No need to apologize! I've been there :-) Thanks for the awesome work.

(And looking forward to your TED talk on fonts.)

PR #1208, opened 2021-10-30
add some missing exports

AaronDavidNewman:
Add some symbols to flow.ts that were previously available in Tables.ts


rvilarl (approved):
LGTM

PR #1210, opened 2021-11-03
Fonts: CONSTANTS & TEXT_FONT

ronyeh:
This PR adds some useful Font related constants and enums (e.g., FontWeight.BOLD). It also adds a `TEXT_FONT` static variable to the Element class (and any subclasses that want to render text).

Currently in the master branch, each Element subclass handles fonts independently, in their own way. Here are excerpts from three random classes:

```typescript
this.font = '10pt Arial';

this.font = {
  family: 'sans-serif',
  size: 8,
  weight: '',
};


this.font = {
  family: 'serif',
  size: 10,
  weight: 'bold italic',
};
```

In the main Fonts PR branch, the text font member variable has been moved up to the Element class, and it is handled consistently across the subclasses. The TEXT_FONT static variable defines the default font for each subclass, instead of inlining things like `'10pt Arial'`. That way, developers who npm install vexflow can programmatically change the default of each element type, if desired.

These additions are used in the main Fonts branch, but I'm submitting them as a separate PR to help with code review. This PR doesn't add any new logic. It just adds the constants, types, and static variables that will be used in the next PR.


src/chordsymbol.ts
```diff
@@ -680,6 +707,9 @@ export class ChordSymbol extends Modifier {
     ctx.openGroup(classString, this.getAttribute('id'));
 
     const start = note.getModifierStartXY(Modifier.Position.ABOVE, this.index);
+    // TEMPORARILY DISABLE. Will be fixed in RONYEH's FONTS PR.
```
ronyeh, line 710: I had to add some of these eslint-disable / ts-ignore to this PR. They will be removed once the Fonts branch is merged.

src/pedalmarking.ts
```diff
@@ -33,12 +33,19 @@ function drawPedalGlyph(name: string, context: RenderContext, x: number, y: numb
...
 export class PedalMarking extends Element {
   /** To enable logging for this class. Set `Vex.Flow.PedalMarking.DEBUG` to `true`. */
-  static DEBUG: boolean;
+  static DEBUG: boolean = false;
 
   static get CATEGORY(): string {
     return 'PedalMarking';
   }
 
+  static TEXT_FONT: Required<FontInfo> = {
```
ronyeh, line 42: The main Element class gets a default TEXT_FONT which specifies the font family, size, weight, and style. Each subclass may optionally override this default.

src/stavetext.ts
```diff
@@ -1,25 +1,33 @@
...
+
   protected options: {
     shift_x: number;
     shift_y: number;
     justification: number;
   };
+
+  protected text: string;
+  protected shift_x?: number;
+  protected shift_y?: number;
```
rvilarl, line 30: If those variables are internal and not accessed from outside (seems to be the case) Consider using cammelCase

rvilarl (approved):

src/stavetext.ts
ronyeh: The constructor takes an options param of the type `options: { shift_x?: number; shift_y?: number; justification?: number }`. Weirdly enough, this class has a shift_x and also a options.shift_x stored internally. Are they redundant? Who knows? Let's revisit this after releasing 4.0. I think folks are probably anxious to get to beta soon. :-)

rvilarl:
@0xfe IMO this is ready to merge.

0xfe:
Looks like there's new conflicts here.

PR #1211, opened 2021-11-03
migration/casting (casting for map() added)

rvilarl:
Minor issue coming from the migration of PianoPlay app. 
Map() returns unknown[]

### DIFF
--- a/tests/gracenote_tests.ts
+++ b/tests/gracenote_tests.ts
@@ -36,11 +36,11 @@ function basic(options: TestOptions): void {
     { keys: ['f/4'], duration: '32' },
     { keys: ['g/4'], duration: '32' },
     { keys: ['a/4'], duration: '32' },
-  ].map(f.GraceNote.bind(f));
+  ].map(f.GraceNote.bind(f)) as GraceNote[];
 
-  const gracenotes1 = [{ keys: ['b/4'], duration: '8', slash: false }].map(f.GraceNote.bind(f));
+  const gracenotes1 = [{ keys: ['b/4'], duration: '8', slash: false }].map(f.GraceNote.bind(f)) as GraceNote[];
 
-  const gracenotes2 = [{ keys: ['b/4'], duration: '8', slash: true }].map(f.GraceNote.bind(f));
+  const gracenotes2 = [{ keys: ['b/4'], duration: '8', slash: true }].map(f.GraceNote.bind(f)) as GraceNote[];
 
   const gracenotes3 = [
     { keys: ['e/4'], duration: '8' },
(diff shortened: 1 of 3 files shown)


tests/gracenote_tests.ts
```diff
@@ -36,25 +36,25 @@ function basic(options: TestOptions): void {
     { keys: ['f/4'], duration: '32' },
     { keys: ['g/4'], duration: '32' },
     { keys: ['a/4'], duration: '32' },
-  ].map(f.GraceNote.bind(f));
+  ].map(f.GraceNote.bind(f)) as GraceNote[];
 
-  const gracenotes1 = [{ keys: ['b/4'], duration: '8', slash: false }].map(f.GraceNote.bind(f));
+  const gracenotes1 = [{ keys: ['b/4'], duration: '8', slash: false }].map(f.GraceNote.bind(f)) as GraceNote[];
```
ronyeh, line 41: Why do we need all these casts? How come these errors don't show up on master or my branch when we are compiling VexFlow?

For example, here we have:
```
[...].map(f.GraceNote.bind(f)) as GraceNote[];
```
The `Factory.GraceNote(noteStruct: GraceNoteStruct): GraceNote` returns a GraceNote, so if we map an array of structs, my TS compiler correctly decides that the return value is `const gracenotes: GraceNote[]`. This can be verified when I mouse over each of those parts of the code (in VS Code, which I assume you are using).

I'm not a fan of all these `as XXX[]` casts if we can avoid them.

ronyeh:
Is this a matter of our *.d.ts files not providing your project enough information on the method signatures of Factory methods? Obviously when we are compiling VexFlow, all of these lines of code pass eslint. There's something curious here. I think your TS compiler is missing some typing information. Can we track it down somehow? At least on the macOS version of VS code, hovering my mouse cursor over a variable or a method provides a tooltip showing the type that the TS compiler thinks it is. 

We need to figure out what typing information is missing, and hopefully provide that through the d.ts files.

ronyeh:
I'm also a bit confused because these are xxx_tests.ts files. When you are compiling PianoPlay, are you also compiling the tests? Is this via a npm link to my Fonts branch?

Tomorrow I'll put up a vexflow-4.0.0-alpha-1.tgz and you can try npm installing it to see if these errors persist.

Do you also have a modified version of OSMD that you are using? If you can provide a tiny hello world example that integrates with OSMD + VexFlow 4 in the same way as PianoPlay, that would be helpful. I can try integrating on my side too to iron out any last bugs. (You don't need to provide your project source.... just the OSMD version that you're using + a barebones integration.)

ronyeh:
Anyways, heading to bed soon... but basically I believe we shouldn't have to change our test code due to an issue with a project npm installing VexFlow. So we need to first find out why there are errors, and how to avoid those errors without adding (possibly) unnecessary casts. Perhaps

tests/gracenote_tests.ts
sschmidTU: In case this doesn't work for @rvilarl , maybe it's also because of the older typescript version (that he was at least using some days ago)?

ronyeh:
FYI while integrating with my own skeleton TS project, I got a suspiciously related error:

```
Cannot find name 'Map'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2015' or later
```

I'll debug the tsconfig.json to see if I can make this go away.

rvilarl:
@ronyeh you are right this is not required.

PR #1212, opened 2021-11-03
Run prettier / eslint on the font data files, to improve consistency and editability of the files.

ronyeh:
_This was extracted from my Fonts PR._

Most of the changes are just changes in the style of quotes, and adding commas. This way, we can edit these files in the future without introducing a huge diff.

### Also in this PR
Roboto Slab and PetalumaScript font data files are temporarily duplicated. The files named `xxx_textmetrics.ts` will be dropped in an upcoming PR, while the almost identical files named `xxx_glyphs.ts` will be kept. We are effectively renaming the files in two steps, to avoid modifying any other source files in this PR. The files are named `xxx_glyphs.ts` to mirror the naming scheme of `bravura_glyphs.ts`, `petaluma_glyphs.ts`, etc, since their file structure is the same (other than the fact that they are missing the actual glyph outlines).


0xfe:
LGTM

PR #1213, opened 2021-11-04
migration: eslintrc config updated

rvilarl:
.eslintrc configuration changed to the one in #1163.

All the changes in src/* and tests/* were done automatically by eslint (there is no need to review them, no visual changes)

This should reduce the number of modified files in #1163 

@ronyeh I decided to do it myself becuase I had the feeling that I was not explaining myself clearly enough.


ronyeh:
Thanks Rodrigo. It seems like a pretty straightforward auto-fix of imports.

I'll make sure to rebase off of this. :-)

PR #1214, opened 2021-11-04
4.0.0

ronyeh:
I'm happy to share that my branch is ready for integration with your projects. Rodrigo has tested it out a few times this week, and I've ironed out most of the issues. I had to do a lot of wrangling of tsconfig.json, package.json, webpack configs, gruntfile, etc.... Fun stuff.

You can see the branch here: https://github.com/ronyeh/vexflow/tree/4.0.0-alpha

To npm install this branch, you have two options:

## 1 – Easy & Fast
```
cd /path/to/your/project
npm uninstall vexflow
npm install https://vexflow-demo.surge.sh/vexflow-4.0.0-alpha-7.tgz
```

OR

## 2 – Build from Source
```
... make a temp directory for storing the 4.0.0 alpha build ...
git clone --depth 1 https://github.com/ronyeh/vexflow.git -b 4.0.0-alpha vex4
cd vex4
npm install
grunt
npm link
cd /path/to/your/project
npm uninstall vexflow
npm link vexflow
```

After you do either option 1 or option 2, try to build your project and let me know the list of issues you encounter!

I'll check this thread and help address any questions or feedback.

I made two small example projects that I have successfully integrated with VexFlow 4. Feel free to clone these projects:
* https://github.com/ronyeh/vexflow-example-typescript
* https://github.com/ronyeh/vexflow-example-javascript

Once we've ironed out the big integration roadblocks, we can ask Mohit to bump the revision number and ship 4.0.0-beta (via dist-tag) to npm.

FYI: this branch is built on top of the enhancements in my Fonts PR branch:
* https://github.com/0xfe/vexflow/pull/1210
* https://github.com/0xfe/vexflow/pull/1216


src/note.ts
```diff
@@ -527,6 +527,11 @@ export abstract class Note extends Tickable {
    * @returns this
    */
   addModifier(modifier: Modifier, index: number = 0): this {
+    // Backwards compatibility with 3.0.9.
+    if (typeof index === 'string') {
+      index = parseInt(index);
```
rvilarl, line 532: Consider logging a deprecated message. Parameter should be `index: number | string = 0`

src/util.ts
```diff
@@ -9,6 +9,10 @@ export class RuntimeError extends Error {
   constructor(code: string, message: string = '') {
     super('[RuntimeError] ' + code + ': ' + message);
     this.code = code;
+    // Keep a debugger; statement here.
+    // This should be removed from the production build, but will stay in the development builds.
+    // eslint-disable-next-line
+    debugger;
```
rvilarl, line 15: Should we have two implementation dpending on mode (production or development)?

rvilarl (changes requested):

Gruntfile.js
```diff
@@ -7,158 +7,194 @@ const InjectPlugin = require('webpack-inject-plugin').default;
 const TerserPlugin = require('terser-webpack-plugin');
 const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
 
-module.exports = (grunt) => {
```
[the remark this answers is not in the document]
ronyeh: I believe I removed this already.  It's not in the current version: https://github.com/ronyeh/vexflow/blob/4.0.0-alpha/Gruntfile.js

src/note.ts
ronyeh: Well we shouldn't accept strings here. The API should officially be index: number = 0.

However, legacy code might actually pass in a string. That's what VexTab did. So I'm just going to parse it here. Sure I can log a warning.

ronyeh:
Temporarily setting this to "draft" while I work with Rodrigo and Aaron on integrating with their projects (e.g., making all the TypeScript types available as expected).

ronyeh:
The current version is ready for integration with your TS and JS projects.

The easiest way is to:
`npm install https://vexflow-demo.surge.sh/vexflow-4.0.0-alpha-6.tgz`

Today I migrated [Music21j](https://github.com/cuthbertLab/music21j) to VexFlow 4. It went pretty smoothly! Mostly I was able to remove all the Vex.Flow. prefixes and import the TS classes directly.

Next steps... after Simon, Michael, Rodrigo, and Aaron have their projects compiled (and mostly working) with this branch, I'm happy to declare that it's "beta ready".

The bulk of the work will be getting all the good stuff reviewed and merged into 0xfe's repo.

I've learned a LOT through supporting these integrations. A ton of work went into coordinating the `tsconfig.json`, `package.json`, and webpack settings in `Gruntfile.js` to make all the JS bundles, TS classes, *.d.ts typings, etc all available for client code to import | require | &lt;script&gt; | &lt;script type="module"&gt;, hahah..... 👎 

Anyways, here are some fun images from the Music21j migration. I'll post the patch as a PR on the Music21j repo.

A
B
C
D

0xfe:
Thanks for all your work here Ron, this really looks great. I've been drowning in some work stuff the last week, and I'll start reviewing as soon as I get some breathing room (targeting this weekend.)

0xfe:
Hey Ron -- at this point, I'm ready to merge, publish as 4.0-beta, and deal with incrementally fixing any issues. Does that work for you?

@rvilarl @mscuthbert @AaronDavidNewman @tommadams for thoughts

rvilarl:
It does work for me!

ronyeh:
Just tested. No visual differences relative to master. There are some new images due to the font tests.

I left some "RONYEH" comments at each line of code that will introduce a visual difference, as discussed in my Font PR. These are all related to bug fixes or font-related improvements that I recommend. I'll submit a small PR shortly after this branch is merged, which will introduce the changes (and visual diffs).

```
Warning: Number of (matching) Current images (1119) is not the same as Reference images (1113). Continuing anyways.
Running 1113 tests with threshold 0.01 (nproc=4)...
Progress : [########################################] 100%
Results stored in ./build/images/diff/results.txt
All images with a difference over threshold, 0.01, are
available in ./build/images/diff, sorted by perceptual hash.


You have        6 warning(s):
Warning: Font.Set_Music_Font_to_Petaluma.Bravura.png missing in ./build/images/reference.
Warning: Font.Set_Music_Font_to_Petaluma.Gonville.png missing in ./build/images/reference.
Warning: Font.Set_Music_Font_to_Petaluma.Petaluma.png missing in ./build/images/reference.
Warning: Font.Set_Text_Font_to_Georgia.Bravura.png missing in ./build/images/reference.
Warning: Font.Set_Text_Font_to_Georgia.Gonville.png missing in ./build/images/reference.
Warning: Font.Set_Text_Font_to_Georgia.Petaluma.png missing in ./build/images/reference.
Success - All diffs under threshold!
```

rvilarl:
@0xfe I have tested with OSMD, let us make the first beta!

ronyeh:
**TODO** for anyone who's free this weekend: :-)
Please help sanity check the 4.0.0-alpha-7 build by doing:

```
npm install https://vexflow-demo.surge.sh/vexflow-4.0.0-alpha-7.tgz
```

You can use your own project, or you can play around with my two sample projects:

TS sample project: https://github.com/ronyeh/vexflow-example-typescript

JS sample project: https://github.com/ronyeh/vexflow-example-javascript


**TODO for 0xfe:**
Bump the revision number to 4.0.0 and publish to npm with the **beta** dist-tag.


I'll watch for any issues and can provide solutions for any issues.

We've had success migrating VexTab, OSMD, and Music21J to VexFlow 4, so I hope other migrations are also straightforward. Smoosic might have the most changes, since the previous TextFont code is now either in Font, TextFormatter, or the loadTextFonts() function exported by `src/fonts/textfonts.ts`. I can work with @AaronDavidNewman to make sure there aren't any issues with his upgrade to VexFlow 4.

demos/fonts/all.html
```diff
@@ -0,0 +1,99 @@
...
+        The build/*.js files are compatible with <script> tags.
+        The build/esm/*.js files are compatible with ESM import statements.
+    -->
+    <script src="../../build/vexflow-debug.js"></script>
+    <script>
+      console.log(Vex);
+      console.log('VexFlow BUILD ' + Vex.Flow.BUILD);
+
+      {
+        // Calling setMusicFont() here is optional, since Bravura is the default font.
```
ronyeh, line 36: I incorporated suggestions from @0xfe regarding additional comments in the demo files. I also removed the big horizontal separator comments. :-)

src/element.ts
```diff
@@ -43,16 +42,30 @@ export abstract class Element {
...
   private context?: RenderContext;
   protected rendered: boolean;
   protected style?: ElementStyle;
   private attrs: ElementAttributes;
   protected boundingBox?: BoundingBox;
   protected registry?: Registry;
 
-  // fontStack and musicFont are both initialized by the constructor via this.setFontStack(...).
-  protected fontStack!: Font[];
-  protected musicFont!: Font;
```
ronyeh, line 55: Removed the per-element music font. It's a nice idea but it worked in unexpected ways in all of my tests. Having one global font stack is simpler for now. Maybe we can revisit this someday if people desire this feature.

src/element.ts
```diff
@@ -43,16 +42,30 @@ export abstract class Element {
...
+  /**
+   * Some elements include text.
+   * The `textFont` property contains information required to style the text (i.e., font family, size, weight, and style).
+   * It is undefined by default, and can be set using `setFont(...)` or `resetFont()`.
+   */
+  protected textFont?: Required<FontInfo>;
```
ronyeh, line 68: Previously, each element implemented their own font instance variable for handling text (e.g., 10pt Arial). However, different elements handled fonts in different ways. I moved it up to Element. It's optional (because not all elements utilize text and text fonts). Elements can override the default text font specified in Element.TEXT_FONT.

src/element.ts
```diff
@@ -225,4 +225,168 @@ export abstract class Element {
...
+   * @param font is 1) a `FontInfo` object or
+   *                2) a string formatted as CSS font shorthand (e.g., 'bold 10pt Arial') or
+   *                3) a string representing the font family (at least one of `size`, `weight`, or `style` must also be provided).
+   * @param size a string specifying the font size and unit (e.g., '16pt'), or a number (the unit is assumed to be 'pt').
+   * @param weight is a string (e.g., 'bold', 'normal') or a number (100, 200, ... 900).
+   * @param style is a string (e.g., 'italic', 'normal').
+   * If no arguments are provided, then the font is set to the default font.
+   * Each Element subclass may specify its own default by overriding the static `TEXT_FONT` property.
+   */
+  setFont(font?: string | FontInfo, size?: string | number, weight?: string | number, style?: string): this {
```
ronyeh, line 255: This is the setFont() API, which consolidates the different ways different elements were setting their own text fonts. I tried to make it as consistent with CSS standards as possible. Even the RenderContexts (CanvasContext & SVGContext) share this `setFont(...)` method signature now.

src/entry/vexflow-core-with-bravura.ts
```diff
@@ -0,0 +1,21 @@
...
+// vexflow-core-with-bravura.ts is the entry point for vexflow-core-with-bravura.js.
+// This version includes only Bravura.
+// It also overrides the `Flow.setMusicFont(...)` function to be async,
+// loading other music fonts (e.g., Petaluma) on the fly.
+
+import { Vex } from '../';
+import { loadMusicFonts } from '../fonts/bundleBravura';
+import { loadTextFonts } from '../fonts/textfonts';
+
+loadMusicFonts();
```
ronyeh, line 15: Simon & Rodrigo were interested in loading fonts dynamically. The way we implemented it over the summer (with magic path aliases) didn't work well when published as an NPM module.

The different bundles we create now each have a separate entry file (see vexflow/src/entry/*). For example, this file (vexflow-core-with-bravura.ts) will eventually be built into vexflow/build/vexflow-core-with-bravura.js. That file bundles only one font, and can dynamically load the other two fonts if requestedvia Flow.setMusicFont(...).

src/flow.ts
```diff
@@ -77,167 +77,214 @@ import { Vibrato } from './vibrato';
...
+   * ```
+   * @returns CASE 1: an array of Font objects corresponding to the provided `fontNames`.
+   * @returns CASE 2: Promise<Font[]> that resolves to the same array of Font objects as above.
+   */
+  static setMusicFont(...fontNames: string[]): Font[] {
+    // Convert the array of font names into an array of Font objects.
+    const fonts = fontNames.map((fontName) => Font.load(fontName));
+    Tables.MUSIC_FONT_STACK = fonts;
+    Glyph.CURRENT_CACHE_KEY = fontNames.join(',');
+    return fonts;
```
ronyeh, line 206: Music engraving fonts are now specified by string (e.g., 'Bravura', 'Petaluma'). Previously we were creating Font objects, and sometimes we'd create multiple Font.Bravura() objects.

This new approach is also more compatible with dynamic font loading. Once a font is dynamically loaded once, it is added to the cache, and future calls to Font.load(fontName) will return the same Font object (See: src/fonts/async.ts).

src/fonts/textfonts.ts
```diff
@@ -0,0 +1,56 @@
...
+      maxSizeGlyph,
+      monospaced: false,
+      bold: false,
+      italic: false,
+      superscriptOffset: 0.66,
+      subscriptOffset: 0.66,
+      serifs: true,
+      description: 'Text font to pair with the Bravura / Gonville music fonts.',
+    });
+  }
```
ronyeh, line 32: @AaronDavidNewman will probably recognize this code. It was extracted from the previous TextFont class, and now parallels the loadMusicFonts() function that we have.

You'll need to look for TextFormatter (as we discussed a while back), as the TextFont class is no longer around.

src/frethandfinger.ts
```diff
@@ -4,17 +4,24 @@
...
-import { FontInfo } from './types/common';
 import { RuntimeError } from './util';
 
 export class FretHandFinger extends Modifier {
   static get CATEGORY(): string {
     return 'FretHandFinger';
   }
 
+  static TEXT_FONT: Required<FontInfo> = {
+    family: 'sans-serif' /* RONYEH: Font.SANS_SERIF */,
```
ronyeh, line 19: 3.0.9 hard coded font families in each individual element. Sometimes it would be 'sans-serif', other times it would be 'Arial'. Sometimes it was capitalized as 'Sans-serif' :-). I'd prefer to use a constant here. But for now I left it as is, because using the new constant will introduce a visual diff.

I'll submit a different PR soon after beta to resolve all of these TODOs.

src/pedalmarking.ts
```diff
@@ -116,11 +123,7 @@ export class PedalMarking extends Element {
     this.custom_depress_text = '';
     this.custom_release_text = '';
 
-    this.font = {
-      family: 'Times New Roman',
-      size: 12,
-      weight: 'italic bold',
```
ronyeh, line 122: We previously allowed weight to include both "italic" and "bold".... The new Font API mirrors the CSS standard.

src/vex.ts
```diff
@@ -37,46 +35,52 @@ export const Vex = {
...
-  StackTrace: (): string | undefined => {
+  static stackTrace(): string | undefined {
     const err = new Error();
     return err.stack;
-  },
-};
+  }
+
+  // Backwards compatability with 3.0.9.
+  static RERR: RuntimeError;
```
ronyeh, line 82: There were a few smaller things like this, where I reintroduced an API that had disappeared sometime between 3.0.9 and today. For example, Vex.RERR was used a lot in VexTab.

tsconfig.json
```diff
@@ -6,19 +6,13 @@
...
     "esModuleInterop": true /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */,
     "skipLibCheck": true /* Skip type checking of declaration files. */,
     "forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */,
     "moduleResolution": "node",
-    "baseUrl": "./src",
-    "paths": {
-      "@bravura": ["fonts/loadStatic"],
-      "@gonville": ["fonts/loadStatic"],
-      "@petaluma": ["fonts/loadStatic"],
-      "@custom": ["fonts/loadStatic"]
```
ronyeh, line 20: These were the magic path aliases that we were using this summer to implement dynamic font loading. However, Rodrigo and I couldn't get it working during our integration phase. The multiple entry files solution we have now is more explicit (there's a little less magic).

ronyeh:
Oh yes I need to mention that the package.json now specifies that the build files are housed in the build/ directory, similar to how many projects keep their builds in dist/ or build/. 

The files that will be uploaded to npm are listed under:

```
"files": [
    "build",
    "src",
    "tests"
  ],
```

If you prefer to copy the build product over to releases/, we will need to figure out a way to organize old releases, possibly in versioned subdirectories or something. And those older versions probably shouldn't be shipped to npm. But overall I think it's cleaner to just keep the build files in build/ going forward.

sschmidTU:
> The files that will be uploaded to npm are listed under:
> 
> ```
> "files": [
>     "build",
>     "src",
>     "tests"
>   ],
> ```

Why should tests be included in the npm release? That seems unnecessary. Npm modules are for distributing libraries, specifically the part of them that makes them importable/usable. If i want to verify Vexflow tests, i'm not looking in one of my 400 modules in the `node_modules` folder, i'm just checking out the repo. Maybe i'm missing something here.

rvilarl:
Thanks Simon, it is just our first npm. I think that it makes sense not to add tests.

ronyeh:
> > The files that will be uploaded to npm are listed under:
> > ```
> > "files": [
> >     "build",
> >     "src",
> >     "tests"
> >   ],
> > ```
> 
> Why should tests be included in the npm release? That seems unnecessary. Npm modules are for distributing libraries, specifically the part of them that makes them importable/usable. If i want to verify Vexflow tests, i'm not looking in one of my 400 modules in the `node_modules` folder, i'm just checking out the repo ;) Maybe i'm missing something here.

Thanks? This is why user/developer feedback is valuable for us. This is my first time publishing a npm package with > 1 users 😁. Removing tests is an easy fix. 

You can inspect the contents of the alpha 7 npm package and let me know if you have any other suggestions on the layout of the files, or if there are any done we might be missing. (maybe I should include the MIT license file?)

sschmidTU:
@ronyeh Good idea, but looks like the `LICENSE` file is automatically included anyways:
> Certain files are always included, regardless of settings:
>
> package.json
> README
> LICENSE / LICENCE
> The file in the "main" field
> README & LICENSE can have any case and extension.

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#license

(Anecdotally, I was searching for `LICENSE` files in the vexflow node_modules folder, and got 853 results ;) )

Gruntfile.js
```diff
@@ -151,77 +160,102 @@ module.exports = (grunt) => {
...
-            // Insert the export line BEFORE the source map comment.
-            const srcMapLinePrefix = '//# sourceMappingURL';
-            const exportVex = 'export default Vex;\n';
-            return content.replace(srcMapLinePrefix, exportVex + srcMapLinePrefix);
+          process: (content) => {
+            // We take a JS file that sets globalThis.Vex, and turn it into a ES module.
+            // To do this, we extract all the classes from Vex.Flow, and re-export them.
+            // We also export Vex both as a named export and as a default export.
+            // THIS IS A HACK, so we should figure out how to use webpack to export both ESM and CJS.
+            const exports = `
```
0xfe, line 230: Not thrilled about this -- can we load this in dynamically from a separate file in the root directory? i.e.,

- put the contents in a file and load it in this `Gruntfile`.
- update dev instructions to call this out as a step
- add a bug to remove this hack (or find a suitable mitigation)

0xfe: Alternatively, can we drop support for CJS?

0xfe:
Hi folks! `4.0.0 beta` pushed to NPM!! 🎉 🎂 🎈 

I'm working through some tail issues and will file some bugs, and start updating docs. I'd also like to publish a blog post with the major changes, some screenshots, and attributions!

Thank you all for your awesome work and all the heavy lifting! Really thrilled to see this land!

rvilarl:
Hi Mohit, are you going to remove the releases directory? We do not really need them now with `npm run reference`

0xfe:
@rvilarl Sounds like the right move for 4.0.0-beta2

0xfe:
Or maybe not, how else would we serve out of unpkg?

Gruntfile.js
ronyeh: I discussed this here: https://github.com/0xfe/vexflow/issues/1222

I have a PR coming for native ESM support that will remove this hack. Hopefully this will be in 4.0 beta 2.

Not sure when you'll want to drop support for CJS. It's the classic old reliable script tag that every web developer learns as a hello world. My proposal is to keep everything in vexflow/build/, and have a `build/esm` folder for the modern module files, and a `build/cjs/` for the pre-bundled files compatible with script tags.

ronyeh:
> Or maybe not, how else would we serve out of unpkg?

From now on, we will serve from unpkg / jsdelivr directly from the build/ folder.

Here are two examples:
```
https://unpkg.com/browse/tone@14.7.77/build/   =>  https://unpkg.com/tone@14.7.77/build/Tone.js
https://unpkg.com/browse/phaser@3.55.2/dist/   =>  https://unpkg.com/phaser@3.55.2/dist/phaser.min.js
```

I would also recommend including the version number directly in the docs, because developers usually copy and paste from the README or tutorials, and not having the version in the URL might subject them to breaking updates (i.e., when we move to VexFlow 5.0).

Using the versioned URLs in our docs is OK because we can do a global search/replace when the time comes to update our docs.

0xfe:
So we commit the `build/` folder instead of `releases/`?

Gruntfile.js
sschmidTU: We also have quite a few hobby developers still using plain JS/HTML with script tags in OSMD. It would be kinda nice to keep CJS support. I believe ronyeh is working on making CJS a different build target as well, so it shouldn't negatively affect the esm build too much.

0xfe: Okay, thanks yeah, I agree with that.

PR #1161, opened 2021-09-28
Allow a RenderContext to be passed directly to the Renderer constructor

tommadams:
fixes #827
fixes #724

This enables users to create their own `RenderContext` implementations.

The required a couple of changes:
 - Move the `CanvasContext` resize logic out of `Renderer` and into `CanvasContext`.
 - Overload the `Renderer` constructor to accept a `RenderContext`.

This PR also fixes a bug in the `Factory` where it would always create an `SVGContext` even if you passed it the element ID for a `HTMLCanvasElement`. Consequently, there are a few image diffs because the `Factory.Draw` test now actually renders something for the canvas test :)

Issue #1157, opened 2021-09-26
 this.element.appendChild is not a function error when using a Vexflow in react-pdf


tommy-banana:
Hi I am using Vexflow successfully in a react app and Vexflow is displaying successfully on the screen.
I am trying to find a way to allow users to download a PDF and each route seems to run into problems one way or another.

I am looking at react-pdf/renderer and they have a couple of methods that should work.

<PDFDownloadLink />

You can see more about this on  https://react-pdf.org/advanced#page-wrapping

But if I use this I get this error from inside the Vexflow code (below) for each Vexflow component:
NB the components are all displaying correctly on the Screen, it is only when trying to download that this error is thrown.

Just wondered if the error looks familiar to anyone or if anyone has any suggestions.

```
TypeError: this.element.appendChild is not a function
new SVGContext
node_modules/vexflow/src/svgcontext.js:41
  38 | this.svgNS = 'http://www.w3.org/2000/svg';
  39 | const svg = this.create('svg');
  40 | // Add it to the canvas:
> 41 | this.element.appendChild(svg);
     | ^  42 | 
  43 | // Point to it:
  44 | this.svg = svg;
View compiled
new Renderer
node_modules/vexflow/src/renderer.js:145
  142 | } else if (this.backend === Renderer.Backends.RAPHAEL) {
  143 |   this.ctx = new RaphaelContext(this.element);
  144 | } else if (this.backend === Renderer.Backends.SVG) {
> 145 |   this.ctx = new SVGContext(this.element);
      | ^  146 | } else {
  147 |   throw new Vex.RERR('InvalidBackend', `No support for backend: ${this.backend}`);
  148 | }
View compiled
(anonymous function)
src/interactives/pitch-composer-master/components/score/index.js:26
  23 | 
  24 | useEffect(() => {
  25 |   if (rendererRef.current == null) {
> 26 |     rendererRef.current = new Renderer(
     | ^  27 |       container.current,
  28 |       Renderer.Backends.SVG
  29 |     );
View compiled
▶ 17 stack frames were collapsed.
updateContainer
src/index.js:30
  27 | 
  28 | this.registerEmojiSource = ({ url, format = 'png' }) => {
  29 |   emojiSource = { url, format };
> 30 | };
     | ^  31 | 
  32 | this.registerHyphenationCallback = callback => {
  33 |   hyphenationCallback = callback;
View compiled
update
src/dom.js:68
  65 | }, [state.url]);
  66 | 
[...]

Issue #655, opened 2018-09-20
enhancement: also accept canvas context instead of HTMLElement.


ritz078:
This will enable usage of this library in web workers using the OffscreenCanvas API. 

More about this has been discussed [here](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/396).

The only limitation right now is that this library depends on DOM access. Once it is enabled to accept canvas context, we can move this to web worker and control the canvas from there. 

I can create a PR.

turnerhayes:
Have you managed to put together a PR for this?

sschmidTU:
Demand for this is also high at OSMD.
Someone should implement Canvas Context or Web Worker support for Vexflow some day, OSMD will follow :)
This would help mobile development with Vexflow/OSMD a lot, because right now we need a DOM element (<div>) to render, and that doesn't integrate well on Mobile.

0xfe:
I think the following would just work. Can someone try it out?

```javascript
ctx = VF.Renderer.bolsterCanvasContext(YOUR_CANVAS_CONTEXT)'

// Create a stave of width 400 at position 10, 40 on the canvas.
var stave = new VF.Stave(10, 40, 400);

// Add a clef and time signature.
stave.addClef("treble").addTimeSignature("4/4");

// Connect it to the rendering context and draw!
stave.setContext(context).draw();
```

turnerhayes:
So the good news is that offscreen canvas does indeed seem to work; I've put together a little repo with a simple demo of it: https://turnerhayes.github.io/vexflow-offscreen-canvas

The bad news is that VexFlow doesn't work (out of the box) in a Worker process; it references `window`. I was able to work around that by doing `var window = self;` in the worker file before importing VexFlow, but it would be better if VexFlow itself were adjusted to work in both main and Worker threads. With that little hack, it seems to work (although I tried adding a bunch of notes in the worker thread to test the use case in which a lot of work is being done on a score, but I don't know enough about how VexFlow works to do that)

PR #1216, opened 2021-11-05
Fonts: DEMOS

ronyeh:
This PR contributes examples in the `vexflow/demos/fonts` folder. It is part of the Font PR that has been extracted for easier code review & merging.

Since these files are independent from the core code, this PR is safe to merge. However, the demos won't actually work until the Font related improvements are merged (in the next couple of PRs).

### DIFF
--- a/demos/fonts/all.html
+++ b/demos/fonts/all.html
@@ -1,0 +1,76 @@
+<!DOCTYPE html>
+<html>
+  <body>
+    <div id="outputBravura"></div>
+    <div id="outputGonville"></div>
+    <div id="outputPetaluma"></div>
+    <!-- The standard vexflow build includes three music fonts. -->
+    <script src="../../build/vexflow.js"></script>
+    <script type="module">
+      console.log('VexFlow BUILD ' + Vex.Flow.BUILD);
+
+      ////////////////////////////////////////////////////////////////////////
+
+      {
+        Vex.Flow.setMusicFont('Bravura');
+        const factory = new Vex.Flow.Factory({
+          renderer: { elementId: 'outputBravura', width: 500, height: 130 },
+        });
+        const score = factory.EasyScore();
+        factory
+          .System()
+          .addStave({
+            voices: [
+              score.voice(score.notes('C#5/q, B4, A4, G#4', { stem: 'up' })),
+              score.voice(score.notes('C#4/h, C#4', { stem: 'down' })),
+            ],
+          })
+          .addClef('treble')
+          .addTimeSignature('4/4');
+        factory.draw();
+      }
+
+      ////////////////////////////////////////////////////////////////////////
+
+      {
+        Vex.Flow.setMusicFont('Gonville');
+        const factory = new Vex.Flow.Factory({
+          renderer: { elementId: 'outputGonville', width: 500, height: 130 },
+        });
+        const score = factory.EasyScore();
+        factory
+          .System()
+          .addStave({
+            voices: [
+              score.voice(score.notes('C#5/q, B4, A4, G#4', { stem: 'up' })),
+              score.voice(score.notes('C#4/h, C#4', { stem: 'down' })),
+            ],
+          })
+          .addClef('treble')
+          .addTimeSignature('4/4');
+        factory.draw();
+      }
+
+      ////////////////////////////////////////////////////////////////////////
+
+      {
+        Vex.Flow.setMusicFont('Petaluma');
+        const factory = new Vex.Flow.Factory({
+          renderer: { elementId: 'outputPetaluma', width: 500, height: 130 },
+        });
+        const score = factory.EasyScore();
+        factory
+          .System()
+          .addStave({
+            voices: [
+              score.voice(score.notes('C#5/q, B4, A4, G#4', { stem: 'up' })),
+              score.voice(score.notes('C#4/h, C#4', { stem: 'down' })),
+            ],
+          })
+          .addClef('treble')
+          .addTimeSignature('4/4');
+        factory.draw();
+      }
+    </script>
+  </body>
+</html>
--- a/demos/fonts/core-with-promise.html
+++ b/demos/fonts/core-with-promise.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+<!-- 
+  Instead of async/await, you can use Promise style chains.
+  For example: Vex.Flow.setMusicFont(...).then(onFulfilledCallback) 
+-->
+<html>
+  <body>
+    <div id="output"></div>
+    <!-- The vexflow core build includes no music fonts. -->
+    <script src="../../build/vexflow-core.js"></script>
+    <script>
+      // Vex.Flow.setMusicFont(...fontNames) is an async function that triggers
+      // the loading of the fonts (stored in vexflow-font-xxxx.js).
+      // It returns a Promise that resolves to the loaded music font(s).
+      Vex.Flow.setMusicFont('Petaluma', 'Bravura', 'Gonville', 'Custom').then(onFontsLoaded);
+
+      function onFontsLoaded(musicFonts) {
+        console.log('Successfully loaded this music font stack:');
+        const [bravura, gonville, petaluma, custom] = musicFonts;
+        console.log(bravura, gonville, petaluma, custom);
+
+        const factory = new Vex.Flow.Factory({ renderer: { elementId: 'output', width: 500, height: 200 } });
+        const score = factory.EasyScore();
+        const system = factory.System();
+        system
+          .addStave({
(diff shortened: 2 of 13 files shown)


demos/fonts/all.html
```diff
@@ -0,0 +1,76 @@
...
+          })
+          .addClef('treble')
+          .addTimeSignature('4/4');
+        factory.draw();
+      }
+
+      ////////////////////////////////////////////////////////////////////////
+
+      {
+        Vex.Flow.setMusicFont('Gonville');
```
ronyeh, line 36: You can change the global music engraving font. The cleanest way is to instantiate a new Factory for each font you want to use on the same webpage. If you change fonts between different sets of voices and staves, but use the same Factory, it won't do what you expect. That's because a single Factory has a loop for drawing all of its internal members in a particular order. 

Anyways, to simplify things, in the Font PR I removed the ability for a single Element to have its own music font. We always use the global music font stack now.

demos/fonts/core-with-promise.html
```diff
@@ -0,0 +1,38 @@
...
+<html>
+  <body>
+    <div id="output"></div>
+    <!-- The vexflow core build includes no music fonts. -->
+    <script src="../../build/vexflow-core.js"></script>
+    <script>
+      // Vex.Flow.setMusicFont(...fontNames) is an async function that triggers
+      // the loading of the fonts (stored in vexflow-font-xxxx.js).
+      // It returns a Promise that resolves to the loaded music font(s).
+      Vex.Flow.setMusicFont('Petaluma', 'Bravura', 'Gonville', 'Custom').then(onFontsLoaded);
```
ronyeh, line 15: async / await is the same as promise / then. If you like using `.then(onFulfilledCallback)`, it works fine.

Also notice here that `Flow.setMusicFont(...)` takes a variable number of font family strings. The first argument is the primary music font. The others are fallbacks, in case a glyph is not found in the first music font.

demos/fonts/core.html
```diff
@@ -0,0 +1,65 @@
...
+          container.removeChild(container.firstChild);
+        }
+
+        // Vex.Flow.setMusicFont(fontName) is an async function that triggers
+        // the loading of the font stored in vexflow-font-${fontName}.js.
+        // Choose between:
+        //   await Vex.Flow.setMusicFont('Bravura')
+        //   await Vex.Flow.setMusicFont('Gonville')
+        //   await Vex.Flow.setMusicFont('Petaluma')
+        await Vex.Flow.setMusicFont(fontName);
```
ronyeh, line 42: For vexflow-core.js, Flow.setMusicFont(fontName) is async. So you'll have to use `await` to make sure the font is loaded before proceeding to render the score.

demos/fonts/with-bravura.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<!-- 
+  vexflow-core-with-bravura.js bundles the Bravura music font. 
+  There is no need to call Vex.Flow.setMusicFont(...).
+-->
+<html>
+  <body>
+    <div id="output"></div>
+    <script src="../../build/vexflow-core-with-bravura.js"></script>
```
ronyeh, line 9: These are new! Simon of OSMD requested support for smaller builds without undesired fonts.

`vexflow-core.js` bundles zero fonts. Load fonts at runtime by calling `Flow.setMusicFont(...fontNames)`.

`vexflow-core-with-bravura.js` bundles only the Bravura font by default. Other fonts can be dynamically loaded with a call to `Flow.setMusicFont(...fontNames)`.

demos/node/fonts.js
```diff
@@ -0,0 +1,46 @@
+// node fonts.js > output.html
+//
+// This demo shows how to use node-canvas to load a font (woff / otf).
```
ronyeh, line 3: When run on the command line via node, the ChordSymbol test cases fall back to Times and Arial, because the web fonts are not available. This demo shows how we can register custom fonts like Roboto Slab and PetalumaScript so that they will appear in node-canvas.

demos/node/import/index.js
```diff
@@ -2,16 +2,12 @@
 
 /* eslint-disable no-console */
 
-import Vex from '../../../build/vexflow-debug.js';
+import Vex from '../../../build/vexflow.module.js';
```
ronyeh, line 5: The current 4.0.0-alpha branch creates a `vexflow.module.js` that has an `export default Vex;` at the bottom.

rvilarl (approved):
Very good idea to separate this block, LGTM.

demos/node/customcontext.js
```diff
@@ -2,17 +2,15 @@
...
 
-const { createCanvas } = require('canvas');
 const Vex = require('../../build/vexflow-debug');
+
 const VF = Vex.Flow;
 
 // A custom Vex.Flow.RenderContext implementation.
-// This is just a stub for demonstration purposes that console.logs all method
-// calls and arguments.
+// This is just a stub for demonstration purposes that console.logs method calls and arguments.
```
rvilarl, line 10: @roney does #1214 address #655?

ronyeh: If I'm reading #655 correctly, I believe this was addressed by tommadams in https://github.com/0xfe/vexflow/pull/1161  He was the one who added that customcontext.js demo. (I just updated it to reflect the current APIs).

rvilarl: What about #1157? Is it also addressed?

rvilarl: @tommadams can you confirm if #655 and/or #1157 are already addressed.

tommadams: Regarding #655: [msdn](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas/getContext#parameters) says that `OffscreenCanvas.getContext('2d')` returns a `CanvasRenderingContext2D` but [@types/offscreencanvas](https://www.npmjs.com/package/@types/offscreencanvas) defines a different type: `OffscreenCanvasRenderingContext2D`. So it looks like it might be the case that the JavaScript runtime treats offscreen contexts as equivalent to canvas contexts but TypeScript treats them as separate types.

We might need to take a dependency on `@types/offscreencanvas` and change the `CanvasContext` constructor to accept both a `CanvasRenderingContext2D` and an `OffscreenCanvasRenderingContext2D`. I can look at that over the weekend.

Regarding #1157: I still don't know what the original problem was. It looked as though the original poster was trying to pass some kind of React object to the `Renderer` constructor but they never actually fully explained what they were doing. I think we can close that one for now.

rvilarl:
@0xfe IMO this is ready to merge.

demos/fonts/all.html
```diff
@@ -0,0 +1,76 @@
...
+    <div id="outputPetaluma"></div>
+    <!-- The standard vexflow build includes three music fonts. -->
+    <script src="../../build/vexflow.js"></script>
+    <script type="module">
+      console.log('VexFlow BUILD ' + Vex.Flow.BUILD);
+
+      ////////////////////////////////////////////////////////////////////////
+
+      {
+        Vex.Flow.setMusicFont('Bravura');
```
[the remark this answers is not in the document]
0xfe: These comments would actually be useful inside the code blocks.

demos/fonts/all.html
```diff
@@ -0,0 +1,76 @@
...
+  <body>
+    <div id="outputBravura"></div>
+    <div id="outputGonville"></div>
+    <div id="outputPetaluma"></div>
+    <!-- The standard vexflow build includes three music fonts. -->
+    <script src="../../build/vexflow.js"></script>
+    <script type="module">
+      console.log('VexFlow BUILD ' + Vex.Flow.BUILD);
+
+      ////////////////////////////////////////////////////////////////////////
```
0xfe, line 12: Not a huge fan of this big comment line :-)

demos/fonts/all.html
0xfe: Your first paragraph is also useful as an embedded comment.

As discussed, +1 for removing per-Element fonts.

0xfe (approved):
Looks great! Some very minor comments, but this is good to merge.

demos/fonts/all.html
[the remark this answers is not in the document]
ronyeh: Sounds good. I will address all of this feedback in the 4.0 branch.

I'll copy these comments into those demo html files.

demos/fonts/all.html
ronyeh: Sounds good. Will remove. Sometimes I put in these "horizontal separators" to help guide the code reader between large code blocks, but here I have extra braces { } anyways, so they're probably not needed.

PR #1219, opened 2021-11-08
Offscreen canvas

### LINKED ISSUE #655: enhancement: also accept canvas context instead of HTMLElement.

tommadams:
Fixes #655.

### DIFF
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -147,7 +147,7 @@ export class CanvasContext extends RenderContext {
 
     // The canvas could be an instance of either HTMLCanvasElement or an OffscreenCanvas.
     // Only HTMLCanvasElement has a style attribute.
-    if (canvas instanceof HTMLCanvasElement) {
+    if (canvas instanceof window.HTMLCanvasElement) {
       canvas.style.width = width + 'px';
       canvas.style.height = height + 'px';
     }
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -9,7 +9,7 @@ import { warn } from './util';
  */
 export class CanvasContext extends RenderContext {
   context2D: CanvasRenderingContext2D | OffscreenCanvasRenderingContext2D;
-  canvas: HTMLCanvasElement | { width: number; height: number };
+  canvas: HTMLCanvasElement | OffscreenCanvas | { width: number; height: number };
 
   /** Height of one line of text (in pixels). */
   textHeight: number = 0;
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -8,7 +8,7 @@ import { warn } from './util';
  * A rendering context for the Canvas backend (CanvasRenderingContext2D).
  */
 export class CanvasContext extends RenderContext {
-  context2D: CanvasRenderingContext2D;
+  context2D: CanvasRenderingContext2D | OffscreenCanvasRenderingContext2D;
   canvas: HTMLCanvasElement | { width: number; height: number };
 
   /** Height of one line of text (in pixels). */
@@ -44,7 +44,7 @@ export class CanvasContext extends RenderContext {
     return [width, height];
   }
 
-  constructor(context: CanvasRenderingContext2D) {
+  constructor(context: CanvasRenderingContext2D | OffscreenCanvasRenderingContext2D) {
     super();
     this.context2D = context;
     if (!context.canvas) {
@@ -132,7 +132,7 @@ export class CanvasContext extends RenderContext {
   }
 
   resize(width: number, height: number): this {
-    const canvasElement = this.context2D.canvas;
+    const canvas = this.context2D.canvas;
     const devicePixelRatio = window.devicePixelRatio || 1;
 
     // Scale the canvas size by the device pixel ratio clamping to the maximum supported size.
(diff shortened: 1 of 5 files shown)


rvilarl (approved):
LGTM

ronyeh:
Hi @tommadams 

On master there is currently an issue with one of the node demos related to instanceof window.HTMLCanvasElement:

```
node demos/node/canvas.js

                if (!(element instanceof window.HTMLCanvasElement)) {
                                         ^
ReferenceError: window is not defined
    at new Renderer (build/vexflow-debug.js:21736:42)
    at Object.<anonymous> (demos/node/canvas.js:11:18)

```

Since we are supporting uses of VexFlow under node, we need to allow for clean room reimplementations of Canvas or Div. For example, [node-canvas](https://www.npmjs.com/package/canvas) has a canvas implementation that fails this instanceof test. (Additionally, `window` is not available in node.)

In my VexFlow 4.0 alpha branch, I added two typeguard methods for checking if an element is HTMLCanvasElement or HTMLDivElement.

https://github.com/ronyeh/vexflow/blob/cbd60f7a993921fb2d81ea523ff5f2a39cdf6e7f/src/typeguard.ts#L14-L39

https://github.com/ronyeh/vexflow/blob/cbd60f7a993921fb2d81ea523ff5f2a39cdf6e7f/src/renderer.ts#L139-L155

Does this implementation look good to you? The first case does the instanceof check, and also supports jsdom. The second case does feature detection on the canvas or div, because if the passed in object includes the methods we care about, we can probably proceed.

```typescript
// eslint-disable-next-line
export const isHTMLCanvas = (element: any): element is HTMLCanvasElement => {
  if (!element) return false;

  const globalObject = typeof window !== 'undefined' ? window : globalThis;
  return (
    // It's either an instance of the HTMLCanvasElement class,
    (typeof globalObject.HTMLCanvasElement === 'function' && element instanceof globalObject.HTMLCanvasElement) ||
    // OR it's pretending to be a <canvas> element. Good enough!
    // Do not rely on .tagName, because node-canvas doesn't provide a tagName.
    (typeof element.getContext === 'function' && typeof element.toDataURL === 'function')
  );
};

// eslint-disable-next-line
export function isHTMLDiv(element: any): element is HTMLDivElement {
  if (!element) return false;

  const globalObject = typeof window !== 'undefined' ? window : globalThis;
  return (
    // It's either an instance of the HTMLDivElement class.
    (typeof globalObject.HTMLDivElement === 'function' && element instanceof globalObject.HTMLDivElement) ||
    // OR it's pretending to be a <div>. See: svgcontext.ts.
    (typeof element.appendChild === 'function' && typeof element.style === 'object')
  );
}
```

So now the instanceof checks become:

```typescript
      // Verify backend and create context
      if (backend === Renderer.Backends.CANVAS) {
        if (!isHTMLCanvas(element)) {
          throw new RuntimeError('BadElement', 'CANVAS context requires an HTMLCanvasElement.');
        }
        const context = element.getContext('2d');
        if (!context) {
          throw new RuntimeError('BadElement', "Can't get canvas context");
        }
        this.ctx = new CanvasContext(context);
      } else if (backend === Renderer.Backends.SVG) {
        if (!isHTMLDiv(element)) {
          throw new RuntimeError('BadElement', 'SVG context requires an HTMLDivElement.');
        }
        this.ctx = new SVGContext(element);
      } else {
        throw new RuntimeError('InvalidBackend', `No support for backend: ${backend}`);
      }
```

src/canvascontext.ts
```diff
@@ -142,10 +142,15 @@ export class CanvasContext extends RenderContext {
...
-    canvasElement.width = width * devicePixelRatio;
-    canvasElement.height = height * devicePixelRatio;
-    canvasElement.style.width = width + 'px';
-    canvasElement.style.height = height + 'px';
+    canvas.width = width * devicePixelRatio;
+    canvas.height = height * devicePixelRatio;
+
+    // The canvas could be an instance of either HTMLCanvasElement or an OffscreenCanvas.
+    // Only HTMLCanvasElement has a style attribute.
+    if (canvas instanceof window.HTMLCanvasElement) {
```
ronyeh, line 150: So in VexFlow 4 we could use `isHTMLCanvas(element)` instead of `instanceof`, to support environments where window is undefined, and cases where the canvas isn't actually an instance of HTMLCanvasElement, but it supplies all the necessary methods and fields.

tests/offscreencanvas_tests.ts
```diff
@@ -0,0 +1,60 @@
...
+import { CanvasContext } from '../src/canvascontext';
+import { Formatter } from '../src/formatter';
+import { Stave } from '../src/stave';
+import { BarlineType } from '../src/stavebarline';
+import { StaveNote } from '../src/stavenote';
+
+const OffscreenCanvasTests = {
+  Start(): void {
+    // At the time of writing, OffscreenCanvas is still an experimental technology.
+    if (window.OffscreenCanvas === undefined) {
```
ronyeh, line 15: It's probably fine since we mostly run tests in the browser, but maybe someday we need a util method for getting the global, to avoid 'window' is not defined scenarios.

0xfe:
Thanks Tom. This CL looks good to me.

@ronyeh, how about we have the `isHTMLCanvas` discussion (and potentially change) in a separate PR? (It's a good discussion, but no need to hold up this PR?)

ronyeh:
Sounds good. I think I can review & test out this PR this afternoon, and can merge it.

That way I can merge the new changes into my 4.0 branch, and make sure you have a clean merge with no conflicts and no visual diffs to review. :-)

ronyeh:
Ran the tests successfully, and checked flow.html to see that the OffscreenCanvas test works as expected. Merged.

tommadams:
Hey, sorry for not replying earlier.

Your solution to `isHTMLCanvas` and `isHTMLDiv` look great, thanks for figuring that out!

PR #1220, opened 2021-11-08
Fix grunt watch with correct use of the webpack `ignored` field.

ronyeh:
@tommadams can you confirm if this patch fixes the grunt watch for you? These are the options I'm currently using on my branch.

Apologies for introducing the bug! :-)


tommadams:
Yep, that works for me. Thanks!

PR #1221, opened 2021-11-08
Minor bug in Fraction found during OSMD integration

rvilarl:
Fraction.GCD should accept 0 as numerator

### DIFF
--- a/src/fraction.ts
+++ b/src/fraction.ts
@@ -19,7 +19,7 @@ export class Fraction {
 
   /** GCD: Greatest common divisor using Euclidean algorithm. */
   static GCD(a: number, b: number): number {
-    if (typeof a !== 'number' || Number.isNaN(a) || a === 0 || typeof b !== 'number' || Number.isNaN(b) || b === 0) {
+    if (typeof a !== 'number' || Number.isNaN(a) || typeof b !== 'number' || Number.isNaN(b)) {
       throw new RuntimeError('BadArgument', `Invalid numbers: ${a}, ${b}`);
     }
 


ronyeh (changes requested):
Right now it throws if b is 0 but a is nonzero. It should also allow that case, but this might require a separate conditional. 

From Wikipedia: "This definition also applies when one of a and b is zero. In this case, the GCD is the absolute value of the non zero integer: gcd(a, 0) = gcd(0, a) = |a|. This case is important as the terminating step of the Euclidean algorithm."

How about having an independent if check above this current line.... Like if(a===0 && b===0) fail....

rvilarl:
GDC(0,0) should also return 0, I believe.

ronyeh:
https://math.stackexchange.com/questions/495119/what-is-gcd0-0#:~:text=Therefore%2C%20since%20every%20natural%20number,0%2C0)%3D0.

A comment in that page suggests gcd(0,n) = n

Even if n is zero. So if a is 0, return b. If b is 0, return a. Otherwise, do what we were previously doing.

rvilarl:
I will change to the euclidian implementation, the result gdc(0,0) is discussed in internet but this approach is as a minimum well documented :)

https://en.wikipedia.org/wiki/Euclidean_algorithm

rvilarl:
see for discussion on gcd(0,0) (the eucledian algorithm is also referenced)

https://math.stackexchange.com/questions/495119/what-is-gcd0-0#:~:text=However%2C%20while%2C%20under%20the%20usual,0%2C0)%3D0.

rvilarl:
Already included in #1214

PR #1229, opened 2021-11-14
api documentation issues fixed

### LINKED ISSUE #510: API documentation doesn't work
The link on the wiki just leads to a file listing that doesn't contain any trace of API documentation.

### LINKED ISSUE #530: API documentation
I am having a very difficult time using vexflow due to either me not finding the proper location for the API documentation or maybe these docs are missing.

Just having an API wiki specifying function arguments would be a huge help...

For example, I cannot find documentation on

1. note durations (from a number representing a fraction of a whole note)
2. creating rests

rvilarl:
fixes #510
fixes #530

**For the reviewers**
- It seems a lot of files but all in `docs/apidocs` are automaticallly generated from sources, no need to review them.
- `typedoc` builds in `build/apidocs`
- the files are then copied by `grunt` to `docs/apidocs`
- `docs/index.html` has now a link to the API documentation.


0xfe:
Thanks @rvilarl. There's a bunch of design here, can you post some screenshots? (or a public link to the docs?)

rvilarl:
Hi @0xfe, **typedoc** was introduced in #874 then **docco** was removed. Typedoc outputs are based on the comments in source code and they are already in master (they are produced under `build/docs` with `grunt` command). This PR is just copying the results to `docs/docsapi/` and linking these files from `docs/index.html` Some screenshots

**Main help page with link to API (highlighted in blue)**
Captura de pantalla de 2021-11-15 05-51-12

**Main API page**
Captura de pantalla de 2021-11-15 06-03-39

**One example: Accidentals**
Captura de pantalla de 2021-11-15 06-04-47

ronyeh:
> This PR is just copying the results to `docs/docsapi/` and linking these files from `docs/index.html`

Is it a good idea to put build products in any folder other than build/? If a developer has already cloned our repo, they could also build the docs themselves. If someone wants to read the api documentation, they probably want to do so on vexflow.com.

Maybe the build/docs/ should just be scp'd up to `vexflow.com/api/` ?

ronyeh:
What if we could use typedoc to build a set of API docs that were browsable within GitHub itself? We wouldn't need to host it anywhere else, and it would then make much sense to copy the typedoc output to `vexflow/docs/api/`.

Would this plugin work? https://www.npmjs.com/package/typedoc-plugin-markdown

It is supposed to generate typedoc output in markdown files. If we checked those *.md files into GitHub, would it be browsable via the GitHub UI? :-)

ronyeh:
OK, I think I will vote in support of this PR, as long as the typedocs are stored in `vexflow/docs/api/`.

I figure we can try it out for now. Every time our method signatures and comments change, it will be reflected in the api documentation that is checked into the repo. If it gets annoying to maintain for whatever reason, we can remove it someday in the future. But I think we can just mostly ignore any changes in docs/api/ and treat them as if they were *.d.ts files, but for humans to look at instead of for the typescript compiler to consume.

rvilarl:
As suggested by Ron I moved apidocs to api. Ready to review.

As already stated above files in docs/api are generated automatically and do not have to be reviewed.

0xfe:
Thanks Rodrigo -- this is perfect, and I like that it remains updated, and we can link to it. Merging -- can you add a link in the wiki?

rvilarl:
@0xfe I have seen that the wiki points to a server public.vexflow.com do you maintain this one?

PR #1230, opened 2021-11-14
Remove releases/ folder

ronyeh:
Coming soon in a separate PR: a script based on `npm ci` that will download any old releases you specify, to support visual diffing between versions.

In a future PR: an interactive script automating the publishing of the NPM package, that is a bit less magical than grunt-brump and grunt-release.


ronyeh:
OK this one is ready to go.

It also fixes the grunt warnings, because it removes the [grunt-release package](https://www.npmjs.com/package/grunt-release) which was last updated 5 years ago. We can automate our NPM release in a different way through an interactive node script (that invokes npm commands and package.json scripts).

0xfe:
Conflict to fix.

So what do we lose with this? Is there any information loss at all from not checking in the releases? Are NPM + git tags sufficient? Are `map` files preserved?

ronyeh:
**What information do we lose?** That's a great question. To be extra safe, you could create a new repo called 0xfe/vexflow-releases/ which includes all the relevant release files.

But assuming that NPM doesn't disappear, we can always get the build files for any revision by doing "npm install vexflow@x.y.z" This means old releases are also browsable via unpkg and jsdeliver. 

* https://unpkg.com/browse/vexflow@3.0.9/
* https://www.jsdelivr.com/package/npm/vexflow

Since I'm the one removing the `releases/`, I can commit to working on a script which pulls down the releases as necessary from npm (via npm install or npm ci). Git tags are also sufficient, assuming we never delete the tag, which would mean we lose the reference to the commit hash. But yeah, if we want to be extra extra safe, we'd want to have a separate vexflow-releases repo. We can then install that repo as an optional submodule in vexflow, so that we could pull down old releases if necessary.

Regarding **source maps**: A comment by @sschmidTU https://github.com/0xfe/vexflow/issues/1231#issuecomment-968757488 suggested that maybe we don't need source maps on NPM. What do you think? :-) In my ESM branch I've temporarily disabled source map generation and 1) it makes webpack build much faster and 2) our build/ folder and npm package is super nice and clean looking. How often do you think people rely on source maps? If they need them, could they just build VexFlow from source and then turn on the sourcemap flag?

Anyways, let me know your opinion on source maps inside the NPM package. I think Simon said he just puts them in the OSMD GitHub release zip file, but does not put them on NPM.

ronyeh:
Tested this, with no visual diffs (via `npm run test:reference`). After this is merged, I'll try to finalize the proper ES module support so Simon can build OSMD with a light version of VexFlow that includes only Gonville. After the ESM PR is merged and tested, I'm happy to declare beta 2. I can work on a script to pull down old release builds from NPM while folks are testing out beta 2.

0xfe:
Ron, instead of a new repo, how about Github releases? I've used that with other projects, and it's really easy to setup and automate.

Re: source maps, we need it somewhere, not necessarily in the NPM packages. We could build and bundle them with the github releases too. The criteria here is that we don't want people to have to build source maps for a version, they should be pre-built and downloadable for reach version.

We can definitely remove them from default builds -- they're only needed for versioned releases.

package.json
```diff
@@ -76,34 +76,30 @@
...
     "opentype.js": "^1.3.4",
     "prettier": "^2.4.1",
     "qunit": "^2.17.2",
     "string-replace-loader": "^3.0.3",
+    "svg2pdf.js": "^2.2.0",
     "ts-loader": "^9.2.6",
     "typedoc": "^0.22.6",
     "typescript": "^4.5.1-rc",
-    "webpack": "^5.59.1",
-    "jspdf": "^2.4.0",
```
rvilarl, line 92: jspdf is being used by a demo: `demo/node/pdf.js`

ronyeh:
Okay. GitHub releases sounds good. I haven't used it before, so will have to learn it.

ronyeh:
This seems like a decent replacement for grunt-release.  It works with GitHub releases. 

https://github.com/release-it/release-it

152,543 weekly downloads and last updated 2 days ago. Seems promising.

0xfe:
Thanks Ron. Merging.

PR #1233, opened 2021-11-15
Use constants in Font class. Fix font related bugs.

ronyeh:
Introduces visual diffs. The outlines of text elements are slightly different, due to use of constant font family:

```
  /** Default sans-serif font family. */
  static SANS_SERIF: string = 'Arial, sans-serif';

  /** Default serif font family. */
  static SERIF: string = 'Times New Roman, serif';
```

Previously, each text element declared a slightly different sans-serif or serif font family.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -362,6 +362,18 @@ export default Vex;`;
     ]
   );
 
+  // `grunt watchDevelop`
+  // This is the fastest way to build vexflow-debug-with-tests.js.
+  // Open tests/flow.html to see the test output.
+  grunt.registerTask(
+    'watchDevelop',
+    `Watch src/ & tests/ for changes and generate a development build.`, //
+    [
+      'clean:build', //
+      'webpack:watchDebugPlusTests',
+    ]
+  );
+
   // `grunt test`
   grunt.registerTask(
     'test',
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,17 @@
 {
   "name": "vexflow",
-  "version": "3.0.9",
+  "version": "4.0.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "vexflow",
-      "version": "3.0.9",
+      "version": "4.0.0",
       "license": "MIT",
+      "dependencies": {
+        "@types/offscreencanvas": "^2019.6.4"
+      },
       "devDependencies": {
-        "@types/offscreencanvas": "^2019.6.4",
         "@typescript-eslint/eslint-plugin": "^5.1.0",
         "@typescript-eslint/parser": "^5.1.0",
         "canvas": "^2.8.0",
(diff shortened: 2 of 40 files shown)


ronyeh:
This one is ready for review. Lots of small visual diffs that are expected. Before we were specifying fonts like Times or sans-serif, or Serif, or Times New Roman, or Arial, etc etc. Now we consolidate mostly to two text font stacks: 

```
class Font {

  /** Default sans-serif font family. */
  static SANS_SERIF: string = 'Arial, sans-serif';

  /** Default serif font family. */
  static SERIF: string = 'Times New Roman, serif';

  ...
}
```

Most of the visual diffs are around the edges of the text, because there are ever so slight differences between Times and Times New Roman, and also tiny differences between Arial and Helvetica.

Beam Complex_Beams_with_Annotations Bravura

Stave Multiple_Line_Stave_Text_Test Bravura


There are differences on any test case that uses the PetalumaScript web font. In our head revision we were pulling the incorrect web font, so the TextFormatter was providing incorrect kerning.

There is also a visual difference on text Annotations, because I tweaked the way we calculate line heights using the tallest glyph (which is actually the lowercase b in many fonts).

This improved the visual layout of lyrics.

Lyrics

Gruntfile.js
```diff
@@ -362,6 +362,18 @@ export default Vex;`;
     ]
   );
 
+  // `grunt watchDevelop`
+  // This is the fastest way to build vexflow-debug-with-tests.js.
+  // Open tests/flow.html to see the test output.
+  grunt.registerTask(
+    'watchDevelop',
```
0xfe, line 369: The camel case here keeps throwing me off. Why don't we use `watch:develop`, `watch:production`, etc.

(Also, let's use `dev|prod` or `development|production` to keep things consistent. I prefer the short form, but not a strong opinion.)

0xfe (commented):
Good fix, Ron, and thanks for the other changes too. Made a minor comment for a future PR.

Gruntfile.js
ronyeh: Thanks for the feedback. I had to improve the build process for the ESM branch, so some of this is already outdated, ha. 

We can revisit this topic when I submit my ESM PR for review, and you can look over the new Gruntfile and let me know what you'd like changed.

Are you against camelCase in in the Gruntfile in general?

I'm not entirely sure a colon works for registerTask(). I've seen it used to separate `task:target` during the config: https://gruntjs.com/configuring-tasks  I've also seen it referenced as a way to pass command line arguments to a task: https://gruntjs.com/creating-tasks#basic-tasks

Anyways, I can investigate to see if/how the colon form can be used here.

0xfe: Not against camelCase in general in the Gruntfile -- mostly for things run in the commandline, where lower-case or lower_case is more common. So I prefer to run `grunt watch-dev` than `grunt watchDev`. Open to even `grunt dev`.

PR #1238, opened 2021-11-22
Add test for issue #411

rvilarl:
@AaronDavidNewman as agreed in #411 I have added a test case in formatter to show this worng behaviour, could you look into a possible solution?

image


0xfe:
The bracket and tuplets should be positioned by shifting `text_line` in the formatter state. I wonder if there was a regression at some point.

PR #1242, opened 2021-11-30
Additive time signatures

rvilarl:
Fixes #987
- Time Signatures now allow '+', '-', '(' and ')'

PR #1247, opened 2021-12-05
fix grunt reference

rvilarl:
Checking in detail the differences in #1242 I have realised that `grunt reference` was broken leaving files from previous versions.

### DIFF
--- a/tools/visual_regression.sh
+++ b/tools/visual_regression.sh
@@ -27,6 +27,9 @@
 # PNG viewer on OSX. Switch this to whatever your system uses.
 # VIEWER=open
 
+# use . as decimal separator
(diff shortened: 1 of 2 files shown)


tools/visual_regression.sh
```diff
@@ -154,7 +157,11 @@ function diff_image() {
   # Calculate the difference metric and store the composite diff image.
   local hash=`compare -metric PHASH -highlight-color '#ff000050' $diff-b.png $diff-a.png $diff-diff.png 2>&1`
 
-  local isGT=`echo "$hash > $THRESHOLD" | bc -l`
+  # Remove scientific notation
```
0xfe, line 160: 👍

PR #1248, opened 2021-12-05
remove NOGLYPHS (only)

rvilarl:
@0xfe as wished I have extracted the removal of NOGLYPHS from #1242.

Fixes #1129
- Fonts regenarated from OTF in order to remove the NOGLYPH which were most probably added manually.
- Valid code enhanced to include Brass Techniques and symbol required by Additive/Complex Signatures. 

Fixes #1188

PR #665, opened 2018-10-15
587 fix dots horizontal position with flag

sug1no:
This PR fix #587.

>// For upstems with flags, the extra space is unnecessary, since it's taken
// up by the flag.

In that case place dots of all noteheads (not only `isInnerNoteIndex() === true`) to right side of flag.

```shell
$ npm test

...

You have 2 warning(s):
  Warning: Grace_Notes.Grace_Note_Slash.svg missing in ./build/images/blessed.
  Warning: Grace_Notes.Grace_Note_Slash_with_Beams.svg missing in ./build/images/blessed.
You have 2 fail(s):
Dot.Basic 100.484
Dot.Multi_Voice 76.7845

```
image

PR #1249, opened 2021-12-05
Parenthesised notehead

### LINKED ISSUE #654: How to represent ghost notes in drums
In drum notation, ghost notes are represented with the note head surrounded by parenthesis (see https://en.wikipedia.org/wiki/Ghost_note#/media/File:Ghost_note_drumming.png). I see that this notation exists for tablatures, however I was wondering how I could use it with "standard" notation ?

### LINKED ISSUE #705: Invalid note initialization object
# Version

1.2.89

# Steps to Reproduce

```index.html
<html>
  <head>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/vexflow/1.2.89/vexflow-min.js"></script>
  </head>
  <body>
    <div>
      Vexflow test
    </div>
    <div>
      <canvas id="vexflow-test"></canvas>
    </div>

    <script>
      var vf = Vex.Flow;
      var factory = new vf.Factory({ renderer: { elementId: 'vexflow-test' } })
      var score = factory.EasyScore()
      score.parse('c3//x')
    </script>
  </body>
</html>
```

# Expected 

the note line should be parsed with success and no exception

# What happened

exception was raised

```
Invalid note initialization object: {"keys":["c/3"],"duration":"8","dots":0,"type":"x","clef":"treble","auto_stem":true}
```

rvilarl:
Fixes #654
- Parenthesised notes added

- `EasyScore` supports now types `p` for parenthesis

**EasyScore Draw_Parenthesised**
EasyScore Draw_Parenthesised Bravura
EasyScore Draw_Parenthesised Gonville
EasyScore Draw_Parenthesised Petaluma


rvilarl:
@0xfe please merge first #1248 to reduce the number of changes in this PR. 
Ready to review and merge.

0xfe:
Thanks Rodrigo. Few things:

- I think we'll need to fine tune the kerning, because the rendering of the parenthesis does not look consistent. Maybe we can find some online examples to get a sense of the right spacing etc.
- What happens if you add other elements? like dots, accidentals, or other articulations. Should this be rendered as a modifier instead of a note head? (so the allocated space can be properly tracked.)
- What happens to the width of the element? (I think it'll need to be adjusted too.)

rvilarl:
I have followed the approach to render both parenthesis separately because the `modifiers` are either right or left of the notehead and I think that it would be inconvenient to have to set two `modifiers`. However I do not get the `modifiers` / `accidentals` moved away from the notehead, `this.width` seems to have no influence.
EasyScore Draw_Others Bravura
EasyScore Draw_Others Gonville
EasyScore Draw_Others Petaluma

rvilarl:
Example from MuseScore
image

AaronDavidNewman:
>> `this.width` seems to have no influence

I think most of the logic will only get the notehead width from its glyph.  We expect everything around the notehead to be a modifier that has its own formatting logic, but noteheads are just glyphs.

>> Maybe we can find some online examples to get a sense of the right spacing etc.

I have seen this (parens around notes) in 2 contexts:  around several notes, in which case it would be a left modifier for some notes, and a right modifier for another.  And in hymnals, where a note is not played on all verses.  In  this case, I think you would want the parens around the accidentals and dots. I'll try to post some examples.

In either case, I think you are stuck having a left and a right modifier.  Or at least, the same modifier class that can either be left or right (depending on the glyph).  Easy score can still treat it like one thing, but I think the formatting will have to be done on each side individually.

Kerning is probably not necessary - you can just give each paren its entire width.  But you will need to consider the placement wrt the other modifiers on the note.

rvilarl:
@AaronDavidNewman Thank!

rvilarl:
@0xfe I have updated the images with the two parenthesis approach. Do you find that ok or do you want me to go for the modifier as suggested by @AaronDavidNewman? If this is the case, what kind of modifier?

0xfe:
Thanks folks. I think this can be built like a NoteHead, we have all the right abstractions for it. And, as far as I can tell, the only thing that matters is `getWidth()`, and we have an ability to customize that through the tables too (see: `getWidth: () => Tables.SLASH_NOTEHEAD_WIDTH` for example.)

For parenthesizing multiple notes, that's a bit more complex and we should treat that differently (probably by using a new type of note, like ClefNote). Let's work on single noteheads for now, and we can figure out a way to do that after.

rvilarl:
@0xfe declaring `getWidth()` and re-arrenging the parenthesis and notehead has the following on the stems. Any suggestion?

EasyScore Draw_Parenthesised Bravura

0xfe:
That's a lot better -- I think that could be fixed easily by using the adding x shifts for parenthetized notes (by the width of the parens + the spacing.) It just needs to be shifted enough for the stem to touch the head. (Note that stem ups will have a negative shift from the stem downs.)

That said, are the accidentals supposed to be inside the parens, or outside?

sschmidTU:
To me accidentals outside of the parens would be fine, because they probably take a lot of space if inside, and space is always an issue. The sample above looks fine to me, though I'm not sure if there's a standard that says accidentals should be included.

AaronDavidNewman:
I agree that this looks much better, and the multi-note case is a different feature.

rvilarl:
> That's a lot better -- I think that could be fixed easily by using the adding x shifts for parenthetized notes (by the width of the parens + the spacing.) It just needs to be shifted enough for the stem to touch the head. (Note that stem ups will have a negative shift from the stem downs.)

Shifting will make the parentesis overlap with the accidentals. Those as the stem will not move.

0xfe:
> Shifting will make the parentesis overlap with the accidentals. Those as the stem will not move.

Look for `stem_up_x_offset` and `stem_down_x_offset`.

rvilarl:
What should I look into to align the noteheads with up and down stem? They are now not align and they cannot share a single stem. Look at the first chord of the second voice in the upper stave.

rvilarl:
I see that this is related to `custom_glyph_code`. Is there any example on how this should work?

rvilarl:
> I see that this is related to `custom_glyph_code`. Is there any example on how this should work?

@0xfe do you know how to use this custom glyph?

0xfe:
Ah, okay, those weren't it. [Here](https://github.com/0xfe/vexflow/blob/9b12482424dd0fc36846b9182519d3289a5e75bc/src/notehead.ts#L210) are the relevant lines (in `notehead.ts`.)

```typescript
getAbsoluteX(): number {
    // If the note has not been preformatted, then get the static x value
    // Otherwise, it's been formatted and we should use it's x value relative
    // to its tick context
    const x = !this.preFormatted ? this.x : super.getAbsoluteX();

    // For a more natural displaced notehead, we adjust the displacement amount
    // by half the stem width in order to maintain a slight overlap with the stem
    const displacementStemAdjustment = Stem.WIDTH / 2;
    const musicFont = Tables.currentMusicFont();
    const fontShift = musicFont.lookupMetric('notehead.shiftX', 0) * this.stem_direction;
    const displacedFontShift = musicFont.lookupMetric('noteHead.displaced.shiftX', 0) * this.stem_direction;

    return (
      x +
      fontShift +
      (this.displaced ? (this.width - displacementStemAdjustment) * this.stem_direction + displacedFontShift : 0)
    );
  }
```

We'll need to add a new metric to the metrics files, and something like:

```typescript
const parenthesizedFontShift = musicFont.lookupMetric('noteHead.parenthesized.shiftX', 0) * this.stem_direction;
```

src/tables.ts
```diff
@@ -933,6 +935,11 @@ const durationCodes: Record<string, any> = {
         getWidth: () => Tables.SLASH_NOTEHEAD_WIDTH,
         position: 'B/4',
       },
+      p: {
+        // Parenthesis drawn by notehead.draw()
+        code_head: 'noteheadDoubleWhole',
+        getWidth: () => Tables.PARENTHESIS_NOTEHEAD_WIDTH,
```
0xfe, line 941: This fixed width works for slashed notes, but I don't think for parenthesis (across all fonts and notehead types) -- you'll need to get the current width, and add the width of two parens + some spacing.

rvilarl:
@0xfe after trying several options without success  I decided to move to the proposal from @AaronDavidNewman: create a new `Modifier`. This one is `Parenthesis`

These are the results:

**EasyScore Draw_Parenthesised Bravura**
EasyScore Draw_Parenthesised Bravura
**EasyScore Draw_Parenthesised Gonville**
EasyScore Draw_Parenthesised Gonville
**EasyScore Draw_Parenthesised Petaluma**
EasyScore Draw_Parenthesised Petaluma

I had to make `Dot` `right_shift` aware. This was a minor change (please look at the chnage in this PR). However there is one visual difference that is not right. **Help would be appreciated**
**Dot Multi_Voice Gonville**
Dot Multi_Voice Gonville
**Current**
Dot Multi_Voice Gonville_Current
**Reference**
Dot Multi_Voice Gonville_Reference

h-sug1no:
For {Dot Multi_Voice} issues, please refer to the #665 (if this fix is still alive, it may be related to the problem).

rvilarl:
> For {Dot Multi_Voice} issues, please refer to the #665 (if this fix is still alive, it may be related to the problem).

Thanks @h-sug1no I have looked further in the details and I found that the main problem was the test introduced in #665. The two voices were not formatted correctly. I have placed all the note1 in voice1 and all the note2 in a voice2 and I have used formatter. Now I believe evrything is ok.

rvilarl:
After modifing the test this is the only visual differences (bravura version):

**Dot Multi_Voice Bravura**
Dot Multi_Voice Bravura
**Current**
**Reference**

Parenthesis tests:
**EasyScore Draw_Parenthesised Bravura**
EasyScore Draw_Parenthesised Bravura
**EasyScore Draw_Parenthesised Gonville**
EasyScore Draw_Parenthesised Gonville
**EasyScore Draw_Parenthesised Petaluma**
EasyScore Draw_Parenthesised Petaluma

h-sug1no:
> Thanks @h-sug1no I have looked further in the details and I found that the main problem was the test introduced in #665. The two voices were not formatted correctly. 

@rvilarl san:

Wow, that's true!
I was only paying attention to the position of the dots.
Now I realize that this test originally drew the beats in pieces.
Great fix!

AaronDavidNewman:
This looks great.  So much better than MuseScore!  Maybe add 2 pixels of padding on the left paren?

src/parenthesis.ts
```diff
@@ -0,0 +1,89 @@
...
+    return true;
+  }
+
+  constructor(position: ModifierPosition) {
+    super();
+
+    this.position = position ?? Modifier.Position.LEFT;
+
+    this.scale = 39;
+    this.setWidth(7);
```
0xfe, line 58: Move these constants to the metrics files please.

src/parenthesis.ts
```diff
@@ -0,0 +1,89 @@
...
+        x_widthL += shift + parenthesis.width;
+      }
+      parenthesis.setXShift(shift);
+    }
+    state.left_shift += x_widthL;
+    state.right_shift += x_widthR;
+
+    return true;
+  }
+
```
0xfe, line 51: Add another static function `build(targetNote: Note)` and move the code from `stavenote.ts` here which builds both the left and right parenthesis.

src/stavenote.ts
```diff
@@ -426,6 +427,12 @@ export class StaveNote extends StemmableNote {
     }
     this.reset();
     this.buildFlag();
+    if (this.noteType === 'p') {
```
0xfe, line 430: Can we move this to `Factory` please? -- it doesn't need to be a `noteType`, otherwise we'll end up with a lot of special cases. See my comment above re: `build(targetNote)`.

src/tables.ts
```diff
@@ -156,6 +156,7 @@ const validNoteTypes: Record<string, { name: string }> = {
   m: { name: 'muted' },
   s: { name: 'slash' },
   g: { name: 'ghost' },
+  p: { name: 'parenthesized' },
```
0xfe, line 159: You can remove this (and the code below.)

src/easyscore.ts
```diff
@@ -175,7 +175,7 @@ export class EasyScoreGrammar implements Grammar {
     return { token: '[0-9whq]+' };
   }
   TYPES(): Rule {
-    return { token: '[rRsSmMhHgG]' };
+    return { token: '[rRsSmMhHgGpP]' };
```
0xfe, line 178: Let's deal with this in a separate PR. Based on the changes below, we might need to call `build()` explicitly.

0xfe (commented):
Thanks Rodrigo. This looks a lot better overall. Two main comments:

- since it's a modifier let's not make it a note type (`p`) -- I'd prefer to eliminate the pattern of automatically adding modifiers where possible. See my comments below.
- move the hardcoded numbers to the metric files, including the left and write spacing, so we can customize them per font.

0xfe:
Also, can you add a method in `factory.ts` too please?

src/tables.ts
rvilarl: done

src/parenthesis.ts
rvilarl: I have called it `addParentheses` rather than build. This is inline to the `getAccidentals` in the other PR. `Build` would also be misleading because to parenthesis are added rather than one parenthesis built.

src/stavenote.ts
rvilarl: Done

src/easyscore.ts
rvilarl: pP removed

rvilarl:
> Also, can you add a method in `factory.ts` too please?

please provide how the function should look like. Do you want to give access to the constructor or the `addParentheses` above. Creating a single `Parenthesis` is not very useful.

src/parenthesis.ts
rvilarl: done

rvilarl:
@0xfe ready for review

0xfe:
Looks great! Merging. Thanks. (For `Factory`, I think we could have a `Parenthesis(note)` that returns both the `Parenthesis` instances.

PR #1251, opened 2021-12-09
Easyscore support for muted, harmony, slash and ghost notes

### LINKED ISSUE #463: GhostNote support in EasyScore
It would be great to implement GhostNotes into EasyScore.

### LINKED ISSUE #705: Invalid note initialization object

rvilarl:
Fixes #705
Fixes #463

- `EasyScore` now also supports  types `h`, `m` (`x` was not a valid type, `m` should be used), `g` (for `GhostNote`)

### DIFF
--- a/src/beam.ts
+++ b/src/beam.ts
@@ -3,6 +3,7 @@
 
 import { Element } from './element';
 import { Fraction } from './fraction';
+import { GhostNote } from './ghostnote';
 import { Note } from './note';
 import { RenderContext } from './rendercontext';
 import { Stem } from './stem';
@@ -555,6 +556,8 @@ export class Beam extends Element {
       // iterate through notes, calculating y shift and stem extension
       for (let i = 1; i < notes.length; ++i) {
         const note = notes[i];
+        // Skip Ghostnotes
+        if (note instanceof GhostNote) continue;
         const adjustedStemTipY =
           this.getSlopeY(note.getStemX(), firstNote.getStemX(), firstNote.getStemExtents().topY, slope) + yShiftTemp;
 
@@ -689,6 +692,8 @@ export class Beam extends Element {
 
     for (let i = 0; i < notes.length; ++i) {
       const note = notes[i];
+      // Skip Ghostnotes
+      if (note instanceof GhostNote) continue;
       const stem = note.getStem();
       if (stem) {
         const stemX = note.getStemX();
--- a/src/easyscore.ts
+++ b/src/easyscore.ts
@@ -9,8 +9,8 @@ import { Music } from './music';
 import { Note } from './note';
 import { Grammar, Match, Parser, Result, Rule, RuleFunction } from './parser';
 import { RenderContext } from './rendercontext';
-import { StaveNote } from './stavenote';
 import { Stem } from './stem';
+import { StemmableNote } from './stemmablenote';
 import { TupletOptions } from './tuplet';
 import { defined, log, RuntimeError } from './util';
 import { Voice } from './voice';
@@ -22,7 +22,7 @@ function L(...args: any[]): void {
 }
 
 // eslint-disable-next-line
-export type CommitHook = (obj: any, note: StaveNote, builder: Builder) => void;
+export type CommitHook = (obj: any, note: StemmableNote, builder: Builder) => void;
 
 export class EasyScoreGrammar implements Grammar {
   builder: Builder;
@@ -42,7 +42,7 @@ export class EasyScoreGrammar implements Grammar {
   }
   PIECE(): Rule {
     return {
-      expect: [this.CHORDORNOTE, this.PARAMS],
+      expect: [this.CHORDORNOTEORGHOST, this.PARAMS],
       run: () => this.builder.commitPiece(),
     };
   }
@@ -57,10 +57,11 @@ export class EasyScoreGrammar implements Grammar {
       expect: [this.DURATION, this.TYPE, this.DOTS, this.OPTS],
     };
   }
-  CHORDORNOTE(): Rule {
+  CHORDORNOTEORGHOST(): Rule {
     return {
       expect: [this.CHORD, this.SINGLENOTE],
       or: true,
+      maybe: true,
     };
   }
   CHORD(): Rule {
@@ -175,7 +176,7 @@ export class EasyScoreGrammar implements Grammar {
     return { token: '[0-9whq]+' };
   }
   TYPES(): Rule {
-    return { token: '[rRsSxX]' };
+    return { token: '[rRsSmMhHgG]' };
   }
   LPAREN(): Rule {
     return { token: '[(]' };
@@ -227,7 +228,7 @@ export class Piece {
 }
 
 export interface BuilderElements {
-  notes: StaveNote[];
+  notes: StemmableNote[];
   accidentals: (Accidental | undefined)[][];
 }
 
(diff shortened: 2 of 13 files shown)


rvilarl:
**EasyScore Draw_Basic_Muted**
EasyScore Draw_Basic_Muted Bravura
EasyScore Draw_Basic_Muted Gonville
EasyScore Draw_Basic_Muted Petaluma

rvilarl:
**EasyScore Draw_Basic_Harmonic Bravura**
EasyScore Draw_Basic_Harmonic Bravura
EasyScore Draw_Basic_Harmonic Gonville
EasyScore Draw_Basic_Harmonic Petaluma

rvilarl:
**EasyScore Draw_Ghostnote_Basic**
EasyScore Draw_Ghostnote_Basic Bravura
EasyScore Draw_Ghostnote_Basic Gonville
EasyScore Draw_Ghostnote_Basic Petaluma

rvilarl:
**EasyScore Draw_Ghostnote_Dotted**
EasyScore Draw_Ghostnote_Dotted Bravura
EasyScore Draw_Ghostnote_Dotted Gonville
EasyScore Draw_Ghostnote_Dotted Petaluma

rvilarl:
No visual differences, ready to review.

src/beam.ts
```diff
@@ -555,6 +556,8 @@ export class Beam extends Element {
       // iterate through notes, calculating y shift and stem extension
       for (let i = 1; i < notes.length; ++i) {
         const note = notes[i];
+        // Skip Ghostnotes
+        if (note instanceof GhostNote) continue;
```
0xfe, line 560: Can we remove these checks? There must be a better abstraction here, no?

src/easyscore.ts
```diff
@@ -57,10 +57,11 @@ export class EasyScoreGrammar implements Grammar {
       expect: [this.DURATION, this.TYPE, this.DOTS, this.OPTS],
     };
   }
-  CHORDORNOTE(): Rule {
+  CHORDORNOTEORGHOST(): Rule {
```
0xfe, line 60: This doesn't feel right, the grammar doesn't change here, and we shouldn't have to special-case ghost notes.

src/easyscore.ts
```diff
@@ -57,10 +57,11 @@ export class EasyScoreGrammar implements Grammar {
...
     return {
       expect: [this.CHORD, this.SINGLENOTE],
       or: true,
+      maybe: true,
```
0xfe, line 64: What's the intent of this? I think this will accept a lot of invalid syntax.

0xfe (changes requested):
Thanks Rodrigo. This feels like a lot of changes with cascading implications: `StaveNote -> StemmableNote`, changes in grammar, changes in tuplets, removing tests for invalid syntax. We should really break this up, so we can work through them individually. Maybe just start with support for muted notes?

src/beam.ts
rvilarl: Beams accept StemmableNotes and GhostNote extends StemmableNote. The abstraction to make them Beam ignore GhostNotes based on the calculations within the for loop would be much less obvious. Would you have a proposal? Today if we leave a GhostNote in the middle of StaveNotes belonging to a Beam, vexflow will not handle it correctly. I believe that this a case to handle even if GhostNotes are not supported by Easyscore.

src/easyscore.ts
rvilarl: The change is the one you have commented below.

src/easyscore.ts
rvilarl: This is the GhostNote. GhostNotes can be now described as `/q/g` they do not need a pitch.

rvilarl:
Support for mute is changing:
```
TYPES(): Rule {
    return { token: '[rRsSxX]' };
  }
```
by
```
TYPES(): Rule {
    return { token: '[rRsSxX]' };
  }
```
And leaving the `drawBasicMutedTest`.
@0xfe Do you find it a good idea to support harmonic notes as well?

rvilarl:
The biggest changes in this PR are:
- to make EasyScore members functions `note(...)`, `beam(...)` and `tuplet(...)` consistent with the associated classes. It was `EasyScore` the only one imposing `StaveNote`.
- generalised `add/get Accidental/Annotation/...` to `Note` I thought about `StemmableNote` but all those elements are `Modifiers` and `AddModifier` was already in `Note`.

@0xfe Should I comment further the changes? They are not many?

src/beam.ts
0xfe: I think the right fix here is for GhostNote not to be stemmable. Or alternatively, `hasStem()` can always return false for GhostNotes -- so in here, you can ignore any notes that don't have stems. Maybe that will work.

src/easyscore.ts
0xfe: This breaks the grammar in a messy way, so I'm not really a fan of it. I'd rather just keep the pitch (despite it not needing it.)

src/easyscore.ts
```diff
@@ -227,7 +228,7 @@ export class Piece {
 }
 
 export interface BuilderElements {
-  notes: StaveNote[];
+  notes: StemmableNote[];
```
0xfe, line 231: Not sure I understand why we're restricting to `StemmableNote` -- what do we gain from this?

src/tuplet.ts
```diff
@@ -273,6 +274,8 @@ export class Tuplet extends Element {
       // y_pos = first_note.getStemExtents().topY - 10;
 
       for (let i = 0; i < this.notes.length; ++i) {
+        // Skip Ghostnotes
+        if (this.notes[i] instanceof GhostNote) continue;
```
0xfe, line 278: Here too, it feels like way too much of an intrusion to explicitly check for GhostNote. I'd rather it break than fail silently.

0xfe:
Thanks Rodrigo -- trying to review this again, there's just too many different things going on at different levels. Can we please break it up, it's going to be much easier to review.

src/tuplet.ts
rvilarl: I am allowing gohstnotes in the middle of beams and tuplets. Should that fail?

src/easyscore.ts
rvilarl: Support for GhostNotes in EasyScore. `GhostNote` is a `StemmableNote`.
Please note that this restriction appeared in `EasyScore`, a `Voice` accepts `Tickable`

src/tuplet.ts
rvilarl: Changed to `hasStem()`

src/easyscore.ts
rvilarl: Done

src/easyscore.ts
rvilarl: Reverted

src/beam.ts
rvilarl: Now with `hasStem()`

rvilarl:
This PR is now divided on two: #1259 & #1260 and is no longer required.

PR #1262, opened 2021-12-18
OSMD VexflowPatch: stave & stavesection

### LINKED ISSUE #1253: OSMD VexflowPatch: Stave
prevent a bug where a modifier width is NaN, leading to a VexFlow error
stave.setSection(section, y, xOffset = 0, fontSize = 12):
add xOffset, fontSize arguments (see stavesection.js)

### LINKED ISSUE #1256: OSMD VexflowPatch: StaveSection
stavesection.draw():
adjust rectangle positioning, make height depend on text height

rvilarl:
Fixes #1253
Fixes #1256 

@sschmidTU could you review this one?

@0xfe the visual regression does not show any difference although there is the difference bellow. Do you agree with this change? To me in makes more sense to use the text meassurement hieght and the question is also if we want to add a padding up / down

**Current**
Stave Multiple_Stave_Barline_Test Bravura
Stave Multiple_Stave_Barline_Test Gonville
Stave Multiple_Stave_Barline_Test Petaluma

**Reference**
Stave Multiple_Stave_Barline_Test Bravura
Stave Multiple_Stave_Barline_Test Gonville
Stave Multiple_Stave_Barline_Test Petaluma

### DIFF
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -91,13 +91,7 @@ class MeasureTextCache {
     const bbox = txt.getBBox();
     svg.removeChild(txt);
 
-    const fontSizeInPt: string = attributes['font-size'];
-
-    // Remove the trailing 'pt' from the font size and scale to convert from points to canvas pixel units.
-    // CSS specifies dpi to be 96 and there are 72 points to an inch: 96/72 == 4/3.
-    const height = Font.convertSizeToPixelValue(fontSizeInPt);
-
-    return { width: bbox.width, height: height };
+    return { width: bbox.width, height: bbox.height };
   }
 }
 
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -91,7 +91,7 @@ class MeasureTextCache {
     const bbox = txt.getBBox();
     svg.removeChild(txt);
 
-    return { width: bbox.width, height: bbox.height };
+    return { x: bbox.x, y: bbox.y, width: bbox.width, height: bbox.height };
   }
 }
 
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -258,12 +258,22 @@ export class CanvasContext extends RenderContext {
 
   measureText(text: string): TextMeasure {
     const metrics = this.context2D.measureText(text);
+
+    let y = 0;
+    let height = 0;
+    if (metrics.fontBoundingBoxAscent) {
+      y = -metrics.fontBoundingBoxAscent;
+      height = metrics.fontBoundingBoxDescent + metrics.fontBoundingBoxAscent;
+    } else {
+      y = -metrics.actualBoundingBoxAscent;
+      height = metrics.actualBoundingBoxDescent + metrics.actualBoundingBoxAscent;
+    }
     // Return x, y, width & height in the same manner as svg getBBox
     return {
       x: 0,
-      y: -metrics.fontBoundingBoxAscent,
+      y: y,
       width: metrics.width,
-      height: metrics.fontBoundingBoxDescent + metrics.fontBoundingBoxAscent,
+      height: height,
     };
   }
 
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -260,7 +260,7 @@ export class CanvasContext extends RenderContext {
     const metrics = this.context2D.measureText(text);
     return {
       width: metrics.width,
-      height: this.textHeight,
+      height: metrics.actualBoundingBoxAscent + metrics.actualBoundingBoxDescent,
     };
   }
 
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -258,9 +258,12 @@ export class CanvasContext extends RenderContext {
 
   measureText(text: string): TextMeasure {
     const metrics = this.context2D.measureText(text);
+    // Return x, y, width & height in the same manner as svg getBBox
     return {
+      x: 0,
+      y: -metrics.fontBoundingBoxAscent,
       width: metrics.width,
-      height: metrics.actualBoundingBoxAscent + metrics.actualBoundingBoxDescent,
+      height: metrics.fontBoundingBoxDescent + metrics.fontBoundingBoxAscent,
     };
   }
 
--- a/src/rendercontext.ts
+++ b/src/rendercontext.ts
@@ -4,6 +4,8 @@
 import { FontInfo } from './font';
 
 export interface TextMeasure {
+  x: number;
+  y: number;
   width: number;
   height: number;
 }
--- a/src/stave.ts
+++ b/src/stave.ts
@@ -310,9 +310,9 @@ export class Stave extends Element {
   }
 
   // Section functions
-  setSection(section: string, y: number, xOffset = 0, fontSize = 10) {
+  setSection(section: string, y: number, xOffset = 0, fontSize?: number) {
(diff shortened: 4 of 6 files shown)


sschmidTU:
I would either set the default fontsize to 10, or not make the font bold. 12pt bold is too big and hard to read in the `Current` version.

@rvilarl by the way, your images don't have a white background, so they're almost unreadable in Github Dark Mode. I guess I have to use Light Mode for this.
image

mscuthbert:
> @rvilarl by the way, your images don't have a white background, so they're almost unreadable in Github Dark Mode. I guess I have to use Light Mode for this.

Yeah, I keep thinking of putting a patch in that draws the images against a white background instead of transparent.  I realize that I haven't been reviewing visual diffs much because they're so hard for me to read.

rvilarl:
> > @rvilarl by the way, your images don't have a white background, so they're almost unreadable in Github Dark Mode. I guess I have to use Light Mode for this.
> 
> Yeah, I keep thinking of putting a patch in that draws the images against a white background instead of transparent. I realize that I haven't been reviewing visual diffs much because they're so hard for me to read.

@sschmidTU @mscuthbert this is a good idea the generated png files are transparent which does not fit with the current dark trend.

rvilarl:
The changes is no font independent. However the svg variants provide a height that is too large. I would appreciate a tip. I do not understand why `SVGtextelement.getbbox()` provides this excesive height more larger than `CanvasRenderingContext2D.measureText()`

image

rvilarl:
> The changes is now font independent. However the svg tests provide a height that is too large. I do not understand why `SVGtextelement.getbbox()` provides this excesive height more larger than `CanvasRenderingContext2D.measureText()`. I would appreciate a tip. :)

@0xfe @ronyeh @AaronDavidNewman any idea?

rvilarl:
**As disscused in #1270** 

> > You usually want max height and actual width with text, for placement. Because text has a baseline. you might want the actual height in some cases, like if you wanted to draw a glyph or a box around it.
> 
> What do you think that we should do here? Should `ctx.measureText(this.text).height` return the actual height of `this.text` or `textFormatter.maxHeight`

**And in this PR**
> The changes is now font independent. However the svg tests provide a height that is too large. I do not understand why `SVGtextelement.getbbox()` provides this excesive height larger than `CanvasRenderingContext2D.measureText()`. I would appreciate a tip. :)

@0xfe @ronyeh @AaronDavidNewman how should we proceed?

ronyeh:
> `ctx.measureText(this.text).height` return the actual height of `this.text` or `textFormatter.maxHeight`

If I recall, the maxHeight is just a hard coded value. It used to be the height of the H character. But it turns out there are some lower case characters that are taller than H, like lowercase b. In any case, using maxHeight is just an approximation and provides an "upper bound" assuming the maxHeight value was correct. 

But ideally, a method called .measureText() should probably return the real height / width measurements, if it is possible to do so correctly and with good enough performance.

rvilarl:
> But ideally, a method called .measureText() should probably return the real height / width measurements, if it is possible to do so correctly and with good enough performance.

There is no performance issue, the function is beying called anyway for the width. The only thing that is pending is to understand why the svg variant calculates the excesive height. :(

rvilarl:
According to https://stackoverflow.com/questions/36146331/javascripts-getbbox-provides-a-box-that-is-taller-than-its-content-text-why

_The bounding box returned by getBBox() is not the tight bounding box around the glyphs. It includes the full em-box height of the text - including allowance for ascenders (tallest characters in font) and descenders (lowest below baseline)._

@0xfe @sschmidTU @ronyeh how do we proceed then?

src/svgcontext.ts
```diff
@@ -91,13 +91,7 @@ class MeasureTextCache {
     const bbox = txt.getBBox();
     svg.removeChild(txt);
 
-    const fontSizeInPt: string = attributes['font-size'];
-
-    // Remove the trailing 'pt' from the font size and scale to convert from points to canvas pixel units.
-    // CSS specifies dpi to be 96 and there are 72 points to an inch: 96/72 == 4/3.
-    const height = Font.convertSizeToPixelValue(fontSizeInPt);
-
```
ronyeh, line 99: What was the reasoning behind removing this chunk of code?

I ported this code in my Fonts PR, but I did not write the original "algorithm".... It is possible that the original author realized that the SVG bounding box was too high. It seems like the approach just uses the font size as a proxy, and turns it into a value in pixel units.

ronyeh:
I spent some time looking around, and indeed the browser's SVG API does not provide a function to retrieve a tight bounding box.

It depends on what text we are sticking in the box. If the text has no descenders (the parts hanging at the bottom of the lowercase y, j, g) then we would need to reduce the height of the rectangle. Similarly, the height of the bbox leaves space for accent marks on upper case letters (Á, É, Ö, Ñ). So if our text has none of that, we'd also need to shorten the containing rectangle.

I put together a demo here:
https://jsfiddle.net/tsq6jxgL/

The basic idea is to change the **dominant-baseline** property of SVG text elements and check the bounding box's Y coordinate to programmatically determine the height of the text. 

Hanging

So maybe we can do this to a text element somewhere offscreen, and determine the height of the capital A? Note that this method ignores accent marks on upper case characters, and also ignores descenders.

I also have no idea how well this is supported across browsers! :-O

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/dominant-baseline
https://en.wikipedia.org/wiki/Ascender_(typography)
https://en.wikipedia.org/wiki/Descender
https://en.wikipedia.org/wiki/Baseline_(typography)


Happy Holidays!

rvilarl:
Hi Ron, thanks for confirming that this is the case. I have had an idea that I will push and comment later today. It is basically making canvas go the svg way.

ronyeh:
What did you think of my JSfiddle demo?

Also, did you see my online comment above on the removed chunk of code?

rvilarl:
> What did you think of my JSfiddle demo?
> 
> Also, did you see my inline comment above on the removed chunk of code?

yes I did! it is impressive that you managed to get the actual height using `'dominant-baseline` I tried with no success. But now and probably to find an alternative to the luck of success I thought that the opposite approach would have some benefits:
- If we have several sections with different texts, ti would be nice to have the rects aligned
- All the data to get the same approach in canvas was already there. There was no need for extra calculations.

I am almost there. I will push and then we decide if it looks fine or it is better to go for the tight approach. And now we know how to do it, it is in your JSfiddle.

rvilarl:
Now done with svg approach also in canvas. I had to modify `TextMeasure` to include `x` & `y` in order to have the required informationfor the location of the rect.

image

The code is now also font size independent. I have included the following test to check it.

image

rvilarl:
Ready to review

src/svgcontext.ts
rvilarl: This one was getting a fixed height rather than returning the ones calculated by `getBBox`

rvilarl:
@0xfe I believe that this one is ready to merge, any concerns from your side?

sschmidTU:
Fantastic fix, Rodrigo! Looks great! Being font size independent is great too.

rvilarl:

0xfe:
Thanks for working through this, folks. @rvilarl, change looks great. For a separate PR, we probably should have some options in `measureText` for whether or not to include the ascender. I can see how that would be useful for things like annotations.

PR #1263, opened 2021-12-18
OSMD VexflowPatch: share noteheads

### LINKED ISSUE #1254: OSMD VexflowPatch: StaveNote
Fix stem/flag formatting. Instead of shifting notes by default, update the stem/flag rendering to render different voices aligned.
Only offset if a note is the same voice, same note.

### LINKED ISSUE #1257: OSMD VexflowPatch: StemmableNote
Add manual flag rendering variable so we can choose not to render flags if notes are sharing a stem.

rvilarl:
fixes #1257
fixes #1254

### DIFF
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -10,6 +10,7 @@
 
 import { Beam } from './beam';
 import { BoundingBox } from './boundingbox';
+import { Dot } from './dot';
 import { ElementStyle } from './element';
 import { Modifier } from './modifier';
 import { ModifierContextState } from './modifiercontext';
@@ -182,7 +183,8 @@ export class StaveNote extends StemmableNote {
 
     // Test for two voice note intersection
     if (voices === 2) {
-      const lineSpacing = noteU.stemDirection === noteL.stemDirection ? 0.0 : 0.5;
+      const lineSpacing =
+        noteU.note.hasStem() && noteL.note.hasStem() && noteU.stemDirection === noteL.stemDirection ? 0.0 : 0.5;
       if (noteU.minLine <= noteL.maxLine + lineSpacing) {
         if (noteU.isrest) {
           // shift rest up
@@ -225,7 +227,7 @@ export class StaveNote extends StemmableNote {
               //offset dot(s) above the shifted note
               //lines + 1 to negative pixels
               noteU.note.getModifiersByType(Category.Dot).forEach((dot) => {
-                dot.setYShift(-10 * (noteL.maxLine - noteU.line + 1));
+                (dot as Dot).setDotShiftY(-noteL.maxLine + noteU.line - 1);
               });
             } else if (lineDiff < 1 && lineDiff > 0) {
               //if the notes are quite close but not on the same line, shift
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -10,7 +10,6 @@
 
 import { Beam } from './beam';
 import { BoundingBox } from './boundingbox';
-import { Dot } from './dot';
 import { ElementStyle } from './element';
 import { Modifier } from './modifier';
 import { ModifierContextState } from './modifiercontext';
@@ -224,11 +223,6 @@ export class StaveNote extends StemmableNote {
                 // shift lower voice right
                 noteL.note.setXShift(xShift);
               }
-              //offset dot(s) above the shifted note
-              //lines + 1 to negative pixels
-              noteU.note.getModifiersByType(Category.Dot).forEach((dot) => {
-                (dot as Dot).setDotShiftY(-noteL.maxLine + noteU.line - 1);
-              });
             } else if (lineDiff < 1 && lineDiff > 0) {
               //if the notes are quite close but not on the same line, shift
               xShift = voiceXShift + 2;
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -190,21 +190,29 @@ export class StaveNote extends StemmableNote {
         } else if (noteL.isrest) {
           // shift rest down
           shiftRestVertical(noteL, noteU, -1);
-        } else if (noteU.stemDirection === noteL.stemDirection) {
-          // upper voice is middle voice, so shift it right
-          xShift = voiceXShift + 2;
-          noteU.note.setXShift(xShift);
         } else {
-          //Vexflowpatch: Instead of shifting notes, remove the appropriate flag
+          //Instead of shifting notes, remove the appropriate flag
           //If we are sharing a line, switch one notes stem direction.
           //If we are sharing a line and in the same voice, only then offset one note
           const lineDiff = Math.abs(noteU.line - noteL.line);
           if (noteU.note.hasStem() && noteL.note.hasStem()) {
             //If we have different dot values, must offset
-            //Or If we have a non-filled in mixed with a filled in notehead, must offset
+            //Or If we have a white mixed with a black notehead, must offset
+            let whiteNoteHeadCount = 0;
+            let blackNoteHeadCount = 0;
+            if (Tables.durationToNumber(noteU.note.duration) === 2) {
+              whiteNoteHeadCount++;
+            } else if (Tables.durationToNumber(noteU.note.duration) > 2) {
+              blackNoteHeadCount++;
+            }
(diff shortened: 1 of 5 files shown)


sschmidTU:
Strange that there's a shift in the first diff, that shouldn't happen. I'm guessing these notes are all in the same voice, which doesn't make sense musically. I'll try to take a look.

The intended use case is of course something like in https://github.com/0xfe/vexflow/issues/1257#issuecomment-996239589:
image

The last 2 diffs look like an improvement.

src/stavenote.ts
```diff
@@ -223,11 +223,63 @@ export class StaveNote extends StemmableNote {
...
+                } else {
+                  //same line, swap stem direction for one note
+                  if (noteL.stemDirection === 1) {
+                    noteL.stemDirection = -1;
+                    noteL.note.setStemDirection(-1);
+                  }
+                }
+              }
+            } //Very close whole notes
+          } else if (!noteU.note.glyph.stem && !noteL.note.glyph.stem && lineDiff < 1.5) {
```
rvilarl, line 282: @0xfe @sschmidTU I was trying to debug these lines in order to resolve the issue and I have realised that they do not get ever run. The function returns on line 151. There is a big FIXME there. It was already in 1.2.93.
Any ideas?

sschmidTU: Not sure what's happening here as I haven't debugged this in Vexflow recently and I don't know what's happening in line 151, but I'm pretty sure these lines were running withour patch in 1.2.93, because the whole patch did what it was supposed to do (see image above).

rvilarl: the problem is now resolved in #1280

rvilarl:
I have built on #1280 it would make sense to merge #1280 in advance. 

The visual difference (only two using #1280 as reference) are the ones below (bravura versions):

**Dot Multi_Voice Bravura**
Dot Multi_Voice Bravura
**Current**
**Reference**
**Three_Voice_Rests Three_Voices____2_Complex Bravura**
Three_Voice_Rests Three_Voices____2_Complex Bravura
**Current**
Three_Voice_Rests Three_Voices____2_Complex Bravura_Current
**Reference**
Three_Voice_Rests Three_Voices____2_Complex Bravura_Reference

rvilarl:
Ready for review :)

rvilarl:

rvilarl:
I will wait until #1280 is merged to make the rebase in this one.

sschmidTU:
I made some corrections to the vexflowpatch on OSMD's side:
https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/dc04dc509eb480ee4b9c4f22f62f63d9942ca18d/src/VexFlowPatch/src/stavenote.js
Also see https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/1098#issuecomment-1007437570.

Now some notes that incorrectly didn't x-shift do x-shift, for example:
Notes not staggering OSMD
fixed:
image

@rvilarl Can you check that these work correctly?

image
X-Shift shouldn't be added to notes that are far enough apart, without a stem overlap:
image
image
a more complex test case:
image

If you need help with modification of the vexflowpatch or something, I can try to help porting it.

rvilarl:
@sschmidTU I believe that we should generate a number of tests for this feature. VexFlow tests.
Some of the ones above do not comply with the time signature of the measure and we get the same duration in rests artificially. I did modify your patch to avoid shared noteheads of different duration but I left out any sharing of rests. I will try to figure out some tests similar to ones above and of course I will be happy to get your help to fix the results.

rvilarl:
@sschmidTU @0xfe I have added an additional test based on the examples from Simon. This is the result:

StaveNote Note_Heads_Placement___Simple Bravura

I suggest to merge this PR and let Simon rework the last patch. Is that OK with you?

rvilarl:
@0xfe Rebased and ready to merge. Visual changes only the ones above.

0xfe:
Thanks Rodrigo -- I'm trying to understand this patch -- I think it needs more tests with a few more corner cases. E.g., mixed up and down-stem notes with half/quarter heads.

Also, not entirely clear how the `renderFlag` changes apply here.

src/stemmablenote.ts
```diff
@@ -243,7 +245,7 @@ export abstract class StemmableNote extends Note {
   }
 
   hasFlag(): boolean {
-    return Tables.getGlyphProps(this.duration).flag && !this.beam;
+    return this.renderFlag && Tables.getGlyphProps(this.duration).flag && !this.beam;
```
0xfe, line 248: `renderFlag` muddies the logic here -- I prefer to have something more specific (like the `!this.beam` in this line, so it's clear that if there's a beam, there's no flag.)

src/stavenote.ts
```diff
@@ -197,9 +197,85 @@ export class StaveNote extends StemmableNote {
...
+            ) {
+              xShift = voiceXShift + 2;
+              if (noteU.stemDirection === noteL.stemDirection) {
+                // upper voice is middle voice, so shift it right
+                noteU.note.setXShift(xShift);
+              } else {
+                // shift lower voice right
+                noteL.note.setXShift(xShift);
+              }
+              if (noteU.note.getModifiersByType('Dot').length > 0) {
```
0xfe, line 219: I think we can wrap this logic into the `Dot` class, so it continues to know where it is. I'd rather call `dot.shiftY()` maybe called by `note.shiftDotsUp(...)` or something.

rvilarl:
Hi Mohit, this was the vexflow patch to share note heads that you did like. It was quite complex to integrate and I did it in two steps #1280 and now this one.
@sschmidTU volunteered to further refine this one once it is merged.

src/stavenote.ts
```diff
@@ -1086,9 +1162,11 @@ export class StaveNote extends StemmableNote {
       this.stem.adjustHeightForFlag();
     }
 
-    ctx.openGroup('stem', undefined, { pointerBBox: true });
-    this.stem?.setContext(ctx).draw();
-    ctx.closeGroup();
+    if (this.stem) {
```
0xfe, line 1165: This looks like a separate issue, right?

0xfe:
Just FYI, overall, this is a good change -- I think if we add a few more tests, and wrap up the dot shift logic, this is good to go.

src/stemmablenote.ts
rvilarl: `renderFlag` is used to avoid the rendering of the flag in shared note heads. Would you prefer something like `sharedNoteHead` and use it directly in `draw`?

src/stavenote.ts
rvilarl: This is to avoid to open the group is there is no stem.

src/stemmablenote.ts
rvilarl: two notes with different pitches from different voices that get to share the stem and flag as a chord. This is coded in the VexFlow patch, I believe.

rvilarl: I have decided to remove this feature.

src/stavenote.ts
rvilarl: I have simplified the code a lot. I have also resolved a bug in `note`. The logic cannot be done within `Dot` because the lines of noteU and noteL are required.

rvilarl:
With the changes from the review, the new test is now sharing the notehead correcly.
StaveNote Note_Heads_Placement___Simple Bravura

rvilarl:
Ready for review.

0xfe:
This looks a lot better, Rodrigo. Any other visual diffs?

rvilarl:
No, only the ones already sent.

0xfe:
Thanks Rodrigo -- can you rebase, then run the tests again please? I want to be sure this doesn't interfere with the commits from the recent merges.

rvilarl:
Good catch! I have to look what happened in the last one! Current version of visual diferences (Bravura) after rebase:

pptr-Dot Multi_Voice Bravura svg_current
pptr-Three_Voice_Rests Three_Voices____2_Complex Bravura svg_current
pptr-StaveNote Note_Heads_Placement___Simple Bravura svg

0xfe:
Thanks Rodrigo. This looks good.

Issue #409, opened 2016-08-01
ModifierContext is too complex. Replace with a simpler formatter.


Silverwolf90:
A `ModifierContext` has a `state` that has 4 properties, each which maps to a `Modifier.Position`.

| `state` property | `Modifier.Position` | Unit |
| --- | --- | --- |
| `top_text_line` | `ABOVE` | staff space |
| `text_line` | `BELOW` | staff space |
| `left_shift` | `LEFT` | pixels |
| `right_shift` | `RIGHT` | pixels |

_Sidenotes:
`text_line` should be renamed to `bottom_text_line`. 
Having different units for horizontal/vertical formatting is not ideal. But we can't escape that for now._

All these properties do is keep track of an incremented value in a specific direction. Isn't this already what the `Formatter` does? It advances a position forward along the x axis (albeit with complex logic to determine the amount...). It feels like the `ModifierContext` duplicates and conflates responsibilities since it tries to manage 4 positions at once. So what if we deprecated the `ModifierContext`, and replaced it with with a new, simpler formatter that takes `Modifier.Position` and advances a position given:
- a `Modifier.Position`
- Objects for formatting
- Config options, ie: spacing

To replace the `ModifierContext`, I think we'd need 4 differently parameterized versions of these (one for each `Modifier.Position`).

In fact, in order to format `GraceNotes` within a `GraceNoteGroup` we use the existing `Formatter` and set its available width to 0 that the minimum width for each note is applies. But this is clearly a hack based on the absence of a simpler formatter.

0xfe:
Yeah, the state units need to be rationalized for sure, but even if you use a formatter, I think you'll still need a context to keep track of the modifiers across voices. The `ModiferContext` does not do any formatting, it's just a container for all the modifiers in a tick (i.e., across voices), and it holds some tracking state for the independent formatters. (It also kickstarts the formatting process, but that's just for ordering, which needs to be global.)

And since each category of modifier (e.g., accidentals, fingering, articulations) have their own distinct formatting logic you'll need a way for them to both format themselves, and juxtapose alongside other modifiers.

I think a simpler (and necessary) cleanup would be to replace the various state properties with a position object that uses the same unit across all axes.

0xfe:
Actually, thinking about this more, I think the units are pretty sane (staff spacing for vertical, and pixels for horizontal.) It actually makes the code more natural to reason about.

Silverwolf90:
[...]

PR #1270, opened 2021-12-23
fix #1265: Fix issues with vertical placement of annotations

### LINKED ISSUE #1265: Problems with Annotation placement
I found this issue while investigating vertical formatting, https://github.com/0xfe/vexflow/issues/411.  There are several problems in this issue so I am trying to address them one at a  time.  This is about the annotations, which do some crazy stuff.  TL;DR: annotation Y placement is incorrect and will take some mildly breaking changes to fix.

Here is an example of some placements as the current code does them.  The text 'topa' means 'top justification, above position' - see below for what that means.

image

1. There are 2 competing parameters, one used during render and one used during formatting.  For formatting, we only consider `ModifierPosition`, either `ABOVE`, or...not above, which is treated like below even though there are horizontal values.  Based on above or below, it acts like other modifiers and calculates its line position and reports to the formatter how much line space it takes up.  Aside from the issue of the parameter, the calculation seems to be almost correct.
2. In the draw method, the position parameter is completely ignored in favor of the `VerticalJustify` parameter.  It uses the line value calculated in the formatter based on position, but then applies it to a direction based on the vertical justify.  This could lead to many combinations of things happening depending on how justification or placement are mixed.
3. The factory Annotation constructor has default of BOTTOM for the vertical justification. The annotation constructor used by itself has a default of TOP, but a position default of 'LEFT', which is treated like BOTTOM!  This is clearly not right.
4. The rendered y position is calculated differently depending on whether vertical justification is bottom or top.  I think this is to take into account the fact that text Y is aligned to the baseline, but it does not consider the correct text Y based on the font.  You can see this issue below.  It also does something different depending on the stem direction on the bottom, but not on the top, not quite sure what's up with that.
5. The formatter calculates Y positions correctly, but for some reason adds different amounts of padding for above vs. below.

Based on all this, I think we can do the following:
1. we should consider `VerticalJustification` and not `ModifierPosition` when calculating Y for both rendering and drawing.  Only one parameter should specify Y placement, and since it could be a breaking change either way, I think it makes sense to keep the annotation-specific parameter.  Even though it's not really 'justification' but placement (vertical justification would justify horizontal text on a vertical axis, which is not what we're doing here).  I guess the intent is justifying relative to the note.
2. If Y justification is set as CENTER or CENTER STEM, we shouldn't report any vertical line change in the note to the formatter.  
3. For 'above' or 'below', we should use the same text height value when rendering as we do when formatting, and the padding should be the same, just in a different direction.

AaronDavidNewman:
fix #1265

Fix some issues with vertical placement and different modifiers, this time focusing on annotation.  Also  fixed some issues found 
in my last PR with articulation placement.  Because of an error in sign, a code leg would never have been traversed.  I added a test case for that and fixed the error.

before:
image

after:
image

Note there are some tiny changes in other test cases, due to the fact we are calculating text height rather than hard-coded estimate.  The text here is slightly closer to the note/stave in the new version, for instance.
before:
image

after:
image


rvilarl:
@AaronDavidNewman there changes specially wit Tabs that I woul not expect.

**Annotation TabNote_Annotations**
Annotation TabNote_Annotations Bravura
Annotation TabNote_Annotations Bravura_Current
Annotation TabNote_Annotations Bravura_Reference

**Annotation Bottom_Annotations_with_Beams**
Annotation Bottom_Annotations_with_Beams Bravura
Annotation Bottom_Annotations_with_Beams Bravura_Current
Annotation Bottom_Annotations_with_Beams Bravura_Reference

**Annotation Standard_Notation_Annotation**
Annotation Standard_Notation_Annotation Bravura
Annotation Standard_Notation_Annotation Bravura_Current
Annotation Standard_Notation_Annotation Bravura_Reference

**Annotation Test_Justification_Annotation_Stem_Down**
Annotation Test_Justification_Annotation_Stem_Down Bravura
Annotation Test_Justification_Annotation_Stem_Down Bravura_Current
Annotation Test_Justification_Annotation_Stem_Down Bravura_Reference

**Annotation Test_Justification_Annotation_Stem_Up**
Annotation Test_Justification_Annotation_Stem_Up Bravura
Annotation Test_Justification_Annotation_Stem_Up Bravura_Current
Annotation Test_Justification_Annotation_Stem_Up Bravura_Reference

**Articulation Articulation___Vertical_Placement**
Articulation Articulation___Vertical_Placement Bravura
Articulation Articulation___Vertical_Placement Bravura_Current
Articulation Articulation___Vertical_Placement Bravura_Reference

**Rhythm Rhythm_Draw___beamed_slash_notes__some_rests**
Rhythm Rhythm_Draw___beamed_slash_notes__some_rests Bravura
Rhythm Rhythm_Draw___beamed_slash_notes__some_rests Bravura_Current
Rhythm Rhythm_Draw___beamed_slash_notes__some_rests Bravura_Reference

**StaveNote Center_Aligned_Note_with_Annotation**
StaveNote Center_Aligned_Note_with_Annotation Bravura
StaveNote Center_Aligned_Note_with_Annotation Bravura_Current
StaveNote Center_Aligned_Note_with_Annotation Bravura_Reference

**Style Basic_Style**
Style Basic_Style Bravura
Style Basic_Style Bravura_Current
Style Basic_Style Bravura_Reference

**Annotation Harmonics**
Annotation Harmonics Bravura
Annotation Harmonics Bravura_Current
Annotation Harmonics Bravura_Reference

**Annotation Fingerpicking**
Annotation Fingerpicking Bravura
Annotation Fingerpicking Bravura_Current
Annotation Fingerpicking Bravura_Reference

**Annotation Bottom_Annotation**
Annotation Bottom_Annotation Bravura
Annotation Bottom_Annotation Bravura_Current
Annotation Bottom_Annotation Bravura_Reference

**Annotation Simple_Annotation**
Annotation Simple_Annotation Bravura
Annotation Simple_Annotation Bravura_Current
Annotation Simple_Annotation Bravura_Reference

**Style TabNote_modifiers_Style**
Style TabNote_modifiers_Style Bravura
Style TabNote_modifiers_Style Bravura_Current
Style TabNote_modifiers_Style Bravura_Reference

**TabTie Tapping**
TabTie Tapping Bravura
TabTie Tapping Bravura_Current
TabTie Tapping Bravura_Reference

src/annotation.ts
```diff
@@ -149,23 +186,20 @@ export class Annotation extends Modifier {
...
     ctx.setFont(this.textFont);
 
     const text_width = ctx.measureText(this.text).width;
-
-    // Estimate text height to be the same as the width of an 'm'.
-    //
-    // This is a hack to work around the inability to measure text height
-    // in HTML5 Canvas (and SVG).
-    const text_height = ctx.measureText('m').width;
+    const text_height = textFormatter.maxHeight + 2;
```
rvilarl, line 202: Why not `ctx.measureText(this.text).height` rather than `ctx.measureText(this.text).width`?
Please note than in #1262 I am trying to fix the implementation of `ctx.measureText(this.text).height`

AaronDavidNewman: You usually want max height and actual width with text, for placement.  Because text has a baseline.  you might want the actual height in some cases, like if you wanted to draw a glyph or a box around it.

AaronDavidNewman: Not sure what's up with the tabs.  It looks like there was some behavior specific to tab annotations.  I'll check on it.

rvilarl:
@AaronDavidNewman do you know how to get the image differences above? I documented it in the WIKI 
https://github.com/0xfe/vexflow/wiki/Visual-Regression-Tests

"Alternatively, you can test against a particular commit (i.e.: master or any other commit). Run the following commands to generate the reference:
```
$ git checkout master
$ npm run reference
```
Then run the following command in your branch:
```
$ npm run test:reference
```

To get the issue linked you have to place the `fix #1265` in the first comment. For example one line at the beginning befor "Fix some issues with vertical placement and different modifiers..." If it is correctly linked you will get it in the **Linked issues** list to the right.

src/annotation.ts
rvilarl: > You usually want max height and actual width with text, for placement. Because text has a baseline. you might want the actual height in some cases, like if you wanted to draw a glyph or a box around it.

What do you think that we should do here? Should `ctx.measureText(this.text).height` return the actual height of `this.text` or `textFormatter.maxHeight`

AaronDavidNewman: >> ctx.measureText(this.text).height return the actual height 

I think so.  Probably the most useful thing would be to return a structure with width, height, maxHeight, and possibly advanceWidth.  90+% of the time, people only care about the width, anyway.

AaronDavidNewman:
> @AaronDavidNewman do you know how to get the image differences above

Does this work on a (windows) PC?  I'm thinking probably not.  Really excited about the headless chrome rendering.

rvilarl:
> > @AaronDavidNewman do you know how to get the image differences above
> 
> Does this work on a (windows) PC? I'm thinking probably not. Really excited about the headless chrome rendering.

What error do you get following the procedure described in the wiki?

0xfe:
Thanks folks -- @rvilarl from the diffs you pasted in, there's two things that stand out:

- Some of the annotations are too far from the top of the stave lines
- Vertically centered annotations are broken

Overall this is a good change, and I wonder if we should fix those issues inside font-specific metrics (maybe by adding some new shifts just for those cases.)

The alternative is to dive right into the whacky world of text font metrics, but I'd really like to avoid that. Instead if we can say some thing like "Vexflow works best with text fonts X, Y, and Z", and optimize for those cases, that would make our lives a lot simpler. Thoughts?

AaronDavidNewman:
> Some of the annotations are too far from the top of the stave

Mohit are you confusing before and after?  The current annotations actually cling a little tighter to the top of the staff than they previously did.

> Vertically centered annotations are broken

Indeed, but I didn't change those :)  Those should be pretty much the same.

The only thing this PR broke (that I've seen so far) is the tab annotations.  Tab notes seem to have a stem, but the stem is never rendered.  I have a fix for that.  Also annotations over rests were taking the stem seriously.  So I just ignore stem in those 2 cases (stem.isRendered only tells you if it's already been rendered, doesn't help for formatting).

> What error do you get following the procedure described in the wiki?

I am stuck on the dependencies (RSVG and ImageMagick for windows 10).  I'm sure it's possible but it seems hard.  I'll try to figure something out, maybe I can borrow a mac from somebody.

rvilarl:
I will give it a try on Windows later today🙂

h-sug1no:
I'm running vrtest on a w10 desktop PC using (vmware (workstation player)+ubuntu).

It would be easier to use linux in a virtual environment like wsl2, virtualbox, or vmware...

sschmidTU:
If I need a Linux environment on Windows, I use VirtualBox (which is free compared to VMWare) + Linux Mint XFCE.
WSL2 works, but is a bit annoying to set up, and WSL2 broke my rsync compared to WSL1 (while WSL1 can't do the visual regression bash script).
My VirtualBox Linux is quite slow compared to native, not sure why with virtualization support, but it works.

AaronDavidNewman:
I think I have all the tab cases worked out now, except for `Bend`.  Here, I think the issue is with `Bend`.  Modifiers treat `top_text_line` or `text_line` inconsistently.  Usually it is treated with regard  to the note, but for modifiers that are always on top/bottom, the `Bend.format` method assumes the text line is relative to the staff.  This causes issues when modifiers are mixed.

I think the solution is to implement the same logic in bend - always make the staff line relative to the highest/lowest `TabNote` depending on whether the placement is on top or bottom.

image

rvilarl:
@AaronDavidNewman in #409 there is some discussion about `top_text_line` and `text_line`

AaronDavidNewman:
I have the bend now considering other modifiers.  There is one test case that is still different, but I think the change is correct:.  This is how it used to look

image

The annotation and bend text are at the same level.  If the annotation were longer, they would collide:
image

I'm not quite sure why you'd want to have a separate annotation instead of just adding the 'T" to the bend text.  The bend text is now stacked on top of the bend (and any other modifiers):

image

rvilarl:
@AaronDavidNewman these two test have still to be looked at (first seems to be ok, just preesenting the complete stave):

**Annotation Fingerpicking**
Annotation Fingerpicking Bravura
**Current**
Annotation Fingerpicking Bravura_Current
**Reference**
Annotation Fingerpicking Bravura_Reference

**Articulation Articulation___Vertical_Placement**
Articulation Articulation___Vertical_Placement Bravura
**Current**
Articulation Articulation___Vertical_Placement Bravura_Current
**Reference**
Articulation Articulation___Vertical_Placement Bravura_Reference

src/annotation.ts
rvilarl: In the end I standarised  in #1262 `measureText` to the svg behaviour. I suggest to change:
```
 const text_width = ctx.measureText(this.text).width;
 const text_height = textFormatter.maxHeight + 2;
```
to
```
const textMeasure = ctx.measureText(this.text);
const text_width = textMeasure.width;
const text_height = textMeasure.height + 2;
```

AaronDavidNewman:
>> these two test have still to be looked at

Right, these are OK.  In both cases I changed the test case to test more things, or fix the broken width.  It is kind of annoying that the 'p' is a little below the line, though.

>> In the end I standarised in #1262 measureText

Is this for a future change?  b/c the current svgcontext is still using `svg.getBbox()` to get the dimensions, rather than the text formatter.

There will be other PRs for the vertical formatting change, so we can change this code in one of those to use SVGContext if that is the preferred way to get text dimensions.

rvilarl:
> > > In the end I standarised in #1262 measureText
> 
> Is this for a future change? b/c the current svgcontext is still using `svg.getBbox()` to get the dimensions, rather than the text formatter.

Sorry you are right #1262 is not merged, I saw the congratulations from Simon and I assumed it was merged. @0xfe are you happy with the approach in #1262 for unifying the measurement of texts in CanvasContext and SvgContext. If #1262 is going to be merged I would recommend the change described above.

AaronDavidNewman:
This is ready for review

AaronDavidNewman:
This is merged and up to date.  Diffs look basically the same.

0xfe:
Awesome stuff Aaron -- made a minor edit to fix a lint error. Merging!

0xfe:
Just looking at guitar bend thing again, it seems like it is inconsistent with how tabs are generally rendered. The `T` (tap) and `Full` (bend) should really be on the same Y if there's no collision.

Can we add some logic not to shift the Y if there's no collision? (or to override the shifting in some situations?)

AaronDavidNewman:
Yeah, bend definitely came out a little worse :(

I can't think of an easy way to check for collisions.  Ideally, the modifier context 'state' would be an array of bounding boxes.  I think this is what @mscuthbert  was alluding to above about Finale.  That is a _lot_ of rewriting, including things that currently work pretty well like accidentals.  So that's the dream.

I can think of 2 possible short-term solutions.  One is to simply switch the order of annotation and bend/vibrato in modifiercontext.   That fixes it, and I don't think it's so awful because many modifiers don't really apply to tab notes, especially if there is a Bend.

The second is to allow bend to 'opt out' of vertical formatting, or allow annotation to 'opt out' of reporting its vertical placement.  In this case the application would be responsible for avoiding collisions by moving the bounding box of the annotation.  We have similar logic with `ChordSymbol`, because sometimes it's OK for a chord change to overlap other notes.

PR #1272, opened 2021-12-24
Bug on  "npm run test:reference" visual regression.

rvilarl:
The new location in build/cjs broke the tests against a reference.

This is a minor change to take into account the location of the files.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -258,7 +258,7 @@ module.exports = (grunt) => {
           {
             expand: true,
             cwd: BUILD_DIR,
-            src: ['*.js', '*.map'],
+            src: ['**/*.js', '**/*.map'],
             dest: REFERENCE_DIR,
           },
         ],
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -258,7 +258,7 @@ module.exports = (grunt) => {
           {
             expand: true,
             cwd: BUILD_DIR,
-            src: ['*.js', '*.map'],
+            src: ['**/*.js', '**/*.map'],
```
0xfe, line 261: Thanks Rodrigo. Can we explicitly add subfolders here?

rvilarl: done

0xfe:
Looks good. Thanks!

ronyeh:
@rvilarl Thanks for catching this! I think I ran tests where yesterday's master was the reference, and my branch was the current. But I didn't consider that my branch would now be the new reference. :-| :-)

PR #1275, opened 2021-12-27
Fix Circular Dependency Issue (using getCategory)

### LINKED ISSUE #1274: 9b12482424dd0fc36846b9182519d3289a5e75bc: "qunit:files" (qunit) task failed.
```
git log --merges

commit 9b12482424dd0fc36846b9182519d3289a5e75bc (HEAD, origin/master, origin/HEAD, master)
Merge: a2570076 c3b67d4b
Author: Mohit Cheppudira <0xfe@users.noreply.github.com>
Date:   Sun Dec 26 07:32:30 2021 -0500

    Merge pull request #1260 from rvilarl/easyscoreMutedHarmonySlashGhost2
    
    Easyscore support for muted, harmony, slash and ghost notes II (ghost)

commit a2570076497367ccba8ddc8a9ccdb299a882c754
Merge: 8cd3b7ea 0a8eec23
Author: Mohit Cheppudira <0xfe@users.noreply.github.com>
Date:   Sun Dec 26 07:17:12 2021 -0500

    Merge pull request #1268 from h-sug1no/vrtest-puppeteer-backend-2
    
    generate_images: add puppeteer(pptr) backend
```
---

  git checkout  .
  git checkout a2570076497367ccba8ddc8a9ccdb299a882c754
  git clean -dfx && npm install && npm run lint && npm run qunit

OK: as follows:
```
...
Running "qunit:files" (qunit) task
Testing tests/flow-headless-browser.html
...
.........................................................................................................................................................................................................................................................................................................................OK
>> 1615 tests completed with 0 failed, 0 skipped, and 0 todo. 
>> 7194 assertions (in 16930ms), passed: 7194, failed: 0

Done.
```

---

  git checkout  .
  git checkout  9b12482424dd0fc36846b9182519d3289a5e75bc
  git clean -dfx && npm install && npm run lint && npm run qunit

NG: as follows:
```
...
Running "qunit:files" (qunit) task
Testing tests/flow-headless-browser.html FF
>> global failure
>> Message: Script error.
>> Actual: null
>> Expected: undefined
>> :0

>> global failure
>> Message: Uncaught ReferenceError: Vex is not defined
>> Actual: null
>> Expected: undefined
>> file:///mnt/sdb/users/sug1no/work3/github/h-sug1no/vexflow.d/vexflow/tests/flow-headless-browser.html:12

Warning: 2 tests completed with 2 failed, 0 skipped, and 0 todo. 
2 assertions (in 8ms), passed: 0, failed: 2 Use --force to continue.

Aborted due to warnings.
```

---

load file:///mnt/sdb/users/sug1no/work3/github/h-sug1no/vexflow.d/vexflow/tests/flow-headless-browser.html with chrome, the following errors are shown in console:

```js
﻿vexflow-debug-with-tests.js:19273 Uncaught ReferenceError: Cannot access 'Note' before initialization
    at Module.Note (vexflow-debug-with-tests.js:19273)
    at Object../src/stemmablenote.ts (vexflow-debug-with-tests.js:24957)
    at __webpack_require__ (vexflow-debug-with-tests.js:48373)
    at Object../src/tabnote.ts (vexflow-debug-with-tests.js:27719)
    at __webpack_require__ (vexflow-debug-with-tests.js:48373)
    at Object../src/dot.ts (vexflow-debug-with-tests.js:4161)
    at __webpack_require__ (vexflow-debug-with-tests.js:48373)
    at Object../src/note.ts (vexflow-debug-with-tests.js:19276)
    at __webpack_require__ (vexflow-debug-with-tests.js:48373)
    at Object../src/notehead.ts (vexflow-debug-with-tests.js:19815)
flow-headless-browser.html:12 Uncaught ReferenceError: Vex is not defined
    at flow-headless-browser.html:12
﻿
```
@rvilarl san:
@0xfe san:

dependency problem?

rvilarl:
fixes #1274

I figured out wich circular dependency was producing the crash. This PR fixes this particular one.

I have learned that:
- methods like `note.isTabNote()` contribute to circular dependencies
- `note instanceof TabNote` produces the same error (also contributes to circular dependencies`

We should probably avoid the methods above.

### DIFF
--- a/src/dot.ts
+++ b/src/dot.ts
@@ -8,7 +8,6 @@ import { Modifier } from './modifier';
 import { ModifierContextState } from './modifiercontext';
 import { Note } from './note';
 import { isStaveNote } from './stavenote';
-import { isTabNote } from './tabnote';
 import { isCategory } from './typeguard';
 import { RuntimeError } from './util';
 
@@ -42,7 +41,7 @@ export class Dot extends Modifier {
         const index = dot.checkIndex();
         props = note.getKeyProps()[index];
         shift = note.getRightDisplacedHeadPx();
-      } else if (isTabNote(note)) {
+      } else if (note.getCategory() === 'TabNote') {
         props = { line: 0.5 }; // Shim key props for dot placement
         shift = 0;
       } else {
@@ -145,7 +144,7 @@ export class Dot extends Modifier {
     const start = note.getModifierStartXY(this.position, this.index, { forceFlagRight: true });
 
     // Set the starting y coordinate to the base of the stem for TabNotes.
-    if (isTabNote(note)) {
+    if (note.getCategory() === 'TabNote') {
       start.y = note.getStemExtents().baseY;
     }
 


ronyeh:
OK, if this fixes the build, I'll support the patch **for now**. But this patch is definitely just a band-aid, because we are reverting to checking the category against a hard coded string, instead of using the isCategory() method (which internally uses instanceof and adds to the dependency chain).

This circular dependency is likely revealing a problem with our architecture. For example, does the **Dot** class really need to check whether its note is a `StaveNote` or a `TabNote`? By checking the type of the note, and doing something slightly different for each case, we are adding a dependency on both the `StaveNote` and `TabNote` classes. This definitely will contribute to circular dependency bugs in the future.

For example:
```
  if (isStaveNote(note)) {
    const index = dot.checkIndex();
    props = note.getKeyProps()[index];
    shift = note.getRightDisplacedHeadPx();
  } else if (isTabNote(note)) {
    props = { line: 0.5 }; // Shim key props for dot placement
    shift = 0;
```

To break the dependency chain, StemmableNote (the ancestor of StaveNote and TabNote) probably needs a method for returning the props / shift for dots. Then, each subclass can implement the method differently.

This way, the Dot class doesn't need to check whether the note object is a StaveNote or TabNote. It breaks the dependency chain.


So maybe we need to check all the places we use these isXXX() typeguards. Maybe using instanceof and these typeguards is just a indicator of potential areas for improvement.

ronyeh:
Also, I'm sort of curious why this problem came up now. :-) The isCategory() type guards have been there for a while, and replaced the previous instanceof checks, so we've always had these circular dependencies.

I wonder what else changed recently that led to this circular dependency causing a problem.

rvilarl:
> So maybe we need to check all the places we use these isXXX() typeguards. Maybe using instanceof and these typeguards is just a indicator of potential areas for improvement.

I agree that both indicate potential areas of improvement. And you have proven in #1276 that the dependency can be eliminated with a refactoring. However I would vote to go for the checking of category with a hardcoded string in order to reduce the dependencies.

I would even suggest to change the category String to an array that would contain the class name and the name of all the ancestor classes.

rvilarl:
> Also, I'm sort of curious why this problem came up now. :-) The isCategory() type guards have been there for a while, and replaced the previous instanceof checks, so we've always had these circular dependencies.
> 
> I wonder what else changed recently that led to this circular dependency causing a problem.

The big difference was moving the modifier related methods from `StaveNote` to `Note`.

rvilarl:
@0xfe Ready to review.

0xfe:
Thanks Rodrigo. Merging this one because it's simpler.

PR #1276, opened 2021-12-27
Fix Circular Dependency Issue

ronyeh:
Fixes  https://github.com/0xfe/vexflow/issues/1274

Remove typeguard methods from dot.ts by moving custom dot behavior out to stavenote.ts and tabnote.ts.

Add support for ESM in flow.html (via a query param esm=true).
This aids in debugging circular dependencies because each source file is loaded separately. Open the browser dev tools / network tab to see the individual files loading.

Loading ESM in flow.html requires a web server. For example, you can run
`npx http-server`
and then browse to: 
`http://127.0.0.1:8080/tests/flow.html?esm=true`


src/dot.ts
```diff
@@ -35,20 +32,7 @@ export class Dot extends Modifier {
...
-        props = note.getKeyProps()[index];
-        shift = note.getRightDisplacedHeadPx();
-      } else if (isTabNote(note)) {
-        props = { line: 0.5 }; // Shim key props for dot placement
-        shift = 0;
-      } else {
-        // note object is not StaveNote or TabNote.
-        throw new RuntimeError('Internal', 'Unexpected instance.');
-      }
+      const { props, shift } = note.getDotPropsAndShift(dot.checkIndex());
```
ronyeh, line 35: The previous block of code is replaced with this single line. Instead of Dot checking the note instance's type, we just let the note instance do its own thing. StaveNote and TabNote implement this method differently. This removes the dependencies.

tests/flow.html
```diff
@@ -99,8 +102,15 @@ <h3>
...
       let loadVexFlow;
-      if (isVersionFourOrNewer) {
-        // When loading version >= 4.0.0, only load vexflow-debug-with-tests.js
+      if (useESM) {
+        // This ESM branch only works if flow.html is accessed from a web server
+        // (e.g., `npx http-server` and browse to http://127.0.0.1:8080/tests/flow.html).
+        loadVexFlow = async () => {
+          const module = await import('../build/esm/entry/vexflow-debug-with-tests.js');
+          window.Vex = module.Vex;
+        };
```
ronyeh, line 111: Adding support for ESM makes debugging circular dependencies a bit easier. Now, each file is loaded separately via import.

tests/font_tests.ts
```diff
@@ -5,21 +5,17 @@
...
-  Flow,
-  Font,
-  FontStyle,
-  FontWeight,
-  PedalMarking,
-  StaveNote,
-  TextBracket,
-  TextNote,
-  Voice,
-} from '../src';
```
ronyeh, line 22: Avoid importing the src/index.ts if possible. Sometimes this helps reduce circular dependencies.

tests/vexflow_test_helpers.ts
```diff
@@ -5,15 +5,13 @@
...
-/* eslint-disable */
+// eslint-disable-next-line
 declare const $: any;
-declare let global: any;
-if (typeof global === 'undefined') {
-  global = window ?? globalThis ?? this;
-}
-/* eslint-enable */
+
+const global = globalObject();
```
ronyeh, line 14: This fix is required when flow.html loads ES modules.

rvilarl:
@ronyeh can this PR be closed now?

ronyeh:
No, I'm keeping it open to remind myself to fix the typeguards.

Also, this PR contains some improvements to flow.html and fixes to ESM that I will break up into a separate PR.

PR #1279, opened 2021-12-30
remove 8 circular dependencies

rvilarl:
This simple change reduces the number of circular dependencies from 25 to 17 (see below). The change also makes the set and get methods more consistent using `Modifier` in both cases.

```
WARNING in Circular dependency detected:
src/articulation.ts -> src/gracenote.ts -> src/stavenote.ts -> src/notehead.ts -> src/note.ts -> src/accidental.ts -> src/gracenotegroup.ts -> src/beam.ts -> src/tuplet.ts -> src/formatter.ts -> src/modifiercontext.ts -> src/articulation.ts

WARNING in Circular dependency detected:
src/ornament.ts -> src/tabnote.ts -> src/dot.ts -> src/gracenote.ts -> src/stavenote.ts -> src/notehead.ts -> src/note.ts -> src/accidental.ts -> src/gracenotegroup.ts -> src/beam.ts -> src/tuplet.ts -> src/formatter.ts -> src/modifiercontext.ts -> src/ornament.ts

WARNING in Circular dependency detected:
src/stave.ts -> src/stavetext.ts -> src/textnote.ts -> src/note.ts -> src/accidental.ts -> src/gracenotegroup.ts -> src/beam.ts -> src/tuplet.ts -> src/formatter.ts -> src/stave.ts

WARNING in Circular dependency detected:
src/stavetext.ts -> src/textnote.ts -> src/note.ts -> src/accidental.ts -> src/gracenotegroup.ts -> src/beam.ts -> src/tuplet.ts -> src/formatter.ts -> src/stave.ts -> src/stavetext.ts

WARNING in Circular dependency detected:
src/stringnumber.ts -> src/stavenote.ts -> src/notehead.ts -> src/note.ts -> src/accidental.ts -> src/gracenotegroup.ts -> src/beam.ts -> src/tuplet.ts -> src/formatter.ts -> src/modifiercontext.ts -> src/stringnumber.ts

WARNING in Circular dependency detected:
src/strokes.ts -> src/note.ts -> src/accidental.ts -> src/gracenotegroup.ts -> src/beam.ts -> src/tuplet.ts -> src/formatter.ts -> src/modifiercontext.ts -> src/strokes.ts

WARNING in Circular dependency detected:
src/tabnote.ts -> src/dot.ts -> src/gracenote.ts -> src/stavenote.ts -> src/notehead.ts -> src/note.ts -> src/accidental.ts -> src/gracenotegroup.ts -> src/beam.ts -> src/tabnote.ts

WARNING in Circular dependency detected:
src/textnote.ts -> src/note.ts -> src/accidental.ts -> src/gracenotegroup.ts -> src/beam.ts -> src/tuplet.ts -> src/formatter.ts -> src/stave.ts -> src/stavetext.ts -> src/textnote.ts
```

### DIFF
--- a/src/note.ts
+++ b/src/note.ts
@@ -593,14 +593,9 @@ export abstract class Note extends Tickable {
     return this;
   }
 
-  // Get all accidentals in the `ModifierContext`
-  getAccidentals(): Modifier[] {
-    return this.checkModifierContext().getMembers('Accidental') as Modifier[];
-  }
-
-  // Get all dots in the `ModifierContext`
-  getDots(): Modifier[] {
-    return this.checkModifierContext().getMembers('Dot') as Modifier[];
+  // Get all modifiers of a specific type in the `ModifierContext`
+  getModifiersByType(type: string): Modifier[] {
+    return this.checkModifierContext().getMembers(type) as Modifier[];
   }
 
   /** Get the coordinates for where modifiers begin. */
--- a/src/note.ts
+++ b/src/note.ts
@@ -1,7 +1,6 @@
 // [VexFlow](https://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 // MIT License
 
-import { Accidental } from './accidental';
 import { Beam } from './beam';
 import { Dot } from './dot';
 import { Font } from './font';
@@ -595,13 +594,13 @@ export abstract class Note extends Tickable {
   }
 
   // Get all accidentals in the `ModifierContext`
-  getAccidentals(): Accidental[] {
-    return this.checkModifierContext().getMembers(Accidental.CATEGORY) as Accidental[];
+  getAccidentals(): Modifier[] {
+    return this.checkModifierContext().getMembers('Accidental') as Modifier[];
   }
 
   // Get all dots in the `ModifierContext`
-  getDots(): Dot[] {
-    return this.checkModifierContext().getMembers(Dot.CATEGORY) as Dot[];
+  getDots(): Modifier[] {
+    return this.checkModifierContext().getMembers('Dot') as Modifier[];
   }
 
   /** Get the coordinates for where modifiers begin. */
--- a/src/note.ts
+++ b/src/note.ts
@@ -2,7 +2,6 @@
 // MIT License
 
 import { Beam } from './beam';
-import { Dot } from './dot';
 import { Font } from './font';
 import { Fraction } from './fraction';
 import { GlyphProps } from './glyph';
@@ -559,22 +558,6 @@ export abstract class Note extends Tickable {
     modifier.setIndex(index);
     this.modifiers.push(modifier);
     this.preFormatted = false;
-    return this;
-  }
-
-  // Helper function to add a dot on a specific key
-  addDot(index: number): this {
-    const dot = new Dot();
-    dot.setDotShiftY(this.glyph.dot_shiftY);
-    this.dots++;
-    return this.addModifier(index, dot);
-  }
-
-  // Convenience method to add dot to all keys in note
-  addDotToAll(): this {
-    for (let i = 0; i < this.keys.length; ++i) {
-      this.addDot(i);
-    }
     return this;
   }
 
--- a/src/note.ts
+++ b/src/note.ts
@@ -562,21 +562,6 @@ export abstract class Note extends Tickable {
     return this;
   }
 
-  // Helper function to add an accidental to a key
-  addAccidental(index: number, accidental: Modifier): this {
-    return this.addModifier(index, accidental);
-  }
-
-  // Helper function to add an articulation to a key
-  addArticulation(index: number, articulation: Modifier): this {
-    return this.addModifier(index, articulation);
-  }
-
-  // Helper function to add an annotation to a key
-  addAnnotation(index: number, annotation: Modifier): this {
-    return this.addModifier(index, annotation);
-  }
-
   // Helper function to add a dot on a specific key
   addDot(index: number): this {
     const dot = new Dot();
--- a/src/note.ts
+++ b/src/note.ts
@@ -535,7 +535,7 @@ export abstract class Note extends Tickable {
    * @param index of the key to modify.
    * @returns this
    */
-  addModifier(modifier: Modifier, index: number = 0): this {
+  addModifier(index: number, modifier: Modifier): this {
     // Backwards compatibility with 3.0.9.
     if (typeof index === 'string') {
       index = parseInt(index);
@@ -564,17 +564,17 @@ export abstract class Note extends Tickable {
 
   // Helper function to add an accidental to a key
   addAccidental(index: number, accidental: Modifier): this {
-    return this.addModifier(accidental, index);
+    return this.addModifier(index, accidental);
   }
 
   // Helper function to add an articulation to a key
   addArticulation(index: number, articulation: Modifier): this {
-    return this.addModifier(articulation, index);
+    return this.addModifier(index, articulation);
   }
 
   // Helper function to add an annotation to a key
   addAnnotation(index: number, annotation: Modifier): this {
-    return this.addModifier(annotation, index);
+    return this.addModifier(index, annotation);
   }
 
   // Helper function to add a dot on a specific key
@@ -582,7 +582,7 @@ export abstract class Note extends Tickable {
     const dot = new Dot();
     dot.setDotShiftY(this.glyph.dot_shiftY);
     this.dots++;
-    return this.addModifier(dot, index);
+    return this.addModifier(index, dot);
   }
 
   // Convenience method to add dot to all keys in note
--- a/src/note.ts
+++ b/src/note.ts
@@ -239,7 +239,6 @@ export abstract class Note extends Tickable {
     stroke_px: number;
   };
   protected duration: string;
-  protected dots: number;
   protected leftDisplacedHeadPx: number;
   protected rightDisplacedHeadPx: number;
   protected noteType: string;
@@ -274,7 +273,6 @@ export abstract class Note extends Tickable {
     this.keyProps = [];
 
     this.duration = parsedNoteStruct.duration;
-    this.dots = parsedNoteStruct.dots;
     this.noteType = parsedNoteStruct.type;
     this.customTypes = parsedNoteStruct.customTypes;
 
@@ -494,7 +492,7 @@ export abstract class Note extends Tickable {
 
   /** Accessor to isDotted. */
   isDotted(): boolean {
-    return this.dots > 0;
+    return this.getModifiersByType('Dot').length > 0;
   }
 
   /** Accessor to hasStem. */
--- a/src/accidental.ts
+++ b/src/accidental.ts
@@ -411,7 +411,7 @@ export class Accidental extends Modifier {
       const modifiedPitches: string[] = [];
 
       const processNote = (t: Tickable) => {
-        // Only StaveNote implements .addAccidental(), which is used below.
+        // Only StaveNote implements .addModifier(), which is used below.
         if (!isStaveNote(t) || t.isRest() || t.shouldIgnoreTicks()) {
           return;
         }
@@ -456,7 +456,7 @@ export class Accidental extends Modifier {
             const accidental = new Accidental(accidentalString);
 
             // Attach the accidental to the StaveNote
-            staveNote.addAccidental(keyIndex, accidental);
+            staveNote.addModifier(keyIndex, accidental);
 
             // Add the pitch to list of pitches that modified accidentals
             modifiedPitches.push(keyString);
--- a/src/articulation.ts
+++ b/src/articulation.ts
@@ -285,7 +285,7 @@ export class Articulation extends Modifier {
         if (position) artic.position = Modifier.PositionString[position];
         return builder.getFactory().Articulation(artic);
       })
-      .map((artic) => note.addModifier(artic, 0));
+      .map((artic) => note.addModifier(0, artic));
   }
 
   /**
--- a/src/dot.ts
+++ b/src/dot.ts
@@ -26,6 +26,27 @@ export class Dot extends Modifier {
     return note.getModifiersByType(Dot.CATEGORY) as Dot[];
   }
 
+  /** Add a dot on the specified keys to the notes. */
+  static buildAndAttach(notes: Note[], options?: { index?: number; all?: boolean }): void {
+    for (const note of notes) {
+      if (options?.all) {
+        for (let i = 0; i < note.keys.length; i++) {
+          const dot = new Dot();
+          dot.setDotShiftY(note.glyph.dot_shiftY);
+          note.addModifier(i, dot);
+        }
+      } else if (options?.index != undefined) {
+        const dot = new Dot();
+        dot.setDotShiftY(note.glyph.dot_shiftY);
+        note.addModifier(options?.index, dot);
+      } else {
+        const dot = new Dot();
+        dot.setDotShiftY(note.glyph.dot_shiftY);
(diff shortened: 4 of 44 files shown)


src/note.ts
```diff
@@ -595,13 +594,13 @@ export abstract class Note extends Tickable {
   }
 
   // Get all accidentals in the `ModifierContext`
-  getAccidentals(): Accidental[] {
-    return this.checkModifierContext().getMembers(Accidental.CATEGORY) as Accidental[];
+  getAccidentals(): Modifier[] {
```
0xfe, line 597: Thanks Rodrigo. This changes the semantics enough not to make these helpers useful. But that said, it seems like there's a bunch of options here:

- Remove it entirely, or,
- Replace this with a generic helper `getModifiers(type: string) -> Modifier[]`, or,
- longer term, start separating the interfaces and implementations.

rvilarl: And what option do you prefer in the first instance?

0xfe: I prefer the second one, and moving `getAccidentals(note: Note): Accidental[]` to a static method in `accidentals.ts`.

rvilarl: Done for `getAccidentals` and `getDots`.
@0xfe two questions related to the logic above:
- Do `addAccidental`, `addAnnotation` and `addArticulartion` make sense as helpers?
- Should `addDot` and `addDotToAll` be static funcstions of `Dot`?

0xfe: 1) I think we should remove them from here, and move them into their respective classes.
2) Yes, agreed. (same as 1)

I think we can keep these APIs consistent (including the new one for `Parenthesis`.) I prefer the calls like:

```
Parenthesis.buildAndAttach(note)
Dot.buildAndAttach(note, index)
Accidental.buildAndAttach(note, index)
...
```

rvilarl: there are two issues: 
- the parameters order of `addAccidental`, `addAnnotation` and `addArticulation` differ from the one in `addModifier` :( I would sugeest to leave untouched perhaps with a deprecated statement.
- while `Parenthesis.BuildAndAttach` and `Dot.buildAndAttach` are fine, `Accidenta.buildAndAttached` would require an additional parameter `type` used in the constructor. I would suggest just to do it for `Dot` and `Parenthesis`
- what about `addDotToAll`? I suggest `Dot.buildAndAttachToAll` this would make sense for Parenthesis as well.

rvilarl: Looking into the dots details, I have realised that we will have to change code like (example from tests)
```
const notes = [
    staveNote({ keys: ['f/4'], duration: '4', stem_direction: Stem.UP }).addDotToAll(),
    staveNote({ keys: ['f/4'], duration: '8', stem_direction: Stem.UP }).addDotToAll(),
    staveNote({ keys: ['f/4'], duration: '16', stem_direction: Stem.UP }).addDotToAll(),
    staveNote({ keys: ['f/4'], duration: '32', stem_direction: Stem.UP }).addDotToAll(),
    staveNote({ keys: ['f/4'], duration: '64', stem_direction: Stem.UP }).addDotToAll(),
    staveNote({ keys: ['f/4'], duration: '128', stem_direction: Stem.UP })
      .addDotToAll()
      .addDotToAll(),
    staveNote({ keys: ['g/4'], duration: '4', stem_direction: Stem.UP }).addDotToAll(),
    staveNote({ keys: ['g/4'], duration: '8', stem_direction: Stem.UP }).addDotToAll(),
    staveNote({ keys: ['g/4'], duration: '16', stem_direction: Stem.UP }).addDotToAll(),
    staveNote({ keys: ['g/4'], duration: '32' }).addDotToAll(),
    staveNote({ keys: ['g/4'], duration: '64', stem_direction: Stem.UP }).addDotToAll(),
    staveNote({ keys: ['g/4'], duration: '128', stem_direction: Stem.UP })
      .addDotToAll()
      .addDotToAll(),
  ];
```
into 
```
const notes = [
    staveNote({ keys: ['f/4'], duration: '4', stem_direction: Stem.UP }),
    staveNote({ keys: ['f/4'], duration: '8', stem_direction: Stem.UP }),
    staveNote({ keys: ['f/4'], duration: '16', stem_direction: Stem.UP }),
    staveNote({ keys: ['f/4'], duration: '32', stem_direction: Stem.UP }),
    staveNote({ keys: ['f/4'], duration: '64', stem_direction: Stem.UP }),
    staveNote({ keys: ['f/4'], duration: '128', stem_direction: Stem.UP }),
    staveNote({ keys: ['g/4'], duration: '4', stem_direction: Stem.UP }),
    staveNote({ keys: ['g/4'], duration: '8', stem_direction: Stem.UP }),
    staveNote({ keys: ['g/4'], duration: '16', stem_direction: Stem.UP }),
    staveNote({ keys: ['g/4'], duration: '32' }),
    staveNote({ keys: ['g/4'], duration: '64', stem_direction: Stem.UP }),
    staveNote({ keys: ['g/4'], duration: '128', stem_direction: Stem.UP }),
  ];
Dot.buildAndAttachToAll(notes[0]);
Dot.buildAndAttachToAll(notes[1]);
Dot.buildAndAttachToAll(notes[2]);
Dot.buildAndAttachToAll(notes[3]);
Dot.buildAndAttachToAll(notes[4]);
Dot.buildAndAttachToAll(notes[5]);
Dot.buildAndAttachToAll(notes[5]);
Dot.buildAndAttachToAll(notes[6]);
Dot.buildAndAttachToAll(notes[7]);
Dot.buildAndAttachToAll(notes[8]);
Dot.buildAndAttachToAll(notes[9]);
Dot.buildAndAttachToAll(notes[10]);
Dot.buildAndAttachToAll(notes[11]);
Dot.buildAndAttachToAll(notes[11]);
```
Do we want to do that? The code is more complex to understand.

rvilarl:
Looking at how this is evolving I would suggest to close the PR without merge.
@0xfe what do you think?

src/note.ts
0xfe: Yeah, I think it actually makes everything cleaner. There's no reason for `StaveNote` to know about `Dot`s, but `Dot`s must know about `StaveNotes`. The user does have to write more code, but the API becomes a bit more consistent, I think. I'd prefer the helpers to come from `Factory`, `System`, etc., and this core API to be sort of "low level."

Also, I think we could remove a lot of circular dependencies with this pattern.

Re: your comments:

1) The parameter orders were all jumbled, and we settled on `addModifier`, and wanted to remove the `addAccidental` etc. If you look at the `addModifier` code, you'll see the checks for inverse parameters (because even the `add*` methods were inversed.) I think it would be good to just get rid of those functions (or replace them with a `console.log` that says it's deprecated.)

2) I went through all the uses of accidental and dot as examples, and we can really remove so much there and clean up the dependencies. I think a more consistent API could be:

### elementInstance.attach(target, options)

```
let accidental = new Accidental(...);
accidental.attach(note, {index: 2});

let dot = new Dot();
dot.attach(note, {index: 2});
dot.attach(note, {all: true});
```

### elementClass.buildAndAttach(target, options)

```
Dot.buildAndAttach(note, {index: 0});
Dot.buildAndAttach(note, {all: true});
Parenthesis.buildAndAttach(note);
```

So `Note` can still have `addDot`, but it does not need a dependency on `Dot()`

### elementClass.getAttached(target)

```
// Replace note.getDots()
Dot.getAttached(note);

// Returns all accidentals 
Accidental.getAttached(note);
```

What do you think? If that makes sense, we can start with something like `Accidental` or `Dot` and see if we can take it out of the dependencies this way?

rvilarl: I think that we will reduce the dependencies this way, no doubt, but I have to play with it a little bit to see the consequences. 
I do like the prototypes you suggested.

rvilarl: q0xfe please move to #1270 while I look into that. @AaronDavidNewman needs to get that PR merged to continue resolving #411

rvilarl: `addModifier` reason for the mess. `Note.addModifier` was changed from `(index, modifier)` to `(modifier,index)` during the TS migration due to the fact that `Tickable` also has a `addModifier` that was not signature compatible with `(index, modifier)`
@0xfe @ronyeh I would suggest to revert this change and rename the function in `Tickable` as the first step. What do you think?

rvilarl: I have reverted the order of parameters. I think this will ease the migrations and there is no good reason for it. I will easily remove `addAnnotation`, `addAccidental` and `addArticulation`

rvilarl: Removed

rvilarl: I will go now for the buildAndAttach change. I believe that this one will reduce the number of dependencies. @0xfe Before I do it, would it make sense to change the signature to allow more than one note?
```
Dot.buildAndAttach([note], {index: 0});
Dot.buildAndAttach([note], {all: true});
Parenthesis.buildAndAttach([note]);
```
Normally parentheses and dots apply to more than one note. The example above would be:
```
const notes = [
    staveNote({ keys: ['f/4'], duration: '4', stem_direction: Stem.UP }),
    staveNote({ keys: ['f/4'], duration: '8', stem_direction: Stem.UP }),
    staveNote({ keys: ['f/4'], duration: '16', stem_direction: Stem.UP }),
    staveNote({ keys: ['f/4'], duration: '32', stem_direction: Stem.UP }),
    staveNote({ keys: ['f/4'], duration: '64', stem_direction: Stem.UP }),
    staveNote({ keys: ['f/4'], duration: '128', stem_direction: Stem.UP }),
    staveNote({ keys: ['g/4'], duration: '4', stem_direction: Stem.UP }),
    staveNote({ keys: ['g/4'], duration: '8', stem_direction: Stem.UP }),
    staveNote({ keys: ['g/4'], duration: '16', stem_direction: Stem.UP }),
    staveNote({ keys: ['g/4'], duration: '32' }),
    staveNote({ keys: ['g/4'], duration: '64', stem_direction: Stem.UP }),
    staveNote({ keys: ['g/4'], duration: '128', stem_direction: Stem.UP }),
  ];
Dot.buildAndAttach([notes], {all: true});
Dot.buildAndAttach([notes[5], notes[11]], {all: true});
``

**Comment by ronyeh** [comment]
Hi @rvilarl, (cc: @0xfe )

Thanks for taking a look at the circular dependencies! I assume these circular dependencies do not break the tests at the moment? 

Some of our circular dependencies are long chains, that I think might have been created due to our typeguard methods.

I read up on circular dependencies, and certain types won't cause problems (e.g., if the referenced classes are used as types, which will be erased by the compiler when converting from TS to JS).

However, `instanceof` can cause nasty circular dependencies that break code at runtime. This is because `obj instanceof Class` requires both **obj** and **Class** to be defined at the moment of the type check. 

If we stop using `instanceof` in the type guard methods (and elsewhere in the code), perhaps it will break up all these big circular chains.

One idea is to just rely on the category strings. (Rodrigo fixed the build last week by doing this.)

However, having the category strings as static members in each class can also cause circular dependencies, since the class needs to be defined first!

So what about `const enum` in TypeScript? Would that be a nice compromise between just using hard-coded strings and directly referencing a static member in the class?

https://www.typescriptlang.org/docs/handbook/enums.html#const-enums

> unlike regular enums they are completely removed during compilation. Const enum members are inlined at use sites. 

So const enums are kind of like `#define` in C.

```
const enum Category {
    Accidental = "Accidental",
    Stave = "Stave",
    Modifier = "Modifier",
}

class A {

    constructor() {
        console.log(Category.Accidental);
        console.log(Category.Stave);
        console.log(Category.Modifier);
    }
...
```

is compiled to:

```
class A {
    constructor() {
        console.log("Accidental" /* Accidental */);
        console.log("Stave" /* Stave */);
        console.log("Modifier" /* Modifier */);
    }
...
```

The TypeScript const enum is completely erased, and the category strings are inserted into all the use sites, as if we had hard-coded them there in the first place.

This helps avoid typos and maintains some of the type-safety-ness of TypeScript, while also breaking the circular dependencies because we are just using strings instead of relying on instanceof.

Does this approach sound like a good idea to you guys? If so, I can submit a PR with the Category const enums, and remove uses of instanceof.

(A PR on improved docs is coming as well.)

File: src/note.ts (hunk 1 above)
Comment (rvilarl) [reply src/note.ts]: I have followed the proposal:
```
Dot.buildAndAttach([note], {index: 0});
Dot.buildAndAttach([note], {all: true});
Parenthesis.buildAndAttach([note]);
```
**and 6 additional circular dependencies are gone!**
**No visual differences!**

My thoughts after playing a bit:

> ### elementInstance.attach(target, options)
> ```
> let accidental = new Accidental(...);
> accidental.attach(note, {index: 2});
> 
> let dot = new Dot();
> dot.attach(note, {index: 2});
> dot.attach(note, {all: true});
> ```

we can do almost the same with `Note.addModifier` I do not see a big benefit here in regards to circular dependencies.

> ### elementClass.buildAndAttach(target, options)
> ```
> Dot.buildAndAttach(note, {index: 0});
> Dot.buildAndAttach(note, {all: true});
> Parenthesis.buildAndAttach(note);
> ```

done, very good against the circular dependencies.

> ### elementClass.getAttached(target)
> ```
> // Replace note.getDots()
> Dot.getAttached(note);
> 
> // Returns all accidentals 
> Accidental.getAttached(note);
> ```

we can do almost the same with `Note.getModifiersByType` I do not see a big benefit here in regards to circular dependencies.

Comment (rvilarl) [reply src/note.ts]: I would leave this PR as it is and make new ones if we want to resolve the remainig dependencies:
```
WARNING in Circular dependency detected:
src/accidental.ts -> src/gracenotegroup.ts -> src/beam.ts -> src/tuplet.ts -> src/formatter.ts -> src/modifiercontext.ts -> src/accidental.ts

WARNING in Circular dependency detected:
src/beam.ts -> src/tuplet.ts -> src/formatter.ts -> src/beam.ts

WARNING in Circular dependency detected:
src/factory.ts -> src/system.ts -> src/factory.ts

WARNING in Circular dependency detected:
src/formatter.ts -> src/beam.ts -> src/tuplet.ts -> src/formatter.ts

WARNING in Circular dependency detected:
src/glyph.ts -> src/tables.ts -> src/glyph.ts

WARNING in Circular dependency detected:
src/gracenotegroup.ts -> src/beam.ts -> src/tuplet.ts -> src/formatter.ts -> src/modifiercontext.ts -> src/accidental.ts -> src/gracenotegroup.ts

WARNING in Circular dependency detected:
src/modifiercontext.ts -> src/accidental.ts -> src/gracenotegroup.ts -> src/beam.ts -> src/tuplet.ts -> src/formatter.ts -> src/modifiercontext.ts

WARNING in Circular dependency detected:
src/notesubgroup.ts -> src/formatter.ts -> src/modifiercontext.ts -> src/notesubgroup.ts

WARNING in Circular dependency detected:
src/system.ts -> src/factory.ts -> src/system.ts

WARNING in Circular dependency detected:
src/tables.ts -> src/glyph.ts -> src/tables.ts

WARNING in Circular dependency detected:
src/tuplet.ts -> src/formatter.ts -> src/beam.ts -> src/tuplet.ts
```

rvilarl: @0xfe ready to review

rvilarl:
Hi @ronyeh,

I missed your answer writing further on the review comment. 
The removal of circular dependencies in this PR is based on refactoring. 

About your idea with enum, I find it very good. I also think that we need more a list of categories to really emulate instanceof. For example, a note is also a tickable and an element.

If you look at the 11 circular dependencies above, you will see that in reallity we have less because, for example,  system -> factory -> system and factory -> system -> factory is the same one.

rvilarl:
In a prerelease integration with OSMD I realised that `Note.dots` was not correctly handled.
@0xfe ready to review, I believe :)

0xfe:
@rvilarl This is great! It really cleans up the API! Thanks, merging.

PR #1280, opened 2021-12-31
simplify format function in Stave Note

### LINKED ISSUE #1283: Issues in tests/flow.html (5,6,7)
**5. Rests and Notes overlapping**
image

**6. Rests overlapping**
image

**7. Lots of overlapping**
image

rvilarl:
Fixes #1283

After realising that the code in `StaveNote.forma()` in #1263 was dead code. I decided to bring it back to life as simple as possible.

The visual differences (Bravura version) are:

**StaveNote Center_Aligned_Note___Multi_Voice Bravura**
StaveNote Center_Aligned_Note___Multi_Voice Bravura
**Current**
StaveNote Center_Aligned_Note___Multi_Voice Bravura_Current
**Reference**
StaveNote Center_Aligned_Note___Multi_Voice Bravura_Reference
**Three_Voice_Rests Auto_Adjust_Rest_Positions___Three_Voices__2 Bravura**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Three_Voices__2 Bravura
**Current**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Three_Voices__2 Bravura_Current
**Reference**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Three_Voices__2 Bravura_Reference
**Three_Voice_Rests Auto_Adjust_Rest_Positions___Two_Voices Bravura**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Two_Voices Bravura
**Current**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Two_Voices Bravura_Current
**Reference**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Two_Voices Bravura_Reference

### DIFF
--- a/src/dot.ts
+++ b/src/dot.ts
@@ -67,8 +67,7 @@ export class Dot extends Modifier {
         const index = dot.checkIndex();
         props = note.getKeyProps()[index];
         // consider right displaced head with no previous modifier
-        if (note.checkModifierContext().getMembers('Parenthesis').length === 0) shift = note.getRightDisplacedHeadPx();
-        else shift = right_shift;
+        shift = note.getFirstDotPx();
       } else if (note.getCategory() === 'TabNote') {
         props = { line: 0.5 }; // Shim key props for dot placement
         shift = right_shift;
--- a/src/dot.ts
+++ b/src/dot.ts
@@ -67,7 +67,7 @@ export class Dot extends Modifier {
         const index = dot.checkIndex();
         props = note.getKeyProps()[index];
         // consider right displaced head with no previous modifier
-        if (right_shift === 0) shift = note.getRightDisplacedHeadPx();
+        if (note.checkModifierContext().getMembers('Parenthesis').length === 0) shift = note.getRightDisplacedHeadPx();
         else shift = right_shift;
       } else if (note.getCategory() === 'TabNote') {
         props = { line: 0.5 }; // Shim key props for dot placement
--- a/src/modifiercontext.ts
+++ b/src/modifiercontext.ts
@@ -132,7 +132,7 @@ export class ModifierContext {
   }
 
   getMembers(category: string): ModifierContextMember[] {
-    return this.members[category];
+    return this.members[category] ?? [];
   }
 
   getWidth(): number {
--- a/src/note.ts
+++ b/src/note.ts
@@ -577,6 +577,24 @@ export abstract class Note extends Tickable {
     };
   }
 
+  getRightParenthesisPx(index: number): number {
+    const props = this.getKeyProps()[index];
+    return props.displaced ? this.getRightDisplacedHeadPx() : 0;
+  }
+
+  getLeftParenthesisPx(index: number): number {
+    const props = this.getKeyProps()[index];
+    return props.displaced ? this.getLeftDisplacedHeadPx() : 0;
+  }
+
+  getFirstDotPx(): number {
+    let px = this.getRightDisplacedHeadPx();
+
+    if (this.checkModifierContext().getMembers('Parenthesis').length !== 0)
+      px += Tables.currentMusicFont().lookupMetric('parenthesis.default.width');
+    return px;
+  }
+
   /** Get the metrics for this note. */
   getMetrics(): NoteMetrics {
     if (!this.preFormatted) {
--- a/src/parenthesis.ts
+++ b/src/parenthesis.ts
@@ -28,8 +28,6 @@ export class Parenthesis extends Modifier {
 
   /** Arrange parentheses inside a ModifierContext. */
   static format(parentheses: Parenthesis[], state: ModifierContextState): boolean {
-    const { right_shift, left_shift } = state;
-
     if (!parentheses || parentheses.length === 0) return false;
 
     let x_widthL = 0;
@@ -40,17 +38,16 @@ export class Parenthesis extends Modifier {
       const note = parenthesis.getNote();
       const pos = parenthesis.getPosition();
       const index = parenthesis.checkIndex();
-      const props = note.getKeyProps()[index];
 
       let shift = 0;
 
       if (pos === ModifierPosition.RIGHT) {
-        shift = (props.displaced ? note.getRightDisplacedHeadPx() : 0) + right_shift;
-        x_widthR += shift + parenthesis.width;
+        shift = note.getRightParenthesisPx(index);
+        x_widthR = x_widthR > shift + parenthesis.width ? x_widthR : shift + parenthesis.width;
       }
       if (pos === ModifierPosition.LEFT) {
-        shift = (props.displaced ? note.getLeftDisplacedHeadPx() : 0) + left_shift;
-        x_widthL += shift + parenthesis.width;
+        shift = note.getLeftParenthesisPx(index);
+        x_widthL = x_widthL > shift + parenthesis.width ? x_widthL : shift + parenthesis.width;
       }
       parenthesis.setXShift(shift);
     }
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -81,7 +81,7 @@ const isInnerNoteIndex = (note: StaveNote, index: number) =>
 
 // Helper methods for rest positioning in ModifierContext.
 function shiftRestVertical(rest: StaveNoteFormatSettings, note: StaveNoteFormatSettings, dir: number) {
-  const delta = (note.isrest ? 0.0 : 1.0) * dir;
+  const delta = dir;
 
   rest.line += delta;
   rest.maxLine += delta;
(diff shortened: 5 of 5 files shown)


rvilarl:
Ready to review

rvilarl:

0xfe:
Thanks Rodrigo. I like the idea of cleaning up that code -- it's definitely a mess, but it looks like the rest repositioning is no longer happening correctly. (It's somewhat broken in the reference too, which was not true before. I'm not sure what change did it, but those resets are not supposed to collide.)

For example, in `Three_Voice_Rests Auto_Adjust_Rest_Positions___Two_Voices Bravura`, the rests in the reference are vertically aligned for the tick. (That feature was not updated for our new font and height logic, so the spacing needs to be corrected -- but either way, it's "more correct" than the new slightly shifted code.)

Not exactly sure what to do here before digging in, I'd first try to see if we can separate and fix the auto-positioning code for the rests, and then try to clean up that part.

rvilarl:
Ok, I think that everything is fixed now. These are the visual differences (Bravura version):

**Dot Multi_Voice Bravura**
Dot Multi_Voice Bravura
**Current**
**Reference**
**StaveNote Center_Aligned_Note___Multi_Voice Bravura**
StaveNote Center_Aligned_Note___Multi_Voice Bravura
**Current**
StaveNote Center_Aligned_Note___Multi_Voice Bravura_Current
**Reference**
StaveNote Center_Aligned_Note___Multi_Voice Bravura_Reference
**Three_Voice_Rests Auto_Adjust_Rest_Positions___Three_Voices__1 Bravura**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Three_Voices__1 Bravura
**Current**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Three_Voices__1 Bravura_Current
**Reference**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Three_Voices__1 Bravura_Reference
**Three_Voice_Rests Auto_Adjust_Rest_Positions___Three_Voices__2 Bravura**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Three_Voices__2 Bravura
**Current**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Three_Voices__2 Bravura_Current
**Reference**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Three_Voices__2 Bravura_Reference
**[Three_Voice_Rests Auto_Adjust_Rest_Positions___Two_Voices Bravura**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Two_Voices Bravura
**Current**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Two_Voices Bravura_Current
**Reference**
Three_Voice_Rests Auto_Adjust_Rest_Positions___Two_Voices Bravura_Reference

@0xfe ready for review.

src/dot.ts
```diff
@@ -41,7 +41,7 @@ export class Dot extends Modifier {
         const index = dot.checkIndex();
         props = note.getKeyProps()[index];
         // consider right displaced head with no previous modifier
-        if (right_shift === 0) shift = note.getRightDisplacedHeadPx();
+        if (note.checkModifierContext().getMembers('Parenthesis').length === 0) shift = note.getRightDisplacedHeadPx();
```
0xfe, line 44: This is a bit too specific, ideally `Dot` should not know about `Parenthesis`. Can we generalize somehow?

0xfe (commented):
Diffs look great. Were there other diffs?

The code LTGM mostly, but see my comment about `getMembers('Parenthesis')`.

src/dot.ts
rvilarl: I have not figure out a way. Previously Dot did not care, it made special calculations on the basis that Dot was always the first Modifier. Now Parenthesis can be first and this has to be checked for.

rvilarl:
> Diffs look great. Were there other diffs?

No other diffs :)

src/dot.ts
0xfe: I think this also masks the case where there is a displaced head with a parenthesis, no?

I'd suggest adding some kind of logic to `StaveNote`, where it can check its own members and return a shift. So only the note knows about whether or not there's a `Parenthesis`.

rvilarl: Hi Mohit, I missed this one. I have been trying different approaches:
- Each element takes care of their shifts. I mean no extra code on dots to get closer to the note. I believe that this is the cleanest approach but then on the example above "Dot Multi_Voice Bravura" all dots are vertically aligned to the right of both notes. Anyway I believe that this example, as it is today, in the fourth group of notes if the notes above had two dots (the second one would overlap with the head of the note below).
- Then I tried to create a ModifierContext for each of the notes, and this works for the dots, but there are other tests that get overlap on the accidentals, so I discarded it.

It is quite a complicated exercise. 

My question would ti be acceptable to get display all the dots to the right of the notes aligned to the left of the head located in the right most position?

rvilarl: The images related to the comment above:

**Current (as in this PR)**
image
**Each element taking care of its own shift. There is something to solve in the 4th group but the question is are the first and the second group ok drawn in this way?**
image

rvilarl: @0xfe @ronyeh @mscuthbert @AaronDavidNewman @sschmidTU should I follow this way? I mean the approach from MuseScore.

ronyeh: I am far from an expert in how engraved music should appear.

However, I do like the pictures where the dots are lined up with each other. It's very pleasing to the eye.

mscuthbert: The MuseScore arrangement looks visually beautiful, but it's incorrect by Gould "Behind Bars":

Screen Shot 2022-01-25 at 08 38 40

That is for Chords with adjacent notes (clusters/7ths etc.).  When they are further apart or have no overlapping notes, then aligning the dots is preferred:

Screen Shot 2022-01-25 at 08 40 53


No automatic algorithm I know of, however, will get the 4th upstemmed group's dots to work properly with the third downstemmed group where there are two notes on the same line (D5) with different numbers of dots AND there isn't room for the downstem D5 to move its dots below (because of the C5 dots).  What an engraver would do in that case is have the dots for the first group appear before the second group's notes appear and move all dots closer to the notes than would be normal:

Screen Shot 2022-01-25 at 08 41 04

The top part may also be moved to the right (contrary to norms of alignment) if it clarifies the dot:

Screen Shot 2022-01-25 at 08 42 43

Once we get to overlapping parts, the rules are even more complex.

Screen Shot 2022-01-25 at 08 43 04

I would NOT expect Vexflow to take care of these situations at this point, but eventually (not for 4.0!) it'd be good to have utility functions to shift different voices at the same tick in different ways.  But I'm not volunteering to do this at this point.

mscuthbert:
Added comment w/ examples from Gould above.

src/dot.ts
rvilarl: Thanks a lot @mscuthbert. What should happen with accidentals in conbination with these dots. I refer, for example, to the last stave above. We have note1, its associated dot, note2. What would happen if note 2 had a #?

mscuthbert: Say the accidental was on C (C#)  -- if the other voice doesn't have a C# then there would be little ambiguity (just ugly) if the sharp were before both groups.  If both groups had a C# then it's even better (even if there is some ambiguity).  If one group had C-natural and the other C# -- then you get into one of those really ugly situations where probably best would be to get out your rarely used squashed # sign that you carved from a wood block for just this occasion, and try to squeeze it in there.  :-D    -- I do have a friend who is a professional engraver whom I could ask about this case if you think we're actually going to support it now.  I'm pretty sure none of the professional software packages do this really.  (I still don't own Dorico though).

ronyeh: Wow I certainly learned something from this thread. Yeah I think VexFlow should just try our best to accommodate the most common use cases, and then allow the user to add a relative shiftX / shiftY to any modifier, which will be applied after regular formatting. That way, a VexFlow user can just eyeball their score and tune by hand. 

This might not work if we are loading music xml and translating it to easy score for automated score creation. However, if we had an API that could say something like, in measure M, voice V, note group N, note index i, addRelativeShift(x, y) to the modifiers, that would allow us to "patch" rendered scores. Then a VexFlow user would have a patch for each musicxml file that they hand tweaked.

rvilarl: @0xfe comment addressed. Doing the change I also fixed a bug on `Parenthesis`. The shifts in format were being added rather than calculating the biggest shift.

**Parenthesised Bravura**
EasyScore Draw_Parenthesised Bravura
**Current**
EasyScore Draw_Parenthesised Bravura_current
**Reference**
EasyScore Draw_Parenthesised Bravura_reference

rvilarl:
@0xfe ready to merge, I believe.

0xfe:
Nice job, Rodrigo. This looks great!

PR #1284, opened 2022-01-05
Documentation

ronyeh:
Work in progress: Move major docs from wiki to repository.

The actual release script will come in a separate PR.


ronyeh:
I have tested the `release-it` tool with this example repository: https://github.com/ronyeh/simple-typescript-library/

It works well, and even manages pre-releases (alpha | beta | rc).

Here are the versions I have released using the tool: https://www.npmjs.com/package/simple-typescript-library?activeTab=versions

Each release takes about 35 seconds.

0xfe:
Thanks Ron -- lets keep a most of the docs in the wiki please. It's much lower effort to manage. (E.g., even the release doc can be in the wiki.)

0xfe:
Also, the release-it tool does seem cool.

ronyeh:
> Thanks Ron -- lets keep a most of the docs in the wiki please. It's much lower effort to manage. (E.g., even the release doc can be in the wiki.)

OK, sounds good.

I did like that the repo docs automatically got a table of contents. But it's nice that wikis are lower-threshold so that anyone can contribute without requiring a PR.

PR #1287, opened 2022-01-06
Improve Gruntfile

ronyeh:
This PR improve our build scripts with:
- release scripts to "automate" the process of publishing on NPM and GitHub
- better ESM support
- improved build performance

- I also improved the wiki documentation for [building & releasing](https://github.com/0xfe/vexflow/wiki/Build%2C-Test%2C-Release).


ronyeh:
Still testing the release scripts and improving the gruntfile / webpack scripts.

Some fun early results of the performance tweaks (tested on a Macbook Pro 2016):

```
$ time grunt

master branch =>   2m40.316s
current branch =>  1m13.565s  to  0m51.830s
```

In addition to webpack improvements, I also turned on the eslint cache. So while the first full build is 1 minute 13 seconds, if I immediately rebuild, it drops to 52 seconds (which is a best case scenario since no files have changed). In practice, it'll be somewhere in between, since I'll have touched some files between builds.

ronyeh:
@rvilarl ready for preliminary feedback. Most of the interesting changes are in the **Gruntfile.js**.

I realized the scripts were split between the package.json (i.e., npm run XXX) and the Gruntfile.js and it was getting confusing. I moved all the scripts into the Gruntfile.js. The npm scripts just call grunt XXX now where possible.

Now a full build takes ~50 seconds on my 2016 laptop, whereas it used to take 2 mins 40 seconds.

The Gruntfile.js also includes the new release script.

Once this branch is merged into master, we can start issuing more releases by typing 'grunt release'.

@rvilarl I still need to add some more comments to Gruntfile.js and clean up a bit, but it's ready for your review at the moment.

Then I'll package up a tgz just to make sure I didn't break anything. :-)

There are no visual differences from current master.

If you want, you can clone this branch and test out the various build scripts to see if the performance improvement is noticeable for you.

ronyeh:
It's probably easiest to read the new Gruntfile.js directly, rather than looking through the diff:

https://github.com/ronyeh/vexflow/blob/test/Gruntfile.js

Let me know if it looks good to you. I tried to make it organized & understandable to a developer wishing to tweak his/her build.

rvilarl:
> It's probably easiest to read the new Gruntfile.js directly, rather than looking through the diff:
> 
> https://github.com/ronyeh/vexflow/blob/test/Gruntfile.js
> 
> Let me know if it looks good to you. I tried to make it organized & understandable to a developer wishing to tweak his/her build.

@ronyeh I find it very well organized and documented. Good job!

rvilarl (approved):

mscuthbert:
Is there a way to make the bare "grunt" or "grunt release" always build the latest source maps?  I think that there's always the possibility otherwise that we're going to make a build where the source-map is out of date and that will cause major problems for users.

ronyeh:
@mscuthbert @0xfe @rvilarl @sschmidTU 

We should make a decision on what to do with source maps. I'm trying to balance the goals of different people, so have been struggling to find the best solution.

I think I've heard from @sschmidTU that he doesn't care for source maps in the final release published to npm. Simon, can you confirm?

However, it sounds like @mscuthbert wants source maps in the npm release. In what situations do appreciate having the source maps? 

In the 4.0 npm release, we will also have the **ES module** files, which are individual un-minified JS files that look like TypeScript files (but with all the typing information removed).

For debugging purposes, you can either use the ES module files, or the vexflow-debug.js, which is also unminified.

From a VexFlow maintainer's perspective, creating source maps each time slows down the build. This is very noticeable when we run `grunt watch`. (On my machine, source maps add at least 5 seconds to each build, so it adds up.) Plus, we test with the `vexflow-debug-with-tests.js` build which is not minified.


**Question 1**: Should we include source maps in the npm release? The npm release will already include vexflow-debug.js, and all the TypeScript source files, and also the ESM build which is basically the TS files without the types.

Here's webpack's section about source maps: https://webpack.js.org/guides/build-performance/#source-maps

sourcemaps

**Question 2**: Assuming we want to keep them.... When should we build source maps? Maybe we should only do it as part of the release process? What about something like this?

```
grunt build:release   <== this creates the source maps
grunt test:release
grunt release
```

I probably won't fold the build and test into the `grunt release` script, because there's a chance the release fails (e.g., if you abort with CTRL+C) and you'll need to run the release again with the same build.

ronyeh:
OK, this is my current plan:

`grunt` makes a complete build but WITHOUT source maps.
`grunt source-map` makes a complete build WITH source maps.

To publish a new release to NPM and GitHub WITH source maps, we do the following:
```
grunt build-test-and-release-to-npm-and-github
````

Internally, this ridiculously long but descriptive script name does the following:
```
grunt source-map
grunt qunit
grunt release
```

I gave it a super long but descriptive name as sort of a joke, but I kind of like it.

Which one do you like better?

`grunt build-test-and-release-to-npm-and-github`

or 

`grunt build-test-release`

**Comment by ronyeh** [comment]
I'm doing some more gruntfile/webpack tweaks and performance testing.

At the moment, I'm starting to lean towards **turning ON** source-maps for all files by default.

The only real downside I can see is that it makes the npm module larger (disk size).

Compilation performance isn't that huge of a hit, if I can tweak the webpack configs correctly. Plus, I'm running a 5 year old machine.

**Comment by 0xfe** [comment]
Ron, awesome stuff overall. Skimming for now, but let me know when it's ready to review.

I prefer `build-test-release`, but not a strong opinion.

For the external interface, I like:

- `build:release` <-- builds with source maps
- `build:release:no-source-maps` <-- builds without source maps.

Either way, I think we shouldn't produce source maps by default, however, we should make it easy for developers to get them if they want.

File: .release-it.json [hunk 1]
```diff
@@ -0,0 +1,31 @@
+{
+  "hooks": {
+    "before:init": [],
+    "after:bump": [
+      "echo add build/ folder",
```
0xfe, line 5: "Adding build/ folder..."

.release-it.json
```diff
@@ -0,0 +1,31 @@
...
+      "git add -f build/",
+      "git commit -m 'Add build/ for release version ${version}.'"
+    ],
+    "after:npm:release": [],
+    "after:git:release": [],
+    "after:github:release": [],
+    "after:release": [
+      "echo Successfully released ${name} ver. ${version} to ${repo.repository}.",
+      "echo remove build/ folder",
```
0xfe, line 14: "Removing build/ folder..."

ronyeh:
> Either way, I think we shouldn't produce source maps by default, however, we should make it easy for developers to get them if they want.

Just to clarify, do you think our NPM release should NOT include source maps by default?

So far, it sounds like @mscuthbert and @AaronDavidNewman like source maps. I want to hear others opinions as well.

I am now leaning toward **including source maps** in the NPM release. The two drawbacks are disk space (~12.5MB) and slightly slower build speed. However, it doesn't affect the final bundle size, and modern computers should be able to provide good enough build performance.

0xfe:
I don't have a strong opinion about it, and I agree that other input is helpful (and also staying consistent with the current best practices are.) 12.5Mb of disk space is not a big deal, IMO, and in the interest of not holding back progress, I'm okay with including source maps.

Most important for me is minimizing the development build cycle period -- time between saving a file and building a debug binary should be as small as possible (which I think is unaffected by this decision.)

ronyeh:
Ready for review. :-)

No visual diffs from current master revision.

Main contributions in Gruntfile.js, but lots of small fixes elsewhere to aid in testing and improve build performance.

I made a small tweak to typedoc config, which caused all the docs to be regenerated. If this makes it impossible to review, I can revert the docs/api/ dir and submit that as a separate follow up PR.

I will add comments to this thread over the next day or two to describe the main things I did.

ronyeh:
For testing this build with your apps, you can use:

```
npm install https://vexflow-demo.surge.sh/vexflow-4.0.1-beta-2022-01-13.tgz
```

Build Performance on my laptop:
```
grunt
Before = 2m22s
After  = 0m46s

grunt watch
After = 2 seconds to compile each edited file, with an additional 7 seconds to report errors (fork-ts-checker-webpack-plugin runs in a background process).
```

entry/vexflow-core.ts
```diff
@@ -51,10 +51,10 @@ Flow.fetchMusicFont = async (fontName: string, fontModuleOrPath?: string | FontM
...
 
     const g = globalObject();
-    const moduleCJS = g['VexFlowFont' + fontName];
-    if (typeof moduleCJS !== 'undefined') {
-      // CJS font modules will set a global variable named: VexFlowFontBravura | VexFlowFontGonville | VexFlowFontPetaluma | etc.
-      fontModule = moduleCJS;
+    const VexFlowFont = g['VexFlowFont'];
+    if (typeof VexFlowFont !== 'undefined' && typeof VexFlowFont[fontName] !== 'undefined') {
+      // CJS font modules will set a global variable named: VexFlowFont.Bravura | VexFlowFont.Gonville | VexFlowFont.Petaluma | etc.
+      fontModule = VexFlowFont[fontName];
```
ronyeh: This is only for `vexflow-core.js` and dynamic font module loading. A CJS font module will set the `VexFlowFont.{FontName}` with the object containing the glyphs and metrics information (e.g., VexFlowFont.Bravura, or VexFlowFont.Leland if you are adding the Leland SMuFL font).

src/fonts/load_all.ts
```diff
@@ -7,11 +7,15 @@ import { loadBravura } from './load_bravura';
 import { loadCustom } from './load_custom';
 import { loadGonville } from './load_gonville';
 import { loadPetaluma } from './load_petaluma';
+// [ADD_MUSIC_FONT]
+// import { loadXXX } from './load_xxx';
```
ronyeh: This is one of the integration points for adding a new music font to VexFlow.

rvilarl:
@ronyeh is #1289 addressed?

ronyeh:
This is the most current build for testing:

```
npm install https://vexflow-demo.surge.sh/vexflow-4.0.1-beta-2022-01-13-B.tgz
```

rvilarl:
> npm install https://vexflow-demo.surge.sh/vexflow-4.0.1-beta-2022-01-13-B.tgz

tested in OSMD looking good!

ronyeh:
```
npm install https://vexflow-demo.surge.sh/vexflow-4.0.1-beta-2022-01-14.tgz
```

PENDING: I think I found one last bug that I'll need to address tomorrow. The version file needs to be generated **after** the release script bumps the version in package.json.

In any case, this PR can be reviewed. I am happy to answer any questions.

Here is the Gruntfile.js:

https://github.com/ronyeh/vexflow/blob/test/Gruntfile.js

rvilarl:
also tested in OSMD looking good!

rvilarl:
@ronyeh the version.ts file is causing me many troubles with rebase. Do you think that we should keep it?

ronyeh:
> @ronyeh the version.ts file is causing me many troubles with rebase. Do you think that we should keep it?

For a rebase, you can just pick either commit. But I can see that it might cause extra steps during a rather long rebase.

I'm currently working on the version file generation right now. I think we can generate it ONLY upon running the release script, so normally it shouldn't be a problem.

Or maybe I can find another solution. I agree with you that updating it on every single build can get annoying.

typedoc.json
```diff
@@ -0,0 +1,10 @@
+{
+  "entryPoints": ["src/index.ts"],
+  "out": "docs/api",
+  "name": "vexflow",
+  "excludeProtected": true,
+  "excludePrivate": true,
+  "disableSources": true,
+  "hideGenerator": true,
+  "intentionallyNotExported": ["OffscreenCanvas", "OffscreenCanvasRenderingContext2D"]
```
rvilarl: Those two interfaces should be exported, they are used in the definition of a public member variable.

tests/key_clef_tests.ts
```diff
@@ -6,7 +6,10 @@
 
 import { MAJOR_KEYS, MINOR_KEYS, TestOptions, VexFlowTests } from './vexflow_test_helpers';
 
-import { ContextBuilder, Glyph, KeySignature, Stave } from '../src';
+import { Glyph } from '../src/glyph';
+import { KeySignature } from '../src/keysignature';
+import { ContextBuilder } from '../src/renderer';
+import { Stave } from '../src/stave';
```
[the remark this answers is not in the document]
rvilarl: Agree perhaps you colud place all these import fixes in another PR to reduce this one. I recommend to do the same in jet another  PR for the API doc files

0xfe:
Thanks Ron -- overall, the PR looks great. The version file, however, is still a bit of a pain, and I'm wondering if it's really worth the effort.

- Do we really need it -- are our version tags insufficient?
- Can it be dynamically generated (say from the commit ID, as a build artifact) instead of actually being committed?
- Can it be an additional git tag instead?

ronyeh:
Thanks for your feedback everyone!

> * Do we really need it -- are our version tags insufficient?
> * Can it be dynamically generated (say from the commit ID, as a build artifact) instead of actually being committed?
> * Can it be an additional git tag instead?

@0xfe regarding the version.ts:

The `Vex.Flow` object exposes a static `BUILD` property. I have found that it's useful to have a simple one-liner that you can call on the Vex.Flow object just to make sure the library is loaded correctly. When you open a dev console and type `Vex.Flow.BUILD` you get something like:
```
{
    "VERSION": "4.0.0",
    "ID": "68fa5901f1bd33077eecb50ebdcb85d292210ff6",
    "DATE": "2022-01-15T06:23:09.650Z"
}
```

If the `DATE` is updated upon every build, it might be useful for developers to know whether they are seeing an old build due to caching or another error (for debugging purposes).

Previously, I used a webpack plugin to insert this info on the fly, during the build process. When I added ESM support (which outputs separate JS files) I realized the version info still had the placeholders (e.g., DATE: "__VEX_BUILD_DATE__", because our ESM build goes through the tsc compiler and not webpack. Thus the version.ts file was born, so I could have the same info compiled into both the webpack/CJS builds and also the tsc/ESM build. :-)

After explaining the above, I think I just convinced myself to **go back to the previous approach of generating this info on the fly** during the build process. I would have to add a script to patch the esm output file, but that should be straightforward.

I'll ping you all when the version.ts issue is resolved in this PR.

typedoc.json
ronyeh: If I remove those, typedoc throws a warning during the build process. The two types are part of the browser APIs, but not yet in a standard so typescript doesn't know about them.

https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas

tests/key_clef_tests.ts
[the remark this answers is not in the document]
ronyeh: If I don't fix it here, the ESM build is broken. I discovered this issue while testing this PR. Since the change is minimal, hopefully it shouldn't make the review process much harder.

0xfe:
Thanks for the explanation, Ron. Generating on the fly sounds good to me! :-)

typedoc.json
rvilarl: I would just export the two interfaces to get rid of the warning.

tests/key_clef_tests.ts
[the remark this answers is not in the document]
rvilarl: Ok for the imports, but in general I would update API on separate PRs it makes the number of files too big. I find it easier to generate document updates PRs. Anyway we all know that API docs are not to be reviewed.

ronyeh:
Ready for review!

Version file issue resolved. Running the build no longer touches `src/version.ts`. Instead the version info is patched in when emitting the CJS (webpack) and ESM (tsc) builds.

The `grunt release` script has been tested with a separate test project, and it works pretty well. I've also run `grunt release:dry-run` multiple times and I'm pretty sure it'll work. :-) The only way to really know is to try it after we merge this PR.

I think we will run:
```
grunt release:beta
```

And then choose the prepatch version (`4.0.1-beta.0`). We can bump the beta prepatch version if necessary, and then promote to rc, and then ship 4.0.1 as the official / "latest" release when ready.

I've also updated the release docs here: https://github.com/0xfe/vexflow/wiki/Build%2C-Test%2C-Release

It includes an extensive section on manual releases, but ideally we will need none of that. :-)

typedoc.json
ronyeh: I wish it were that easy. (If it is, please send me the fix!)

When I was working with you on the OSMD build, we were having integration issues because we were including the type definitions as a dependency.

I then tried copying the type definitions directly into VexFlow. I ran into issues with that, so I just pared it down to the two interfaces that we reference in VexFlow.

If I export the interfaces, it will generate documentation for them. However, that documentation will be incorrect since my interface definitions (in canvascontext.ts) are incomplete.

These types are built into the browser, so the documentation shouldn't really be generated by us anyways. The API is here: https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas  The spec is here: https://html.spec.whatwg.org/multipage/canvas.html#offscreencanvasrenderingcontext2d

So I'm inclined to leave this as is. Unless you don't like the **intentionallyNotExported** field. If that is the case, I can just remove that field and the two warnings will show up again when geenerating typedoc.

ronyeh:
Current build for testing:
```
npm install https://vexflow-demo.surge.sh/vexflow-4.0.1-beta-2022-01-18.tgz
```

No visual difference when compared to master @ 54966e3d29d255e6af8e776ddb9fa154da135eb5

typedoc.json
rvilarl: See #1297 I have included the API doc changes. This might also reduce the number of files here once you rebase.

ronyeh: I pulled your branch and diffed against my current docs. They aren't the same, so it'll cause docs conflicts on any rebase or merge. This PR also includes a typedoc.json which is used to ignore the two types, and adjust the footer.

ronyeh: filter

To simplify the review process, you can use the File filter dropdown and uncheck the *.html files.

They were all modified because I added a `typedoc.json` file to contain the options, which includes removing the footer on each page to make it look cleaner.

ronyeh:
Thanks to @rvilarl for suggesting ways to simplify the review of this PR.

To make the process easier, you can first review & merge my changes to the API docs:

#1299

That should reduce the number of files in this PR by approximately 188.

ronyeh:
Current beta for testing:

```
npm install  https://vexflow-demo.surge.sh/vexflow-4.0-beta-2022-01-22.tgz
```

I haven't changed anything. Just doing a sanity check to make sure nothing broke due to merge. I tested it successfully with my example project: https://github.com/ronyeh/vexflow-example-javascript

rvilarl (approved):

0xfe (approved):

0xfe:
Looks good, Ron. Approved -- can you fix the conflicts and merge please?

ronyeh:
@0xfe sounds good. will do!

PR #1288, opened 2022-01-07
Fix issues of vertical placement: tuplet

AaronDavidNewman:
With tuplet, annotation, and accidental CS all merged it will look like this:

image

I ran visual regression tests (which I figured out finally after 3 years how to do!) and there were no changes.  The logic here only kicks in if there are modifiers that take up vertical space on one of the notes the tuplet covers, and currently none of our other test cases do this.

### DIFF
--- a/src/tuplet.ts
+++ b/src/tuplet.ts
@@ -273,13 +273,23 @@ export class Tuplet extends Element {
       // y_pos = first_note.getStemExtents().topY - 10;
 
       for (let i = 0; i < this.notes.length; ++i) {
-        const top_y =
-          this.notes[i].getStemDirection() === Stem.UP
-            ? this.notes[i].getStemExtents().topY - 10
-            : this.notes[i].getStemExtents().baseY - 20;
-
-        if (top_y < y_pos) {
-          y_pos = top_y;
+        const note = this.notes[i];
+        let modLines = 0;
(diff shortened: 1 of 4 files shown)


rvilarl:
@AaronDavidNewman this fixes #411 then, am I right?

AaronDavidNewman:
Partially :)  There will be 2 more pull requests for string number and FretHandFinger that I have yet to do.  Annotations are in a different pull request and accidentals are already merged.

src/tuplet.ts
```diff
@@ -273,19 +273,32 @@ export class Tuplet extends Element {
       // y_pos = first_note.getStemExtents().topY - 10;
 
       for (let i = 0; i < this.notes.length; ++i) {
```
0xfe, line 275: Can you add a comment around what this loop is doing?

0xfe:
Thanks for this Aaron. Can we take this opportunity to move those three constants (4, 10, 20) into the metric files please?

src/tuplet.ts
AaronDavidNewman: done

AaronDavidNewman:
This is ready to merge now

0xfe:
LGTM. Merging.

PR #1290, opened 2022-01-10
fix openGroup optional parameters

rvilarl:
Minor fix: the first parameter should be optional. This parameter is expected to be undefined or a string within the function.

@sschmidTU thanks to find this out!

No visual differences.

### DIFF
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -103,7 +103,7 @@ export class CanvasContext extends RenderContext {
   }
 
   // eslint-disable-next-line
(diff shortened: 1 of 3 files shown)


sschmidTU (approved):
Looks good. Did the same with the 3.0.9 definitions at DefinitelyTyped.

rvilarl:
fill in SvgContext also fixed.

0xfe (commented):
Thanks for the fix, Rodrigo. Merging.

PR #1296, opened 2022-01-16
Fix vertical placement of stringnumber and fret number.  For issue #411

AaronDavidNewman:
This is the last fix for 411.  It adds the stringno and fret number.  This is what it looks like now.  I run the diffs and there are only some minor differences in the stringnumber tests.

I added some permutations to the formatting test case.  I also removed the bracket from the test.  There is no way to align that as it stands, because it only has the first and last note in its definitions.  Also, it might span multiple measures.  So formatting that will have to be the responsibility of the application.

image


src/stringnumber.ts
```diff
@@ -5,14 +5,16 @@
...
 import { ModifierContextState } from './modifiercontext';
 import { Note } from './note';
 import { Renderer } from './renderer';
 import { isStaveNote } from './stavenote';
 import { Stem } from './stem';
-import { isStemmableNote } from './stemmablenote';
+import { isStemmableNote, StemmableNote } from './stemmablenote';
+import { Tables } from './tables';
 import { RuntimeError } from './util';
+import { Stave } from './stave';
```
rvilarl, line 17: Stave does not need to be imported, delete the line.

src/stringnumber.ts
```diff
@@ -173,41 +196,44 @@ export class StringNumber extends Modifier {
   draw(): void {
     const ctx = this.checkContext();
     const note = this.checkAttachedNote();
+    const stave = note.checkStave();
```
rvilarl, line 199: This line is not required.

rvilarl (approved):

AaronDavidNewman:
ready for merge

rvilarl (approved):

0xfe:
Can you fix the conflict with `bend.ts` please?

AaronDavidNewman:
ok, fixed

0xfe:
Thanks Aaron. Merging!

PR #1297, opened 2022-01-18
fix typedoc warnings

rvilarl:
Fix the warnings from typedoc document generation. Two interfaces used on a public variable were not exproted.

As expected, no visual differences.


ronyeh:
I don't agree with this approach, unfortunately.

In this PR, you exported the two types (which are not VexFlow types), which adds two documentation pages.

However, our two type definitions are not complete, so the docs are not in sync with the docs at:

https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.offscreencanvas.html

https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.offscreencanvasrenderingcontext2d.html

https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas

I really think we should just not export those two types, because they are part of the (upcoming) browser spec.

ronyeh:
docs

The above gif shows the difference between our two approaches.

By exporting the types, you added two documentation pages (green links).

However, the OffScreenCanvas types should be treated like the regular Canvas types. They are external types (built into the browser API) and they should not have any links and standalone pages.

ronyeh:
I have another question: Do you submit PRs without the docs/api/ directory? I thought you added typedoc into the build process so that the API docs could always be in sync with the state of the repository.

Sometimes I pull down a recent merge, and when I run `grunt`, new API doc pages appear. That seems to suggest that recent PRs did not include the docs.

If you think having the docs makes it harder to review, perhaps we should remove typedoc from the normal build process.

Perhaps, we should just generate typedoc for releases only? I can remove typedoc from the regular `grunt` and only include it whenever we run the release script. Do you prefer this solution? Otherwise, every time we touch files in src/, we should also generate the API docs and submit them in the same PR. That's the only way to keep the docs in sync with the source.

rvilarl:
> I have another question: Do you submit PRs without the docs/api/ directory? I thought you added typedoc into the build process so that the API docs could always be in sync with the state of the repository.

It depends, if it is a few I pack them in, otherwise I do an independent typedoc update. 

> Sometimes I pull down a recent merge, and when I run `grunt`, new API doc pages appear. That seems to suggest that recent PRs did not include the docs.
> 
> If you think having the docs makes it harder to review, perhaps we should remove typedoc from the normal build process.
> 
> Perhaps we should just generate typedoc for releases only?

I would leave it and make it ofter.

> I can remove typedoc from the regular `grunt` and only include it whenever we run the release script. Do you prefer this solution? Otherwise, every time we touch files in src/, we should also generate the API docs and submit them in the same PR. That's the only way to keep the docs in sync with the source.

No please leave it as it is. With `git add` it is possible to select what to commit.

As an example: 
- this PR has 39 API docs and only 1 file to check with a minor change. I find that fine.
- #1287 has a lot of changes and a lot of API docs. Although API docs are not to be reviewed it is very inconvenient to move up and down going through all the changes to see the sources that have to be reviewed. I use github interface to review.

ronyeh:
Thanks for your feedback. I added this PR: 

https://github.com/0xfe/vexflow/pull/1299

That should reduce the number of changed files in the Gruntfile by approximately 188.

Yes, generating the API docs automatically every time I type grunt is a bit of extra overhead. It also seems to break my `git diff` (or at the very least it adds a lot of noise to the diff).

We can discuss whether it's better to just generate docs upon each release. That way we can do it on demand (and you can submit a PR for updated docs whenever convenient). Also, is there a way for us to version our API docs? Let's say we release 4.0.1, but a week later we release 4.1.0 with a few new methods. The docs will now only reflect the latest release.

PR #1299, opened 2022-01-20
Tweaks to API docs.

ronyeh:
Add typedoc.json.
Remove the footer.
Generate docs for any types added in recent merges (e.g., Parenthesis).

This can be merged BEFORE the Gruntfile.js PR https://github.com/0xfe/vexflow/pull/1287

PR #1300, opened 2022-01-20
Swap parameter order of addModifier(...)

ronyeh:
Most of the work was handled by a regex:

```
Search for:
.addModifier\((\d), (.*)\)([^)]?)$

Replace with:
.addModifier($2, $1)$3
```

It didn't handle 33 instances (e.g., when there were two calls to .addModifier() chained on a single line, and when there were extra comment characters at the end of the line).


ronyeh:
This one is based directly off the Gruntfile PR branch: https://github.com/0xfe/vexflow/pull/1287

I did this to avoid merge conflicts from having too many separate PRs in flight.

Please review & merge https://github.com/0xfe/vexflow/pull/1299 and https://github.com/0xfe/vexflow/pull/1287 before taking a look at this one.

src/articulation.ts
```diff
@@ -285,7 +285,7 @@ export class Articulation extends Modifier {
         if (position) artic.position = Modifier.PositionString[position];
         return builder.getFactory().Articulation(artic);
       })
-      .map((artic) => note.addModifier(0, artic));
+      .map((artic) => note.addModifier(artic, 0));
```
rvilarl, line 288: Remove the '0' this is the default value.

src/dot.ts
```diff
@@ -33,16 +33,16 @@ export class Dot extends Modifier {
...
       } else if (options?.index != undefined) {
         const dot = new Dot();
         dot.setDotShiftY(note.glyph.dot_shiftY);
-        note.addModifier(options?.index, dot);
+        note.addModifier(dot, options?.index);
       } else {
         const dot = new Dot();
         dot.setDotShiftY(note.glyph.dot_shiftY);
-        note.addModifier(0, dot);
+        note.addModifier(dot, 0);
```
rvilarl, line 45: Default value

src/note.ts
```diff
@@ -532,30 +532,23 @@ export abstract class Note extends Tickable {
...
-      throw new RuntimeError(
-        'WrongParams',
-        'Call signature to addModifier not supported, use addModifier(modifier: Modifier, index) instead.'
-      );
+      throw new RuntimeError('WrongParams', 'Incorrect call signature. Use ' + signature + ' instead.');
     }
     modifier.setNote(this);
     modifier.setIndex(index);
-    this.modifiers.push(modifier);
-    this.preFormatted = false;
```
rvilarl, line 558: `this.preFormatted = false`should stay to make sure that the new Modifier is preformatted on the next call to `preFormat`

src/tickable.ts
```diff
@@ -278,7 +278,7 @@ export abstract class Tickable extends Element {
    * @param mod the modifier
    */
   // eslint-disable-next-line
-  addModifier(index: number, mod: Modifier): this {
+  addModifier(mod: Modifier, index: number = 0): this {
```
rvilarl, line 281: I believe that you can remove now `index` and the `eslint-disable...`

tests/accidental_tests.ts
```diff
@@ -161,11 +161,11 @@ function formatAccidentalSpaces(options: TestOptions): void {
     new StaveNote({
       keys: ['e##/5'],
       duration: '8d',
-    }).addModifier(0, new Accidental('##')),
+    }).addModifier(new Accidental('##'), 0),
```
rvilarl, line 164: No change required, I would leave the explicit `0` with `Accidental`. Perhaps not in this case but indeed with Chords.

tests/annotation_tests.ts
```diff
@@ -116,13 +116,13 @@ function simple(options: TestOptions, contextBuilder: ContextBuilder): void {
         { str: 4, fret: 9 },
       ],
       duration: 'h',
-    }).addModifier(0, new Annotation('T')),
+    }).addModifier(new Annotation('T'), 0),
```
rvilarl, line 119: Lots of unnecessary 0s in this file. Up to you.

tests/articulation_tests.ts
```diff
@@ -68,10 +68,10 @@ function drawArticulations(options: TestOptions): void {
...
     f.StaveNote({ keys: ['a/4'], duration: 'q', stem_direction: 1 }),
   ];
-  notesBar1[0].addModifier(0, new Articulation(sym1).setPosition(4));
-  notesBar1[1].addModifier(0, new Articulation(sym1).setPosition(4));
-  notesBar1[2].addModifier(0, new Articulation(sym1).setPosition(3));
-  notesBar1[3].addModifier(0, new Articulation(sym1).setPosition(3));
+  notesBar1[0].addModifier(new Articulation(sym1).setPosition(4), 0);
+  notesBar1[1].addModifier(new Articulation(sym1).setPosition(4), 0);
+  notesBar1[2].addModifier(new Articulation(sym1).setPosition(3), 0);
+  notesBar1[3].addModifier(new Articulation(sym1).setPosition(3), 0);
```
rvilarl, line 74: 0s not really required with `Articulation`. I would remove all of them in this file. Up to you.

tests/bach_tests.ts
```diff
@@ -64,17 +64,17 @@ function minuet1(options: TestOptions): void {
   system.addConnector('singleRight');
   system.addConnector('singleLeft');
 
-  id('m1a').addModifier(0, f.Fingering({ number: '5' }));
+  id('m1a').addModifier(f.Fingering({ number: '5' }), 0);
```
rvilarl, line 67: 0s not required in this file. I will no longer comment on this topic on the next files.

rvilarl (changes requested):

0xfe:
Ron, since this change is so pervasive (and will surprise a lot of users), can we also add a console error to `addModifier()` if the first parameter is a number (or not a modifier)? Something like: `Warning: addModifier signature changed to (modifier, index), please update your code.`

ronyeh:
> Ron, since this change is so pervasive (and will surprise a lot of users), can we also add a console error to `addModifier()` if the first parameter is a number (or not a modifier)? Something like: `Warning: addModifier signature changed to (modifier, index), please update your code.`

Our code already has this check:

https://github.com/0xfe/vexflow/blob/master/src/note.ts#L547-L554

We can also list it prominently on the changelog for 4.0.

Another potentially big breaking change is that we eliminated a bunch of helper methods from StaveNote in an effort to reduce circular dependencies: https://github.com/0xfe/vexflow/blob/00ec15c67ff333ea49f4d3defbd9e22374c03684/src/stavenote.js#L795-L833

I think I can re-introduce some or most of the helper methods that we removed without adding any new circular dependencies. I can also add warnings in those methods, because those helper methods also had index first, modifier second:
https://unpkg.com/browse/vexflow@3.0.9/src/stavenote.js

rvilarl:
Hi Ron, the helper methods were agreed with Mohit to be removed. In the end, most of them add an addtional call in the stack and the Dots one were as suggested with Mohit move to Dot with buildAndAttach. It makes no sense to change that back.
I believe we should limit this PR to the signature change.

ronyeh:
> Hi Ron, the helper methods were agreed with Mohit to be removed. In the end, most of them add an addtional call in the stack and the Dots one were as suggested with Mohit move to Dot with buildAndAttach. It makes no sense to change that back. I believe we should limit this PR to the signature change.

Sounds good! Yes, I agree this PR should only handle the parameter swap.

src/note.ts
ronyeh: The super class implementation of addModifier (in Tickable) accomplishes those two lines:

https://github.com/ronyeh/vexflow/blob/f88ca8309b3e9a424675936e0d7f3a335f2dfb8d/src/note.ts#L549-L552

So I added a call to the superclass's implementation instead of duplicating the code:
```
super.addModifier(modifier, index);
```

https://github.com/ronyeh/vexflow/blob/f88ca8309b3e9a424675936e0d7f3a335f2dfb8d/src/tickable.ts#L281-L285

0xfe:
Ron, is this ready for review?

ronyeh:
@0xfe ready for review! If you're looking at the diffs via the GitHub web UI, you can use the File Filter dropdown and uncheck the *.html files (which are all in docs/api/).

src/articulation.ts
ronyeh: I'm going to leave the 0s there for now, to be explicit.

src/tickable.ts
ronyeh: DONE! This actually restores the Tickable API to be the same as version 3.0.9. Not sure when the unused index argument showed up.

3.0.9: https://unpkg.com/browse/vexflow@3.0.9/src/tickable.js

ronyeh: Actually, I see now why we added the extra unused argument. After I made your suggested edit, the tests now throw 12 errors, because we call `getTickables()` and then call `addModifier()` on the tickable with an index argument. 

To fix this, we will have to cast each Tickable to Note.

I'd like to avoid forcing developers to add extra casts `(tickable as Note).addModifier(mod, index)`.

In this case, I think it's convenient to have Tickable have the same API as Note, but just ignore the second argument.

src/note.ts
rvilarl: This is fine

rvilarl (approved):

0xfe:
Lots of toil here -- thanks for the change, Ron. Merging.

PR #1302, opened 2022-01-21
Issue 1301: Fix vertical placement of Bend and Vibrato in tab staves

### LINKED ISSUE #1301: Fix bend/vibrato vertical placement
Bend and annotation now update the modifier context with its height.  One side-effect of this is that the vibrato now appears above the bend, and the 'tap' appears above the bend instead of at the same level.

In the case of vibrato, it already checks for a bend and moves itself down if it detects one.  It just needs to add another line, since the bend is increasing the line height.

For annotation and bend, it doesn't really make sense to do that because an annotation could be anything, there is no way to tell it is the 'Tap' for a bend.  So we can add a specific tap for a bend.  

Is a 'Tap' always the letter 'T'?  Not a guitar person.,..

Should be:

image

image

Is currently:
image
image

AaronDavidNewman:
Fix #1301.

Vibrato places itself below Bend if there's a bend.  Bend has optional 'tap' parameter that puts the tap to the left of the bend, instead of using an Annotation:

image

image

Visual regression showed only the expected changes.

### DIFF
--- a/src/bend.ts
+++ b/src/bend.ts
@@ -147,18 +147,16 @@ export class Bend extends Modifier {
   getText(): string {
     return this.text;
   }
+  getTextHeight(): number {
+    const textFormatter = TextFormatter.create(this.textFont);
+    return textFormatter.maxHeight;
+  }
 
   /** Recalculate width. */
   protected updateWidth(): this {
+    const textFormatter = TextFormatter.create(this.textFont);
     const measureText = (text: string) => {
-      let textWidth: number;
-      const ctx = this.getContext();
-      if (ctx) {
-        textWidth = ctx.measureText(text).width;
-      } else {
-        textWidth = Tables.textWidth(text);
-      }
-      return textWidth;
+      return textFormatter.getWidthForTextInPx(text);
     };
 
     let totalWidth = 0;
--- a/src/bend.ts
+++ b/src/bend.ts
@@ -6,7 +6,6 @@ import { FontInfo } from './font';
 import { Modifier } from './modifier';
 import { ModifierContextState } from './modifiercontext';
 import { TextFormatter } from './textformatter';
-import { Tables } from './tables';
 import { TabNote } from './tabnote';
 import { RuntimeError } from './util';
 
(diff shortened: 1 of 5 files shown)


src/bend.ts
```diff
@@ -5,7 +5,7 @@ import { Element } from './element';
 import { FontInfo } from './font';
 import { Modifier } from './modifier';
 import { ModifierContextState } from './modifiercontext';
-import { Stave } from './stave';
+import { TextFormatter } from './textformatter';
 import { Tables } from './tables';
```
rvilarl, line 9: Tables is no longer required, delete the line.

tests/vibrato_tests.ts
```diff
@@ -11,6 +11,7 @@ const VibratoTests = {
   Start(): void {
     QUnit.module('Vibrato');
     const run = VexFlowTests.runTests;
+    run('Vibrato with Bend', withBend);
```
rvilarl, line 14: This test seems to be called twice. Delete one instance.

tests/vibrato_tests.ts
```diff
@@ -103,8 +104,8 @@ function withBend(options: TestOptions, contextBuilder: ContextBuilder): void {
       positions: [{ str: 2, fret: 10 }],
       duration: 'q',
     })
-      .addModifier(0, new Bend('Full', false))
-      .addModifier(0, new Vibrato().setVibratoWidth(60)),
+      .addModifier(0, new Vibrato().setVibratoWidth(60))
+      .addModifier(0, new Bend('Full', false)),
```
rvilarl, line 108: Does the call order have any effect? If not it seems that vibrato_tests.ts has no change and should stay as it was.

rvilarl (approved):

tests/vibrato_tests.ts
AaronDavidNewman: Changed it back.   Order doesn't matter.

tests/vibrato_tests.ts
AaronDavidNewman: oops...ok

AaronDavidNewman:
ready for merge

0xfe:
LGTM. Thanks for this. We should add this to the changelog.

PR #1303, opened 2022-01-28
adjust puppeteer options for linux

rvilarl:
fixes qunit with linux


ronyeh:
I saw this https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#setting-up-chrome-linux-sandbox

Is there a easy way to set up a sandbox in Linux for VexFlow testing? 

For our purposes, we are just using puppeteer to load flow.html in the qunit task, so I guess it's safe to disable the sandbox to make testing easier. I'll test to make sure this works on my machine, and will merge the PR if it does.

ronyeh:
Works fine on macOS.

PR #1304, opened 2022-01-29
Gruntfile.js improvements

ronyeh:
- skip an unnecessary git commit during the release process
- patch an issue in the parse-path npm package so we can release to the correct repo https://github.com/IonicaBizau/parse-path/pull/32
- small clean up

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -377,12 +377,12 @@ module.exports = (grunt) => {
     // Requires the CJS build to be present in the `build/cjs/` directory (See: grunt build:cjs).
     // The grunt-contrib-qunit package uses puppeteer to load the test page.
     qunit: {
+      files: ['tests/flow-headless-browser.html'],
       options: {
         puppeteer: {
           args: ['--no-sandbox', '--disable-setuid-sandbox'],
         },
       },
-      files: ['tests/flow-headless-browser.html'],
     },
     copy: {
       // grunt copy:reference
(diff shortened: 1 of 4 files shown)


ronyeh:
The patch fixes an edge case where the git repo's owner name is parsed incorrectly because it is recognized as a number, so the parse-path package thinks it is a port number instead of a GitHub user name.

```
git@github.com:0xfe/vexflow.git
```

0xfe:
Thanks Ron. Merging this for now.

It seems like parse-path is unmaintained. Does release-it know about this bug?

ronyeh:
@0xfe Yes I have notified release-it of the bug in one of their dependencies.

https://github.com/release-it/release-it/issues/864


```
vexflow
└─┬ release-it@14.12.3          155454 weekly downloads, last published 7 days ago
  └─┬ git-url-parse@11.6.0      2605676 weekly downloads, last published 5 months ago
    └─┬ git-up@4.0.5            2896490 weekly downloads, last published 7 months ago
      └─┬ parse-url@6.0.0       2915686 weekly downloads, last published 7 months ago
        └── parse-path@4.0.3    2931343 weekly downloads, last published 12 months ago. <== THE BUG IS HERE
```

The odd thing is that the 4 dependencies are all authored the same guy.

https://www.npmjs.com/package/git-url-parse
https://www.npmjs.com/package/git-up
https://www.npmjs.com/package/parse-url
https://www.npmjs.com/package/parse-path

I'm okay with patching the issue for now, since this edge case only affects us when we run the release script.

I'll keep an eye on this though. Perhaps someday we'll roll our own release script. I think it won't be too bad. We'd just need to find the right sequence of `npm`, `git`, and `gh` CLI commands to handle the version bumping, publishing to npm, and releasing to GitHub.

PR #1325, opened 2022-02-05
test for #1295: grace notes & articulations

### LINKED ISSUE #1295: Grace note position is affected by other modifiers / annotations
HI all,

I'm working on using Vexflow for creating some snare drum etudes. 
For some reason, the grace note's position changes if the "main" note has a modifier or annotation. 
This creates a lot of layout mess with snare drum etudes, where accents, grace notes (aka flams) and annotations (aka stickings) are used frequently.

Is this by design? is there a way to avoid this? (ie have all the grace notes in the same relative position to their "main" notes)

Check out fiddle:
https://jsfiddle.net/b681dk9r/ 

Cheers,
Nitzan

rvilarl:
I have first created a new test to observe the problem:
**Articulation GraceNote_Articulation**
Articulation GraceNote_Articulation Bravura

The code related to the problem is at the end on `Articulation.format`:
```js
   const width = articulations
      .map((articulation) => articulation.getWidth())
      .reduce((maxWidth, articWidth) => Math.max(articWidth, maxWidth));

    state.left_shift += width / 2;
    state.right_shift += width / 2;
``` 
I have tried to approaches.

1) **Take into account the actual width**
```js
    const width = articulations
      .map((articulation) => articulation.getWidth())
      .reduce((maxWidth, articWidth) => Math.max(articWidth, maxWidth));

    state.left_shift = Math.max(state.left_shift, width / 2);
    state.right_shift = Math.max(state.right_shift, width / 2);
    
```
2) **Remove it completely**

```js
    /*
    const width = articulations
      .map((articulation) => articulation.getWidth())
      .reduce((maxWidth, articWidth) => Math.max(articWidth, maxWidth));

    state.left_shift += width / 2;
    state.right_shift += width / 2;
    */
```

### DIFF
--- a/src/articulation.ts
+++ b/src/articulation.ts
@@ -258,12 +258,6 @@ export class Articulation extends Modifier {
       }
     });
 
-    const width = articulations
-      .map((articulation) => articulation.getWidth())
-      .reduce((maxWidth, articWidth) => Math.max(articWidth, maxWidth));
-
-    state.left_shift += width / 2;
-    state.right_shift += width / 2;
     return true;
   }
 
--- a/src/articulation.ts
+++ b/src/articulation.ts
@@ -258,6 +258,12 @@ export class Articulation extends Modifier {
       }
     });
 
+    const width = articulations
+      .map((articulation) => articulation.getWidth())
+      .reduce((maxWidth, articWidth) => Math.max(articWidth, maxWidth));
+
+    state.left_shift += width / 2;
+    state.right_shift += width / 2;
     return true;
   }
 
--- a/tests/articulation_tests.ts
+++ b/tests/articulation_tests.ts
@@ -33,7 +33,6 @@ const ArticulationTests = {
     run('Articulation - Fermata Above/Below', drawFermata, { sym1: 'a@a', sym2: 'a@u' });
     run('Articulation - Inline/Multiple', drawArticulations2, { sym1: 'a.', sym2: 'a.' });
     run('TabNote Articulation', tabNotes, { sym1: 'a.', sym2: 'a.' });
-    run('GraceNote Articulation', graceNotes);
   },
 };
 // Helper function for creating StaveNotes.
@@ -426,42 +425,6 @@ function tabNotes(options: TestOptions, contextBuilder: ContextBuilder): void {
   voice.draw(ctx, stave);
 
   ok(true, 'TabNotes successfully drawn');
-}
-
-function graceNotes(options: TestOptions): void {
-  const f = VexFlowTests.makeFactory(options, 700, 130);
-  const stave = f.Stave({ x: 10, y: 10, width: 650 });
-
-  const gracenotes = [{ keys: ['b/4'], duration: '8', slash: false }].map(f.GraceNote.bind(f));
-
-  const notes = [
-    f
-      .StaveNote({ keys: ['c/5'], duration: '4', auto_stem: true })
-      .addModifier(f.GraceNoteGroup({ notes: gracenotes }).beamNotes(), 0),
-    f
-      .StaveNote({ keys: ['c/5'], duration: '4', auto_stem: true })
-      .addModifier(f.GraceNoteGroup({ notes: gracenotes }).beamNotes(), 0),
-    f
-      .StaveNote({ keys: ['c/5'], duration: '4', auto_stem: true })
-      .addModifier(f.GraceNoteGroup({ notes: gracenotes }).beamNotes(), 0),
-    f
-      .StaveNote({ keys: ['c/5'], duration: '4', auto_stem: true })
-      .addModifier(f.GraceNoteGroup({ notes: gracenotes }).beamNotes(), 0),
-  ];
-  notes[1].addModifier(new Articulation('a-').setPosition(3), 0);
-  notes[2].addModifier(new Articulation('a-').setPosition(3), 0);
-  notes[3].addModifier(new Articulation('a-').setPosition(3), 0);
-  notes[2].addModifier(new Articulation('a>').setPosition(3), 0);
-  notes[3].addModifier(new Articulation('a>').setPosition(3), 0);
-  notes[3].addModifier(new Articulation('a@a').setPosition(3), 0);
-
-  const voice = f.Voice().setStrict(false).addTickables(notes);
-
-  new Formatter().joinVoices([voice]).formatToStave([voice], stave);
-
-  f.draw();
-
-  ok(true, 'GraceNoteBasic');
 }
 
 VexFlowTests.register(ArticulationTests);
--- a/tests/articulation_tests.ts
+++ b/tests/articulation_tests.ts
@@ -33,6 +33,7 @@ const ArticulationTests = {
     run('Articulation - Fermata Above/Below', drawFermata, { sym1: 'a@a', sym2: 'a@u' });
     run('Articulation - Inline/Multiple', drawArticulations2, { sym1: 'a.', sym2: 'a.' });
     run('TabNote Articulation', tabNotes, { sym1: 'a.', sym2: 'a.' });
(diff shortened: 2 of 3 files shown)


rvilarl:
Results:
**Original**
Articulation GraceNote_Articulation Bravura
**Approach 1**
Articulation GraceNote_Articulation Bravura_current
**Approach 2**
Articulation GraceNote_Articulation Bravura_current

Both approaches produce a lot of visual differences (21 & 126). Although the approach 2 produces more differences, it causes no overlap and the distribution of notes in the staves is more even. As a consequence I have gone for the removal of the code. This should also increase the performance :).

AaronDavidNewman:
It works ok in these examples, because the articulations won't usually collide with modifiers that are to the left-right of the note, even if they overlap the same `x`.  In the last sample, if the grace note were up an octave for instance, it looks like it would collide.

My thought was to subtract the greatest of note head width or flag width, if > 0 then  that is the thing we divide by 2.  But since the articulations are never that much bigger than note head anyway, maybe it isn't worth it.

rvilarl:
> My thought was to subtract the greatest of note head width or flag width, if > 0 then that is the thing we divide by 2. But since the articulations are never that much bigger than note head anyway, maybe it isn't worth it.

This would be approach 1, above. But I have to say that the result of approach 2 (the one in this PR) gives a better visual result. Perhaps we should consider moving the grace notes, in general, a bit to the left.

AaronDavidNewman:
I was thinking:

``` javascript
  const noteHeadWidth = note.somehowGetMaxNoteheadWidth(); // I'm sure there's a function to do this....
... 
width = Math.max(width - noteHeadWidth, 0);  // using your width calculation
```

Which would be 0 if the articulations were smaller than the note head, and just the extent they exceed the notehead otherwise.   

That said, I think your approach is fine for articulation.  but I think the logic above would work for any horizontally-centered modifier.

rvilarl:
> I was thinking:
> 
> ```js
>   const noteHeadWidth = note.somehowGetMaxNoteheadWidth(); // I'm sure there's a function to do this....
> ... 
> width = Math.max(width - noteHeadWidth, 0);  // using your width calculation
> ```
> 
> Which would be 0 if the articulations were smaller than the note head, and just the extent they exceed the notehead otherwise.

Understood, and I believe that the approach 1 is equivalent. It calculates the required space to the left and to the right as half the size of the articulation and it only modifies `state.left_shift` and `state.right_shift` if half the width of the articulation requires more space.

```js
    const width = articulations
      .map((articulation) => articulation.getWidth())
      .reduce((maxWidth, articWidth) => Math.max(articWidth, maxWidth));

    state.left_shift = Math.max(state.left_shift, width / 2);
    state.right_shift = Math.max(state.right_shift, width / 2);
    
```
Please note that the last two lines do not use `+=`

rvilarl:
@0xfe this is also ready for review.🙂

0xfe:
What are the other visual diffs here? There's two things I'm nervous about here:

1) We're removing the code that manages horizontal spacing when articulations get close together and tight.

2) This seems like we'd have to deal with this for other modifiers too (e.g., Annotation), and a better fix would be in GraceNoteGroup.

rvilarl:
@AaronDavidNewman @0xfe I have to investigate this topic, I do not understand why `Articulation.format` is modifying the position of the grace notes. `GraceNoteGroup.format` is called before `Articulation.format` in `ModifierContext.preFormat()`

AaronDavidNewman:
> I do not understand why Articulation.format is modifying the position of the grace notes. GraceNoteGroup.format

There is some interaction that goes on with grace notes and modifiers, because grace notes are also StaveNotes.  I didn't investigate yet, but one possibility is that the preFormat method of grace note ends up considering the width of articulation.  Another possibility is that the notes are pre-formatted correctly, but moved during the formatting of the measure, due to the extra width on that tick context.

rvilarl:
I believe that it is the second one. Being pushed for being a StaveNote by the TickContext.

AaronDavidNewman:
I think you are right.  So there's your answer - the grace note is formatted like a tickable with 0 ticks.  So it will always end up on the left edge its parent note, which is the note 'x' - the left modifiers.  Even if the modifiers are formatted after the grace note in the ModifierContext, they will be formatted by the time the tick formatting happens.

>> Which would be 0 if the articulations were smaller than the note head, and just the extent they exceed the notehead otherwise.

> Understood, and I believe that the approach 1 is equivalent. 

I'm still not convinced these are the same.  I'll have some time to dig into this today and I can convince myself :) 

> Please note that the last two lines do not use +=

Won't that ignore the width of accidentals that format before the grace notes in ModifierContext?  I think the `+=` is correct, just for a smaller value than we were calculating before (and 0 in many cases).

AaronDavidNewman:
I think you are on to something with the order.  

We are always going to place articulations centered above or below a stemmable note head.  So there's no point in considering the width of anything else when placing them.  Even if there are accidentals, the accidentals should avoid the articulations (and in practice I don't think they would ever collide), not the other way around.  So we can move these to the first modifier in modifier context.

Since articulations are always centered over a stemmable note, we should only report their width as the overlap with the note head (probably we should consider the flag here, also).  Often this is 0 or some small number.  

It makes sense to have grace notes as 'tickables' in their own tick context, because you would always want them to be after the previous notes, in a set of justified voices.  Having them as a tickable accomplishes that.  So I don't think there's anything wrong with gracenotegroup.ts in that sense.

So given all that, I tried this, on top of what you had.  ModifierContent.ts:

``` javascript
    // The ordering below determines when different members are formatted and rendered.
    StaveNote.format(members[Category.StaveNote] as StaveNote[], state);
    Articulation.format(members[Category.Articulation] as Articulation[], state);
    Parenthesis.format(members[Category.Parenthesis] as Parenthesis[], state);
```

articulation.ts
``` javascript
 let maxGlyphWidth = 0;
...
    articulations.forEach((articulation) => {
...
      maxGlyphWidth = Math.max(note.getGlyph().getWidth(), maxGlyphWidth);
...
   });
...
    const overlap = Math.max(width - maxGlyphWidth, 0);

    state.left_shift += overlap / 2;
...
```

That gives us this.  I think this might be the best we can do (I added the accidental on of thee test cases).

image

rvilarl:
What about adding a test with annotations. Some lyrics. I think this will push the grace notes too far away.

AaronDavidNewman:
Now that I think of it, we can keep the modifier context order the same, and expand this to:

``` javascript
const overlap = Math.min(Math.max(width - maxGlyphWidth, 0), Math.max(width - state.left_shift * 2, 0));
```
That should work for any modifier that is centered above or below the note and stacked vertically.  I think this is what you were going for with solution 1.

image

Annotations don't have to be stacked vertically or tied to stemmable notes, in theory.  But for all our test cases this looked OK.  Curious to see how many diffs this causes.

rvilarl:
> Won't that ignore the width of accidentals that format before the grace notes in ModifierContext?

INo, the accidentals are considered in the value of `state.left_shift`
```js
    const width = articulations
      .map((articulation) => articulation.getWidth())
      .reduce((maxWidth, articWidth) => Math.max(articWidth, maxWidth));

    state.left_shift = Math.max(state.left_shift, width / 2);
    state.right_shift = Math.max(state.right_shift, width / 2);
    
```

rvilarl:
> image
> 
> Annotations don't have to be stacked vertically or tied to stemmable notes, in theory. But for all our test cases this looked OK. Curious to see how many diffs this causes.

I think that the annotation does not have a visual impact because of the `#`. Could you try placing `word` on the second note?

AaronDavidNewman:
> I think that the annotation does not have a visual impact because of the #.

image

>  you can make then a PR to solve it as already proprosed.

Sure!

rvilarl:
@AaronDavidNewman go for it

PR #1327, opened 2022-02-10
Add SVG classes (+ ids partly) to beams, stems within beams, staveties (OSMD VexflowPatch)

sschmidTU:
Without this PR, staveties, beams and stems within beams are just random `<path>` elements in the SVG Dom, not inside a `<g>` node with a class (and sometimes id), which non-beamed stems and most other elements like key signatures and clefs are at this point:
image
Instead, these 3 path nodes will be within `<g class='vf-stem/beam'>`  groups too now.

There are a few differences in how the SVG classes are created.
For staveties, we just save the SVG node as an attribute 'el' to the stavetie, so the SVG node doesn't need an ID -> you can get it by `stavetie.getAttribute('el')`.
  (alternatively, we could set the id by taking the ID of the start note + `-tie` if you want to get the info from the SVG alone)
Same for beams.
For stems within beams, the id of the note is added to the stem id, because there can be multiple stems within the beam, so we need a way to correlate them to their note, to say which stem belongs to which note.

Merging this would basically complete OSMD's Vexflowpatch integration, at least when #1263 is merged (and a few more bugs are fixed).

------

I couldn't run tests, but there should be no visual or functional changes other than adding the SVG classes.

My `grunt`, `grunt test` and `grunt reference` processes keep getting killed for some reason.

image
(I don't think my Ctrl+C is sticky)
Maybe it's running out of RAM? But then it should use Swap or something. I've configured the VM to have 1.5GB of RAM, which I thought should be enough for a Linux machine running a console running npm/grunt tasks.

`grunt test` aborted with this error (likely just a result of getting killed):
```console
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^

RpcIpcMessagePortClosedError: Cannot send the message - the message port has been closed for the process 2749.
    at /home/simon/vexflow/vexflow-patch/node_modules/fork-ts-checker-webpack-plugin/lib/rpc/rpc-ipc/RpcIpcMessagePort.js:47:47
```
ran on Linux Mint 20.2 in a VM (VirtualBox on Windows host).
Looks like the process is getting killed, which also happened when running `grunt` the first time, though that succeeded the second time.

### DIFF
--- a/src/beam.ts
+++ b/src/beam.ts
@@ -851,7 +851,9 @@ export class Beam extends Element {
     this.notes.forEach((note) => {
       const stem = note.getStem();
       if (stem) {
+        ctx.openGroup('stem', note.getAttribute('id') + '-stem');
         stem.setContext(ctx).draw();
+        ctx.closeGroup();
       }
     }, this);
   }
--- a/src/beam.ts
+++ b/src/beam.ts
@@ -879,6 +879,7 @@ export class Beam extends Element {
         if (lastBeamX) {
           const lastBeamY = this.getSlopeY(lastBeamX, firstStemX, beamY, this.slope);
 
+          this.setAttribute('el', ctx.openGroup('beam'));
           ctx.beginPath();
           ctx.moveTo(startBeamX, startBeamY);
           ctx.lineTo(startBeamX, startBeamY + beamThickness);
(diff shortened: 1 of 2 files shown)


rvilarl (approved):
@0xfe I have run the tests and there is no visual differences

ronyeh:
I tested this in flow.html with the Chrome inspector to verify the `<g>` groups are added.
I also verified that there are no visual diffs.

Thanks, Simon!

PR #1328, opened 2022-02-13
fix #1295 by more accurately reporting articulation and annotation width

### LINKED ISSUE #1295: Grace note position is affected by other modifiers / annotations

AaronDavidNewman:
Prevent grace notes from being pushed too far away from the note by annotations and articulations, because those modifiers reported incorrect width (too high).  For centered, vertically stacked modifiers, subtract the note width from the modifier width and report the overlap.

old:
image

new:
image

This causes 298 visual differences.  Most of them are not visible to me, if I didn't see the diff.  Our perceptive hash is super sensitive to text changes.  

image

Some new test cases take up less room, but I didn't see any collisions.  This is an example:
old:
image
new:
image

### DIFF
--- a/src/annotation.ts
+++ b/src/annotation.ts
@@ -73,6 +73,7 @@ export class Annotation extends Modifier {
   static format(annotations: Annotation[], state: ModifierContextState): boolean {
     if (!annotations || annotations.length === 0) return false;
     let width = 0;
+    let maxGlyphWidth = 0;
     for (let i = 0; i < annotations.length; ++i) {
       const annotation = annotations[i];
       const textFormatter = TextFormatter.create(annotation.textFont);
@@ -81,6 +82,8 @@ export class Annotation extends Modifier {
       let verticalSpaceNeeded = textLines;
 
       const note = annotation.checkAttachedNote();
+      maxGlyphWidth = Math.max(note.getGlyph().getWidth(), maxGlyphWidth);
+
       const stave: Stave | undefined = note.getStave();
       const stemDirection = note.hasStem() ? note.getStemDirection() : Stem.UP;
       let stemHeight = 0;
@@ -146,8 +149,9 @@ export class Annotation extends Modifier {
         annotation.setTextLine(state.text_line);
       }
     }
-    state.left_shift += width / 2;
-    state.right_shift += width / 2;
+    const overlap = Math.min(Math.max(width - maxGlyphWidth, 0), Math.max(width - state.left_shift * 2, 0));
+    state.left_shift += overlap / 2;
+    state.right_shift += overlap / 2;
     return true;
   }
 
--- a/src/annotation.ts
+++ b/src/annotation.ts
@@ -72,8 +72,10 @@ export class Annotation extends Modifier {
   /** Arrange annotations within a `ModifierContext` */
   static format(annotations: Annotation[], state: ModifierContextState): boolean {
     if (!annotations || annotations.length === 0) return false;
-    let width = 0;
-    let maxGlyphWidth = 0;
+    let leftWidth = 0;
+    let rightWidth = 0;
+    let maxLeftGlyphWidth = 0;
+    let maxRightGlyphWidth = 0;
     for (let i = 0; i < annotations.length; ++i) {
       const annotation = annotations[i];
       const textFormatter = TextFormatter.create(annotation.textFont);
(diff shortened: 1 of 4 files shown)


src/annotation.ts
```diff
@@ -146,8 +149,9 @@ export class Annotation extends Modifier {
         annotation.setTextLine(state.text_line);
       }
     }
-    state.left_shift += width / 2;
-    state.right_shift += width / 2;
+    const overlap = Math.min(Math.max(width - maxGlyphWidth, 0), Math.max(width - state.left_shift * 2, 0));
```
rvilarl, line 152: why is `state.right_shift` not part of the equation?

tests/articulation_tests.ts
```diff
@@ -24,6 +26,7 @@ const ArticulationTests = {
   Start(): void {
     QUnit.module('Articulation');
     const run = VexFlowTests.runTests;
+    run('Articulation - Grace Notes', graceNotes);
```
rvilarl, line 29: It would make more sense to place this test now on gracenote test becuase it combines grace notes not only with articulations but also with annotations and accidentals.

I suggest that you revert this test file from master and I open my PR again.

rvilarl (changes requested):

tests/articulation_tests.ts
AaronDavidNewman: ok

src/annotation.ts
AaronDavidNewman: fixed

tests/gracenote_tests.ts
```diff
@@ -106,7 +110,43 @@ function basic(options: TestOptions): void {
...
+  notes[3].addModifier(new Articulation('a@a').setPosition(3), 0);
+
+  const voice = f.Voice().setStrict(false).addTickables(notes);
 
+  new Formatter().joinVoices([voice]).formatToStave([voice], stave);
+
+  f.draw();
+
+  ok(true, 'GraceNoteBasic');
+}
```
rvilarl, line 149: `beamNotes()` is not required I also think that leaving the modifiers close to the notes makes it easier to understand. This is how I left it:
```js
function graceNoteModifiers(options: TestOptions): void {
  const f = VexFlowTests.makeFactory(options, 700, 130);
  const stave = f.Stave({ x: 10, y: 10, width: 650 });

  const gracenotes = [{ keys: ['b/4'], duration: '8', slash: false }].map(f.GraceNote.bind(f));

  const notes = [
    f
      .StaveNote({ keys: ['c/5'], duration: '4', auto_stem: true })
      .addModifier(f.GraceNoteGroup({ notes: gracenotes }), 0),
    f
      .StaveNote({ keys: ['c/5'], duration: '4', auto_stem: true })
      .addModifier(f.GraceNoteGroup({ notes: gracenotes }), 0)
      .addModifier(new Articulation('a-').setPosition(3), 0),
    f
      .StaveNote({ keys: ['c/5'], duration: '4', auto_stem: true })
      .addModifier(f.GraceNoteGroup({ notes: gracenotes }), 0)
      .addModifier(new Articulation('a-').setPosition(3), 0)
      .addModifier(new Accidental('#')),
    f
      .StaveNote({ keys: ['c/5'], duration: '4', auto_stem: true })
      .addModifier(f.GraceNoteGroup({ notes: gracenotes }), 0)
      .addModifier(new Articulation('a-').setPosition(3), 0)
      .addModifier(new Annotation('words')),
    f
      .StaveNote({ keys: ['c/5'], duration: '4', auto_stem: true })
      .addModifier(f.GraceNoteGroup({ notes: gracenotes }), 0)
      .addModifier(new Articulation('a-').setPosition(3), 0)
      .addModifier(new Articulation('a>').setPosition(3), 0),
    f
      .StaveNote({ keys: ['c/5'], duration: '4', auto_stem: true })
      .addModifier(f.GraceNoteGroup({ notes: gracenotes }), 0)
      .addModifier(new Articulation('a-').setPosition(3), 0)
      .addModifier(new Articulation('a>').setPosition(3), 0)
      .addModifier(new Articulation('a@a').setPosition(3), 0),
  ];

  const voice = f.Voice().setStrict(false).addTickables(notes);

  new Formatter().joinVoices([voice]).formatToStave([voice], stave);

  f.draw();

  ok(true, 'GraceNoteModifiers');
}

```

rvilarl: This is also why in new and old images above the number of notes is different.

src/annotation.ts
```diff
@@ -146,8 +149,12 @@ export class Annotation extends Modifier {
...
       }
     }
-    state.left_shift += width / 2;
-    state.right_shift += width / 2;
+    const overlap = Math.min(
+      Math.max(width - maxGlyphWidth, 0),
+      Math.max(width - (state.left_shift + state.right_shift), 0)
+    );
+    state.left_shift += overlap / 2;
+    state.right_shift += overlap / 2;
```
rvilarl, line 157: I just realised that this calculations are ignoring `this.horizontalJustification`, are not day? Also shifted notes. The second problem affects articulations and accidentals as well.

rvilarl: Of course the previuos ones were also ignoring that.

AaronDavidNewman: You are right.  The horizontal formatting has never considered horizontal justification, and none of our test cases exercise this.  

I didn't think about the left and right shift being different, but that certainly could be.  We could consider a left and right Max in the for loop, and then a left and right overlap.  That should reduce to the same thing in the case of being centered. 2x as many calculations but the same logic.

0xfe:
Just jumping in here. Aaron, is this ready for review, or should I hold off? (I noticed Rodrigo has some changes requested.)

AaronDavidNewman:
afaik all the changes have been made

rvilarl (approved):

0xfe (approved):
Looks great. Merging.

PR #1332, opened 2022-02-23
Formatter Whitespace_and_justify: keep num_beats constant

rvilarl:
Reviewing the tests' output I thought that in the test below there was a rendering error: Some notes were missing. 
The problem is that the test did not keep num_beats contant. 
I suggest this change to avoid that someone else also thinks that the rendering is wrong. Bellow the Bravura version:

 **Formatter Whitespace_and_justify Bravura**
Formatter Whitespace_and_justify Bravura
**current**
Formatter Whitespace_and_justify Bravura_current
**reference(master)**
Formatter Whitespace_and_justify Bravura_reference


0xfe:
Good catch!

PR #1334, opened 2022-02-27
Release script

ronyeh:
Fix the release script:
* Do not assume that the typescript compiler `tsc` is installed globally.
* Patch the buggy `parse-path` package before release-it is imported via `require()`.


ronyeh:
I believe this fixes the release script so that VexFlow can be released to NPM and GitHub from a completely fresh clone of the repo.

0xfe:
Thanks, looks good. Will try a release.

PR #1337, opened 2022-03-02
Update README.

ronyeh:
In preparation for VexFlow 4 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,57 +1,35 @@
-# VexFlow 4 (beta)
+# VexFlow 4
 
-A JavaScript / TypeScript library for rendering music notation.
-
-Copyright (c) 2010 Mohit Muthanna Cheppudira
+VexFlow is an open-source music notation rendering library. It is written in TypeScript and compiled to ES6. VexFlow supports HTML
+Canvas and SVG and runs right in the browser.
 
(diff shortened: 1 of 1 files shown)


mscuthbert:
Can I suggest one little change:

```javascript
// skip this next line if using the <script> tag:
import Vex from 'vexflow';
```
we'll have a lot of users who are very new to web development and might not know this.  (We might even consider for the `<script>` examples to not give the URL but show the full:

```
<script src="https://unpkg.com/vexflow/build/cjs/vexflow-debug.js"></script>
```

Note also that I've left off the version number in the link (so it won't install v4 until it's fully released).  This way people who just want to use *some* Vexflow will not be hardcoding a beta version or something that has a security hole, etc.   And then we don't need to update the README with each minor version.  We could also do `vexflow@^4/build` and then only need to update here every once in a while.

Otherwise all approved and looks good to me.

ronyeh:
Thanks for the feedback!

I agree, and I think I will concentrate on the simplest case for the README (regular script tag deployment). The README will also mention that there are other ways of integrating (npm install, typescript, etc.). We can link out to the wiki for more in depth walk throughs for each method of integrating VexFlow.

PR #1339, opened 2022-03-03
Address bug 1335, right padding for longer last note in measure

AaronDavidNewman:
diffs and more info to follow

### DIFF
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -774,19 +774,6 @@ export class Formatter {
     let targetWidth = adjustedJustifyWidth;
     const distances = calculateIdealDistances(targetWidth);
     let actualWidth = shiftToIdealDistances(distances);
-    // Calculate right justification by finding max of (configured value, min distance between tickables)
-    // so measures with lots of white space use it evenly, and crowded measures use at least the configured
-    // space.
-    const calcMinDistance = (targetWidth: number, distances: Distance[]) => {
-      let mdCalc = targetWidth / 2;
-      if (distances.length > 1) {
-        for (let di = 1; di < distances.length; ++di) {
-          mdCalc = Math.min(distances[di].expectedDistance / 2, mdCalc);
-        }
-      }
-      return mdCalc;
-    };
-    const minDistance = calcMinDistance(targetWidth, distances);
 
     // Just one context. Done formatting.
     if (contextList.length === 1) return 0;
@@ -794,8 +781,16 @@ export class Formatter {
     // right justify to either the configured padding, or the min distance between notes, whichever is greatest.
     // This * 2 keeps the existing formatting unless there is 'a lot' of extra whitespace, which won't break
     // existing visual regression tests.
-    const paddingMax = configMaxPadding * 2 < minDistance ? minDistance : configMaxPadding;
-    const paddingMin = paddingMax - (configMaxPadding - configMinPadding);
+    const paddingMaxCalc = (curTargetWidth: number) => {
+      let lastTickablePadding = configMaxPadding * 2;
+      const lastTickable = lastContext && lastContext.getMaxTickable();
+      if (lastTickable) {
+        lastTickablePadding = lastTickable.getVoice().softmax(lastContext.getMaxTicks().value()) * curTargetWidth;
+      }
+      return configMaxPadding * 2 < lastTickablePadding ? lastTickablePadding : configMaxPadding;
+    };
+    let paddingMax = paddingMaxCalc(targetWidth);
+    let paddingMin = paddingMax - (configMaxPadding - configMinPadding);
     const maxX = adjustedJustifyWidth - paddingMin;
 
     let iterations = maxIterations;
@@ -803,6 +798,8 @@ export class Formatter {
     // without going over
     while ((actualWidth > maxX && iterations > 0) || (actualWidth + paddingMax < maxX && iterations > 1)) {
       targetWidth -= actualWidth - maxX;
+      paddingMax = paddingMaxCalc(targetWidth);
+      paddingMin = paddingMax - (configMaxPadding - configMinPadding);
       actualWidth = shiftToIdealDistances(calculateIdealDistances(targetWidth));
       iterations--;
     }
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -786,10 +786,15 @@ export class Formatter {
       let lastTickablePadding = 0;
       const lastTickable = lastContext && lastContext.getMaxTickable();
       if (lastTickable) {
+        const voice = lastTickable.getVoice();
+        // If the number of actual ticks in the measure <> configured ticks, right-justify
+        // because the softmax won't yield the correct value
+        if (voice.getTicksUsed().value() !== voice.getTotalTicks().value()) {
+          return configMaxPadding;
+        }
         const tickWidth = lastTickable.getWidth();
         lastTickablePadding =
-          lastTickable.getVoice().softmax(lastContext.getMaxTicks().value()) * curTargetWidth -
-          (tickWidth + leftPadding);
+          voice.softmax(lastContext.getMaxTicks().value()) * curTargetWidth - (tickWidth + leftPadding);
       }
       return configMaxPadding * 2 < lastTickablePadding ? lastTickablePadding : configMaxPadding;
     };
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -789,7 +789,7 @@ export class Formatter {
         const voice = lastTickable.getVoice();
         // If the number of actual ticks in the measure <> configured ticks, right-justify
         // because the softmax won't yield the correct value
-        if (voice.getTicksUsed().value() !== voice.getTotalTicks().value()) {
+        if (voice.getTicksUsed().value() > voice.getTotalTicks().value()) {
           return configMaxPadding;
         }
         const tickWidth = lastTickable.getWidth();
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -778,6 +778,16 @@ export class Formatter {
 
     // Just one context. Done formatting.
     if (contextList.length === 1) return 0;
+    const calcMinDistance = (targetWidth: number, distances: Distance[]) => {
+      let mdCalc = targetWidth / 2;
+      if (distances.length > 1) {
+        for (let di = 1; di < distances.length; ++di) {
+          mdCalc = Math.min(distances[di].expectedDistance / 2, mdCalc);
+        }
+      }
+      return mdCalc;
+    };
+    const minDistance = calcMinDistance(targetWidth, distances);
 
     // right justify to either the configured padding, or the min distance between notes, whichever is greatest.
     // This * 2 keeps the existing formatting unless there is 'a lot' of extra whitespace, which won't break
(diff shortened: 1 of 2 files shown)


RyoSusami:
It may be a misunderstanding because I still have a poor understanding of this area, but I will report it just in case.
Regarding the case where the space at the end of the bar is large in the Bach demo, isn't Stave.defaultPadding on line 222 of system.ts correctly Stave.rightPadding?

0xfe:
Thanks Aaron. This is an interesting approach, and probably the right one. However, I think it should only be applied in some cases where the extra spacing is required for readability, not all the cases. (I'm not exactly sure what all those cases are -- certainly the example in the bug is one of them.)

AaronDavidNewman:
> Stave.defaultPadding on line 222 of system.ts correctly Stave.rightPadding? 

@RyoSusami in the music font metrics, 'stave.padding' is the left padding - the (relative) start X of the first note.  'stave.endPaddingMin' and 'stave.endPaddingMax' is the right padding.  'stave.defaultPadding' is the sum of these.

If you have same width for the stave and the voice, the notes will be crowded.  If you add `defaultPadding` (sum of left and right) to the staff width vs. the justify width, the notes will fit.    When you 'formatToStave' this is how we calculate the widths of voice vs. staff.

Mostly this is the case when space is limited - this PR is mostly about when there is plenty of white space, where does the right padding come from.

AaronDavidNewman:
@rvilarl  @0xfe I made a small tweak, we should check the diffs again.  I subtracted the width of the tick and the left padding from the value, since the distance from the right edge of the note to the next note will be those things.  This looks better in many cases. 

For the automatic accidental test cases, I think the issue is the test case.  The width of the measure is 11 beats, but the time signature is 4 beats so the softmax value there is too high.  Maybe this can be one of the reasons for just right-justifying it.  Others that I've checked look better with the change.

rvilarl:
@AaronDavidNewman you are right there are new and more cases looking better, for example the one below (bravura, current first, then reference)
Accidental Accidental_Padding Bravura_current
Accidental Accidental_Padding Bravura_reference

I still found a few where the space is still too big (two examples below):
Accidental Automatic_Accidentals___C_major_scale_in_Ab Bravura_current
Accidental Automatic_Accidentals___C_major_scale_in_Ab Bravura_reference
Annotation Placement Bravura_current
Annotation Placement Bravura_reference

AaronDavidNewman:
This change only applies the new logic if number of ticks in the measure matches the configured value. This should fix the accidental cases where there are 11 beats in a 4/4 measure.

I need to look into the issue with the chord change tests.  The issue is the extra width of the annotations.  I think in this case, the earlier 4.0 logic looked better, not sure how easy it will be to detect this.

AaronDavidNewman:
The issue with the chord changes is a different problem, with the chord module itself.  The width reported by chordsymbol is too large, because it doesn't take account the note width.

That will be a separate PR.  So I think we can ignore that issue.

AaronDavidNewman:
This still looks too big to me:

image

I think the issue is the softmax factor.  With 100, this measure yields 1/2 note having about 3x the width of a 1/4 note.  Here's what it looks like if the 1/2 note is first.  Softmax factor of 100 works well justifying voices to the same width, but things get stretchy if there is a lot of space.

image

Here's the same example with a softmax factor of 10:
image

so it's really the same problem.  With this PR, the issue is consistent at least.  

We could globally reduce default softmax, but that would change _everything_.  The key thing would be to check some of the formatter multi-voice test cases and see if they still look good with a lower softmax factor.  Maybe we could dynamically change the softmax factor based on the number of voices (this is a configurable parameter but I doubt people want to be manually setting that everywhere).

rvilarl:
@0xfe @AaronDavidNewman @ronyeh taking into account the info below, shouldn't we discuss and completely re-factor the `preFormat` function?

Leaving the softmax topic aside,some of the improvements have been lost, for example the one below where there are now no diffs
Accidental Accidental_Padding Bravura_reference

there are improvements
Barline Simple_BarNotes Bravura_current
Barline Simple_BarNotes Bravura_reference
Barline Style_BarNotes Bravura_current
Barline Style_BarNotes Bravura_reference

there are still a lot of cases though where we are not getting the expected result
Accidental Automatic_Accidentals___Multi_Voice_Inline Bravura_current
Accidental Automatic_Accidentals___Multi_Voice_Inline Bravura_reference
Accidental Automatic_Accidentals___No_Accidentals_Necessary Bravura_current
Accidental Automatic_Accidentals___No_Accidentals_Necessary Bravura_reference
Annotation Bottom_Annotation Bravura_current
Annotation Bottom_Annotation Bravura_reference
Annotation Placement Bravura_current
Annotation Placement Bravura_reference
Beam Lengthy_Beam Bravura_current
Beam Lengthy_Beam Bravura_reference
Beam Sixteenth_Beam Bravura_current
Beam Sixteenth_Beam Bravura_reference
Bend Double_Bends Bravura_current
Bend Double_Bends Bravura_reference
Bend Double_Bends_With_Release Bravura_current
Bend Double_Bends_With_Release Bravura_reference
ChordSymbol Bottom_Chord_Symbols Bravura_current
ChordSymbol Bottom_Chord_Symbols Bravura_reference
ChordSymbol Bottom_Stem_Down_Chord_Symbols Bravura_current
ChordSymbol Bottom_Stem_Down_Chord_Symbols Bravura_reference
ChordSymbol Chord_Symbol_Font_Size_Tests Bravura_current
ChordSymbol Chord_Symbol_Font_Size_Tests Bravura_reference
ChordSymbol Chord_Symbol_Kerning_Tests Bravura_current
ChordSymbol Chord_Symbol_Kerning_Tests Bravura_reference
ChordSymbol Double_Bottom_Chord_Symbols Bravura_current
ChordSymbol Double_Bottom_Chord_Symbols Bravura_reference
ChordSymbol Top_Chord_Symbols Bravura_current
ChordSymbol Top_Chord_Symbols Bravura_reference
ChordSymbol Top_Chord_Symbols_Justified Bravura_current
ChordSymbol Top_Chord_Symbols_Justified Bravura_reference
Dot Multi_Voice Bravura_current
Dot Multi_Voice Bravura_reference
EasyScore Draw_Parenthesised Bravura_current
EasyScore Draw_Parenthesised Bravura_reference

This is not complete, there are 100s from changes

ronyeh:
Thanks, @AaronDavidNewman  for working on this. I think this is an important topic to address, since good formatting is central to a good VexFlow developer experience.

Once we are all happy with the results of this new formatter, let me know. We can merge this into the release candidate, and have it included with the VexFlow 4 release.

ronyeh:
I'm not very familiar with the formatter :-). It seems like ultimately, our formatter needs to try its best to set the TickContext's x value to a good location within the measure.

```
context.setX(...);
```

In a future PR, perhaps we can add a easy feature to let the developers **hand tweak** a `TickContext` x / y location, after the fact? Basically, we add an API to set an absolute x / y value for any `TickContext`. This override is applied after the formatting is done, but before display. I think this would be valuable because no automatic algorithm can ever account for all edge cases. The music engraver should always have the final say (similar to how a graphic designer might manually kern the font).



Anyways, I'd like to share my process for quickly hacking on VexFlow and comparing the changes against the head revision. My guide is here: https://github.com/0xfe/vexflow/wiki/Debugging-with-the-ESM-Build

It's also summarized below:


# Use the ESM Build for Rapid Iteration


This requires two clones of the VexFlow repo.

In repo A, check out the master branch and run the following command:

```
grunt test:browser:esm
```

This compiles the ES modules build in watch mode. It loads `flow.html` in a local web server so that you can use the compiled library in `vexflow/build/esm/`.


In repo B, check out a recent PR. The Github CLI command is:


```
gh pr checkout https://github.com/0xfe/vexflow/pull/1339
```

In repo B, build the ESM version and open the `flow.html` test page:

```
grunt test:browser:esm
```

The second server usually starts at port 8081.


We now have two different versions in adjacent browser tabs. Quickly flip between the two tabs to see any differences.

Edit the VexFlow source in one of the repos, and wait a moment for the ESM watch to recompile. Refresh the tab for the repo that you just modified.

The ESM build includes individual unminimized JS files! This allows us to use the browser's developer tools to set breakpoints in the compiled ESM files.

AaronDavidNewman:
This is ready for review.  In the interest of reducing the number of changes, I added the old logic back in the case where the number of ticks actually in the measure did not match the number of ticks in the measure (e.g. ,3 or 5 beats in a 4 beat measure).  This covers a large number of our test cases.  (I'm having trouble getting the correct image reference images on my Linux box, so I don't have an exact number).

If the number of beats in the measure is correct, and there is enough room, the distance from the last note and the default first note of the next measure will about the same is the distance after the note would be, if it occurred earlier in the measure.

We still need to address the issue of the softmax value being too high.  Until we do that, the gap following a long note like that of the original in issue 1335 will be too large (whether the last note of the measure or not).

I also think we can ignore the changes to the chord change module.  This also need a new PR similar to what we did with articulation and annotation, to subtract the note width from centered chords.
old:
image
new (s/b same):
image

old:
image
new (s/b same):
image

old:
image
new (s/b same - last note is shortest value):
image

old:
image
new (s/b different - more space for last 1/4 note):
image

It is interesting how seldom the last note is the longest in the measure.  In looking at all genres of music in my house, only a few bars from a long piece will have measures like that.

ronyeh:
I won't have a chance to look at this until maybe early next week. @rvilarl Shall we agree that this is the highest numbered PR that will go into VexFlow 4.0.1? Any PR with a number that is > 1339 should be bumped to release 4.1 (unless it is a critical bug fix). I think there is value in getting a release candidate out there, and then shipping VexFlow 4 so more people can give us feedback.

Regarding formatting, I think a longer term strategy might be to allow a developer to swap out the default formatter function by calling:

```typescript
// Provide a formatter name, or a function that can format Elements on a stave.
Vex.Flow.setFormatter( f: string | Function )
```

So we could do stuff like:
```
Vex.Flow.setFormatter('default') // Use the default formatter. Normally, calling this is not necessary.
Vex.Flow.setFormatter('simple'); // Use the simple formatter which is appropriate for quick testing.
Vex.Flow.setFormatter(myFormatterFunction); // Use a custom formatter function.
```

rvilarl:
>  This covers a large number of our test cases. (I'm having trouble getting the correct image reference images on my Linux box, so I don't have an exact number).

@AaronDavidNewman find attached a zip with all the differences (bravura only) There are still unexpected outputs for example on Text tests
[bravura.zip](https://github.com/0xfe/vexflow/files/8231524/bravura.zip)

rvilarl:
> @rvilarl Shall we agree that this is the highest numbered PR that will go into VexFlow 4.0.1? Any PR with a number that is > 1339 should be bumped to release 4.1 (unless it is a critical bug fix). 

Agree. It could be that others are also merged in the time frame and I would not stop that. But once all <= 1339 are merged we should release. We are all wishing to have a release :)

AaronDavidNewman:
I would not say they're unexpected.  The changes aren't really related to the test cases though, so it's a bit of noise.

Mostly I would say they are better.  Even with the wee size, you can tell the last note is longer than the others here:

image

rvilarl:
You are right

AaronDavidNewman:
@rvilarl  what am I looking for?  The last note is further from the right, the first note position is fixed so the other notes in the measure will be changed also.  All expected.   Using getBBox(), you can tell that the last note is 90 px from the right side, the 3rd note is 121 px from the last note (because of the justification, third note is wider)

rvilarl:
> @rvilarl what am I looking for? The last note is further from the right, the first note position is fixed so the other notes in the measure will be changed also. All expected. Using getBBox(), you can tell that the last note is 90 px from the right side, the 3rd note is 121 px from the last note (because of the justification, third note is wider)

Sorry @0xfe ready to review

0xfe:
Thanks for this change. I think it's a good improvement over the current formatting. Re: `softmaxFactor`, it's a tough one -- the default is `100`, but it's meant to be customized because different values look better with different combinations of note durations.

I think if we explored hard enough we could find some pattern which we could use to automatically calculate the `softmaxFactor`, but I haven't figured it out yet.

PR #1340, opened 2022-03-04
feat(beam): Add method to set partial beam direction

gristow:
The direction of partial beams (beams at a 16th note level subdivision or smaller that do not connect to another beam) can convey important musical information to performers about how to group notes together. For instance, when in 6/8, the two versions below imply different rhythmic accentuation:

image

This PR adds `beam.setPartialBeamDirectionAt(noteIndex, beamDirection)` and `beam.unsetPartialBeamDirectionAt(noteIndex)`, as well as a `partialBeamDirections` option on `EasyScore.beam`.


src/beam.ts
```diff
@@ -74,6 +76,15 @@ export class Beam extends Element {
   private beam_count: number;
   private unbeamable?: boolean;
 
+  /**
+   * Overrides to default beam directions for secondary-level beams that do not
+   * connect to any other note. See further explanation at
+   * `setPartialBeamSideAt`
+   */
+  private forced_partial_directions: {
```
ronyeh, line 84: Can we make this camelCase (i.e. `forcedPartialDirection`)?

Same with beam_count => beamCount above. I think JS style guides generally avoid underscores in variable names, except with UPPER_CASE_CONSTANTS.

ronyeh: I just realized beam_count isn't part of your PR, haha... and it looks like beam.ts has a lot of underscore_variables already. I do think new contributions should move toward camelCase, and it's possible that in some future (breaking) VexFlow, we'll move everything to camelCase.

gristow: Yes, tons of snake_case in beam.ts currently. Most of them are for private class members. Shall I go ahead and convert those as well?

ronyeh: We could convert them in a subsequent PR. It'll make this one easier to review/test. I'm not a music expert, but the new feature looks cool!

gristow: Ok, great -- I'll change the snake_cases that I added to camelCase and commit, but leave the existing variables/class members unchanged. Let me know if there's anything else you see that needs attention. Thanks!

src/easyscore.ts
```diff
@@ -1,6 +1,7 @@
 // [VexFlow](https://vexflow.com) - Copyright (c) Mohit Muthanna 2010.
 // MIT License
 
+import { PartialBeamDirection } from '.';
```
rvilarl, line 4: `from '.'` should be `from './beam`

gristow: Good catch -- that was a VS Code autocomplete import, I'll watch for that in the future.

ronyeh: I think at some point I'll suggest a VSCode workspace settings change, so that it will prefer to auto import files with the .js extension, which is more explicit and also supported by TypeScript. By default it opts for the shortest import, and currently importing from '.' can break our ESM build.

ronyeh:
This looks good to me! I checked out the branch and tested it. No visual diffs, except for the new test case of course.

Welcome back, @gristow !

PR #1341, opened 2022-03-05
feat(annotation): Apply styles from setStyle when rendering

gristow:
This PR enables rendering styles applied using the base `Element` class's `setStyle()` method when drawing annotations.


rvilarl (approved):

0xfe:
Looks good, thanks!

PR #1343, opened 2022-03-06
feat(StaveConnector) add getType(), tighten setType() typings

gristow:
This PR:
* Adds a `getType()` method to match the existing `setType()` method on `StaveConnector`
* Tightens the parameters definition on `setType()` to use a union of acceptable types, instead of all strings and numbers. (This also improves intellisense completion, see below.)

Previously:
image

After Change:
image


src/staveconnector.ts
```diff
@@ -30,6 +30,13 @@ function drawBoldDoubleLine(ctx: RenderContext, type: number, topX: number, topY
   ctx.fillRect(topX - thickLineOffset, topY, variableWidth, botY - topY);
 }
 
+/**
+ * see {@link StaveConnector.type} & {@link StaveConnector.typeString}
+ */
+export type StaveConnectorType =
+  | typeof StaveConnector['type'][keyof typeof StaveConnector['type']]
+  | keyof typeof StaveConnector['typeString'];
```
0xfe, line 38: I like this PR, but I'm having a hard time grokking this type definition. @ronyeh thoughts on this?

gristow: @0xfe --just pushed a revision that makes the StaveConnectorType definition cleaner and sets the types of StaveConnector.type and StaveConnector.typeString by using the `StaveConnectorType` definition, rather than the other way around. Hopefully it's a lot more readable/grokkable.

0xfe: Thanks @gristow -- this is definitely better. Can you add a comment explaining this type so future me understands it :-)

0xfe:
Looks great, merging!

PR #1344, opened 2022-03-06
feat(glyph): Expose getXShift() and getYShift() methods

gristow:
This pr adds `getXShift()` and `getYShift()` methods to match the setters in the `Glyph` class. 

(Since glyphs can be rendered directly to a point or rendered to a staff, calculating their rendered bounding box for purposes of interactivity, etc... has to be done outside of the glyph class. Adding these methods makes this calculation possible.)


src/glyph.ts
```diff
@@ -427,11 +427,19 @@ export class Glyph extends Element {
     return this;
   }
 
+  getXShift() {
```
rvilarl, line 430: Return type missing `:number`

src/glyph.ts
```diff
@@ -427,11 +427,19 @@ export class Glyph extends Element {
...
+    return this.x_shift;
+  }
+
   setXShift(x_shift: number): this {
     this.x_shift = x_shift;
     return this;
   }
 
+  getYshift() {
```
rvilarl, line 439: Return type missing `:number`

rvilarl (changes requested):

PR #1348, opened 2022-03-14
Fix 1347 - Annotations are too close together

### LINKED ISSUE #1347: Annotation notes can be too close together
Notes with annotations can now be too close together.  This was probably caused by recent annotation changes, but I'm not quite sure how.  Those changes affect the overall measure width, but the individual tick context should still be considered when formatting the notes.

I noticed this when I ported the latest changes to Smoosic.  If I modify the annotation lyric test slightly I see it:

image

AaronDavidNewman:
fix #1347 

This is a cut/paste issue in my last annotation change set.  The centered width was being divided by 2 twice, so the width was too low.
was;
image

is:
image

I'm having trouble with the visual regression, I get an error in 'Buiding to ./build/esm'  on Ubuntu linux:

image


rvilarl:
> I'm having trouble with the visual regression, I get an error in 'Buiding to ./build/esm' on Ubuntu linux

@AaronDavidNewman try deleting node_modules `rm -r node_modules/` and installing again `npm install`

src/annotation.ts
```diff
@@ -69,6 +69,12 @@ export class Annotation extends Modifier {
     centerStem: AnnotationVerticalJustify.CENTER_STEM,
   };
 
+  // Use the same padding for annotations as note head so the
+  // words don't run into each other.
+  static get minAnnotationPadding(): number {
+    const musicFont = Tables.currentMusicFont();
+    return musicFont.lookupMetric('glyphs.noteHead.minPadding');
+  }
```
rvilarl, line 77: `StaveNote::minNoteheadPadding()`makes the same calculation. You could use it. 
Is this calculation good for `TabNotes`?

sschmidTU: @AaronDavidNewman looks like your typescript (that's building the ESM) doesn't know the `??` operator, I've seen it happen in other systems, maybe an old version? A workaround is rewriting `??` to ` x ? y : z`, but maybe you can use a newer typescript version.

rvilarl: > > StaveNote.minNoteheadPadding()makes the same calculation. You could use it.
> > Is this calculation good for TabNotes?
> 
> I didn't want to add a dependency to StaveNote from Annotation. I think it will look good for any type of notes, it's to gap vs. other text mostly (lyrics) so words aren't squished together.

I see, what about calling `Tables.currentMusicFont().lookupMetric('glyphs.noteHead.minPadding')` directly on the two instances below?
 
> > try deleting node_modules rm -r node_modules/ and installing again npm install
> 
> This happened even on a fresh install. I was using `npm run reference`.

what version of npm do you use? `npm --version` Do you have typescript installed globally? `tsc --version`

AaronDavidNewman: npm version: 6.14.6.  I see the version referenced in the package is 8.3.  node version is 12.18.3.

When I installed this, the latest node (16.something) didn't work.  It had something to do with a dependency on the 'canvas' package.  Maybe this is not an issue anymore.  I'll try updating node also.

rvilarl: try `sudo npm install -g npm@latest`

AaronDavidNewman: OK, finally got it working.   Running npm install with latest node and npm did the trick.
It is pretty much all like this.  The 120+ notes with annotations take up slightly more room.  So no surprises:

was:
Grace_Notes With_Articulation_and_Annotation_on_Parent_Note Bravura_reference

is now (grace note before words is a little further to the right):
Grace_Notes With_Articulation_and_Annotation_on_Parent_Note Bravura_current

pptr-Annotation Standard_Notation_Annotation Petaluma svg


pptr-canvas_Annotation Harmonics Bravura

0xfe:
Look good, thanks.

PR #1349, opened 2022-03-19
Make the CJS and ESM builds export the same thing.

ronyeh:
Huge thanks to @gristow for discovering that our CJS and ESM builds don't quite export the same thing. :-D

I'm still working on it, but I think we can have a clean solution for this. The actual files we need to modify are in `vexflow/entry/` because those are the entry files for the various libs that are produced by webpack (CJS). Those are also the entry files that will be compiled by TypeScript compiler into ESM entry files. See the `exports` field in `package.json` for a list of all the entry files for require / import.

In this PR, I will clean up `src/index.ts` to no longer have a default export. For compatibility & flexibility, all the entry files will have both named exports and a default export. Additionally, we will verify that CJS and ESM code can both import modules in the same manner (e.g., a named import of `StaveNote`).

I will update you all when there's a build you can test with!


ronyeh:
I think this is ready for a code review & comments.

However, I still need to do more testing so I'm leaving it marked as a draft PR.

Some notes:
* src/index.ts is a bit cleaner: https://github.com/ronyeh/vexflow/blob/cjs-esm/src/index.ts
* See the exports at the bottom of the entry files in `vexflow/entry/`. For example: https://github.com/ronyeh/vexflow/blob/cjs-esm/entry/vexflow.ts

I'll do some testing with my own projects, and then put up a build on my server so you can try with your projects. Hopefully we can get this merged early next week so we can test a release candidate for the VexFlow 4 release!

gristow:
Thanks, @ronyeh! I'll do some testing this afternoon/evening!

gristow:
I built and tested this out on the dev branch of uTheory this afternoon -- it works beautifully in both the ESM & CJS contexts, and I like the cleanliness of using all named exports in src/index.ts, too. Looks good to me!

ronyeh:
If anyone else needs to test this build, try this:

```
npm install https://vexflow-demo.surge.sh/vex4_2022_03_20.tgz
```

ronyeh:
No visual diffs from current master.

I've tested it with a few example projects. I think this one is good to go.

entry/vexflow-petaluma.ts
```diff
@@ -16,5 +16,7 @@ loadCustom();
 Flow.setMusicFont('Petaluma', 'Custom');
 loadTextFonts();
 
+// Re-export all exports from index.ts.
 export * from '../src/index';
-export default Vex;
+// Also collect all exports into a default export for CJS projects.
```
rvilarl, line 21: Vex is no longer exprted, please remove line 7:
`import { Vex } from '../src/vex';`

rvilarl (changes requested):

rvilarl:
> No visual diffs from current master.
> 
> I've tested it with a few example projects. I think this one is good to go.

I have tested it with OSMD, I also think that it is good to go.

entry/vexflow-petaluma.ts
ronyeh: Thanks! Done.

PR #1350, opened 2022-03-21
Rename Justification to TextJustification.

ronyeh:
While working on the most recent CJS/ESM PR, I noticed that we were exporting a generically named enum called `Justification` from the TextNote file: 
https://github.com/0xfe/vexflow/blob/b23bb465b2e52f2e4c50af269b79225d395f3f5b/src/textnote.ts#L10-L14

StaveLine also has an identical "enum": https://github.com/0xfe/vexflow/blob/b23bb465b2e52f2e4c50af269b79225d395f3f5b/src/staveline.ts#L65-L69

In this PR, I renamed Justification to TextJustification so that it is more clear what the enum is used for. This is similar to the convention used in other classes, like:

https://github.com/0xfe/vexflow/blob/b23bb465b2e52f2e4c50af269b79225d395f3f5b/src/annotation.ts#L20-L32

`AnnotationVerticalJustify` and `AnnotationHorizontalJustify` are long names, but it is clear what they are for.

Summary:
* Renamed `Justification` => `TextJustification` in the textnote.ts file.
* staveline.ts now imports and uses the `TextJustification` enum, to reduce duplication.

## A comment on naming:

The word "justification" can be a bit unclear. In the text formatting world, justify usually means that a block of text is aligned to both the left margin and the right margin at the same time. So the spacing between words is automatically adjusted (via a "formatter" hehe) so that every line has the same width.

We use "justification" in many places inside VexFlow. I think in most cases, the word "align" would be clearer. For example: 
* HorizontalAlign
* VerticalAlign

Anyways, that's a much bigger breaking change, so I'm going to leave it for a future discussion.

### DIFF
--- a/src/flow.ts
+++ b/src/flow.ts
@@ -75,7 +75,7 @@ import { TabTie } from './tabtie';
 import { TextBracket, TextBracketPosition } from './textbracket';
 import { TextDynamics } from './textdynamics';
 import { TextFormatter } from './textformatter';
-import { Justification, TextNote } from './textnote';
+import { TextJustification, TextNote } from './textnote';
 import { TickContext } from './tickcontext';
 import { TimeSignature } from './timesignature';
 import { TimeSigNote } from './timesignote';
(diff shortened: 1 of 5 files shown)


ronyeh:
This change is actually backwards compatible with 3.0.9: https://github.com/0xfe/vexflow/blob/00ec15c67ff333ea49f4d3defbd9e22374c03684/src/textnote.js#L12-L19

`TextNote.Justification.LEFT`, `TextNote.Justification.CENTER`, `TextNote.Justification.RIGHT` still all return the same values as in 3.0.9.

Issue #285, opened 2015-03-25
Styling modifiers, staveModifiers, glyphs


gristow:
In working on a GUI project, I'm finding the need to color & style modifiers & specific glyphs on a pretty regular basis. For instance, when a user clicks on a barline, I'd like to be able to color it to show it as selected.

There's some consistency in how we do this within StaveNote & NoteHead, using the setStyle() functions -- my suggestion is to expand this to implement setStyle() within modifiers, staveModifiers and glyphs so that specific styling can be set at any level.

If this is something others would be interested in, I'm happy to implement & send a PR.

0xfe:
Yes, interested. I'd like to have some consistency in the styling of all the elements.

Really, the end goal (as I see it) is to support out-of-band styling via stylesheets, which will significantly cut down the code required to style scores. To do this we need:

* an internal `style` object (key-value map) for each element (it could even have styles like "xShift", "yShift", etc.)
* `setStyle()`: which sets these styles.
* A style-aware `draw`, which applies the styles as they are rendered.
* internal `id` and `class` fields, similar to CSS.
* a global `StyleManager` class, to which all elements are auto-registered when they are created, to allow for applying styles based on `id` and `class`.
* a stylesheet format and parser which uses `StyleManager` to apply styles in a stylesheet.

I think your adding a consistent `setStyle` aligns with this larger goal. So please go ahead and implement if you're still interested, but keep the larger goal in mind as you design the API. Thanks.

(FYI, I'm still out on vacation with spotty Internet, and will be slow to respond.)

gristow:
Yes, I agree, CSS-like styling is the way to go. Having a `style` object on each class will allow us to interact with these objects just as we would applying styles to DOM objects in ECMAScript. We can also implement something similar to [Element.classlist](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) for adding & removing classes to an object to keep the script as similar as possible to working directly with DOM objects.

Within `SVGContext` we can implement it as actual CSS, with the option to write inline CSS to the SVG  object or refer to an external stylesheet.  This will clean up our SVG objects significantly (we should see a rendering speed boost), and will give those using `SVGContext` access to SVG elements programmatically with `getElementById`.  We can allow users to access to the VexFlow objects in a similar way via `StyleManager`.

Assuming we use as a starting point SVG's support for CSS, we'll want to aim to support these properties (many of which are supported in Canvas but not yet implemented in `canvasContext`):
*`fill` -- already implemented
[...]

PR #1357, opened 2022-03-30
setGroupStyle

rvilarl:
> @0xfe @gristow would it make sense to arrange elements genericaly in a tree?
> with that we could find childs for example by className we would have a root element without parent (probably the score), childs of the scores would be the measures, child of the measures would be the notes and then we could define a setGroupStyle() that would apply to the referenced element and all its childs.
>@ronyeh @AaronDavidNewman what do you think?

POC to show the idea I described in #285 quoted above.
I have also realised that if we set a `Registry` see below,
https://github.com/0xfe/vexflow/blob/46af63bb5eb52c66d3a30d978b3a08d04eecf5c6/src/element.ts#L111
we can use the registry member fuctions to look for elements and appy styles.

### DIFF
--- a/src/element.ts
+++ b/src/element.ts
@@ -67,7 +67,6 @@ export abstract class Element {
     return Category.Element;
   }
 
-  protected parent?: Element;
   protected children: Element[] = [];
   protected static ID: number = 1000;
   protected static newID(): string {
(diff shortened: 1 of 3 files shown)


rvilarl:
@0xfe although I mentioned the word POC, this one is ready to review and merge. Then we only have to decide which elements we want to associated. There are no visual differences.

0xfe:
Hi Rodrigo -- I'm generally not a fan of this pattern because it becomes difficult to manage. It's better for parents to identify their children, than for children to identify their parents, otherwise things can get inconsistent quickly. Especially here since we're duplicating the relationships that are already in the classes.

I'd rather have a `setChildElements` or `addChildElement` method instead.

rvilarl:
> I'd rather have a `setChildElements` or `addChildElement` method instead.

Hi Mohit, done, no visual differences.

0xfe:
Looks good. Thanks for the changes.

PR #1372, opened 2022-04-04
fix unnecessary shift

### LINKED ISSUE #1365: OSMD: unnecessary x-shift
> * [ ]  Unnecessary x-shift from note alignment in test_note_alignment_long_distance
>   image

rvilarl:
fixes #1365
Only one visual difference:

**StaveNote Note_Heads_Placement___Simple Bravura**
**current**
StaveNote Note_Heads_Placement___Simple Bravura_current
**reference**
StaveNote Note_Heads_Placement___Simple Bravura_reference

### DIFF
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -250,7 +250,7 @@ export class StaveNote extends StemmableNote {
                 }
               }
             } //Very close whole notes
-          } else {
+          } else if (lineDiff < 1) {
             xShift = voiceXShift + 2;
             if (noteU.stemDirection === noteL.stemDirection) {
               // upper voice is middle voice, so shift it right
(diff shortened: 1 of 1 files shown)


src/stavenote.ts
```diff
@@ -259,6 +259,12 @@ export class StaveNote extends StemmableNote {
               // shift lower voice right
               noteL.note.setXShift(xShift);
             }
+          } else if (noteU.note.hasStem()) {
+            noteU.stemDirection = -noteU.note.getStemDirection();
+            noteU.note.setStemDirection(noteU.stemDirection);
```
0xfe, line 264: Thanks Rodrigo. We should only change the stem direction in controlled situations. Can we move this into the `if (voices != 2)` branch? (otherwise we'll have weird issues with  stems changing direction in other situations.)

rvilarl: Hi Mohit, this is already, I believe, in a very controlled situation. All these conditions have to be fulfilled:
- there are two voices
- only one of the notes has stem (the other one is a whole)
- they are not in the same line
- they overlap (the stem overlaps with the whole)

There are no other visual differences in all the tests (this is the only one where the change of stem direction was applied)

0xfe: Thanks Rodrigo -- I didn't realize that this was inside the `voices === 2` branch.

0xfe:
Looks good, merging.

PR #1376, opened 2022-04-15
Add last note padding test case using softmax

AaronDavidNewman:
This addresses issue 1370.  Added a test case to formatter test cases to demonstrate softmax and how it affects measures with extra whitespace.  I don't think there is any other action we need to take, but this seems to come up a lot.

image

### DIFF
--- a/tests/formatter_tests.ts
+++ b/tests/formatter_tests.ts
@@ -152,6 +152,7 @@ function rightJustify(options: TestOptions): void {
 function penultimateNote(options: TestOptions): void {
   const f = VexFlowTests.makeFactory(options, 500, 550);
   const score = f.EasyScore();
+  const staffWidth = 310;
   let system: System | undefined = undefined;
   let voices: Voice[] = [];
   let notes: StemmableNote[] = [];
@@ -159,7 +160,12 @@ function penultimateNote(options: TestOptions): void {
   let stave: Stave | undefined = undefined;
   let y = 10;
   const draw = (softmax: number) => {
-    system = f.System({ width: 309.87, y, formatOptions: { align_rests: true }, details: { softmaxFactor: softmax } });
+    system = f.System({
+      width: staffWidth,
+      y,
+      formatOptions: { align_rests: true },
+      details: { softmaxFactor: softmax },
+    });
     notes = [];
     voices = [];
     note = score.notes('C4/8/r', { clef: 'bass' })[0];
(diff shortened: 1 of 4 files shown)


rvilarl:
@AaronDavidNewman shouldn't we change the default from 100 to 5 as well?

tests/formatter_tests.ts
```diff
@@ -147,6 +149,48 @@ function rightJustify(options: TestOptions): void {
...
+    voices.push(score.voice(notes).setMode(2));
+    notes = [];
+    stave = system.addStave({ voices: voices });
+    stave.addClef('bass');
+    stave.addTimeSignature('2/4');
+    voices = [];
+    f.draw();
+    y += 100;
+  };
+  draw(100);
```
rvilarl, line 186: It would be good to add a text indicating which softmax value was used.

tests/formatter_tests.ts
```diff
@@ -119,7 +121,7 @@ function buildTickContexts(): void {
 }
 
 function rightJustify(options: TestOptions): void {
-  const f = VexFlowTests.makeFactory(options, 1200, 300);
+  const f = VexFlowTests.makeFactory(options, 1200, 150);
```
rvilarl, line 124: why did you change this value?

tests/formatter_tests.ts
```diff
@@ -132,7 +134,7 @@ function rightJustify(options: TestOptions): void {
   const renderTest = (time: VoiceTime, n: number, duration: string, duration2: string, x: number, width: number) => {
     const formatter = f.Formatter();
 
-    const stave = f.Stave({ x, y: 40, width });
+    const stave = f.Stave({ x, y: 20, width });
```
rvilarl, line 137: Why did you change this value?

rvilarl (changes requested):

tests/formatter_tests.ts
AaronDavidNewman: Trying to conserve vertical space.  Will it mess with the visual regression?

AaronDavidNewman:
> 

That would change _all_ the test cases. Do we want to do that?

I'm also open to making it easier to set the global softmax.  It seems like it is defined in 2 places.  It seemse like it should be defined in one place, either be in tables, or in the font metrics.

AaronDavidNewman:
image

image

tests/formatter_tests.ts
rvilarl: That is ok I thought that it might have been overseen.

rvilarl (approved):

rvilarl:
> > shouldn't we change the default from 100 to 5 as well?
> 
> That would change _all_ the test cases. Do we want to do that?
> 
> I'm also open to making it easier to set the global softmax. It seems like it is defined in 2 places. It seemse like it should be defined in one place, either be in tables, or in the font metrics.

I would go for that, @0xfe what do you think?

AaronDavidNewman:
I am going to cancel this pull request and resubmit because I screwed up my local branches.

PR #1379, opened 2022-04-17
exclude hidden notes from formatting

### LINKED ISSUE #1369: OSMD: Rests disappear
> * [ ]  In OSMD Function Test - Invisible Notes, rests disappear:
>   (diff:)
>   image
> 
> I assume something goes wrong when unifying visible and invisible rests (Musescore:): image
> 
> Also, the half rest in measure 5 is missing now: image
>

rvilarl:
fixes #1369

The notes which are hidden are not to be considered in the formatting calculations.

No visual differences.

### DIFF
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -164,64 +164,41 @@ export class StaveNote extends StemmableNote {
       });
     }
 
-    let voices = notesList.length;
-    if (voices < 2) return true;
-
-    let noteU = notesList[0];
+    let voices = 0;
+    let noteU = undefined;
     let noteM = undefined;
     let noteL = undefined;
-    let draw = noteU.note.render_options.draw ? noteU.note.render_options.draw : true;
+    const draw = [false, false, false];
 
-    if (voices === 2) {
-      // 2 voices & notesList[0] hidden, return
-      if (!draw) return true;
-      noteL = notesList[1];
-      draw = noteL.note.render_options.draw ? noteL.note.render_options.draw : true;
-      // 2 voices & notesList[1] hidden, return
-      if (!draw) return true;
-      // 2 voices to process
-      voices = 2;
-    } else if (!draw) {
-      // notesList[0] hidden
-      noteU = notesList[1];
-      draw = noteU.note.render_options.draw ? noteU.note.render_options.draw : true;
-      // notesList[0 & 1] hidden, return
-      if (!draw) return true;
-      // only notesList[0] hidden, voices == 2
-      else {
-        noteL = notesList[2];
-        voices = 2;
-        draw = noteL.note.render_options.draw ? noteL.note.render_options.draw : true;
-        // notesList[0 & 2] hidden, return
-        if (!draw) return true;
-        // else 2 voices to process
-      }
-    } else {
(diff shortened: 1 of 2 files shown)


rvilarl:
@0xfe this is ready to go, it avoids shifting notes due to overlaps with hidden notes.

0xfe:
Can you add some tests for this? Not clear what these changes are doing without tests.

src/stavenote.ts
```diff
@@ -164,11 +164,64 @@ export class StaveNote extends StemmableNote {
...
+    let noteL = undefined;
+    let draw = noteU.note.render_options.draw ? noteU.note.render_options.draw : true;
+
+    if (voices === 2) {
+      // 2 voices & notesList[0] hidden, return
+      if (!draw) return true;
+      noteL = notesList[1];
+      draw = noteL.note.render_options.draw ? noteL.note.render_options.draw : true;
+      // 2 voices & notesList[1] hidden, return
+      if (!draw) return true;
```
0xfe, line 181: There are a lot of these `if (!draw) return true`, and it's pretty hard to reason about. Can you group or restructure them so there are fewer code paths to navigate?

rvilarl: This is the function shifting notes if they overlap. The calculation was ignoring that some of the notes might be hidden and hidden botes do not overlap.
I have put a comment on each return. The one above tells tha we have two voices but one of them is hidden, as a consquence no shift is required and we can return.

rvilarl: I changed the approach and now there is only one return and I believe it is much clearer. Thanks for the comment I got lost trying to follow the original approach.

rvilarl:
> Can you add some tests for this? Not clear what these changes are doing without tests.

@0xfe I have made a variant of
**StaveNote Note_Heads_Placement___Simple**
StaveNote Note_Heads_Placement___Simple Bravura

called **StaveNote Note_Heads_Placement___Hidden**
which hides the whole notes in the first and second measures and the quarter note and one half rest in the last measure.
Result with current `StaveNote` modifications
StaveNote Note_Heads_Placement___Hidden_Notes Bravura
Results without modifications
StaveNote Note_Heads_Placement___Hidden_Notes Bravura
Here you can see that the notes are shifted in the measure 1 & 2  and that in measure 4 the stem direction of the first note is changed and the half rest that was visible was hidden. This is all not required when the hidden notes are taken into consideration.

src/stavenote.ts
```diff
@@ -164,11 +164,41 @@ export class StaveNote extends StemmableNote {
...
 
-    const voices = notesList.length;
+    let voices = 0;
+    let noteU = undefined;
+    let noteM = undefined;
+    let noteL = undefined;
+    const draw = [false, false, false];
+
+    for (let i = 0; i < notesList.length; i++) {
+      draw[i] = notesList[i].note.render_options.draw == false ? false : true;
```
0xfe, line 174: FYI, you can remove this loop:

`const draw = notesList.map((n) => !(n.note.render_options.draw))`

0xfe (approved):
Thanks Rodrigo, merging.

PR #1382, opened 2022-04-24
Consider existing modifiers when calculating chord symbol width

AaronDavidNewman:
Like the changes made to articulation and annotation.  Consider the width of the note and also existing modifiers, and subtract from the width of the chord symbol.  The width should be the overlap, not the sum of all the widths.

This is a new test case but without this change, the chord symbols would be further apart (this is using min padding, I realize most real-world music would use more):

image

I also updated some jazz ornament metrics and added ornaments to factory.


AaronDavidNewman:
This is odd, double bottom test case takes up _more_ room for some reason and is cut off.

image

AaronDavidNewman:
This last issue was a formatting bug.  Since the bug is in formatter, I need to redo the test cases.
I also added ornament to factory.

rvilarl (changes requested):

AaronDavidNewman:
This is ready now.  The test cases were too narrow b/c the GC was scaled up by 1.5 after the factory was created.  There was no formatting issue.

rvilarl (approved):

rvilarl:
Visual differences (allways in order 1 diff 2 current 3 reference)
ChordSymbol Bottom_Chord_Symbols Bravura
ChordSymbol Bottom_Chord_Symbols Bravura_current
ChordSymbol Bottom_Chord_Symbols Bravura_reference
ChordSymbol Bottom_Stem_Down_Chord_Symbols Bravura
ChordSymbol Bottom_Stem_Down_Chord_Symbols Bravura_current
ChordSymbol Bottom_Stem_Down_Chord_Symbols Bravura_reference
ChordSymbol Chord_Symbol_Font_Size_Tests Bravura
@AaronDavidNewman here it seems that current requires more space, this seems to be a failure
ChordSymbol Chord_Symbol_Font_Size_Tests Bravura_current
ChordSymbol Chord_Symbol_Font_Size_Tests Bravura_reference
ChordSymbol Chord_Symbol_Kerning_Tests Bravura
ChordSymbol Chord_Symbol_Kerning_Tests Bravura_current
ChordSymbol Chord_Symbol_Kerning_Tests Bravura_reference
ChordSymbol Double_Bottom_Chord_Symbols Bravura
ChordSymbol Double_Bottom_Chord_Symbols Bravura_current
ChordSymbol Double_Bottom_Chord_Symbols Bravura_reference
ChordSymbol Top_Chord_Symbols Bravura
ChordSymbol Top_Chord_Symbols Bravura_current
ChordSymbol Top_Chord_Symbols Bravura_reference
ChordSymbol Top_Chord_Symbols_Justified Bravura
ChordSymbol Top_Chord_Symbols_Justified Bravura_current
ChordSymbol Top_Chord_Symbols_Justified Bravura_reference
Ornament Jazz_Ornaments Bravura
@AaronDavidNewman this indeed much better but in the 1st measure it gets to tight with accidentals and also the last ornaments go beyond the staff. On the last measure there are rough collisions.
Ornament Jazz_Ornaments Bravura_current
Ornament Jazz_Ornaments Bravura_reference

AaronDavidNewman:
The chord change test cases are correct.  There is a 'reportWidth' flag that can be set to false, if the program doesn't want to consider the width of the chord at all.  This is sometimes what you want, if a chord change applies to lots of notes (which is sometimes the case with jazz transcriptions).  But I was abusing here, I think, to make the test cases look nice.  Now I have turned it off, so the widths became longer since they do in fact overlap.

For the ornaments, the long drop/long 'doit' are supposed to go beyond the measure - that's expected.  You'd use this type of ornament when you wanted the drop to run into the following measure.  Otherwise you'd use one of the short ones.

The short ones, they have metrics that handle these parameters and I adjusted those.

image

rvilarl:
@0xfe LGTM ready to merge.

0xfe:
Looks great Aaron. Thanks for the changes. And thanks for the review Rodrigo.

PR #1383, opened 2022-04-24
add softmax test with more extra padding

AaronDavidNewman:
Show effects of softmax with larger amounts of whitespace in a measure.  See discussion in PR 1376 which was cancelled and replaced with this one.

image

image

image

@rvilarl  wrote in previous PR:
>>> shouldn't we change the default from 100 to 5 as well?

>> That would change all the test cases. Do we want to do that?
>> I'm also open to making it easier to set the global softmax. It seems like it is defined in 2 places. It seemse like it should be  defined in one place, either be in tables, or in the font metrics.

I would go for that, @0xfe what do you think?


0xfe:
Tests look good. Also, single place for global softmax sounds good, so long as we can still do local overrides.

PR #1384, opened 2022-04-29
Refactor/sofmaxfactor

rvilarl:
The first commit refactors softmaxFactor leaving it in `Formatter` only,  no longer duplicated in `Voice`. This first commits has an impact on the interfaces but produces no visual difference.

The second commit changes the default from 100 to 5. This value was suggested as a good one by @AaronDavidNewman this produces a lot of visual differences and I believe that the results are better.

Ready to review.

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -487,12 +487,12 @@ export class Factory {
     return multiMeasureRest;
   }
 
-  Voice(params?: { time?: VoiceTime | string; options?: { softmaxFactor: number } }): Voice {
+  Voice(params?: { time?: VoiceTime | string }): Voice {
     const p = {
       time: '4/4',
       ...params,
     };
-    const voice = new Voice(p.time, p.options);
+    const voice = new Voice(p.time);
     this.voices.push(voice);
     return voice;
   }
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -655,7 +655,7 @@ export class Formatter {
 
     // Calculate the "distance error" between the tick contexts. The expected distance is the spacing proportional to
     // the softmax of the ticks.
-    function calculateIdealDistances(adjustedJustifyWidth: number): Distance[] {
+    function calculateIdealDistances(adjustedJustifyWidth: number, softmaxFactor: number): Distance[] {
       const distances: Distance[] = contextList.map((tick: number, i: number) => {
         const context: TickContext = contextMap[tick];
         const voices = context.getTickablesByVoice();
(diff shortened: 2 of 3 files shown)


0xfe:
Thanks Rodrigo. It's hard to tell if the visual diffs are because of the softmax factor or the code changes.

I think we should break this up into two PRs.

First PR, change softmax factor to 5, and then eyeball tests (which should all be different.)

Second PR (this one), only refactor, and make sure that there are no visual diffs.

rvilarl:
Ok. I made it in two steps, the two commits. The first is the refactor and there was no differences. I will split.

AaronDavidNewman:
Hi Rodrigo, I had already done global softmax a couple of days ago but didn't have a chance to submit it as a PR.  It seems like you don't need all these changes.  Have a look at:

https://github.com/AaronDavidNewman/vexflow/commit/c2462fbdc63bdc279479c856965f64c8b9742a80

I created a global variable in formatter with accessors, to prevent circular dependency with voice and formatter.  Once you do this, applications can set the default softmax to whatever they like, without having to recompile.  And setting it on a per-measure basis still works the same as before.  You shouldn't have to make it passed all over the place. It is set when the formatter is created, either the passed-in value or the global default, and stays the same during formatting.

0xfe:
Thanks folks. Rodrigo, can we use Aaron's approach instead? -- it's definitely a lot less invasive :-)

rvilarl:
> Second PR (this one), only refactor, and make sure that there are no visual diffs.

This one is now only refactor and there are no visual differences

> Hi Rodrigo, I had already done global softmax a couple of days ago but didn't have a chance to submit it as a PR. It seems like you don't need all these changes. 

Sorry :( I wanted to go for the 5 default and I look into the details of the duplication you mentioned, see below. 

> Thanks folks. Rodrigo, can we use Aaron's approach instead? -- it's definitely a lot less invasive :-)

@0xfe @AaronDavidNewman I made it more invasive for the following reasons:
- I could not imagine that different voices were using different values.
- I considered that the parameter was more related to the formatter.
- The formatter was setting all the voices to the same value.

As a consequence I thought that the value shoud be stored only once in the formatter and not copied in each voice and that it should be provided to the voices when formatting was required. Have a look at the approach and I leave it up to you to decide what you prefer.

AaronDavidNewman:
@rvilarl I think you and I are thinking about it in the same way.  I was trying to minimize change but not really trying for the best possible refactoring.  

You're right, there is no reason at all to have it as a constructor parameter in `voice`.  So I think your idea to remove that is good and makes it a more clear that softmax constant is a 'formatter' responsibility, and not a 'voice' responsibility.

The responsibilities of 'voice' class are a little muddled to begin with.  It seems an ideal 'voice' would only be accessed through the formatter, but that is very different from the interface we have.   That's why I like the idea of keeping setSoftmaxFactor in voice, even though it is pretty convoluted logically.

rvilarl:
@AaronDavidNewman thanks for the explanation. :) I get that you prefer to leave softmaxFactor in Voice. That is fine. Should I close this PR and you make one?

AaronDavidNewman:
@rvilarl Sorry that I didn't communicate this very well.  I removed the softmax parameter from 'voice' in my changeset and pushed it.  So if you want we can close yours and we can use that one.

rvilarl:
Closed, we continue with #1386

Issue #1346, opened 2022-03-12
CJS/ESM Build Alignment


gristow:
The output of the CJS and ESM builds from `src/index.ts` have different shapes because, in the Webpack UMD build process, the default export of Vex in `src/index.ts` overwrites the named exports from the same file.

As a result, when working in an ESM context this will work:
```typescript
import { StaveNote } from 'vexflow';
const sn = new StaveNote();
```

But the parallel CJS code will not work:
```typescript
const StaveNote = require('vexflow').StaveNote;
const sn = new StaveNote(); // Uncaught Reference Error: StaveNote is not defined
```

I'm hitting this issue, personally, using Vite which transforms ESM import style code to CJS require style code in node to provide the server-side render. (And I would guess given the rise in SSR frameworks we'll soon see others with similar issues.)

A quick fix would be to merge Flow onto the default export in `src/index.ts`:
```typescript
// Export Vex as a named export AND as a default export, to support both ways of importing.
export { Vex };
export default {
  ...Vex,
  ...Flow,
};
```

This alters the default export to become an object, instead of the Vex class -- but, since we only use static props on the Vex class anyway, this doesn't break any tests. (And it shouldn't break any implementations.)

Thoughts? (Especially @ronyeh since you did the bulk of the ESM build work.)

I've got a working branch with these changes -- but will hold off on a PR unless/until we think this makes sense as a way to go.

0xfe:
Thanks @gristow. I don't have a strong opinion here, the quick fix sounds okay to me. But @ronyeh might have other suggestions.

ronyeh:
Sorry for the delay. I was busy this week with deadlines.

TLDR: I agree that we should strive for maximum flexibility, so that VexFlow is compatible with as many tools as possible. However, we should make sure this PR doesn't **also** change the shape of the ESM build. Ideally the ESM import and CJS require will return identical objects (if possible).

---

I left the CJS export untouched while building the ESM support. I didn't want to break any of the existing CJS code, so the shape of the CJS output is the same as it has always been. Calling `require('vexflow')` gives you the `Vex` object. To get the `StaveNote` class, you currently still need to do `Vex.Flow.StaveNote` like in VexFlow 3.0.9.

So yes, the ESM build provides much more flexibility vs the CJS build.

Here is some test code which works with the ESM build:

```javascript
// ESM Example

import { Vex } from 'vexflow';
import { StaveNote } from 'vexflow';

// Get the default export.
[...]

PR #1385, opened 2022-05-01
stringnumber: radius protected, circle configurable. 

### LINKED ISSUE #1364: OSMD: String numbers colliding
> * [ ]  String numbers colliding vertically in test_grace_note_fingerings_and_strings
>   image

rvilarl:
fixes #1364

I added one test without circle (this one collides as described in #1364 without the changes in this PR).
StringNumber String_Number_In_Notation___radius_0 Bravura

which is a variant of the standard with radius 8
StringNumber String_Number_In_Notation Bravura

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -464,10 +464,11 @@ export class Factory {
     return fingering;
   }
 
-  StringNumber(params: { number: string; position: string }): StringNumber {
+  StringNumber(params: { number: string; position: string; radius?: number }): StringNumber {
     const stringNumber = new StringNumber(params.number);
     stringNumber.setPosition(params.position);
     stringNumber.setContext(this.context);
+    if (params.radius != undefined) stringNumber.radius = params.radius;
     return stringNumber;
   }
 
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -464,11 +464,11 @@ export class Factory {
     return fingering;
   }
 
-  StringNumber(params: { number: string; position: string; drawCircle?: boolean }): StringNumber {
+  StringNumber(params: { number: string; position: string }, drawCircle = true): StringNumber {
     const stringNumber = new StringNumber(params.number);
     stringNumber.setPosition(params.position);
     stringNumber.setContext(this.context);
-    if (params.drawCircle != undefined) stringNumber.setDrawCircle(params.drawCircle);
+    stringNumber.setDrawCircle(drawCircle);
     return stringNumber;
   }
 
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -464,11 +464,11 @@ export class Factory {
     return fingering;
   }
 
-  StringNumber(params: { number: string; position: string; radius?: number }): StringNumber {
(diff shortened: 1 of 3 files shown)


src/factory.ts
```diff
@@ -464,10 +464,11 @@ export class Factory {
     return fingering;
   }
 
-  StringNumber(params: { number: string; position: string }): StringNumber {
+  StringNumber(params: { number: string; position: string; radius?: number }): StringNumber {
     const stringNumber = new StringNumber(params.number);
     stringNumber.setPosition(params.position);
     stringNumber.setContext(this.context);
+    if (params.radius != undefined) stringNumber.radius = params.radius;
```
0xfe, line 471: Not a fan of manipulating the instance outside an API call.

0xfe (changes requested):
I don't understand this change -- it doesn't match the PR title. Why not just disallow (or special case) radius `0`?

src/factory.ts
rvilarl: Me neither but `radius` has no setter / getter and it is a public member. OSMD sets radius to 0 in order to present stringNumbers without circles.

What do you think about my considerations above?:

> I wonder if it really makes sense to allow to set different radius:
> 
> - Radii bigger than 8 overlap on chords and smaller than 7 overlap with the number.
> - the this.setWidth(20); // ??? is also somehow related to the radius.
> 
> I would recommend changing the class to allow the configuration of wheter a circle should be drawn.

With this I mean amking radius protected and adding a function or an option to say if the circle should be drawn.

rvilarl:
> I don't understand this change -- it doesn't match the PR title. Why not just disallow (or special case) radius `0`?

The PR title refers to the associated issue #1346 there you can see that the string numbers (Roman numbers) collide with the notes. I think we agree, I will make radius protected and allow the special case without circle. I did not do that in the first instance because it changes the interface.

rvilarl:
> > I don't understand this change -- it doesn't match the PR title. Why not just disallow (or special case) radius `0`?
> 
> The PR title refers to the associated issue #1364 there you can see that the string numbers (Roman numbers) collide with the notes. I think we agree, I will make radius protected and allow the special case without circle. I did not do that in the first instance because it changes the interface.

@0xfe done, ready to review.

src/factory.ts
rvilarl: New approach pushed.

src/factory.ts
```diff
@@ -464,10 +464,11 @@ export class Factory {
     return fingering;
   }
 
-  StringNumber(params: { number: string; position: string }): StringNumber {
+  StringNumber(params: { number: string; position: string; drawCircle?: boolean }): StringNumber {
```
0xfe, line 467: It's not clear from this if the circle is drawn or not by default. Can you make it `drawCircle: boolean = true`? (You can remove the `if undefined` below too, in that case.)

rvilarl: This has to be done outside params, done.

PR #1386, opened 2022-05-02
Softmax default: set softmax default in exactly one place

AaronDavidNewman:
See also: https://github.com/0xfe/vexflow/pull/1384

Since formatter class always sets the voice class softmax to match its own, there's no reason to have a separate softmax parameter.  Additionally, softmax default can now be set globally by using the setter in formatter.ts.  For instance, setting it in vexflow_test_helpers.ts to a value other than the default (100) will affect all the test cases where it is not set explicitly in formatter options.


src/annotation.ts
```diff
@@ -97,7 +97,7 @@ export class Annotation extends Modifier {
         maxLeftGlyphWidth = Math.max(glyphWidth, maxLeftGlyphWidth);
         leftWidth = Math.max(leftWidth, textWidth) + Annotation.minAnnotationPadding;
       } else if (annotation.horizontalJustification === AnnotationHorizontalJustify.RIGHT) {
-        maxRightGlyphWidth = Math.max(glyphWidth, maxLeftGlyphWidth);
+        maxRightGlyphWidth = Math.max(glyphWidth, maxRightGlyphWidth);
```
rvilarl, line 100: I think that you included this change by mistake.

src/chordsymbol.ts
```diff
@@ -13,10 +13,11 @@ import { Font, FontInfo, FontStyle, FontWeight } from './font';
 import { Glyph } from './glyph';
 import { Modifier } from './modifier';
 import { ModifierContextState } from './modifiercontext';
+import { Note } from './note';
 import { StemmableNote } from './stemmablenote';
 import { Tables } from './tables';
 import { TextFormatter } from './textformatter';
-import { Category } from './typeguard';
+import { Category, isStemmableNote } from './typeguard';
 import { log } from './util';
```
rvilarl, line 22: I think that you included these changes by mistake. Also the other ones in this file.

tests/chordsymbol_tests.ts
```diff
@@ -8,14 +8,18 @@ import { TestOptions, VexFlowTests } from './vexflow_test_helpers';
 import { Accidental } from '../src/accidental';
 import { ChordSymbol } from '../src/chordsymbol';
 import { Factory } from '../src/factory';
+import { Font, FontGlyph } from '../src/font';
```
rvilarl, line 11: The changes in this file were also probably added by mistake.

src/formatter.ts
```diff
@@ -19,6 +19,16 @@ import { isNote, isStaveNote } from './typeguard';
 import { defined, log, midLine, RuntimeError } from './util';
 import { Voice } from './voice';
 
+// Default softmax.  Can be set on measure level with FormatterOptions
+// or globally by changing the default
+export let softmaxDefault = 100;
```
rvilarl, line 24: I think that this declaration should be in `Tables` as a static member. (I would probaby go for softmaxFactor rather than softmaxDefault if we agree that it makes no sense to have different values on different voices)

src/formatter.ts
```diff
@@ -19,6 +19,16 @@ import { isNote, isStaveNote } from './typeguard';
...
+export const setSoftmaxDefault = (value: number) => {
+  softmaxDefault = value;
+};
+export const getSoftmaxDefault = () => {
+  return softmaxDefault;
+};
+
```
rvilarl, line 31: I think that those two function should be members of `Flow`.

rvilarl (changes requested):

src/annotation.ts
AaronDavidNewman: The code in green is what is currently in the vexflow repo.  I'm not seeing it as a change.

tests/chordsymbol_tests.ts
AaronDavidNewman: Maybe I missing something?  Maybe you are a PR behind?  I see the code here:

https://github.com/0xfe/vexflow/blob/master/tests/chordsymbol_tests.ts

It only shows up here because I did a merge to my vexflow repository.

src/formatter.ts
AaronDavidNewman: I think softmaxDefault and the setter has to be in the same file.  And I didn't want to put it in Tables.js because that is not accessible from outside of vexflow.

I could put both the variable and the setters in flow - will that work?

src/formatter.ts
AaronDavidNewman: I get a lint error if softmax and accessor functions are not in the same file.  I can move both to flow, or possibly comment away the lint error.

src/annotation.ts
rvilarl: Don't you see 6 files changed in this PR?

src/formatter.ts
rvilarl: My proposal is to do it like `setMusicFont` / `getMusicFont`

`static SOFTMAX_DEFAULT = 100` in `Tables` or `SOFMAX_FACTOR` if we agree on my oppnion in the discussion above.

src/formatter.ts
rvilarl: My proposal is to do it like `setMusicFont` / `getMusicFont`
```
setSoftmaxDefault (value: number) {
  Tables.SOFTMAX_DEFAULT = value;
};
getSoftmaxDefault() {
  return Tables.SOFTMAX_DEFAULT;
};
```
in `Flow` or set/getSoftmaxFactor if we agree in the discussion above.

src/annotation.ts
AaronDavidNewman: > Don't you see 6 files changed in this PR?

yeah but 4 of them are from a merge

src/formatter.ts
AaronDavidNewman: > My proposal is to do it like setMusicFont / getMusicFont


It seems like this should work, but I'm getting this strange error.  Any ideas why?  I notice that there is nowhere in the test files or the code base where 'getMusicFont' is actually used, I wonder if that would product a similar error.  (it is late here so maybe I am overlooking something obvious.  Surprising that it compiles fine)

Also, this makes it so setting softmax for the measure occurs in a different module that setting softmax globally, which is a little odd.

anyway, here is the error that introduces.
```
barnote.ts:79 Uncaught ReferenceError: Cannot access 'Beam' before initialization
    at Module.Beam (barnote.ts:79:1)
    at Object../src/flow.ts (flow.ts:247:13)
    at __webpack_require__ (bootstrap:19:1)
    at Object../src/formatter.ts (textfonts.ts:52:1)
    at __webpack_require__ (bootstrap:19:1)
    at Object../src/tuplet.ts (tuning.ts:70:1)
    at __webpack_require__ (bootstrap:19:1)
    at Object../src/beam.ts (barnote.ts:79:1)
    at __webpack_require__ (bootstrap:19:1)
    at Object../src/index.ts (gracetabnote.ts:28:1)
Beam @ barnote.ts:79
./src/flow.ts @ flow.ts:247
__webpack_require__ @ bootstrap:19
./src/formatter.ts @ textfonts.ts:52
__webpack_require__ @ bootstrap:19
./src/tuplet.ts @ tuning.ts:70
__webpack_require__ @ bootstrap:19
./src/beam.ts @ barnote.ts:79
__webpack_require__ @ bootstrap:19
./src/index.ts @ gracetabnote.ts:28
__webpack_require__ @ bootstrap:19
(anonymous) @ make namespace object:7
(anonymous) @ vexflow-debug-with-tests.ts:19
(anonymous) @ vexflow-debug-with-tests.ts:19
webpackUniversalModuleDefinition @ universalModuleDefinition:9
(anonymous) @ universalModuleDefinition:10
23:07:47.525 flow.html?module=Bach%20Demo:139 Uncaught ReferenceError: Vex is not defined
    at flow.html?module=Bach%20Demo:139:18
(anonymous) @ flow.html?module=Bach%20Demo:139
load (async)
(anonymous) @ flow.html?module=Bach%20Demo:38
loadScript @ flow.html?module=Bach%20Demo:36
loadVexFlow @ flow.html?module=Bach%20Demo:125
(anonymous) @ flow.html?module=Bach%20Demo:137

```

src/annotation.ts
rvilarl: I think that in order ot fix that all you need to do is:
```
git checkout master
git pull upstream master
git checkout softmax-default
git rebase master
git push -f
```
perhaps you have a different name for Mohit's repository, then you have to replace `upstream` above. Check using:

`git remote -v`

this is what I get:
```
aarondavidnewman        https://github.com/aarondavidnewman/vexflow.git (fetch)
aarondavidnewman        https://github.com/aarondavidnewman/vexflow.git (push)
origin  https://github.com/rvilarl/vexflow.git (fetch)
origin  https://github.com/rvilarl/vexflow.git (push)
ronyeh  https://github.com/ronyeh/vexflow.git (fetch)
ronyeh  https://github.com/ronyeh/vexflow.git (push)
sschmidtu       https://github.com/sschmidtu/vexflow.git (fetch)
sschmidtu       https://github.com/sschmidtu/vexflow.git (push)
upstream        https://github.com/0xfe/vexflow.git (fetch)
upstream        https://github.com/0xfe/vexflow.git (push)
wassertim       https://github.com/wassertim/vexflow.git (fetch)
wassertim       https://github.com/wassertim/vexflow.git (push)
```

ronyeh: In the GitHub user interface for reviewing this PR, I see 6 files changed.

If you changed only 2 files in this PR, then ideally we would see 2 changed files (which would make it easier to review).

If you have been regularly merging your branch from 0xfe:master, shouldn't this change not appear in the PR? Or maybe you can take Rodrigo's rebase approach.

But generally for branches with involved changes, I prefer using the merge approach. I only use rebase for simple changes that can be reapplied with no (or very minimal) conflicts.

src/formatter.ts
ronyeh: Whenever I see "Cannot access 'XXXX' before initialization", I assume it's a circular dependency.

It might compile correctly, but when running, it means that class A imported class B, which during initialization requires class A (it might be a few dependencies down, but eventually it requires class A).

Unfortunately, class A hasn't been fully initialized yet since it was still in the middle of importing class B, so some of its static properties are still undefined.

Sorry I haven't been following this PR closely. Are we trying to add a static property to Tables to customize the softmax value?

If all external accesses are through the Flow class, I don't see why we can't keep the actual property inside Tables.


FYI: I implemented getMusicFont(). I thought I did use it somewhere (maybe in the demos but not in the unit tests). If I skipped it, it's my fault, haha.

Let me know if you need help debugging this (esp if you think it's a circular dependency).

AaronDavidNewman: It is definitely a circular dependency.  flow requires beam requires tuplet which requires formatter.  So we can't add a dependency from formatter to flow without further changes.  I am going back to softmax default as global variable in formatter.  Not everything needs to be part of a class.

src/formatter.ts
AaronDavidNewman: @rvilarl  see above, I tried this but it introduced a circular dependency when 'formatter' is dependent on 'flow'.  I couldn't think of an easy way around this.

rvilarl: @AaronDavidNewman did you see the PR that I made to your repository?

AaronDavidNewman:
will do later today, thanks.

AaronDavidNewman:
OK I merged it.  But now formatter still has its own softmax definition, .  so it's back to being defined in 2 places?  How do you now set the global value of softmax?

rvilarl:
> OK I merged it. But now formatter still has its own softmax definition, . so it's back to being defined in 2 places? How do you now set the global value of softmax?

@AaronDavidNewman something went wrong with the merge, please do the following

git checkout softmax-default
git reset 5cf536d
git reset --hard
git push -f

the global value is now set with either
`Tables.SOFTMAX_FACTOR = value`
inside the libraray or with 
`Flow.SOFTMAX_FACTOR = value`
on apps using the libray.

AaronDavidNewman:
OK, I made formatter use the value in Tables, and the setting in Flow still works from outside.

rvilarl (approved):
Later we can decide if we want to keep the variable locally in Voice or simply use the one in Table.

0xfe:
This works, thanks folks.

PR #1387, opened 2022-05-04
chords no longer need to be sorted

### LINKED ISSUE #104: Problem with chord accidental on wrong line.
Based on my understanding the following code should add a flat to the A, but it is rendered on the B
chord
 

	var notes = [
    	new Vex.Flow.StaveNote({ keys: ["B/4", "A/4"], duration: "q",  stem_direction: 1})
  	];	
  	
	//Add flat to the A  
  	notes[0].addAccidental(1, new Vex.Flow.Accidental('b'));

rvilarl:
Fixes #104 
This addresses the problem once discussed in #104.
I decided to go an implement option 2. 

No visual differences.

### DIFF
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -13,14 +13,14 @@ import { BoundingBox } from './boundingbox';
 import { ElementStyle } from './element';
 import { Modifier } from './modifier';
 import { ModifierContextState } from './modifiercontext';
-import { Note, NoteStruct } from './note';
+import { KeyProps, Note, NoteStruct } from './note';
 import { NoteHead } from './notehead';
 import { Stave } from './stave';
 import { Stem, StemOptions } from './stem';
 import { StemmableNote } from './stemmablenote';
 import { Tables } from './tables';
 import { Category } from './typeguard';
-import { defined, log, midLine, RuntimeError, warn } from './util';
+import { defined, log, midLine, RuntimeError } from './util';
 
 function showDeprecationWarningForNoteHeads(): void {
   // eslint-disable-next-line
@@ -133,9 +133,9 @@ export class StaveNote extends StemmableNote {
     const notesList: StaveNoteFormatSettings[] = [];
 
     for (let i = 0; i < notes.length; i++) {
-      const props = notes[i].getKeyProps();
-      const line = props[0].line;
-      let minL = props[props.length - 1].line;
+      const props = notes[i].sortedKeyProps;
+      const line = props[0].keyProps.line;
+      let minL = props[props.length - 1].keyProps.line;
       const stemDirection = notes[i].getStemDirection();
       const stemMax = notes[i].getStemLength() / 10;
       const stemMin = notes[i].getStemMinimumLength() / 10;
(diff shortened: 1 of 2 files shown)


0xfe:
Can you expand more in the PR description (and also in the comments) please? This is a pretty hard breaking I think, because the notes that were provided are going to be reordered, which breaks the indices that they use, which is going to be very surprising for the user.

I'd rather we provide a warning (and maybe also a helper function to sort the chords separately.)

rvilarl:
I will provide more comments. But this is exactly the opposite, they used to be reorder and this was very confusing. There was a console log to warn that, now they are no longer externally reordered, only internally.

src/stavenote.ts
```diff
@@ -133,9 +133,10 @@ export class StaveNote extends StemmableNote {
     const notesList: StaveNoteFormatSettings[] = [];
 
     for (let i = 0; i < notes.length; i++) {
-      const props = notes[i].getKeyProps();
-      const line = props[0].line;
-      let minL = props[props.length - 1].line;
+      // formating relys on sortedKeyNotes in order to calculate line & minL
```
ronyeh, line 136: Typo:

``
`
// Formatting uses sortedKeyProps to calculate line and minL.
```

**Comment by rvilarl** [comment]
> I quickly looked at this. It seems like now you are still sorting, but keeping the sorted list in: `sortedKeyProps`, is that correct?

Yes this is correct. Some internal routines require the pitches to be sorted.

> I just want to make sure it doesn't break any existing code. I noticed that you changed some of the test code. If we leave the test code as is (no changes), will anything break due to this PR?

It does not break any existing code. If the pitches are sorted it also works. I modified the test to show that even if they are not sorted, the result is the same.

**Comment by rvilarl** [comment]
@0xfe is there anything missing in this PR? I think that this a very important improvement. Formats like MusicXML do not impose this pitch order.

**Comment by 0xfe** [comment]
Thanks Rodrigo. Change looks good. I'll merge this, however you can also add Ron's comment rewording separately.

File: tests/accidental_tests.ts [hunk 2]
```diff
@@ -238,26 +238,27 @@ function basic(options: TestOptions): void {
...
-      .addModifier(accid('n'), 5)
-      .addModifier(accid('bb'), 6),
+      .StaveNote({ keys: ['e/4', 'f/4', 'a/4', 'c/5', 'e/5', 'g/5', 'd/4'], duration: '2' })
+      .addModifier(accid('##'), 6)
+      .addModifier(accid('n'), 0)
+      .addModifier(accid('bb'), 1)
+      .addModifier(accid('b'), 2)
+      .addModifier(accid('#'), 3)
+      .addModifier(accid('n'), 4)
+      .addModifier(accid('bb'), 5),
```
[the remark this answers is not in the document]
rvilarl: Hi Ron I was looking with my mobile and I missed these comments sorry. As explained below, this is just changed to proof that order is not required.

src/stavenote.ts
rvilarl: Done directly on master.

rvilarl:
Review comment updated directly on master

PR #1395, opened 2022-05-27
leland support

### LINKED ISSUE #1391: Leland font
@AaronDavidNewman I believe that the only difference between vexflow_smoosic and vexflow is the support to leland font. Would you like to include it here and use vexflow directly?

rvilarl:
fixes #1391 
I have also adapted the font generators to round the decimals. This produces no visual differences.

Leland is included in the test set.


ronyeh:
Thanks. I can probably review this PR starting next Tuesday.

0xfe:
Can we put the rounding changes into a separate PR please? (also, use `toFixed(3)` instead of rounding. The precision came straight from the font files, and I'm guessing they support high zoom levels.)

rvilarl:
> Can we put the rounding changes into a separate PR please? (also, use `toFixed(3)` instead of rounding. The precision came straight from the font files, and I'm guessing they support high zoom levels.)

Hi Mohit, the decimals come from a bounding box calculation. The glyphs themselves did not have decimals. ToFixed returns a string and uses strings in general for the calculation.

rvilarl:
@0xfe as explained in the comment above I believe that the glyphs are now more consistent rounding the bounding boxes as well. The usage of math.round is also consitent with `fix`. Do you agree or do you want still me to modify something?

0xfe:
Rodrigo, can we separate the rounding changes in a different PR please, and leave this just for adding Leland? (otherwise we're mixing topics, and it's harder to do the review.)

rvilarl:
@0xfe done ready for review

src/fonts/load_all.ts
```diff
@@ -16,6 +17,7 @@ export function loadAllMusicFonts(): void {
   loadGonville();
   loadPetaluma();
   loadCustom();
+  loadLeland();
```
ronyeh, line 20: Adding `loadLeland()` to this `load_all.ts` file means that the main vexflow.js build will get larger, since it now includes Leland. Are we all agreeing to this? I'm fine with making the main vexflow.js larger (we can tag this as the 4.1 release).... but I assume we should have a limit, no? We can't include _every single_ SMuFL font.... :-)

ronyeh: Tagging @AaronDavidNewman @0xfe @sschmidTU @mscuthbert ...

I think @AaronDavidNewman said he's fine with making the vexflow.js build larger.

I am also OK with this (for now) as long as we have the individual `vexflow-{single_font_name}.js` variants. Anyone who wants a smaller library can use `vexflow-bravura.js` or `vexflow-gonville.js`, etc...

tests/vexflow_test_helpers.ts
```diff
@@ -185,6 +185,7 @@ export class VexFlowTests {
     Bravura: ['Bravura', 'Gonville', 'Custom'],
     Gonville: ['Gonville', 'Bravura', 'Custom'],
     Petaluma: ['Petaluma', 'Gonville', 'Bravura', 'Custom'],
+    Leland: ['Leland', 'Bravura', 'Gonville', 'Custom'],
```
ronyeh, line 188: We probably don't need 'Gonville' here. Since Bravura is a complete SMuFL font, removing Gonville shouldn't have any impact on the tests.

rvilarl: I also thought so, but I made the test and `vexWiggleArpeggioUp` is only defined in `Gonville`.

:( No change done.

rvilarl: @0xfe I have seen that `vexWiggleArpeggioUp` is a 90 degrees rotated version of `wiggleArpeggiatoUp` does Vexflow provide runtime rotation or do theses glyphs have to be rotated offline?

rvilarl: They have to be rotated. See https://github.com/w3c/smufl/issues/218#issuecomment-1146567666

rvilarl:
@0xfe  any thoughts on this one?

ronyeh:
OK I'm testing this PR right now.

ronyeh:
`flow.html` tests on a 2016 Macbook Pro:

BEFORE:
```
Tests completed in 15783 milliseconds.
7221 assertions of 7221 passed, 0 failed.
```

AFTER:
```
Tests completed in 19990 milliseconds.
8857 assertions of 8857 passed, 0 failed.
```

It adds 4 seconds to the tests on my machine. This is fine, but perhaps not scalable if we ever plan to add even more SMuFL fonts.

I'm personally happy with sticking with these four fonts for the time being: Bravura, Gonville, Petaluma, Leland.

Bravura is the only "complete" font with all of the SMuFL glyphs.

If there is ever a strong desire from the VexFlow community to add even more SMuFL fonts, we should figure out "the right way" to do it.

Perhaps the eventual right way (e.g., VexFlow 5, haha) is to just have:
* `vexflow.js` which contains zero fonts, but includes all the logic for arranging and formatting the glyphs. This is the vexflow-core.js of today.
* `vexflow-font-XXXX-metrics.js` which contains the hints for each font (i.e., our "metrics" files)
* `XXXX.otf` which contains the glyphs for each font in OpenType format. We drop our "glyphs" files completely, and instead load OTF files at runtime via the FontFace API.

ronyeh:
OK, I think I'm going to go ahead and merge this.

I tested and it looks OK. The only issue I discovered was that `vexflow-core.js` (the lib without any fonts) couldn't find Leland at runtime. This is because this `fontModules` mapping needs to be updated as well:

https://github.com/0xfe/vexflow/blob/c93de0ae99072a171774d8a445cd23729b8d631f/entry/vexflow-core.ts#L17-L22

I'll just patch it manually after the merge.

PR #1401, opened 2022-06-02
additional noteheads supported

### LINKED ISSUE #1400: Noteheads (Unofficial MusicXML Test Suite 22a)
There are a few notehead missing to support this test.

image

rvilarl:
fixes #1400 

Some alignment still needs to be fine adjust but it is good enough to make a first review.

Visual differences (Bravura), double whole glyphs were missing
StaveNote StaveNote_Draw___Harmonic_And_Muted Bravura
StaveNote StaveNote_Draw___Harmonic_And_Muted Bravura_current
StaveNote StaveNote_Draw___Harmonic_And_Muted Bravura_reference

New tests (Bravura)
NoteHead Various_Note_Heads_1 Bravura
NoteHead Various_Note_Heads_2 Bravura


src/flow.ts
```diff
@@ -328,8 +328,8 @@ export class Flow {
     return Tables.clefProperties(clef);
   }
   // eslint-disable-next-line
-  static keyProperties(key: string, clef?: string, params?: any): any {
-    return Tables.keyProperties(key, clef, params);
+  static keyProperties(duration:string, key: string, clef?: string, params?: any): any {
```
ronyeh, line 331: This changes the method signature. If you need to add this parameter, we should do it in a backwards compatible way if at all possible (unless there is a very compelling reason to change the method signature).

rvilarl: In VexFlow there are two methods to specify the note heads:
- using `type` which affects all the note heads in a chord (this method takes into account the duration to select the appropriate note head ie: black, half...)
- using and additional parameter on the 'key' string (this was only thought for custom noteheads and does not take into accout the duration)
however it is possible to have different types of note heads that should take into account the duration.

As a consequence I made the second method duration aware.

src/tables.ts
```diff
@@ -156,34 +156,10 @@ const validNoteTypes: Record<string, { name: string }> = {
...
-  S1: { code: 'noteheadSquareWhite' },
-  S2: { code: 'noteheadSquareBlack' },
-
-  /* Rectangle */
-  R1: { code: 'vexNoteHeadRectWhite' }, // no smufl code
-  R2: { code: 'vexNoteHeadRectBlack' }, // no smufl code
+  d: { name: 'diamond' },
+  cx: { name: 'circle x' },
+  tu: { name: 'triangle up' },
+  td: { name: 'triangle down' },
```
rvilarl: Add missing type definitions.

src/flow.ts
rvilarl: Done. Changed to addtional param.

src/tables.ts
rvilarl: Done

rvilarl:
@ronyeh comments addressed

tools/fonts/config/valid_codes.js
```diff
@@ -52,21 +52,51 @@ module.exports = {
...
   noteheadBlack: 'vb',
+  noteheadXDoubleWhole: 'vxx',
   noteheadXWhole: 'v92',
   noteheadXHalf: 'v95',
   noteheadXBlack: 'v3e',
-  noteheadCircleX: 'v3b',
+  noteheadCircleXDoubleWhole: 'vxx',
+  noteheadCircleXWhole: 'vxx',
+  noteheadCircleXHalf: 'vxx',
+  noteheadCircleX: 'vxx',
```
ronyeh, line 62: Can you explain the significance of 'vxx' vs null in this `valid_codes.js` file? I see that some of the fields are null, but others are 'vxx'. 

How come noteheadCircleX changes from v3b to vxx in this case?

rvilarl: > Can you explain the significance of 'vxx' vs null in this `valid_codes.js` file? I see that some of the fields are null, but others are 'vxx'. 

In the generation process of the Gonville and Custom fonts they end up on different log messages:
- null -> No Glyph for...
- vxx -> No Outline for...
But there is no impact on the generated fonts. See fontgen-gonville-custom.js 

> How come noteheadCircleX changes from v3b to vxx in this case?

The size of this circlex head for Gonville does not much the size of the other durations' circlex heads in Bravura. I decided to disable it and fall back to Bravura for the complete set.

rvilarl: The circle x, which is now smaller, and also `noteheadDiamondWhole`, which was pointing to a wrong Glyph, originates other visual difference in Gonville that I forgot to document (I only looked into Bravura).

NoteHead Drum_Chord_Heads Gonville
NoteHead Drum_Chord_Heads Gonville
current
NoteHead Drum_Chord_Heads Gonville_current
reference
NoteHead Drum_Chord_Heads Gonville_reference
NoteHead Various_Heads Gonville
NoteHead Various_Heads Gonville
current
NoteHead Various_Heads Gonville_current
reference
NoteHead Various_Heads Gonville_reference
Percussion Percussion_Basic2 Gonville
Percussion Percussion_Basic2 Gonville
current
Percussion Percussion_Basic2 Gonville_current
reference
Percussion Percussion_Basic2 Gonville_reference
Percussion Percussion_Notes Gonville
Percussion Percussion_Notes Gonville
current
Percussion Percussion_Notes Gonville_current
reference
Percussion Percussion_Notes Gonville_reference
Percussion Percussion_Snare1 Gonville
Percussion Percussion_Snare1 Gonville
current
Percussion Percussion_Snare1 Gonville_current
reference
Percussion Percussion_Snare1 Gonville_reference

ronyeh:
I'll be away from computer for next 9 days or so. Will be able to merge PRs and do releases when I am back to my laptop.

ronyeh:
I'm currently reviewing & testing this PR.

tools/fonts/config/valid_codes.js
```diff
@@ -52,21 +52,51 @@ module.exports = {
...
+  noteheadSquareBlack: 'vxx',
+  noteheadTriangleDownDoubleWhole: 'vxx',
+  noteheadTriangleDownWhole: 'vxx',
+  noteheadTriangleDownHalf: 'vxx',
+  noteheadTriangleDownBlack: 'vxx',
+  noteheadTriangleUpDoubleWhole: 'vxx',
   noteheadTriangleUpWhole: 'v49',
   noteheadTriangleUpHalf: 'v93',
   noteheadTriangleUpBlack: 'v40',
-  noteheadDiamondWhole: 'v46',
```
ronyeh, line 62: Just a quick question here. This changed from v46 to v27. Since I'm not familiar with these v codes, I'm just noting this to make sure it was an intentional change. :-)

ronyeh:
OK I examined the test output and it looks OK.

I'm going to merge this, but there needs to be more work on getting the noteheads to match up with the stems, especially in the Leland font.

Some of the note head sizes also have gotten smaller, so now they no longer connect with the stems properly.

See: the test cases that start with "Various" on the test page: `flow.html?module=NoteHead`

Let's get the stems + noteheads fixed before shipping the next release!

tools/fonts/config/valid_codes.js
rvilarl: yes it was, the code was not correctly mapped. You can look at the codes with

tools/fonts/inspector/gonville.html

Issue #97, opened 2013-02-05
Add support for cross stave beams.


0xfe:
This would require small changes to the beam and stem drawing code.

Also, ghost notes will have to be added to partner staves for synchronization.

apendua:
Does anyone know what's the current status of this ticket?

dfcreative:
Yep, very interested too )

mscuthbert:
If you haven't seen the Steinberg development blog, it's fun reading in general for anyone developing VexFlow but this post is particularly relevant: http://blog.steinberg.net/2015/03/development-diary-part-10/  The takeaway is that making the beams on cross-staff beaming is really hard and except for Lilypond, none of the current software gets it right.  I suggest merging this pull request when we can get the code working with the latest VexFlow version and then making the remaining problems be separate tickets to work on later.

dfcreative:
Seems that it is very requested feature, maybe we should set bounty on that?

Silverwolf90:
Sorry I totally dropped the ball on this. I simply don't have time to contribute actively anymore :/ [I've pushed my branch with my work](https://github.com/Silverwolf90/vexflow/tree/feature/cross-stave-beams). It's *really* old (~2 years). And it's a **massive** hack that barely works, and is without any sophistication, ie: does not avoid the beam-corners that Daniel identifies in that blog post.

0xfe:
No problem -- thanks for everything so far!

mscuthbert:
Maybe beam corners in cross-staff beaming aren't as important as people think they are.  Here from two consecutive measures of a Brahms Piano part (Cl. sonata, op. 120 nr 1 mvmt 2), Breitkopf and Härtel -- superstar engraver still does two consecutive, nearly identical cross-staff beam corners differently:

![screen shot 2016-07-16 at 18 48 29](https://cloud.githubusercontent.com/assets/3521479/16897665/e789378a-4b85-11e6-9d97-705645020329.png)

If this is the best of the best screwing up manually, I think we can accept "pretty good" automatedly. :-)

(note also the impossible rhythm in the r.h. soprano of the piano)

Silverwolf90:
I'm doing some refactoring around beams/stems at the moment. I suspect cross-stave beams will be a bit easier to implement when I'm done. Maybe I'll give it another shot!

SalahAdDin:
Holly good, it's really awesome.

0xfe:
Closing old issue.

[...]

PR #1409, opened 2022-06-14
cross stave voices

### LINKED ISSUE #1373: Design issue: Notes in a voice cannot be placed on different staves (using System)
This is not allowed by VexFlow and it seems that the solution is to make two voices out of one adding GhostNotes. However this is an impediment to have slurs or beams between staves (see #434)

I am collaborating with @jaredjj3 on https://github.com/stringsync/vexml. This is the application that I am using to produce the jsfiddle examples in the OSMD issues from musicxml files.

There I am facing currently this design issue with **test_clef_end_measure.xml** from OSMD

@AaronDavidNewman once vexml is more mature we will be able to generate real scores for testing. As you commented once it would be good to have more real score testing. Or as @mscuthbert did use http://lilypond.org/doc/v2.20/input/regression/musicxml/collated-files.html

@0xfe Do we want to make a design change to allow notes from one voice on different staves?

rvilarl:
Fixes #1373

It is also a first step to allow cross stave beams #97 in this regard I have added a test where the voice is crossing staves. What is missing is to locate the beam between the staves.

**Formatter Vertical_alignment___cross_stave Bravura**
Formatter Vertical_alignment___cross_stave Bravura

### DIFF
--- a/src/system.ts
+++ b/src/system.ts
@@ -27,6 +27,15 @@ export interface SystemStave {
   debugNoteMetrics?: boolean;
 }
 
+interface StaveInfo {
+  stave: Stave;
+  noJustification: boolean;
+  options: StaveOptions;
+  spaceAbove: number;
+  spaceBelow: number;
+  debugNoteMetrics: boolean;
+}
+
 /**
  * Formatting for systems created/drawn from factory:
  *
@@ -68,14 +77,16 @@ export class System extends Element {
   protected formatter?: Formatter;
   protected startX?: number;
   protected lastY?: number;
-  protected parts: Required<SystemStave>[];
+  protected partStaves: StaveInfo[];
+  protected partVoices: Voice[];
   protected connector?: StaveConnector;
-  protected debugNoteMetricsYs?: { y: number; voice: Voice }[];
+  protected debugNoteMetricsYs?: { y: number; stave: Stave }[];
 
   constructor(params: SystemOptions = {}) {
     super();
     this.setOptions(params);
-    this.parts = [];
+    this.partStaves = [];
+    this.partVoices = [];
   }
 
   /** Set formatting options. */
@@ -123,8 +134,8 @@ export class System extends Element {
    */
   addConnector(type: StaveConnectorType = 'double'): StaveConnector {
     this.connector = this.factory.StaveConnector({
-      top_stave: this.parts[0].stave,
-      bottom_stave: this.parts[this.parts.length - 1].stave,
+      top_stave: this.partStaves[0].stave,
+      bottom_stave: this.partStaves[this.partStaves.length - 1].stave,
       type,
     });
     return this.connector;
@@ -163,16 +174,28 @@ export class System extends Element {
     };
 
     const ctx = this.getContext();
-    p.voices.forEach((voice) =>
+    p.voices.forEach((voice) => {
       voice
         .setContext(ctx)
         .setStave(stave)
         .getTickables()
-        .forEach((tickable) => tickable.setStave(stave))
-    );
+        .forEach((tickable) => tickable.setStave(stave));
+      this.partVoices.push(voice);
+    });
 
-    this.parts.push(p);
+    this.partStaves.push(p);
     return stave;
+  }
+
+  /**
+   * Add voices to the system with stave already assigned.
+   */
+  addVoices(voices: Voice[]) {
+    const ctx = this.getContext();
+    voices.forEach((voice) => {
+      voice.setContext(ctx);
+      this.partVoices.push(voice);
+    });
   }
 
   /** Format the system. */
@@ -184,32 +207,39 @@ export class System extends Element {
 
     let y = this.options.y;
     let startX = 0;
-    let allVoices: Voice[] = [];
     let allStaves: Stave[] = [];
-    const debugNoteMetricsYs: { y: number; voice: Voice }[] = [];
+    const debugNoteMetricsYs: { y: number; stave: Stave }[] = [];
 
-    // Join the voices for each stave.
-    this.parts.forEach((part) => {
+    this.partStaves.forEach((part) => {
       y = y + part.stave.space(part.spaceAbove);
       part.stave.setY(y);
-      formatter.joinVoices(part.voices);
       y = y + part.stave.space(part.spaceBelow);
       y = y + part.stave.space(this.options.spaceBetweenStaves);
       if (part.debugNoteMetrics) {
-        debugNoteMetricsYs.push({ y, voice: part.voices[0] });
+        debugNoteMetricsYs.push({ y, stave: part.stave });
         y += 15;
       }
-      allVoices = allVoices.concat(part.voices);
       allStaves = allStaves.concat(part.stave);
 
       startX = Math.max(startX, part.stave.getNoteStartX());
     });
 
+    // Re-assign Stave to update y position
+    this.partVoices.forEach((voice) => {
+      voice.getTickables().forEach((tickable) => {
+        const stave = tickable.getStave();
+        if (stave) tickable.setStave(stave);
+      });
+    });
+
+    // Join the voices
+    formatter.joinVoices(this.partVoices);
+
     // Update the start position of all staves.
-    this.parts.forEach((part) => part.stave.setNoteStartX(startX));
-    if (this.options.autoWidth) {
-      justifyWidth = formatter.preCalculateMinTotalWidth(allVoices);
-      this.parts.forEach((part) => {
+    this.partStaves.forEach((part) => part.stave.setNoteStartX(startX));
+    if (this.options.autoWidth && this.partVoices.length > 0) {
+      justifyWidth = formatter.preCalculateMinTotalWidth(this.partVoices);
+      this.partStaves.forEach((part) => {
         part.stave.setWidth(justifyWidth + Stave.rightPadding + (startX - this.options.x));
       });
     } else {
@@ -217,7 +247,10 @@ export class System extends Element {
         ? this.options.width - (startX - this.options.x)
         : this.options.width - (startX - this.options.x) - Stave.defaultPadding;
     }
-    formatter.format(allVoices, this.options.noJustification ? 0 : justifyWidth, this.options.formatOptions);
+    if (this.partVoices.length > 0) {
+      formatter.format(this.partVoices, this.options.noJustification ? 0 : justifyWidth, this.options.formatOptions);
+    }
+    formatter.postFormat();
 
     for (let i = 0; i < this.options.formatIterations; i++) {
       formatter.tune(options_details);
@@ -244,7 +277,11 @@ export class System extends Element {
     }
 
     this.debugNoteMetricsYs.forEach((d) => {
-      d.voice.getTickables().forEach((tickable) => Note.plotMetrics(ctx, tickable, d.y));
+      this.partVoices.forEach((voice) => {
+        voice.getTickables().forEach((tickable) => {
+          if (tickable.getStave() === d.stave) Note.plotMetrics(ctx, tickable, d.y);
+        });
+      });
     });
   }
 }
--- a/src/system.ts
+++ b/src/system.ts
@@ -121,6 +121,32 @@ export class System extends Element {
     }
   }
 
+  /** Set origin X. */
+  getX(): number {
+    return this.options.x;
+  }
+
+  /** Get origin X. */
+  setX(x: number) {
+    this.options.x = x;
+    this.partStaves.forEach((s) => {
+      s.stave.setX(x);
+    });
+  }
+
+  /** Set origin y. */
+  getY(): number {
+    return this.options.y;
+  }
+
+  /** Get origin y. */
+  setY(y: number) {
+    this.options.y = y;
+    this.partStaves.forEach((s) => {
+      s.stave.setY(y);
+    });
+  }
+
   /** Set associated context. */
   setContext(context: RenderContext): this {
     super.setContext(context);
@@ -239,8 +265,9 @@ export class System extends Element {
     this.partStaves.forEach((part) => part.stave.setNoteStartX(startX));
     if (this.options.autoWidth && this.partVoices.length > 0) {
       justifyWidth = formatter.preCalculateMinTotalWidth(this.partVoices);
+      this.options.width = justifyWidth + Stave.rightPadding + (startX - this.options.x);
       this.partStaves.forEach((part) => {
-        part.stave.setWidth(justifyWidth + Stave.rightPadding + (startX - this.options.x));
+        part.stave.setWidth(this.options.width);
       });
     } else {
       justifyWidth = this.options.noPadding
--- a/src/system.ts
+++ b/src/system.ts
@@ -28,7 +28,6 @@ export interface SystemStave {
 }
 
 interface StaveInfo {
-  stave: Stave;
   noJustification: boolean;
   options: StaveOptions;
   spaceAbove: number;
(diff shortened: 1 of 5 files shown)


rvilarl:
I also discovered a bug while adding this feature and I added the following test:

**Beam Complex_Beams_with_Articulations_two_Staves Bravura**
pptr-Beam Complex_Beams_with_Articulations_two_Staves Bravura svg
**current**
pptr-Beam Complex_Beams_with_Articulations_two_Staves Bravura svg_current
**reference, as rendered without the src modifications in this PR*
pptr-Beam Complex_Beams_with_Articulations_two_Staves Bravura svg_reference

rvilarl:
Ready to review! No visual differences!

rvilarl:
I have included set/getX and set/getY to facilitate the repositioning of the system's origin when `autowidth` is used. Doing that I have found also a bug in the calculated bounding box when autowidth is `true` 
@0xfe this is ready for review.

src/system.ts
```diff
@@ -239,8 +265,9 @@ export class System extends Element {
     this.partStaves.forEach((part) => part.stave.setNoteStartX(startX));
     if (this.options.autoWidth && this.partVoices.length > 0) {
       justifyWidth = formatter.preCalculateMinTotalWidth(this.partVoices);
+      this.options.width = justifyWidth + Stave.rightPadding + (startX - this.options.x);
       this.partStaves.forEach((part) => {
-        part.stave.setWidth(justifyWidth + Stave.rightPadding + (startX - this.options.x));
+        part.stave.setWidth(this.options.width);
```
AaronDavidNewman, line 270: Kind of a picky point, but I feel like `options` isn't the right place to store this value.  options is stuff set in the constructor.  Probably what makes more sense is to change how bounding box is computed if autowidth is set.

```
let computedWidth = this.options.width;
if (this.options.autoWidth && this.partVoices.length > 0) {
  computedWidth = ...
} else { ...
}

```

And then use `computedWidth` for bounding box.  The logic here all looks correct, though.

rvilarl: @AaronDavidNewman Thanks for the review!
I started with a similar approach to your suggestion making the computation close to
```js
this.boundingBox = new BoundingBox(this.options.x, this.options.y, this.options.width, this.lastY - this.options.y);
```
but I decided to change for the following reasons:
- the calculation was duplicating code already there to calculate `justifyWidth`
- `set/getX` `set/getY` had to store the value in `this.options` to make sure that the values are used in `format`
- `setOptions` is a public method. So the values can be changed after construction
- when `options.autoWidth` is set `options.width`is set with a 500 (default value) which is of no use

The logic works! I have a PR prepared for vexml with the following code inspired on your suggestion in https://github.com/0xfe/vexflow/issues/1411

```js
let prevSystem: VF.System | undefined;
systems.forEach((s) => {
  if (prevSystem) {
    let x = prevSystem.getX() + prevSystem.getBoundingBox()!.getW();
    let y = prevSystem.getY();
    if (x > 1000) {
      x = 0;
      y += prevSystem.getBoundingBox()!.getH() + 50;
      s.addConnector('singleLeft');
    } 
    s.setX(x);
    s.setY(y);
  } else {
    s.addConnector('singleLeft');
  }
  s.format();
  prevSystem = s;
});
factory.draw());

```

rvilarl: @AaronDavidNewman I believe that the line 217 has a bug but I left it untouched:
https://github.com/0xfe/vexflow/blob/3cfc2f6f8dd6c807177276549f9a087aabb509e8/src/system.ts#L216-L218
I believe that it should be:
```js
justifyWidth = this.options.noPadding
  ? this.options.width - (startX - this.options.x)
  : this.options.width - (startX - this.options.x) - Stave.defaultPadding;
```
what do you think?

AaronDavidNewman: I also left this alone.  I don't understand the use case for 'no padding', if it is only about the right padding then your way is correct.

rvilarl: I changed it the current calculation could result in negative numbers

rvilarl: I am not sure, I leave it as it is.

rvilarl:
@0xfe ready to review

rvilarl:
@0xfe the last commit is only a test improvement. Ready to review.

rvilarl:
@ronyeh now that you finished noteheads, this is quite interesting, I think.

src/system.ts
```diff
@@ -27,6 +27,14 @@ export interface SystemStave {
   debugNoteMetrics?: boolean;
 }
 
+interface StaveInfo {
```
0xfe, line 30: I don't understand why we need this separately? Why not use `SystemStave` above?

src/system.ts
```diff
@@ -68,14 +76,18 @@ export class System extends Element {
...
+  protected partVoices: Voice[];
   protected connector?: StaveConnector;
-  protected debugNoteMetricsYs?: { y: number; voice: Voice }[];
+  protected debugNoteMetricsYs?: { y: number; stave: Stave }[];
 
   constructor(params: SystemOptions = {}) {
     super();
     this.setOptions(params);
-    this.parts = [];
+    this.partStaves = [];
```
0xfe, line 88: Here too, not fully clear why these three are separate?

0xfe (commented):
Thanks Rodrigo. Some thoughts.

src/system.ts
rvilarl: I left `SystemStave` so as not to modify the interface of `System.addStave`. However the relation `voices` to `staves` is not N to 1, one `Voice` can have `Notes` in different `Staves`, as a consequence, they must be sotred separately to allow the N to M relation. Then `SatveInfo` is only used in `System.format` for this reason I placed it apart. This also allows to return `this.partStaves` directly in `System.getStaves()`.

src/system.ts
rvilarl: See above

rvilarl:
@ronyeh @0xfe once we merge this one I think that we can release 4.1

rvilarl:
@0xfe is the answer to your comments ok? Do you want to follow another approach?

0xfe:
Hi Rodrigo -- sorry, I haven't had a chance to think this through deeply. Cross-stave notation has a lot of corner cases, and this needs a bit more thought in the design so we know we can eventually support them. The three things to think about are: 1) up/down stems under a single beam, with different durations, 2) handling collisions across voices (accept, ignore, reformat, etc.), and 3) formatting.

I think (3) might "just work", and for (2) we can support some simple cases. For (1) though, I'm not sure we can robustly handle it right now.

rvilarl:
@0xfe sorry I do not know how to proceed now.

rvilarl:
> Yes, sorry, understood. I meant to say that I need more time to review this properly -- I think we'll need to think about a design at a bit of a higher level, because this specific change feels a little like a stopgap, not necessarily something that we can build into a more robust solution.

@0xfe The design responds to the following...
Requiremets:
1. The relation Stave to Voice is M:N and not 1:N
2. The interface of the class must be backwards compatible

Analysis:
- `protected parts: Required<SystemStave>[];` constraints the relation Stave to Voice to 1:N. In order to resolve that: the storage of Staves and Voices have been located in different arrays. The information associated to Staves has been placed in third array with relation 1:1 to the Stave array based on the index within the array.
- `addStave(params: SystemStave): Stave` interface stays unchanged. An empty array of voices is a valid parameter.
- a new method `addVoices(voices: Voice[])` has been added to allow the addition of Voices independently.
 
> 
> One way to go about this (without blocking on me) would be to have a lot of tests covering all the cases we want this to cover. E.g., if you can add tests covering cross stave notation across different combinations of beams, stem directions, and durations, we can have some grounding on the requirements.

I have already includede the three scenarios that I wanted to test:
```
run('Vertical alignment - cross stave (beam up)', crossStave, { stem1: 'up', stem2: 'up' }); 
run('Vertical alignment - cross stave (beam down)', crossStave, { stem1: 'down', stem2: 'down' });
run('Vertical alignment - cross stave (beam middle)', crossStave, { stem1: 'down', stem2: 'up' });
```
The third one is not placing the beam coreectly but this fix is not part of this PR

rvilarl:
@0xfe any feedback on the above?

rvilarl:
> Hi Rodrigo -- sorry, I haven't had a chance to think this through deeply. Cross-stave notation has a lot of corner cases, and this needs a bit more thought in the design so we know we can eventually support them. The three things to think about are: 1) up/down stems under a single beam, with different durations, 2) handling collisions across voices (accept, ignore, reformat, etc.), and 3) formatting.
> 
> I think (3) might "just work", and for (2) we can support some simple cases. For (1) though, I'm not sure we can robustly handle it right now.

I have added tests to address the concern 1). It was already clear and now more evident that the middle beam is not working properly. I am not aiming to correct this issue in this PR, I am aiming to allow a single voice to move between different staves. I already stated on the 22nd of June:

> I have expanded the test to show the different beam positions (up, down, middle) on cross stave voices. The one in the middle needs still some changes in beam/stem/?. However I believe that it makes sense to have the test case already to show the failure.

New tests without middle beam

Formatter Vertical_alignment___cross_stave__beam_up1_ Bravura
Formatter Vertical_alignment___cross_stave__beam_up2_ Bravura
Formatter Vertical_alignment___cross_stave__beam_up3_ Bravura
Formatter Vertical_alignment___cross_stave__beam_up4_ Bravura
Formatter Vertical_alignment___cross_stave__beam_down1_ Bravura
Formatter Vertical_alignment___cross_stave__beam_down2_ Bravura
Formatter Vertical_alignment___cross_stave__beam_down3_ Bravura
Formatter Vertical_alignment___cross_stave__beam_down4_ Bravura

New tests with middle beam

Formatter Vertical_alignment___cross_stave__beam_middle1_ Bravura
Formatter Vertical_alignment___cross_stave__beam_middle2_ Bravura
Formatter Vertical_alignment___cross_stave__beam_middle3_ Bravura
Formatter Vertical_alignment___cross_stave__beam_middle4_ Bravura

@0xfe @ronyeh I believe that a change to allow cross tab voices is required, however let me know if the approach is in your oppinion wrong and I will close the PR.

sschmidTU:
> New tests with middle beam
> 
image

There seems to be a bug that the last stem is too long, though you're probably aware of that, Rodrigo.
Otherwise, looking good so far! (haven't taken a detailed look yet though)

rvilarl:
> There seems to be a bug that the last stem is too long, though you're probably aware of that, Rodrigo. Otherwise, looking good so far! (haven't taken a detailed look yet though)

This PR does not aim to resolve the bugs with the middle beam. This should be addressed in another PR.

AaronDavidNewman:
@rvilarl this looks really good!  I think I can contribute best by adding some test cases.  I am going to pull this in to Smoosic and try some things.   I don't think I understand it well enough yet to have any opinions on the code or how you did it.

I think this feature deserves its own test cases.  If we keep existing beaming and formatting test cases, it's a little easier to see that any changes we make will be backwards-compatible.  And formatting test cases are already getting pretty crowded.

I do have one question though - have we always supported a middle beam?  Or did you add that also?  I didn't see any test cases with a middle beam that existed before.

rvilarl:
> @rvilarl this looks really good! I think I can contribute best by adding some test cases. I am going to pull this in to Smoosic and try some things. I don't think I understand it well enough yet to have any opinions on the code or how you did it.

@AaronDavidNewman thanks a lot for your feedback! I can drive you through the changes if you want. 
 
> I think this feature deserves its own test cases. If we keep existing beaming and formatting test cases, it's a little easier to see that any changes we make will be backwards-compatible. And formatting test cases are already getting pretty crowded.

You are right they are getting crowded.

> I do have one question though - have we always supported a middle beam? Or did you add that also? I didn't see any test cases with a middle beam that existed before.

Before it was not possible to have a Voice moving through different Staves using System. The representation of the middle Beam just requires to use stem_down in the upper stave and stem_up in the lower one. I have not change the rendering of Beams. Yo can see in the test results that this representation is buggy.

AaronDavidNewman:
I'm going to try to push my new test case file to you, which also incorporates your existing cases.  I notice you didn't have any cross-clef, and all your crossover notes were the last 2.  Probably because this (cross-clef) doesn't work so well with EasyScore.  So I made it possible to choose direction, staff, and clef for each note.

Middle clef seems to just be pretty broken, especially if there is more than one change in direction:

image

AaronDavidNewman:
@rvilarl  it looks like I can't push changes from my fork of vexflow to your fork.  I have my changes on a branch here, you can check it out if you like.  Basically, I created a new file for cross-stave changes, and I added a couple of test cases.  I also made it easier to create different scenarios, you can specific each tick index which beam direction and which stave.

https://github.com/AaronDavidNewman/vexflow/tree/refactor/system

rvilarl:
@AaronDavidNewman thanks a lot, it is a pitty that the PR is not getting through :(

rvilarl:
@AaronDavidNewman thanks for the tests, they are now merged. I have also reverted formatter_tests.ts.
@0xfe the new tests look like this. As I have already explained the aim of this PR is to allow voices to cross staves the rendering bugs of the beams are to be addressed in another PR (@AaronDavidNewman perhaps you could look into that)

Tests added(bravura)
CrossBeam Single_clef_mixed_1 Bravura
CrossBeam Single_clef_mixed_2 Bravura
CrossBeam Mixed_clef_voice_middle Bravura
CrossBeam Vertical_alignment___cross_stave__beam_down1_ Bravura
CrossBeam Vertical_alignment___cross_stave__beam_down2_ Bravura
CrossBeam Vertical_alignment___cross_stave__beam_down3_ Bravura
CrossBeam Vertical_alignment___cross_stave__beam_down4_ Bravura
CrossBeam Vertical_alignment___cross_stave__beam_middle1_ Bravura
CrossBeam Vertical_alignment___cross_stave__beam_middle2_ Bravura
CrossBeam Vertical_alignment___cross_stave__beam_middle3_ Bravura
CrossBeam Vertical_alignment___cross_stave__beam_middle4_ Bravura
CrossBeam Vertical_alignment___cross_stave__beam_up1_ Bravura
CrossBeam Vertical_alignment___cross_stave__beam_up2_ Bravura
CrossBeam Vertical_alignment___cross_stave__beam_up3_ Bravura
CrossBeam Vertical_alignment___cross_stave__beam_up4_ Bravura

rvilarl:
As I already commented in https://github.com/stringsync/vexml/issues/54 this PR is resolving a design problem and a bug:
- without this PR voices can not move between staves, which is indeed a valid use case. `system` in VexFlow is not allowing that and I had to avoid its usage in the current state of this library. Once https://github.com/0xfe/vexflow/pull/1409 is merged I will make a PR using `system`which reduces the complexity in this library.
- I also discussed with Aaron on how to use `system` with `autowidth` in https://github.com/0xfe/vexflow/issues/1411. I have extended the interface of `System` with get/setX & get/setY to simplify this use case. Doing that I discovered a bug, `system` was not providing the recalculated width on its bounding box when `autowidth` was used.
- Well and two other bugs, not all the modifercontexts were only tracked properly with one voice and a call to postFormat was missing. :)

How should we proceed?

rvilarl:
I have the feeling that I have to make this change in smaller steps to improve the progress.

0xfe:
Hi Rodrigo -- Thanks for working on this, and thanks Aaron for reviewing and helping. I've not had the time to engage here, sorry, may day job has had me consumed. I do think breaking it up will help, and also really exploring the range of corner cases (with tests.) If Aaron, or others would like to do the initial reviews for this, it'll make it easier for me to get through it too.

AaronDavidNewman:
I have pulled these changes into Smoosic, and I'm starting to dig into it.  I should have some solid feedback by the end of this weekend, work willing.

AaronDavidNewman:
I put my feedback in the original issue, let me know what you think!

Issue #1415, opened 2022-06-22
Fix stems + noteheads


ronyeh:
https://github.com/0xfe/vexflow/pull/1401#issuecomment-1163651219

Recent PRs have caused some stems to be detached from the noteheads. We need to update the metrics files (esp for Leland) so that the stems and noteheads connect smoothly.

ronyeh:
@rvilarl if you have time can you look into this? I noticed that in the new test cases from recent merges, some note heads don't connect well to their stems. I think we need to adjust the font metrics files (especially in Leland) to make the note heads match up again. For example, a couple of note heads in Gonville changed recently. I think that may have resulted in the note head not aligning with the stems anymore.

I think this issue should be resolved before the 4.1 release, since it is a regression, in some ways.

rvilarl:
@ronyeh this is only adjustment of metric files, I think that it can be done after 4.1, I would not wait for that. My problem is that I doubt on how some noteheads should be connected to the stem. I will look into it. @AaronDavidNewman could you help here? 
@ronyeh by the way have you looked into the open PRs?

rvilarl:
These are the results @ronyeh @AaronDavidNewman what are the must fix according to you?

**Leland**
**NoteHead Various_Heads**
![pptr-NoteHead Various_Heads Leland svg](https://user-images.githubusercontent.com/22865285/176285358-b4f19fb1-776c-4b89-8218-ef9e93fd2067.png)
**NoteHead Various_Heads_1**
![pptr-NoteHead Various_Note_Heads_1 Leland](https://user-images.githubusercontent.com/22865285/176285480-18b8330f-461d-4379-840d-8e035708df23.svg)
**NoteHead Various_Heads_2**
![pptr-NoteHead Various_Note_Heads_2 Leland](https://user-images.githubusercontent.com/22865285/176285545-d8a2dced-95e3-4cc6-99ef-ec3773860911.svg)

**Bravura**
**NoteHead Various_Heads**
![pptr-NoteHead Various_Heads Bravura svg](https://user-images.githubusercontent.com/22865285/176285362-e1caa711-26c7-4230-b3aa-b39b2bdce1d7.png)
**NoteHead Various_Heads_1**
![pptr-NoteHead Various_Note_Heads_1 Bravura](https://user-images.githubusercontent.com/22865285/176285484-64ec6d35-2851-41e9-9300-24eb59c820ed.svg)
**NoteHead Various_Heads_2**
[...]

PR #1420, opened 2022-07-01
notehead: refactor any in glyphProps

rvilarl:
As discussed in #1415 first step removing any from tables.ts

No visual differences.

### DIFF
--- a/src/glyph.ts
+++ b/src/glyph.ts
@@ -12,7 +12,7 @@ import { Category } from './typeguard';
 import { defined, RuntimeError } from './util';
 
 export interface GlyphProps {
-  code_head?: string;
+  code_head: string;
   ledger_code_head?: string;
   dot_shiftY: number;
   position: string;
--- a/src/glyph.ts
+++ b/src/glyph.ts
@@ -13,7 +13,7 @@ import { defined, RuntimeError } from './util';
 
 export interface GlyphProps {
   code_head?: string;
-  leger_code_head?: string;
+  ledger_code_head?: string;
   dot_shiftY: number;
   position: string;
   rest: boolean;
--- a/src/glyph.ts
+++ b/src/glyph.ts
@@ -14,26 +14,26 @@ import { defined, RuntimeError } from './util';
 export interface GlyphProps {
   code_head?: string;
   leger_code_head?: string;
-  dot_shiftY?: number;
-  position?: string;
-  rest?: boolean;
-  line_below?: number;
-  line_above?: number;
-  stem_beam_extension?: number;
-  stem_up_extension?: number;
-  stem_down_extension?: number;
-  stem?: boolean;
+  dot_shiftY: number;
+  position: string;
+  rest: boolean;
+  line_below: number;
+  line_above: number;
+  stem_beam_extension: number;
+  stem_up_extension: number;
+  stem_down_extension: number;
+  stem: boolean;
   code?: string;
   code_flag_upstem?: string;
   code_flag_downstem?: string;
   flag?: boolean;
   width?: number;
   text?: string;
-  tabnote_stem_down_extension?: number;
-  tabnote_stem_up_extension?: number;
-  beam_count?: number;
+  tabnote_stem_down_extension: number;
+  tabnote_stem_up_extension: number;
+  beam_count: number;
   shift_y?: number;
-  getWidth?(a?: number): number;
+  getWidth(a?: number): number;
 }
 
 export interface GlyphOptions {
--- a/src/glyph.ts
+++ b/src/glyph.ts
@@ -7,34 +7,33 @@ import { Element } from './element';
 import { Font, FontGlyph } from './font';
 import { RenderContext } from './rendercontext';
 import { Stave } from './stave';
-import { Stem } from './stem';
 import { Tables } from './tables';
 import { Category } from './typeguard';
 import { defined, RuntimeError } from './util';
 
 export interface GlyphProps {
-  code_head: string;
-  dot_shiftY: number;
-  position: string;
-  rest: boolean;
-  line_below: number;
-  line_above: number;
-  stem_beam_extension: number;
-  stem_up_extension: number;
-  stem_down_extension: number;
-  stem: Stem;
-  code: string;
-  code_flag_upstem: string;
-  code_flag_downstem: string;
-  flag: boolean;
-  width: number;
-  text: string;
-  tabnote_stem_down_extension: number;
-  tabnote_stem_up_extension: number;
-  beam_count: number;
-  shift_y: number;
-  getWidth(a?: number): number;
-  getMetrics(): GlyphMetrics;
+  code_head?: string;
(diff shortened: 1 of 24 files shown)


rvilarl:
@ronyeh could you review?

src/glyph.ts
```diff
@@ -7,34 +7,33 @@ import { Element } from './element';
...
-  width: number;
-  text: string;
-  tabnote_stem_down_extension: number;
-  tabnote_stem_up_extension: number;
-  beam_count: number;
-  shift_y: number;
-  getWidth(a?: number): number;
-  getMetrics(): GlyphMetrics;
+  code_head?: string;
+  leger_code_head?: string;
```
sschmidTU, line 16: it's `ledger` by the way, not `ledger. like `ledger note`.

rvilarl: I also thought about this one, and I found:

https://en.m.wikipedia.org/wiki/Ledger_line

Perhaps we should unify.

rvilarl:
I also have a proposal:
this.glyph and this.code (variables used to store the value returned by glyphProps) are confusing. I would suggest to rename them to this.glyphProps

src/glyph.ts
AaronDavidNewman: @sschmidTU Beginner music books in the US use ledger.  So Hal Leonard does not agree :)  But I went with leger b/c that's what Smufl uses, and I feel like we're a Smufl shop.

rvilarl:
> I also have a proposal: this.glyph and this.code (variables used to store the value returned by getGlyphProps) are confusing. I would suggest to rename them to this.glyphProps

Done

rvilarl:
Doing the change I realised that `ClefNote` should inherit from `GlyphNote`

rvilarl:
No vvisual differences and ready for review

src/glyph.ts
0xfe: I prefer using `ledger` in this codebase (mainly to be consistent.)

rvilarl: Ok I understand everything ledger except for the Smufl glyph IDs.

rvilarl:
@0xfe leger topic noted. Anything else? This is ready for review.

src/glyph.ts
rvilarl: Done

src/multimeasurerest.ts
```diff
@@ -227,7 +227,7 @@ export class MultiMeasureRest extends Element {
       x += glyphs[2].width + spacing;
     }
     for (let i = 0; i < n1; ++i) {
-      Glyph.renderGlyph(ctx, x, yTop, rest_scale, rest.glyph_code);
+      if (rest.glyph_code) Glyph.renderGlyph(ctx, x, yTop, rest_scale, rest.glyph_code);
```
0xfe, line 230: This looks like a potentially silent bug. Shouldn't there always be a `glyph_code` if we're here?

0xfe (commented):
Hi Rodrigo -- I like that ClefNote is not based on GlyphNote, however I'm not sure that this change to eliminate `any` is worth it.

We've replaced a lot of required struct members to optional, and now it becomes very difficult to reason about in code because it's hard to know if an instance of that struct will have a member set or not, so there's a littering of `if (glyphProps.foo) { ... }`, which is either unnecessary code, or creates silent failures.

I think it makes the codebase more complicated.

rvilarl:
> Hi Rodrigo -- I like that ClefNote is not based on GlyphNote, 

Ok I can revert this one.

> however I'm not sure that this change to eliminate `any` is worth it.
> We've replaced a lot of required struct members to optional...

I believe that elminating any is what reveals silent failures. I have only made optional those members which where optionally defined. If they are required, we should make them required and make sure that they are defined which is today, I believe, not the case.

src/multimeasurerest.ts
rvilarl: Slash notes do not define a `code_head` this is why `glyph_code` is also optional. I could define `code_head` for slashes as 'N/A' or '' and make the attribute not optional. Would you prefer this approach?

rvilarl: code_head no longer optional.

rvilarl:
ClefNote inherits from Note
@0xfe ready to review

0xfe:
On Sat., Jul. 9, 2022, 11:43 Rodrigo Vilar, ***@***.***>
wrote:

> Hi Rodrigo -- I like that ClefNote is not based on GlyphNote,
>
> Ok I can revert this one.
>

Sorry typo, I meant _now_ not _not_ -- I like this change. Should prob be a
separate PR.

> however I'm not sure that this change to eliminate any is worth it.
> We've replaced a lot of required struct members to optional...
>
> I believe that elminating any is what reveals silent failures.
>

Removing any this way actually masks them, not reveals them. With the way
it is right now (i.e., with any), at least incorrect code will fail loudly.
With these changes, you'll need to check for the existence of each optional
anytime the struct is used, and then do something specific about it -- the
easiest is to ignore the paths, which causes the silent failures.

Does that make sense?


ave only made optional those members which where optionally defined. If

rvilarl:
>> however I'm not sure that this change to eliminate any is worth it.
>> We've replaced a lot of required struct members to optional...
>>
>> I believe that elminating any is what reveals silent failures.
>>
> Removing any this way actually masks them, not reveals them. With the way
> it is right now (i.e., with any), at least incorrect code will fail loudly.
> With these changes, you'll need to check for the existence of each optional
> anytime the struct is used, and then do something specific about it -- the
> easiest is to ignore the paths, which causes the silent failures.
> 
> Does that make sense?

@0xfe I think that we are consideering reveal in two different scenarios. I am considering compile-time (this is how I discovered that stem was a boolean) and you are considering run-time. If you want to fail them as before I can use the `as` or the `!` operators rather than checking with `if`. Please note that this PR also resolves the overlap of `GlyphProps` and `Glyph` which is very confusing.

0xfe:
Yes, I understand that removing `any` gives us compile-time checking for free -- I'm not disputing that. I'm saying that the tradeoff here is creating lots of optional fields, that requires us to scatter run-time checks (with `if` or `!`) across the codebase which can mask errors, which IMO is worse.

(Right now the `any` will at least fail loudly at runtime if we got something wrong -- so long as we have good test coverage, we can keep the code robust.)

So if we want to eliminate `any`, we should think about this tradeoff. All the optionals here signal to me that there's some hidden substructures that we're not seeing.  E.g., is there a pattern where we can break up `glyphProps` into different interfaces or something?

rvilarl:
I believe that there are two options:
1. separate `durationCodes` in two arrays one with the `common` elements and make it of type `GlyphProps` rather than `Partial<GlyphProps>`
2.  fill all the defaults in `getGlyphProps`and leave the `durationCodes` as `Partial<GlyphProps>`

@0xfe what do you think?

rvilarl:
Closed to focus on #1409 in order to release 4.1

PR #1430, opened 2022-08-30
Add getters for six attributes

### LINKED ISSUE #1429: add public methods in several places
(I'm willing to implement this myself.)  

music21j has been upgraded to use Vexflow 4.  :-)  We're thrilled.  (Thanks @ronyeh and others)

Here are some of the places where music21j currently needs to use `@ts-ignore` because we're reaching into private/protected attributes.

Necessary for functioning of music21j.  Only one place:

* Formatter.tickContexts  (AlignmentContexts)

<s>We also currently alter `StaveOptions.line_config` which isn't protected but does not have a public interface.</s> Ah it does! we'll use that.

Here are some protected elements we access for testing the integration of the two libraries:

* KeySignature.glyphs (did the right glyph get appended?)
* StaveTie.notes (are the ties placed and attached to the right notes)
* Stave.measure (getMeasure() -- setMeasure() already exists)
* TextNote.text (are the right TextNotes in the right place?)

Happy to add these if there's no objection.  If setMeasure() wasn't already a public interface, I'd suggest renaming it to `setMeasureNumber` for clarity (and change .measure to .measureNumber internally), but it's already out there, unless we want to add an alias.

mscuthbert:
Fixes #1429

These are places where the music21j project have needed to break TypeScript encapsulation to read or test conversions.

(wow VF 4!  What a joy it is to edit the code now!)

### DIFF
--- a/tests/textnote_tests.ts
+++ b/tests/textnote_tests.ts
@@ -127,15 +127,17 @@ function superscriptAndSubscript(options: TestOptions): void {
       .addModifier(f.Accidental({ type: '#' }), 1),
   ]);
 
-  const voice2 = score.voice([
+  const notes2 = [
     f.TextNote({ text: Flow.unicode.flat + 'I', superscript: '+5', duration: '8' }),
     f.TextNote({ text: 'D' + Flow.unicode.sharp + '/F', duration: '4d', superscript: 'sus2' }),
     f.TextNote({ text: 'ii', superscript: '6', subscript: '4', duration: '8' }),
     f.TextNote({ text: 'C', superscript: Flow.unicode.triangle + '7', subscript: '', duration: '8' }),
     f.TextNote({ text: 'vii', superscript: Flow.unicode['o-with-slash'] + '7', duration: '8' }),
     f.TextNote({ text: 'V', superscript: '7', duration: '8' }),
-  ]);
+  ];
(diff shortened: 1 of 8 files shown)


rvilarl (approved):

tests/textnote_tests.ts
```diff
@@ -127,15 +127,17 @@ function superscriptAndSubscript(options: TestOptions): void {
...
     f.TextNote({ text: 'D' + Flow.unicode.sharp + '/F', duration: '4d', superscript: 'sus2' }),
     f.TextNote({ text: 'ii', superscript: '6', subscript: '4', duration: '8' }),
     f.TextNote({ text: 'C', superscript: Flow.unicode.triangle + '7', subscript: '', duration: '8' }),
     f.TextNote({ text: 'vii', superscript: Flow.unicode['o-with-slash'] + '7', duration: '8' }),
     f.TextNote({ text: 'V', superscript: '7', duration: '8' }),
-  ]);
+  ];
+  equal(notes2[0].getText(), Flow.unicode.flat + 'I', 'TextNote.getText() return .text');
 
+  const voice2 = score.voice(notes2);
```
rvilarl, line 140: What is the aim of this change?

rvilarl: @mscuthbert what does the equal test?

mscuthbert: It just tests that getText() does what it is supposed to. I figured I shouldn't add a public API without at least one test of it.

PR #1432, opened 2022-08-31
System is not handling ModifierContexts correctly with more than one Stave

### LINKED ISSUE #1427: System is not handling ModifierContexts correctly with more than one Stave
I plan to resolve this issue in two PRs:
1. define a test showing the bug.
2. correct the bug.

rvilarl:
This is the second step and resolves #1427: A new test showing the issue. Also related to #1409 simplification.

Before change:

Beam Complex_Beams_with_Articulations_two_Staves Bravura

After change:
Beam Complex_Beams_with_Articulations_two_Staves Bravura

### DIFF
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -52,6 +52,13 @@ export interface AlignmentContexts<T> {
   resolutionMultiplier: number;
 }
 
+export interface AlignmentModifierContexts {
+  list: Map<Stave | undefined, number[]>;
+  map: Map<Stave | undefined, Record<number, ModifierContext>>;
+  array: ModifierContext[];
+  resolutionMultiplier: number;
+}
+
 type addToContextFn<T> = (tickable: Tickable, context: T, voiceIndex: number) => void;
 type makeContextFn<T> = (tick?: { tickID: number }) => T;
 
@@ -67,13 +74,21 @@ function createContexts<T>(
   makeContext: makeContextFn<T>,
   addToContext: addToContextFn<T>
 ): AlignmentContexts<T> {
-  const resolutionMultiplier = Formatter.getResolutionMultiplier(voices);
+  let resolutionMultiplier = 0;
 
   // Initialize tick maps.
   const tickToContextMap: Record<number, T> = {};
   const tickList: number[] = [];
   const contexts: T[] = [];
 
+  if (voices.length == 0)
+    return {
+      map: tickToContextMap,
+      array: contexts,
+      list: tickList.sort((a, b) => a - b),
+      resolutionMultiplier,
+    };
+  resolutionMultiplier = Formatter.getResolutionMultiplier(voices);
   // For each voice, extract notes and create a context for every
   // new tick that hasn't been seen before.
   voices.forEach((voice, voiceIndex) => {
@@ -177,7 +192,7 @@ export class Formatter {
   protected totalShift: number;
   protected tickContexts?: AlignmentContexts<TickContext>;
   protected formatterOptions: Required<FormatterOptions>;
-  protected modifierContexts?: AlignmentContexts<ModifierContext>;
+  protected modifierContexts: AlignmentModifierContexts[];
   protected voices: Voice[];
   protected lossHistory: number[];
   protected durationStats: Record<string, { mean: number; count: number }>;
@@ -413,7 +428,7 @@ export class Formatter {
 
     // Arrays of tick and modifier contexts.
     this.tickContexts = undefined;
-    this.modifierContexts = undefined;
+    this.modifierContexts = [];
 
     // Gaps between contexts, for free movement of notes post
     // formatting.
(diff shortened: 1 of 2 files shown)


src/formatter.ts
```diff
@@ -560,12 +575,60 @@ export class Formatter {
   }
 
   /** Create a `ModifierContext` for each tick in `voices`. */
-  createModifierContexts(voices: Voice[]): AlignmentContexts<ModifierContext> {
-    const fn: addToContextFn<ModifierContext> = (tickable: Tickable, context: ModifierContext) =>
-      tickable.addToModifierContext(context);
-    const contexts = createContexts(voices, () => new ModifierContext(), fn);
-    this.modifierContexts = contexts;
-    return contexts;
+  createModifierContexts(voices: Voice[]) {
```
0xfe, line 578: What's different here from `createContexts`? Is there no way to share the code?

src/formatter.ts
```diff
@@ -52,6 +52,13 @@ export interface AlignmentContexts<T> {
   resolutionMultiplier: number;
 }
 
+export interface AlignmentModifierContexts {
```
0xfe, line 55: Seems like we can merge this with the above, no?

src/formatter.ts
rvilarl: The difference is required by Systems with several Staves. In these cases:
- There is only one TickContext for each tick. This TickContext is shared by all the staves.
- There are N ModifierContext for each tick, where N is the number of staves. ModiferContexts are not shared between Staves.

This is the reason why `AlignmentModifierContexts` below is defining map keyed by Stave. 

I see not easy way to share the code.

src/formatter.ts
rvilarl: See my comment above.

src/formatter.ts
```diff
@@ -67,13 +74,21 @@ function createContexts<T>(
...
   const contexts: T[] = [];
 
+  if (voices.length == 0)
+    return {
+      map: tickToContextMap,
+      array: contexts,
+      list: tickList.sort((a, b) => a - b),
+      resolutionMultiplier,
+    };
+  resolutionMultiplier = Formatter.getResolutionMultiplier(voices);
```
0xfe, line 91: Maybe restructure here to keep `resolutionMultiplier` a const.

```typescript
  // let resolutionMultiplier = 0; <-- REMOVE THIS

  const tickToContextMap: Record<number, T> = {};
  const tickList: number[] = [];
  const contexts: T[] = [];

  if (voices.length == 0)
    return {
      map: tickToContextMap,
      array: contexts,
      list: tickList.sort((a, b) => a - b),
      resolutionMultiplier: 0,
    };
  const resolutionMultiplier = Formatter.getResolutionMultiplier(voices);
```

src/formatter.ts
0xfe: sounds good, thanks

src/formatter.ts
```diff
@@ -560,12 +575,60 @@ export class Formatter {
...
-      tickable.addToModifierContext(context);
-    const contexts = createContexts(voices, () => new ModifierContext(), fn);
-    this.modifierContexts = contexts;
-    return contexts;
+  createModifierContexts(voices: Voice[]) {
+    if (voices.length == 0) return;
+    const resolutionMultiplier = Formatter.getResolutionMultiplier(voices);
+
+    // Initialize tick maps.
+    const tickToContextMap: Map<Stave | undefined, Record<number, ModifierContext>> = new Map();
```
AaronDavidNewman, line 583: The issue as I understand it is that 2 tickables at the same tickcontext can have staves at different y values.  So the MC should be different so the y offset is calculated correctly if it is a staff offset vs. a notehead offset.  But it's also true that stave does not have to be defined here, so we are mapping in Stave | undefined.  That seems kinda gross :)   Can we map instead on the y value of the staves?  If a tickable does not have a stave, we just use 0.  Then it can just be `Record<number, Record <number, ModifierContext>>` .

rvilarl: Yes we could do that, but theoretically we can move the Staves with setY making it difficult to know which is the associated Stave. Anyway I am happy to change if this is a prefered approach.

src/formatter.ts
rvilarl: done

src/formatter.ts
AaronDavidNewman: Now that I think of it, I don't prefer it.  Indexing by 'y' seems like a side-effect, and like you say, the staves could start out with same y and change later.  

It seems like this could be simpler though. Even the original code seems too complicated...

``` javascript
const postFormatContexts = (contexts: AlignmentContexts<ModifierContext> | AlignmentContexts<TickContext>) =>
      contexts.list.forEach((tick) => contexts.map[tick].postFormat());
```
Can't it just be this?
``` javascript
contexts.array.forEach((mc) => mc.postFormat());
```

We should only need a key from the `staff.attrs.id` and `integerTicks`.  So like 'vf-101-1024'.  If there is no stave it is just '-1024'.  Then we can have just `tickToContextMap: Record<string, ModifierContext>` to keep track of MC we've already created.  And once we've called `tickable.addToModifierContext(context);` for all tickables, we don't care about that relationship anymore.  We just need to call `postFormat` on all the MCs, which we can do from an array.

It's late here, so I may be missing something...

rvilarl: @0xfe do you have a preference?

rvilarl: @AaronDavidNewman I have simplified this function removing `postFormatContext` in general the template functions using AlignmentContexts<T> are no longer required to be templates and could be re-writen.

rvilarl:
Comments addressed, ready to merge.

src/formatter.ts
rvilarl: @AaronDavidNewman i believe that the approach you proposed did not work because the TickContexts have to be formatted in tick order. I just thought that this probably not the case for ModifierContexts, I have to try.

AaronDavidNewman (approved):

AaronDavidNewman:
This is good, I finally understand the issue.  The `postFormat`  for the modifier contexts needs to be called in order of ticks.  Otherwise the stem extents aren't calculated correctly and that is why the articulations collide with the beam.

Minor performance thing - I think you can defer sorting the ticks until just before `postFormat` is called.  Otherwise you are sorting once per join, vs. once at the end.

``` javascript
    tickList.forEach((instance) => {
      instance.sort((a, b) => a - b);
    });
```

src/formatter.ts
AaronDavidNewman: For the record, what I said above is incorrect.  We need to call `postFormat` on the MC _in_ _tick_ _order_.  That is the bug here, postFormat was called for all the ticks in one stave, and then all the ticks in a different stave.

rvilarl:
> This is good, I finally understand the issue. The `postFormat` for the modifier contexts needs to be called in order of ticks. Otherwise the stem extents aren't calculated correctly and that is why the articulations collide with the beam.
> 
> Minor performance thing - I think you can defer sorting the ticks until just before `postFormat` is called. Otherwise you are sorting once per join, vs. once at the end.
> 
> ```js
>     tickList.forEach((instance) => {
>       instance.sort((a, b) => a - b);
>     });
> ```

@AaronDavidNewman I was also thinking about that. I think that the order is only required for the TickContexts and not for the ModifierContexts. If the teory is good this would also have an impact on performance. I will give it a try.

rvilarl:
@AaronDavidNewman I was right, the list is not required for ModifierContexts and also modifierContexts was not required to an array.

rvilarl:
Ready to merge now with some simplifications discussed with @AaronDavidNewman

rvilarl:
@0xfe i will now merge. Your comments are addressed and I have one approval. We can revert it if required.

PR #1440, opened 2022-09-24
fix colon in tuplets and dot in debug bars

### LINKED ISSUE #1439: Ratioed tuplets doesn't show dots
Hi all,

As the title suggests, I'm trying to draw some basic tuplets with brackets and ratio, however when I set ratioed to true it draws the ratio without the dots(ie 3 2 instead of 3:2)

Here's a basic fiddle based on Ronyeh's tuplets tutorial

https://jsfiddle.net/01ned978/

Cheers,
Nitzan

rvilarl:
fixes #1439

### DIFF
--- a/src/rendercontext.ts
+++ b/src/rendercontext.ts
@@ -100,7 +100,7 @@ export function drawDot(ctx: RenderContext, x: number, y: number, color = '#F55'
 
   // draw a circle
   ctx.beginPath();
-  ctx.arc(x, y, 3, 0, Math.PI * 2, true);
+  ctx.arc(x, y, 3, 0, Math.PI * 2, false);
   ctx.closePath();
   ctx.fill();
   ctx.restore();
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -442,12 +442,14 @@ export class SVGContext extends RenderContext {
     const x0 = x + radius * Math.cos(startAngle);
     const y0 = y + radius * Math.sin(startAngle);
 
-    // Handle the edge case where arc length is greater than or equal to
-    // the circle's circumference:
-    //   https://html.spec.whatwg.org/multipage/canvas.html#ellipse-method-steps
+    // svg behavior different from canvas.  Don't normalize angles if
+    // we are drawing a circle because they both normalize to 0
+    const tmpStartTest = normalizeAngle(startAngle);
+    const tmpEndTest = normalizeAngle(endAngle);
     if (
       (!counterclockwise && endAngle - startAngle >= TWO_PI) ||
-      (counterclockwise && startAngle - endAngle >= TWO_PI)
+      (counterclockwise && startAngle - endAngle >= TWO_PI) ||
+      tmpStartTest === tmpEndTest
     ) {
       const x1 = x + radius * Math.cos(startAngle + Math.PI);
       const y1 = y + radius * Math.sin(startAngle + Math.PI);
@@ -461,9 +463,8 @@ export class SVGContext extends RenderContext {
       const x1 = x + radius * Math.cos(endAngle);
       const y1 = y + radius * Math.sin(endAngle);
 
-      startAngle = normalizeAngle(startAngle);
-      endAngle = normalizeAngle(endAngle);
-
+      startAngle = tmpStartTest;
+      endAngle = tmpEndTest;
       let large: boolean;
       if (Math.abs(endAngle - startAngle) < Math.PI) {
         large = counterclockwise;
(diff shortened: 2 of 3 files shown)


rvilarl:
image

image

rvilarl:
@AaronDavidNewman thanks for the help!

AaronDavidNewman:
Suggestion: since 'counterclockwise' has no equivalent in SVG and since this code is clearly broken, how about we just remove it from the signature?  If you are worried about backwards-compatibility, we can throw an exception if it is specified as 'true' and make it default to false, and mark that it will be deprecated in the future.

rvilarl:
> Suggestion: since 'counterclockwise' has no equivalent in SVG and since this code is clearly broken, how about we just remove it from the signature? If you are worried about backwards-compatibility, we can throw an exception if it is specified as 'true' and make it default to false, and mark that it will be deprecated in the future.

The code does not seem to be broken, I have tried:
```js
      ctx.beginPath();
      ctx.arc(colon_x, this.y_pos - this.point * 0.08, colon_radius, Math.PI * 2, 0, true);
      ctx.closePath();
      ctx.fill();
      ctx.beginPath();
      ctx.arc(colon_x, this.y_pos + this.point * 0.12, colon_radius, Math.PI * 2, 0, true);
      ctx.closePath();
      ctx.fill();
```
and it also shows the colons properly. If we remove it, we would leave the canvas and the svg renderers inconsistent.

AaronDavidNewman:
>> The code does not seem to be broken, 

It's dependent on the order of angles, though.  If you send the larger angle in first, you have to set 'counterclockwise' with svg, but not canvas.

I think we could fix it in svgcontext by just setting 'counterclockwise' to true/false if startAngle is larger/smaller than endAngle.

``` typescript
  drawArc(context, 40, 25, 5, 0, Math.PI, true);
  drawArc(context, 60, 25, 5, 0, Math.PI, false);
  drawArc(context, 80, 25, 5, 0, 2 * Math.PI, true);
  drawArc(context, 100, 25, 5, 0, 2 * Math.PI, false);
  drawArc(context, 120, 25, 5, Math.PI, 0, false);
  drawArc(context, 140, 25, 5, Math.PI, 0, true);
  drawArc(context, 160, 25, 5, 2 * Math.PI, 0, false);
  drawArc(context, 180, 25, 5, 2 * Math.PI, 0, true);
  drawArc(context, 200, 25, 5, 3 * Math.PI, 0, false);
  drawArc(context, 220, 25, 5, 3 * Math.PI, 0, true);
```
canvas:
image

svg:
image

AaronDavidNewman:
Sorry for all the spam.  The following svg 'arc' function will render same as canvas for circle, semi-circle, regardless of angle order.  Please use this.

``` typescript
arc(x: number, y: number, radius: number, startAngle: number, endAngle: number, counterclockwise: boolean): this {
    const x0 = x + radius * Math.cos(startAngle);
    const y0 = y + radius * Math.sin(startAngle);

    // svg behavior different from canvas.  Don't normalize angles if
    // we are drawing a circle because they both normalize to 0
    const tmpStartTest = normalizeAngle(startAngle);
    const tmpEndTest = normalizeAngle(endAngle);
    if (
      (!counterclockwise && endAngle - startAngle >= TWO_PI) ||
      (counterclockwise && startAngle - endAngle >= TWO_PI) ||
      tmpStartTest === tmpEndTest
    ) {
      const x1 = x + radius * Math.cos(startAngle + Math.PI);
      const y1 = y + radius * Math.sin(startAngle + Math.PI);
      // There's no way to specify a completely circular arc in SVG so we have to
      // use two semi-circular arcs.
      this.path += `M${x0} ${y0} A${radius} ${radius} 0 0 0 ${x1} ${y1} `;
      this.path += `A${radius} ${radius} 0 0 0 ${x0} ${y0}`;
      this.pen.x = x0;
      this.pen.y = y0;
    } else {
      const x1 = x + radius * Math.cos(endAngle);
      const y1 = y + radius * Math.sin(endAngle);

      startAngle = tmpStartTest;
      endAngle = tmpEndTest;
      let large: boolean;
      if (Math.abs(endAngle - startAngle) < Math.PI) {
        large = counterclockwise;
      } else {
        large = !counterclockwise;
      }
      if (startAngle > endAngle) {
        large = !large;
      }

      const sweep = !counterclockwise;

      this.path += `M${x0} ${y0} A${radius} ${radius} 0 ${+large} ${+sweep} ${x1} ${y1}`;
      this.pen.x = x1;
      this.pen.y = y1;
    }
    return this;
  }
```

rvilarl:
done

rvilarl:
All comments included, ready to merge. @AaronDavidNewman if you approve I will merge.

AaronDavidNewman (approved):

AaronDavidNewman:
lgtm

Issue #1447, opened 2022-10-03
Refactor & dynamic styling (step 3)


rvilarl:
I believe that we are have way through and I would like to make styling more consistent. From what I read in this issue, I understand:
- that the styling options should be stored in `Element.attrs`.
- that `Registry` should be the above mentioned StyleManager.

I would also like to include the feature that setting a style attribute in a renderer SVG Element would change the style on the fly without needing to re-render.

_Originally posted by @rvilarl in https://github.com/0xfe/vexflow/issues/285#issuecomment-1264413940_

rvilarl:
> * that the styling options should be stored in `Element.attrs`.
> * that `Registry` should be the above mentioned StyleManager.

@0xfe are these two points right?

PR #1448, opened 2022-10-07
minor style fixes

rvilarl:
- Remove unneccesary check of `this.style!=undefined` before calling `applyStyle` / `restoreStyle` this already checked within these functions.
- Allow ´setStyle´ to also accept `undefined`
- Access `this.style` through `set/getStyle`function
- `flagStyle` not required, `flag` contains its own `ElementStyle`


rvilarl:
@0xfe I will merge this one. It is minor changes without visual differences and will allo me to make a PR with my proposal for #1447

PR #1444, opened 2022-10-02
Refactor & dynamic styling (step 1 Element.attrs as map of strings)

rvilarl:
Fixes #1443 

On the first step to address #285 I have made `Element.attrs` and `Registry` consistent. Both accept `string | undefined`
- `Element.attrs.class` is now a string with a space joined array of class names.

PR #1449, opened 2022-10-07
Refactor & dynamic styling (step 3 styles in Element.attrs)

rvilarl:
#fixes 1447

@0xfe this PR shows my proposal to store the ElementStyles in Element.attrs. I have decided to go for the prefix 'style_' rather than a single string which was much more complex. This is just to ilustrate my proposal.

It also contains the changes from #1444 please focus on the style functions only.


rvilarl:
Now with #1444 merged, it should be easier to review.
Ready to review.

rvilarl:
Why do I make this question?

There are different member variables storing very similar information:
- `Element.style: ElementStyle`
- `Element.attrs: ElementAttributes`
- `Element.textFont: required<FontInfo>`
- `SVGContex.attributes: Attributes`
where the types are defined like that:
```
interface ElementStyle {
  shadowColor?: string;
  shadowBlur?: number;
  fillStyle?: string;
  strokeStyle?: string;
  lineWidth?: number;
}
interface ElementAttributes {
  [name: string]: string | undefined;
  id: string;
  type: string;
  class: string;
}
interface FontInfo {
  family?: string;
  size?: number | string;
  weight?: string | number;
  style?: string;
}
export type Attributes = {
  [name: string]: string | number | undefined;
  'font-family'?: string;
  'font-size'?: string | number;
  'font-style'?: string;
  'font-weight'?: string | number;
};
```
We see already that `element.FontInfo` members are translated into `font-*` members of `SVGContext.attributes`

My suggestion is to translate the members of `Element.styles` into `style-*` members of `Elements.attrs`

rvilarl:
@AaronDavidNewman @mscuthbert @sschmidTU what do you think?

rvilarl:
Other possibilities would be:
- to use `ElementStyle` and `FontInfo` in `SVGContext` rather than `Attributes`
- to merge `ElementStyle` and `FontInfo`

one of the above or both.

rvilarl:
This topic did not get much attention, I close it.

PR #1455, opened 2022-10-14
Resolve issues related to integration of 4.1.0-alpha.0

### LINKED ISSUE #1454: Issues detected on the integration of 4.1.0-alpha.0 in OSMD
> @rvilarl Looks good, thanks for merging! Just one issue:

> I'm getting an error `BadGlyph: Glyph noteheadCircleX does not exist in font.
>    at Glyph.lookupGlyph (glyph.js:120:1)`
> with OSMD Function Test - Notehead Shapes, and the notes don't render.
> They should look like this:
> image

> Same in `test_percussion_display_step_from_xml.musicxml`.

> Also, as with Vexflow 1.2.93 before we patched it out, the Chrome console gives a warning `Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true. See: https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently`.

> We patched that in Vexflow 1.2.93 in https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/1242.
Basically, it was this line in renderer.js:
`this.ctx = Renderer.bolsterCanvasContext(this.element.getContext('2d', {willReadFrequently: true}));`

> You could either set that as a default as well, or offer an option. There seems to be no downside to this, though the performance boost also seems to be minimal.

> I can also merge before these issues are resolved, but then I'll add this to the issue list in #1139.

_Originally posted by @sschmidTU in https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/1247#issuecomment-1279087435_

rvilarl:
fixes #1454 

Two changes:
- Add Gonville noteheadCircleX
- Resolve warning with Canva2D


sschmidTU:
Thanks!
Just for reference, this is the warning with Canvas2D in Chrome:
image

For more details, see https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/issues/1242

rvilarl:
Ready for review.

sschmidTU:
@rvilarl I've tested the circle noteheads and they work great again with the PR.

However, the warning about `willReadFrequently` has not disappeared, and I'm a bit at a loss why. The place where you added the attribute is the only place where the `CanvasRenderingContext2D` is created in vexflow, as far as I can see.

I tried debugging it, but I couldn't find a fix or another place where to set the attribute.
It's possible that the context2d is overwritten in resize(), because on osmd's develop we currently also set the attribute there (which does not fit vexflow 4's signatures). Though this is just speculation.

sschmidTU:
Actually, this code doesn't get executed, so that explains it.
image

Getting no console log.

sschmidTU:
Actually, nevermind, this works.
I just didn't replace the build correctly. I thought OSMD was using the files in `node_modules/vexflow/src`, but it's actually using the build in `node_modules/vexflow/build/esm` (javascript), which makes it a bit harder to hotfix. But no biggie.

sschmidTU (approved):
Looks good to me, tested, works like a charm!
(both the notehead and the chrome warning are fixed)

rvilarl:
Thanks @sschmidTU I will merge, this is really a minor topic.

PR #1456, opened 2022-10-16
fix applyStyle calls location

rvilarl:
Make the combination openGorup/applyStyle closeGroup/restoreStyle in the code set


rvilarl:
This is a simple consitency exercise important for a PR to come that will finally allow to set the styles in SVGs taking into account the styles alreay set in the associated groups. I will merge it. No visual differences.

PR #1457, opened 2022-10-16
styles refactored to consider groups

### LINKED ISSUE #596: svgcontext: remove some attributes when possible
svgcontext generates some additional attributes.

please refer to the following uri:
~~https://jsfiddle.net/jc5tut2f/~~

EDIT by @ronyeh: the example for version 3.0.9 is now here: https://jsfiddle.net/etko73nj/

```xml
<?xml version="1.0" encoding="UTF-8"?>
<svg width="500" height="500">
   <path stroke-width="1" fill="none" stroke="#999999" stroke-dasharray="none"
     font-family="Arial" font-size="10pt" font-weight="normal" font-style="normal"
     d="M10 80L410 80" />
...
   <rect stroke-width="0.3" fill="black" stroke="black" stroke-dasharray="none"
    font-family="Arial" font-size="10pt" font-weight="normal" font-style="normal"
    x="10" y="79.5" width="1" height="41" />
...
   <path stroke-width="0.3" fill="black" stroke="none" stroke-dasharray="none"
    font-family="Arial" font-size="10pt" font-weight="normal" font-style="normal"
    x="410" y="79.5" width="1" height="41" d="..." />
...
   <text stroke-width="0.3" fill="black" stroke="none" stroke-dasharray="none"
    font-family="Arial" font-size="12pt" font-weight="normal" font-style="normal"
    x="79.90079999999999" y="70"
    width="1" height="41">Render this</text>
...
</svg>
```

* the following attributes can be removed?:

- [x] {path, rect}.font-{family, weight, style, size} : #628
- [x] path.{x, y, width, height} : #628
- [x] text.{width, height} : #628

### LINKED ISSUE #1451: Refactor & dynamic styling (step 4)
After prototyping a bit I believe that we should follow one of the following approaches:
1. allow to set styles only for groups
2. add an id to all the SVGElements associated to an Element

I would go for 1. it is less complex and would reduce significantly the size of the of the SVGs. 2. would also require to use suffixes on the IDs as one Element can have for example several paths.

@0xfe @AaronDavidNewman @ronyeh @mscuthbert @sschmidTU what do you think?

rvilarl:
fixes #1451
fixes #596

I'm verry proud of this PR, hope you like it. I have followed the approach:

> 3) Where possible, when a group is open, apply attributes to the group element instead of its children.

_Originally posted by @gristow in https://github.com/0xfe/vexflow/issues/596#issuecomment-386156535_

There is no visual differences and the size of the SVGs have been reduced by 5,2MB

### DIFF
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -6,11 +6,16 @@ import { Font, FontInfo, FontStyle, FontWeight } from './font';
 import { GroupAttributes, RenderContext, TextMeasure } from './rendercontext';
 import { normalizeAngle, prefix, RuntimeError } from './util';
 
-// eslint-disable-next-line
-export type Attributes = Record<string, any>;
+export type Attributes = {
+  [name: string]: string | number | undefined;
+  'font-family'?: string;
+  'font-size'?: string | number;
+  'font-style'?: string;
+  'font-weight'?: string | number;
+};
 
 /** For a particular element type (e.g., rect), we will not apply certain presentation attributes. */
-const ATTRIBUTES_TO_IGNORE: Record<string /* element type */, Attributes /* ignored attributes */> = {
+const ATTRIBUTES_TO_IGNORE: Record<string /* element type */, Record<string, boolean> /* ignored attributes */> = {
   path: {
     x: true,
     y: true,
@@ -83,10 +88,10 @@ class MeasureTextCache {
     }
 
     txt.textContent = text;
-    txt.setAttributeNS(null, 'font-family', attributes['font-family']);
-    txt.setAttributeNS(null, 'font-size', attributes['font-size']);
-    txt.setAttributeNS(null, 'font-style', attributes['font-style']);
-    txt.setAttributeNS(null, 'font-weight', attributes['font-weight']);
+    if (attributes['font-family']) txt.setAttributeNS(null, 'font-family', attributes['font-family']);
+    if (attributes['font-size']) txt.setAttributeNS(null, 'font-size', `${attributes['font-size']}`);
+    if (attributes['font-style']) txt.setAttributeNS(null, 'font-style', attributes['font-style']);
+    if (attributes['font-weight']) txt.setAttributeNS(null, 'font-weight', `${attributes['font-weight']}`);
     svg.appendChild(txt);
     const bbox = txt.getBBox();
     svg.removeChild(txt);
@@ -118,6 +123,8 @@ export class SVGContext extends RenderContext {
   parent: SVGGElement;
   // The stack of groups.
   groups: SVGGElement[];
+  // The stack of attributes associated with each group.
+  protected groupAttributes: Attributes[];
 
   backgroundFillStyle: string = 'white';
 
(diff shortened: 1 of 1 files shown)


rvilarl:
Below an example:

image

rvilarl:
@AaronDavidNewman @mscuthbert would you like to have a look?
Ready to review

src/svgcontext.ts
```diff
@@ -118,6 +123,8 @@ export class SVGContext extends RenderContext {
   parent: SVGGElement;
   // The stack of groups.
   groups: SVGGElement[];
+  // The stack of attributes associated with each group.
+  protected groupAttributes: Attributes[];
```
rvilarl, line 127: `groupAttributes` is a stack that contains the current attributes in each nested group. The active ones, this means, the ones in the group and the ones inhereted from the other groups in the stack.

src/svgcontext.ts
```diff
@@ -148,7 +155,8 @@ export class SVGContext extends RenderContext {
     };
 
     this.state = {
-      scale: { x: 1, y: 1 },
+      scaleX: 1,
+      scaleY: 1,
```
rvilarl, line 159: I have refactored those two in order to avoid a more complex definition of `Attributes`

src/svgcontext.ts
```diff
@@ -160,6 +168,10 @@ export class SVGContext extends RenderContext {
       ...defaultFontAttributes,
     };
 
+    this.groupAttributes = [];
+    this.applyAttributes(svg, this.attributes);
+    this.groupAttributes.push({ ...this.attributes });
+
```
rvilarl, line 174: I have used SVG to place the initial attributes of SVGContext, the ones defined in the constructor.

src/svgcontext.ts
```diff
@@ -342,7 +358,12 @@ export class SVGContext extends RenderContext {
       if (attrNamesToIgnore && attrNamesToIgnore[attrName]) {
         continue;
       }
-      element.setAttributeNS(null, attrName, attributes[attrName]);
+      if (
+        attributes[attrName] &&
+        (this.groupAttributes.length == 0 ||
+          attributes[attrName] != this.groupAttributes[this.groupAttributes.length - 1][attrName])
```
rvilarl, line 364: Only apply attribute if it is not already defined or it has a different value in the stack of groups.

rvilarl:
My comments above hopefully help to review the PR.

src/svgcontext.ts
mscuthbert: Looking good, but one thing to consider is whether you want to use a more React/Lit-Element style of:

```
this.groupAttributes = [...this.groupAttributes, ...this.attributes];
```

where groupAttributes is treated as immutable, and thus Proxy objects could set listeners for it to do something more complex when altered.  Just a thought for the future.

mscuthbert (approved):
What an important and useful PR (not just for saving space, but for ease of manipulating later).  Major kudos.

src/svgcontext.ts
rvilarl: I left it untouched to be consistent with `groups` and `state_stack`. I can make a PR to change the three if you want.

rvilarl:
@mscuthbert Thanks a lot!

Issue #1458, opened 2022-10-16
svg grouping


rvilarl:
All the prototype that I have done to get #1457 ready have highlighted some issues / questions:
- There are still elements with more than 3 digits precission. Should we make the rounding directly in SVGContext rather that in Glyph to make sure that all the elements get rounded? I think we should.
- Does it make sense to have a group that always contains only one SVGElement? I think that it would make more sense to get rid of the group and assign an ID to the element (path / rect / ...)
- Do we want make new classes for the elements that we want to color independently in order to be able to access them as an SVGElement (dynamic coloring)? For example ledger lines or lines on the staves as requested in #1433 I would go for the ledger lines with staff lines I am unsure.
- Are the elements rightly grouped? i.e.: Should StaveModifiers, Clef, and/or Time/KeySignature be grouped under Stave? I am unsure, this would for example make all this elements grey taking the default of the Stave.

@AaronDavidNewman @mscuthbert @sschmidTU what is your oppinion?

AaronDavidNewman:
> Should StaveModifiers, Clef, and/or Time/KeySignature be grouped under Stave? 

It is preferred.  For SVG at least, there is a performance benefit if the modifiers are under the parent element.  You can remove and replace a stave by just removing the stave group, for instance.

It looks they are that way already.  Tempo is an outlier, in that it is not part of a group in svg, even though it is rendered under the stave.  

It would be good to have tempo as part of a group, because you can remove/replace tempo element without having to redraw anything else (change time signature, for instance, usually requires re-rendering the notes).

rvilarl:
> > Should StaveModifiers, Clef, and/or Time/KeySignature be grouped under Stave?
> 
> It is preferred. For SVG at least, there is a performance benefit if the modifiers are under the parent element. You can remove and replace a stave by just removing the stave group, for instance.
> 
> It looks they are that way already. Tempo is an outlier, in that it is not part of a group in svg, even though it is rendered under the stave.
> 
> It would be good to have tempo as part of a group, because you can remove/replace tempo element without having to redraw anything else (change time signature, for instance, usually requires re-rendering the notes).

I agree that it makes more sense. However if we group the StaveModifiers under Stave the Modifiers also get the defaults from this group and the StaveBarlines end up grey, which I personally prefer, but also the clef and key/timemeassures, wich I doubt.

[...]

PR #1460, opened 2022-10-16
move rounding to SVGContext

rvilarl:
addesses #1458 1st bullet

>- There are still elements with more than 3 digits precission. Should we make the rounding directly in SVGContext rather that in Glyph to make sure that all the elements get rounded? I think we should.

A lot of not visible minor changes. The same problem as when we introduced the rounding in Glyph. I will put some diff images examples below.

Test SVGs reduded by 900 KB

### DIFF
--- a/src/glyph.ts
+++ b/src/glyph.ts
@@ -8,7 +8,6 @@ import { Font, FontGlyph } from './font';
 import { RenderContext } from './rendercontext';
 import { Stave } from './stave';
 import { Stem } from './stem';
-import { Tables } from './tables';
 import { Category } from './typeguard';
 import { defined, RuntimeError } from './util';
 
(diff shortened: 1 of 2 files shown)


rvilarl:
Rests Dotted Bravura
Formatter Annotations Bravura
Articulation Articulation___Vertical_Placement Bravura

rvilarl:
Ready to review

AaronDavidNewman:
Did you observe how this affected the 'path' elements in the DOM? For instance, ties and slurs should have 3 decimals, where they had more before.

rvilarl:
> Did you observe how this affected the 'path' elements in the DOM? For instance, ties and slurs should have 3 decimals, where they had more before.

Yes this is coming from #1457 where the grouping is done better but the precision is more than 3 digits

image

And this is the result of this PR (grouping improvements are still not merged)

image

AaronDavidNewman (approved):
works for me

rvilarl:
@AaronDavidNewman thanks a lot!

PR #1461, opened 2022-10-16
dynamic fillstyle

rvilarl:
This is finally the dynamic set of styles applied to fillStyle only. 
I have also changed the test **StaveNote › Interactive Mouseover StaveNote** to call `note.setStyle` directly.

No visual changes the modified test provides the same dynamic behaviour.

### DIFF
--- a/src/element.ts
+++ b/src/element.ts
@@ -141,12 +141,26 @@ export abstract class Element {
    */
   setStyle(style: ElementStyle | undefined): this {
     this.style = style;
+    const svg = this.getSVGElement();
+    if (svg) {
+      if (style?.fillStyle) {
+        svg.setAttribute('fill', style.fillStyle);
+        for (let i = 0; i < svg.children.length; i++) svg.children[i].setAttribute('fill', '');
+      }
+    }
     return this;
   }
 
   /** Set the element & associated children style used for rendering. */
   setGroupStyle(style: ElementStyle): this {
     this.style = style;
+    const svg = this.getSVGElement();
+    if (svg) {
+      if (style?.fillStyle) {
+        svg.setAttribute('fill', style.fillStyle);
+        for (let i = 0; i < svg.children.length; i++) svg.children[i].setAttribute('fill', '');
+      }
+    }
     this.children.forEach((child) => child.setGroupStyle(style));
     return this;
   }
--- a/src/element.ts
+++ b/src/element.ts
@@ -145,7 +145,9 @@ export abstract class Element {
     if (svg) {
       if (style?.fillStyle) {
         svg.setAttribute('fill', style.fillStyle);
-        for (let i = 0; i < svg.children.length; i++) svg.children[i].setAttribute('fill', '');
+      }
+      if (style?.strokeStyle) {
+        svg.setAttribute('stroke', style.strokeStyle);
       }
     }
     return this;
@@ -158,7 +160,9 @@ export abstract class Element {
     if (svg) {
       if (style?.fillStyle) {
         svg.setAttribute('fill', style.fillStyle);
-        for (let i = 0; i < svg.children.length; i++) svg.children[i].setAttribute('fill', '');
+      }
+      if (style?.strokeStyle) {
+        svg.setAttribute('stroke', style.strokeStyle);
       }
     }
     this.children.forEach((child) => child.setGroupStyle(style));
--- a/tests/stavenote_tests.ts
+++ b/tests/stavenote_tests.ts
@@ -402,14 +402,14 @@ function drawBasic(options: TestOptions, contextBuilder: ContextBuilder): void {
         item.addEventListener(
           'mouseover',
           () => {
-            note.setStyle({ fillStyle: 'green' });
+            note.setStyle({ fillStyle: 'green', strokeStyle: 'green' });
           },
           false
         );
         item.addEventListener(
           'mouseout',
           () => {
-            note.setStyle({ fillStyle: 'black' });
+            note.setStyle({ fillStyle: 'black', strokeStyle: 'black' });
           },
           false
         );
--- a/tests/stavenote_tests.ts
+++ b/tests/stavenote_tests.ts
@@ -392,12 +392,6 @@ function drawBasic(options: TestOptions, contextBuilder: ContextBuilder): void {
   ];
   expect(noteStructs.length * 2);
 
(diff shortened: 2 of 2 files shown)


rvilarl:
@ronyeh This is my last PR for version 4.1. I would like to get all the open PRs with label 4.1 in this version.
Ready to review

rvilarl:
@AaronDavidNewman @mscuthbert @sschmidTU I would like to get your feedbacck on this one.

tests/stavenote_tests.ts
```diff
@@ -392,21 +392,27 @@ function drawBasic(options: TestOptions, contextBuilder: ContextBuilder): void {
   ];
   expect(noteStructs.length * 2);
 
-  const colorDescendants = (parentItem: SVGElement, color: string) => () =>
-    parentItem.querySelectorAll('*').forEach((child) => {
-      child.setAttribute('fill', color);
-      child.setAttribute('stroke', color);
```
ronyeh, line 398: I notice the original code sets the 'stroke' as well. Do we want to support stroke style in your PR? Or do you think fill is sufficient?

src/element.ts
```diff
@@ -141,12 +141,26 @@ export abstract class Element {
    */
   setStyle(style: ElementStyle | undefined): this {
     this.style = style;
+    const svg = this.getSVGElement();
+    if (svg) {
+      if (style?.fillStyle) {
+        svg.setAttribute('fill', style.fillStyle);
```
ronyeh, line 147: If the attribute is called `fill` then why don't we use the same naming scheme? I think calling it `fill` is simpler than `fillStyle`.

So the resulting line of code would look like:

```
svg.setAttribute('fill', style.fill);
```

src/element.ts
```diff
@@ -141,12 +141,26 @@ export abstract class Element {
...
+        for (let i = 0; i < svg.children.length; i++) svg.children[i].setAttribute('fill', '');
```
ronyeh, line 148: Can we add braces { } here even though it's a single line of code? I think always including braces is a simpler/safer coding style, and is easier to read.

src/element.ts
```diff
@@ -141,12 +141,26 @@ export abstract class Element {
...
   }
 
   /** Set the element & associated children style used for rendering. */
   setGroupStyle(style: ElementStyle): this {
     this.style = style;
+    const svg = this.getSVGElement();
+    if (svg) {
+      if (style?.fillStyle) {
+        svg.setAttribute('fill', style.fillStyle);
+        for (let i = 0; i < svg.children.length; i++) svg.children[i].setAttribute('fill', '');
```
ronyeh, line 161: Same here. I think having extra braces around the single line of code is nicer, in case we want to add more lines of code someday. It also makes it easier for me to scan the code quickly (especially during code review).

src/element.ts
ronyeh: Ignore the above comment. I forgot that the SVG specification and the Canvas2D specification are different! I think in VexFlow, we are trying to adhere to the CanvasRenderingContext2D API as much as possible.

So here we are trying to support `CanvasRenderingContext2D.fillStyle`.

ronyeh (changes requested):
I find it slightly weird that element.ts's `setStyle()` now includes special code to handle SVG. Does this break anything if we are rendering to the canvas API instead?

Are we planning to support the dynamic fillstyle in canvas as well? I guess in canvas, we would have to re-draw the element, or trigger a redraw of the entire canvas whenever the fillStyle changes.

Anyways, it kind of bugs me that the special SVG code lives inside `element.ts`. Maybe this should be in the render context instead? So the code would look something like:

```
this.getContext().setFillStyle(....)
```

And then the canvascontext.ts and svgcontext.ts would each have their own implementations to support dynamic fillStyle updates. It's OK if the canvascontext.ts implementation does nothing for now. It could be a future PR.

tests/stavenote_tests.ts
rvilarl: I was thinking about agreeing on how to proceed with fill and then add others

src/element.ts
rvilarl: the for statement is no longer required after #1457, removed

rvilarl:
@ronyeh Thanks for your review. 

> I find it slightly weird that element.ts's `setStyle()` now includes special code to handle SVG. Does this break anything if we are rendering to the canvas API instead?

No, the elements do not have an associated SVGElement when canvas is used.
 
> Are we planning to support the dynamic fillstyle in canvas as well? I guess in canvas, we would have to re-draw the element, or trigger a redraw of the entire canvas whenever the fillStyle changes.

No canvas does not support dynamic styling. This is also, I believe, the reason why there is no canvas variant of the mouseover tests.

> Anyways, it kind of bugs me that the special SVG code lives inside `element.ts`. Maybe this should be in the render context instead? So the code would look something like:
> 
> ```
> this.getContext().setFillStyle(....)
> ```
> 

I would not do that, the context would then need the element to get the svgelement.

> And then the canvascontext.ts and svgcontext.ts would each have their own implementations to support dynamic fillStyle updates. It's OK if the canvascontext.ts implementation does nothing for now. It could be a future PR.

rvilarl:
@ronyeh @AaronDavidNewman @mscuthbert @sschmidTU @0xfe what is your opinion? Do you like this approach? I would like to get your oppinion in order to either merge or close and make a new alpha release.

mscuthbert:
> @ronyeh @AaronDavidNewman @mscuthbert @sschmidTU @0xfe what is your opinion? Do you like this approach? I would like to get your oppinion in order to either merge or close and make a new alpha release.

I really like the idea, but I also think that it's backwards incompatible enough that it might be worth making the dynamic part a boolean  (most cautious would be make the default false):

```
setStyle(style: ElementStyle | undefined, dynamic: boolean = true): this {
   ...
   if (svg && dynamic) ...
```

this way there is a way to get the older behavior if people want it.  For instance, if they are drawing the score in one place, then manipulating it to draw it in a second place.

But yeah, I can immediately see about 20 places this will simplify my code.

AaronDavidNewman:
I don't see any problems with it.  I am maybe not the best source since I don't use dynamic styles (vs. css) and if I did I would apply to svg element directly.

rvilarl:
There is no concensus here, let us leave it for now.

PR #1462, opened 2022-10-17
Make stavetempo use text formatter and not getBBox

AaronDavidNewman:
for performance reasons

### DIFF
--- a/src/stavetempo.ts
+++ b/src/stavetempo.ts
@@ -6,6 +6,7 @@ import { Glyph } from './glyph';
 import { Stave } from './stave';
 import { StaveModifier, StaveModifierPosition } from './stavemodifier';
 import { Tables } from './tables';
+import { TextFormatter } from './textformatter';
 import { Category } from './typeguard';
 
 export interface StaveTempoOptions {
@@ -75,11 +76,12 @@ export class StaveTempo extends StaveModifier {
     const y = stave.getYForTopText(1) + this.shift_y;
 
     ctx.save();
+    const textFormatter = TextFormatter.create(this.textFont);
 
     if (name) {
       ctx.setFont(this.textFont);
       ctx.fillText(name, x, y);
-      x += ctx.measureText(name).width;
+      x += textFormatter.getWidthForTextInPx(name);
     }
 
     if (duration && bpm) {
@@ -87,9 +89,9 @@ export class StaveTempo extends StaveModifier {
       ctx.setFont({ ...this.textFont, weight: 'normal', style: 'normal' });
 
       if (name) {
-        x += ctx.measureText(' ').width;
+        x += textFormatter.getWidthForTextInPx('(');
         ctx.fillText('(', x, y);
-        x += ctx.measureText('(').width;
+        x += textFormatter.getWidthForTextInPx(')');
       }
 
       const code = Tables.getGlyphProps(duration);
--- a/src/stavetempo.ts
+++ b/src/stavetempo.ts
@@ -86,12 +86,14 @@ export class StaveTempo extends StaveModifier {
 
     if (duration && bpm) {
       // Override the weight and style.
-      ctx.setFont({ ...this.textFont, weight: 'normal', style: 'normal' });
+      const noteTextFont = { ...this.textFont, weight: 'normal', style: 'normal' };
+      ctx.setFont(noteTextFont);
+      const noteTextFormatter = TextFormatter.create(noteTextFont);
 
       if (name) {
-        x += textFormatter.getWidthForTextInPx('(');
+        x += noteTextFormatter.getWidthForTextInPx('|');
         ctx.fillText('(', x, y);
-        x += textFormatter.getWidthForTextInPx(')');
+        x += noteTextFormatter.getWidthForTextInPx('(');
       }
 
       const code = Tables.getGlyphProps(duration);
@@ -112,11 +114,10 @@ export class StaveTempo extends StaveModifier {
         ctx.fillRect(x - scale, y_top, scale, stem_height);
 
         if (code.flag) {
-          Glyph.renderGlyph(ctx, x, y_top, options.glyph_font_scale, code.code_flag_upstem, {
+          const flagMetrics = Glyph.renderGlyph(ctx, x, y_top, options.glyph_font_scale, code.code_flag_upstem, {
             category: 'flag.staveTempo',
           });
-
-          if (!dots) x += 6 * scale;
+          x += (flagMetrics.width * Tables.NOTATION_FONT_SCALE) / flagMetrics.font.getData().resolution;
         }
       }
 
@@ -127,7 +128,6 @@ export class StaveTempo extends StaveModifier {
         ctx.arc(x, y + 2 * scale, 2 * scale, 0, Math.PI * 2, false);
         ctx.fill();
       }
-
       ctx.fillText(' = ' + bpm + (name ? ')' : ''), x + 3 * scale, y);
     }
 


AaronDavidNewman:
visual differences don't show anything but there are some in tempo text.  There is a little more space between tempo text and bpm.

old:
image

new:
image

src/stavetempo.ts
```diff
@@ -75,21 +76,22 @@ export class StaveTempo extends StaveModifier {
     const y = stave.getYForTopText(1) + this.shift_y;
 
     ctx.save();
+    const textFormatter = TextFormatter.create(this.textFont);
 
     if (name) {
       ctx.setFont(this.textFont);
       ctx.fillText(name, x, y);
-      x += ctx.measureText(name).width;
+      x += textFormatter.getWidthForTextInPx(name);
```
rvilarl, line 84: I have made the following change to check what is the difference:
```
      ctx.fillText(name, x, y);
      ctx.fillRect(x, y + 8, ctx.measureText(name).width, 2);
      ctx.fillRect(x, y + 13, textFormatter.getWidthForTextInPx(name), 2);
      x += textFormatter.getWidthForTextInPx(name);
```
and I get

image

@AaronDavidNewman I would say that `meassureText` provides the right `width`.

rvilarl:
I have been further investigating the topic and I believe that the bug is in the SVGContext variant of `ctx.measureText()` which is providing a width 0 to spaces located alone, at the beginning of a string and at the end of a string.
The follwing:
```
console.log('tempo');
console.log(ctx.measureText(' ').width);
console.log(ctx.measureText('A').width);
console.log(ctx.measureText('AA').width);
console.log(ctx.measureText('A ').width);
console.log(ctx.measureText(' A').width);
console.log(ctx.measureText('A A').width);
```
returns:

**(canvas)**
4.6640625
13.472946166992188
26.945892333984375
18.137008666992188
18.137008666992188
31.609954833984375
**(svg)**
**0 -- wrong**
14
27.472946166992188
**14 -- wrong**
**14 -- wrong**
30.077835083007812

rvilarl:
I also found two bugs in StaveTempo one related with the update of x and another related to the problem above. They can be resolved like that:
```diff
      // Draw stem and flags
      if (code.stem) {
        let stem_height = 30;

        if (code.beam_count) stem_height += 3 * (code.beam_count - 1);

        stem_height *= scale;

        const y_top = y - stem_height;
        ctx.fillRect(x - scale, y_top, scale, stem_height);

        if (code.flag) {
          Glyph.renderGlyph(ctx, x, y_top, options.glyph_font_scale, code.code_flag_upstem, {
            category: 'flag.staveTempo',
          });

-          if (!dots) x += 6 * scale;
+          x += code.getWidth() * scale;
        }
      }

      // Draw dot
      for (let i = 0; i < dots; i++) {
        x += 6 * scale;
        ctx.beginPath();
        ctx.arc(x, y + 2 * scale, 2 * scale, 0, Math.PI * 2, false);
        ctx.fill();
      }

-      ctx.fillText(' = ' + bpm + (name ? ')' : ''), x + 3 * scale, y);
+      x += ctx.measureText(' ').width;
+      ctx.fillText('= ' + bpm + (name ? ')' : ''), x + 3 * scale, y);
    }

    ctx.restore();
    return this;
  }
}
```

rvilarl:
@AaronDavidNewman please check if these resolve the problem with your score.

rvilarl:
I suggest to divide the topic:
1. resolve the bug
2. decide what we do in general with `measureText` and `TextFormatter`. Have you measure the processing time difference between the two?

AaronDavidNewman:
@rvilarl every one of these red triangles is a call to measureText/getBBox in svgcontext.  Every time it is called, the entire svg area is repainted or at least needs to be evaluated.  In Smoosic I've moved all my getBBox to after the score is rendered, and it made it about 10x faster.  But there are still some modifiers that use measureText.

image
image

rvilarl:
Perhaps the problem is that we are creating the element to calculate the width within the svg where we are drawing. Creating somewhere else might have a better performance. The problem I see is that we will not be able to be accurate programatically without using the SVGElement interface.

AaronDavidNewman:
It's the call to getBBox, not the rendering of the element.  This is a general principle of graphics programming and applies to html as well.  You can draw stuff all you want, and it's very fast, but once you go to measure something you've drawn, you need to block everything and compute the geometry.

https://developers.google.com/speed/docs/insights/browser-reflow

> we will not be able to be accurate programatically without using the SVGElement interface.

Why not, if we have metrics from the font?  None of the other positioning/formatting in vexflow has to render the glyphs to know how big they will be or where they should be positioned.  Text is just glyphs.  

For large blocks of text, kerning may factor into it but not for something like tempo and lyrics.  And you may need to measure if there are complicated transformations (like rotations of svg groups).  But all we're doing here is measuring the extent of the text, and that is completely dependent on the dimensions of the glyphs.

AaronDavidNewman:
@rvilarl I investigated this a little bit.  I'm not quite sure why you and I see different things with the same code, but I see why the text looks a little funky in some of the examples.
The main problem is that our default font is 'Times New Roman', but we don't have metrics for this in the system.  So it uses the fallback which is RobotoSlab which is considerably 'bigger'.  If I change the Bach example to use RobotSlab it looks like this with the code I just pushed:

image

So you can see, the spacing looks much better.  I also changed the flag width to use the actual flag width.  It shouldn't use `code.getWidth` because that was already used on line 103.

I did not want to change all the default fonts to be Roboto just for this bug.  I have the times metrics with Smoosic, but I wasn't sure about the legality of doing anything with this font to include it with VexFlow.

Here is the logic with dot, no dot, and flag + dot.

image

rvilarl:
Look at flow.html tests related with:
- Bach Demo
- Stave - Tempo Test

The current canvas and svg results look different in my computer:
- dots are not right located in both
- the separation to the equal is less in svg

image
image

image
image

rvilarl:
The tempo tests should be quite independent from the font. There is only the space and the equal.

AaronDavidNewman:
Are you running my branch?  I don't understand why we're seeing different things.  This is what the tempo tests look like for me in svg.  I will try to pull the vexflow tip + my change and see if it is different.

image

rvilarl:
I was running master plus your stavetempo file. Now your branch and I get the same. What browser do you use? I use chrome

image

rvilarl:
@AaronDavidNewman I understand the explanation, calculation based on RobotoSlab. However the calculation are still quite inaccurate and this is what I meant when I said: 
> we will not be able to be accurate programatically without using the SVGElement interface.

image

it is also still a bug that the SVGContext does not print the heading space. However canvas does

image

AaronDavidNewman:
> we will not be able to be accurate programatically without using the SVGElement interface.

Why not?  Browsers don't render text with an element, measure the element, and render again.  They calculate the extent based on the metrics in the font file and base the layout on that, and these are exactly the same as what we have.  

Also, keep in mind that measuring metrics of a 'space character' in svg (or html) might not give you what you expect, since space in xml isn't always preserved.  It's better to decide how much space you want between words and just change the x by that amount.

> the SVGContext does not print the heading space

I think this is the thing about space not being preserved.  If you use 'tspan' element, you should be able to turn that off with an attribute but I haven't tried it.

Here are the renderings on various platforms.  I think this is ready to merge.

Linux chrome:
image

Linux FF:
image

Chrome windows:
image

FF Windows:
image

Safari Macbook (M1):
image

Chrome macbook:
image

rvilarl:
> Here are the renderings on various platforms. I think this is ready to merge.

Ok

PR #1468, opened 2022-11-06
use TextFormatter for annotation and stavesection

AaronDavidNewman:
This change has 2 parts:  It adds a generic serif and sans-serif text metrics to the text formatter for the built-in browser fonts (Arial and Times).  Also, it uses the text formatter for annotation and stavesection (rehearsal numbers).  I also added text formatter tests similar to the ones @rvilarl  used, both for string width and box around the text to test the bounding box accuracy.

In the next set of these, I hope to add the kerning information to the tool that generates the text metrics, this should fix any remaining issues of estimation vs. measuring.

There are about 450 changes in the visual regression tests, most are very small.  I'll include some examples.

diff:
Vibrato Vibrato_with_Bend Bravura

reference:
Vibrato Vibrato_with_Bend Bravura_reference

new:
Vibrato Vibrato_with_Bend Bravura_current

diff:
pptr-Annotation Placement Leland svg

reference:
pptr-Annotation Placement Leland svg_reference

new:
pptr-Annotation Placement Leland svg_current

These didn't show up in the regression tests but there are small differences in the section numbers.

reference:
image

new:
image

New text formatter tests, green line is estimation.  Note that I didn't see the measure text function that give the min/max Y, just the box.  This function could probably be rewritten in such a way that we could assume e.g. y=-5 meant the y was 5 below the baseline.  If it is, the measured should be the same accuracy as the estimate.

green line is estimate, blue is measured.
image

In SVG, rect does not honor 'stroke' - is this is known issue?  It works fine in Canvas.

SVG:
image

Canvas:
image


AaronDavidNewman:
I just noticed that I left some code in font_tests that isn't used.  I'll remove it.

src/annotation.ts
```diff
@@ -86,7 +86,7 @@ export class Annotation extends Modifier {
       const annotation = annotations[i];
       const textFormatter = TextFormatter.create(annotation.textFont);
       // Text height is expressed in fractional stave spaces.
-      const textLines = (5 + textFormatter.maxHeight) / Tables.STAVE_LINE_DISTANCE;
+      const textLines = (2 + textFormatter.getYForStringInPx(annotation.text).height) / Tables.STAVE_LINE_DISTANCE;
```
rvilarl, line 89: Why did we move from max height of the font to height of the actual text? I think it is a good change for smufl fonts where some characters are very high (ie clefs) compared to actual text.

src/fonts/sans_bold_text_metrics.ts
```diff
@@ -0,0 +1,862 @@
+export const SansFontBold = {
+  glyphs: {
+    '0': {
+      x_min: 86,
+      x_max: 1038,
+      y_min: -25,
+      y_max: 1472,
+      ha: 1497,
```
rvilarl, line 8: These values can be obtained. as I proposed, in run time with `canvas.measureText` which returns:
```
/** The dimensions of a piece of text in the canvas, as created by the CanvasRenderingContext2D.measureText() method. */
interface TextMetrics {
    /** Returns the measurement described below. */
    readonly actualBoundingBoxAscent: number;
    /** Returns the measurement described below. */
    readonly actualBoundingBoxDescent: number;
    /** Returns the measurement described below. */
    readonly actualBoundingBoxLeft: number;
    /** Returns the measurement described below. */
    readonly actualBoundingBoxRight: number;
    /** Returns the measurement described below. */
    readonly fontBoundingBoxAscent: number;
    /** Returns the measurement described below. */
    readonly fontBoundingBoxDescent: number;
    /** Returns the measurement described below. */
    readonly width: number;
}
```
what is the increase in size of the library witth these metrics?

src/stavesection.ts
```diff
@@ -51,32 +52,32 @@ export class StaveSection extends StaveModifier {
...
+    const height = textHeight + 2 * padding; // add top & bottom padding
 
     //  Seems to be a good default y
     const y = stave.getYForTopText(1.5) + this.shift_y;
     const x = this.x + shift_x;
     if (this.drawRect) {
       ctx.beginPath();
-      ctx.setLineWidth(rectWidth);
-      ctx.rect(x, y + textMeasurements.y - paddingY, width, height);
+      ctx.rect(x, y - height + headroom, width, height);
```
rvilarl, line 77: line width is no longer  set to 2

src/stavesection.ts
```diff
@@ -51,32 +52,32 @@ export class StaveSection extends StaveModifier {
...
       ctx.stroke();
     }
-    ctx.fillText(this.section, x + paddingX, y);
+    ctx.fillText(this.section, x + padding, y - padding);
```
rvilarl, line 80: y padding has been removed

rvilarl (changes requested):

src/annotation.ts
AaronDavidNewman: My plan is to get rid of `maxSizeGlyph` as a member of the text metrics, and get the max height value directly from the font data.  Picking some arbitrary character seemed...arbitrary.  But when I used this value, it was higher than what we had been using, and the annotations took to much space.  And also people had complained about the much extra whitespace.  Using the size of the actual text tightens it up a little bit.

src/fonts/sans_bold_text_metrics.ts
AaronDavidNewman: It is about 15K for the fonts we have now (the 5 numbers per glyph).  So you could think of it like a permanent/persistent cache for the canvas method.

src/stavesection.ts
AaronDavidNewman: Since the width/height is coming from the actual text and not the max glyph, you need to add padding when placing the text for both x and y.  The width and height from the rectangle already account for the padding.  'headroom' is added for text that drops below the baseline, since 'y' in the text element always refers to the baseline.  So the character 'C' for instance usually drops a bit below the baseline.

src/annotation.ts
AaronDavidNewman: Maybe for annotation we could add padding to options?  That would let you decide how much space goes between annotations, like for multiple verses.  Annotations above/below usually have to share space with articulations/ornaments so we want to use as little space as possible by default.

src/stavesection.ts
rvilarl: I mean that the call to setLineWidth is no longer there, is that on purpose?

src/fonts/sans_bold_text_metrics.ts
rvilarl: OK

src/stavesection.ts
rvilarl: But you are using 
`const height = textHeight + 2 * padding;`
will not all the padding be left above?

AaronDavidNewman: The padding * 2 is on the rectangle, because the padding goes on top and bottom, left and right.  But when placing the text, you only consider padding * 1 since text element always places in the lower left.

Visually, there is more space on the top of the boxes because that's how the fonts were designed.  Whether you are measuring or getting the height from the font metrics, the max y tends to return higher than the highest point on the glyph.

src/stavesection.ts
AaronDavidNewman: Oh...right, because it's called on line 61.  This call was redundant.

PR #1471, opened 2022-11-15
issue 1469 - get top Y for top annotation

AaronDavidNewman:
Small change, when annotation placement is 'top', get the final entry in the `note.Ys()`, since that will have the (numerically lowest, screen highest) Y value.  Previously, it was always getting entry [0] which is the numerically highest, staff-lowest pitch.

old:
image

new test case:
image

I am upgrading my Ubuntu VM right now so I can't do visual diff.  But this will only affect cases with multiple pitch keys and stem up, and we didn't have any test cases for that.

### DIFF
--- a/src/annotation.ts
+++ b/src/annotation.ts
@@ -284,7 +284,8 @@ export class Annotation extends Modifier {
       const yb = stave.getYForBottomText(this.text_line);
       y = yt + (yb - yt) / 2 + text_height / 2;
     } else if (this.verticalJustification === AnnotationVerticalJustify.TOP) {
-      y = note.getYs()[0] - (this.text_line + 1) * Tables.STAVE_LINE_DISTANCE;
+      const yAr = note.getYs();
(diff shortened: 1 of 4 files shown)


rvilarl:
LGTM

rvilarl:
@AaronDavidNewman I checked the output after merging and this has a bug. Two examples:

Annotation Harmonics Bravura
Annotation Harmonics Bravura_current
Annotation Harmonics Bravura_reference
Annotation Simple_Annotation Bravura
Annotation Simple_Annotation Bravura_current
Annotation Simple_Annotation Bravura_reference

AaronDavidNewman:
Huh...maybe Ys for tab clef are not sorted?  If so, @StevenSchuerstedt had it correct that we should just reduce to the smallest value.

rvilarl:
Should I revert the changes in annotation.ts & chordsymbol.ts?

AaronDavidNewman:
No, I will push another PR shortly.  I am running the visual regression tests now.

PR #1472, opened 2022-11-15
Rename/metrics

rvilarl:
@AaronDavidNewman I just realised that the new metrics were .js files. Fixed.


AaronDavidNewman:
Ah...thanks.  I'll fix that next time I generate text fonts.

PR #1478, opened 2022-11-27
refactor any in GlyphProps

rvilarl:
Time to try that again.
continued from #1420

### DIFF
--- a/src/annotation.ts
+++ b/src/annotation.ts
@@ -90,7 +90,7 @@ export class Annotation extends Modifier {
       let verticalSpaceNeeded = textLines;
 
       const note = annotation.checkAttachedNote();
-      const glyphWidth = note.getGlyph().getWidth();
+      const glyphWidth = note.getGlyphProps().getWidth();
       // Get the text width from the font metrics.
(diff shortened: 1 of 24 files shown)


rvilarl:
Ready to review, I believe that this approach is more clear. Look into #1420 for previous discussions.

mscuthbert:
Could `getGlyph()` remain as a deprecated synonym for `getGlyphProps()` until v5?  It seems fine to improve and refine behavior between major releases but removing a major public interface mid major version where there is no cross-version unified behavior is unusual.

I know it's an internal interface, but   .clef.clef.X looks like it'd be prone to errors in the future.

Everything else looks really good to me.

rvilarl:
> Could `getGlyph()` remain as a deprecated synonym for `getGlyphProps()` until v5? It seems fine to improve and refine behavior between major releases but removing a major public interface mid major version where there is no cross-version unified behavior is unusual.

Done
 
> I know it's an internal interface, but .clef.clef.X looks like it'd be prone to errors in the future.

Agree but I beleive that the problem is rather in Clef having a clef member.

> Everything else looks really good to me.

@mscuthbert Thanks for the review!

PR #1479, opened 2022-11-27
Usage of FontFace rather than paths

### LINKED ISSUE #1127: Improvements to Glyph Rendering in SVGContext / CanvasContext

### LINKED ISSUE #1459: Measuring/estimating text in VexFlow quickly and accurately

rvilarl:
fixes #1127
fixes #1459

Personally I see two main benefits:
- the library is smaller.
- the produced SVG are also smaller.

```
-rw-r--r--. 1 rodrigo rodrigo 1911241 nov 27 22:06 reference/cjs/vexflow-debug.js
-rw-r--r--. 1 rodrigo rodrigo 1380174 nov 27 22:15 build/cjs/vexflow-debug.js
```
StaveNote with two noteheads with one accidental each.
```svg
<g class="vf-stavenote" id="vf-auto1300">
 <g class="vf-notehead" id="vf-auto1348" pointer-events="bounding-box">
  <text stroke="none" x="258.331" y="120"></text>
  <text stroke="none" x="232.347" y="120"></text>
 </g>
 <g class="vf-notehead" id="vf-auto1349" pointer-events="bounding-box">
  <text stroke="none" x="258.331" y="110"></text>
  <text stroke="none" x="245.339" y="110"></text>
 </g>
</g>
```


ronyeh:
As this is a big PR, it'll take time to properly review. Thanks for looking into this though!

Can you write up some extra info in this thread to help us with this review? What are the main tradeoffs?

This handles all of our font needs, whether we are using it to render chord symbols or music engraving symbols??

rvilarl:
> As this is a big PR, it'll take time to properly review. Thanks for looking into this though!
> 
> Can you write up some extra info in this thread to help us with this review? What are the main tradeoffs?

The changes are simple and the **review will be much quicker than what you anticipate** :)

- Concept: `xxx_glyphs.ts` files containing no path ('o' member) contains now a 'unicode' value. These unicodes will be printed with `FillText`.
- `Font` has two new methods to measure text: `textMetrics` and `measureText`. Those replace `TextFormatter`. This is related to most of the changes. This method is quick enough and a text cache is no longer required.
- Images generation with `jsdom`, I only managed to get the fonts loaded by node using `RegisterFont`.
 
> This handles all of our font needs, whether we are using it to render chord symbols or music engraving symbols??

Yes both, and also, to write text. The SMuFL OTF files include  characters for many languages (hindi,...)

mscuthbert:
I don't ever manipulate the points of the path after its drawn, so that won't be a breaking change to me, but I do tend to change the fill-color (and opacity, etc.) of the SVG.  Is that done with the same interface in this PR?  It looks like an incredible  space savings (especially since I can then load one copy of Bravura [or Gootville] for the whole site instead of two, but I wonder if it's backwards compatible enough to go into v4 or if it's more of a v5 thing?

rvilarl:
Draft until #1476 is resolved, I want to integrate this PR in this one.

src/annotation.ts
```diff
@@ -84,15 +83,15 @@ export class Annotation extends Modifier {
     let maxRightGlyphWidth = 0;
     for (let i = 0; i < annotations.length; ++i) {
       const annotation = annotations[i];
-      const textFormatter = TextFormatter.create(annotation.textFont);
+      const measure = Font.measureText(annotation.text, annotation.textFont!);
```
mscuthbert, line 86: perhaps refactor from `measure` to `measurement`?  `measure` has a very specific implication in music applications.

PR #1477, opened 2022-11-27
custom noteheads refactored

rvilarl:
fixes #1415 
location is now calculated on run time and the associated metrics are no longer required.

PR #1481, opened 2022-12-06
Fix/clef code and tempo notes

rvilarl:
Resolves bugs found in #1476 

> Looking in detail at all the tests, I found two bugs:
> 
> * clef code used for percussion was wrong, now fixed.
> * stavetempo had not been aligned with [custom noteheads refactored #1477](https://github.com/0xfe/vexflow/pull/1477), now fixed

I have also removed the threshold in the visual regression. The threshold was not revealing the problem with StaveTempo.

### DIFF
--- a/tools/visual_regression.sh
+++ b/tools/visual_regression.sh
@@ -147,32 +147,19 @@ function diff_image() {
   cp $fileA $diff-a.png
   cp $fileB $diff-b.png
 
-  # Calculate the difference metric and store the composite diff image.
-  local hash=`compare -metric PHASH -highlight-color '#ff000050' $diff-b.png $diff-a.png $diff-diff.png 2>&1`
-
-  # Remove scientific notation
-  local hash6=`printf "%.6f" $hash`
-  local THRESHOLD6=`printf "%.6f" $THRESHOLD`
-
-  local isGT=`echo "$hash6 > $THRESHOLD6" | bc -l`
-  if [ "$isGT" == "1" ]
-  then
-    # Add the result to results.text
-    echo $name $hash >$diff.fail
-    # Threshold exceeded, save the diff and the original, current
-    cp $diff-diff.png $DIFF/$name.png
-    cp $diff-a.png $DIFF/$name'_'$ANAME.png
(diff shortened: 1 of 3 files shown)


rvilarl:
Percussion clef code fixed.
Reference
Clef Clef_End_Test Bravura_reference
Current
Clef Clef_End_Test Bravura_current
Stave Tempo notes fixed
Reference
Stave Tempo_Test Bravura_reference
Current
Stave Tempo_Test Bravura_current

rvilarl:
Ready to review.

tools/visual_regression.sh
```diff
@@ -147,32 +147,19 @@ function diff_image() {
...
-    echo "  Image diff stored in $DIFF/$name.png"
-    # $VIEWER "$diff-diff.png" "$diff-a.png" "$diff-b.png"
-    # echo 'Hit return to process next image...'
-    # read
-  else
-    echo $name $hash >$diff.pass
-  fi
+  # Store the composite diff image.
+  # Add the result to results.text
+  echo $name $hash >$diff.fail
```
ronyeh, line 152: You removed the hash variable on line 151, but here we are echo-ing the $hash.

ronyeh:
By "removing the threshold" do you mean that now, if anything is different (even 1 pixel) we will render out the diffs?

I am in favor of this change. It's good to see all changes, even if they are minimal, to be aware of what a PR affects.

ronyeh:
I approve this PR, assuming you address my comment on `visual_regression.sh`. Make sure the script is bug free. :-)

rvilarl:
> By "removing the threshold" do you mean that now, if anything is different (even 1 pixel) we will render out the diffs?
> 
> I am in favor of this change. It's good to see all changes, even if they are minimal, to be aware of what a PR affects.

This is right and I agree with you it is better to see everything that changed. I missed the StaveTempo problem above because the difference was below the threshold.

tools/visual_regression.sh
rvilarl: Addressed and checked.

rvilarl:
@ronyeh thanks for the review, merging.

PR #1484, opened 2022-12-10
Easily compare images in the current/ and reference/ folders

ronyeh:
I built a tool to help us compare the output of two VexFlow versions.

This can help us understand the changes we are introducing with each PR.

I still need to test a bit more, but try it out and let me know what you think!

Load `vexflow/tools/compare/index.html` in your local browser. Then drag a folder with the `current` and `reference` images onto the page. You can then type some search terms to narrow down the list to the images you want to inspect (e.g., "gonv bach" will find the Bach demo tests in the Gonville font).


rvilarl:
Looks very good!
- Overlapping images are still not shown correctly, right?
- It would be useful to have a label indicating which are the current and the reference ones.

 Thanks @ronyeh

ronyeh:
Open `vexflowtools/compare/index.html` and load in the `images/` folder.

There are 4 different view modes, accessible via number keys:

1. show current and reference side by side
2. alternate between current and reference images (so you can flip back and forth).
3. stack/overlay the two images. The top image is translucent. Use left/right arrow keys to flip back and forth.
4. image diff

There probably are small bugs, but this is good enough to help us review Gonville SMuFL.

Once we review the changes, I am comfortable merging the new Gonville as long as we agree to continue to improve it so that it more closely matches:
A) the example on the Gonville homepage: https://www.chiark.greenend.org.uk/~sgtatham/gonville/
B) the 4.1.0 release

If the visual changes from 4.1.0 are noticeable enough, and we can't (or won't) fix, then the next official release will be 5.0.

rvilarl (approved):

ronyeh:
Thanks @rvilarl for reviewing & merging. Next we can tackle #1480 ... I'll try to use this tool to see if there are any visual diffs that we can "fix" or minimize.

ronyeh:
By the way @rvilarl I updated docs/api. Please pull down these changes before any future PR merges. :-)

rvilarl:
@ronyeh the empty README was on purpose?

ronyeh:
:-) I am writing it right now. I will submit it in a later PR.

PR #1492, opened 2022-12-16
Gootville support

### LINKED ISSUE #1475: Gootville SMuFL

rvilarl:
fix #1475

Gootville support added, if we do not want to increase the core, I would still added as a separated font.


rvilarl:
@AaronDavidNewman @mscuthbert @sschmidTU what do you think about gootville?

mscuthbert:
I think it's a good idea, but I'm still trying to get my own systems on 4.0 w/ small payloads to pay this much attention to the future loads.  (I'll probably use Bravura once I can find a way not to load the font twice, since it's already in my payload)

mscuthbert:
I do think though: support Gonville or Gootville, but not both -- they're too similar for someone to really choose between the two.  And my preference is to follow Dorico and call Gootville Gonville (unless the Gonville or Gootville authors object) -- Gonville as SMuFL hasn't been around long enough for people to get totally attached to it; Gootville should be considered the SMuFL version of Gonville w/ a few tweaks.

rvilarl:
@mscuthbert my proposal in #1475 was to move to Gootville but I understood that there were quite a few concerns. This is why I smuflized Gonville in https://github.com/rvilarl/gonville I do not know how to proceed now. I have always liked Gootville but there are differences compared to Gootville, see #1475

ronyeh:
Currently, VexFlow 4 builds to multiple targets. Each SMuFL font would be in its own standalone vexflow library (e.g., vexflow-gonville.js and vexflow-gootville.js).

One possibility is to treat VexFlow as a testing sandbox for all the SMuFL fonts. Then `vexflow.js` would be the monolithic library that includes all SMuFL fonts. We would support creating individual targets that includes just one (or a few) fonts.

MuseScore also calls Gootville "Gonville". Although I'm not sure anyone has asked Simon Tatham what he thinks about Gootville vs Gonville. Rodrigo remapped the Gonville glyphs to the SMuFL code points, so perhaps our version of Gonville is "more true" to the original designer's vision?

In any case, maybe "**handle every SMuFL font known to humanity**" is a good approach for VexFlow?

We can bump the major revision to 5.0 so we can handle this shift in philosophy. The build script could accept a flag for which fonts you want to keep. Of course, we should only include/redistribute fonts that are licensed under the SIL Open Font License. If the need arises, we could also make sure VexFlow builds fine with commercial SMuFL fonts (although we would not redistribute those fonts, but instead only provide helper methods to load and draw the glyphs from the commercial OTF).

If we move toward a FontFace API approach, and stop including the glyph paths in our vexflow library, this might solve @mscuthbert 's problem of loading Bravura twice. Ideally, we should just use FontFace to load the Bravura.otf once, and then vexflow is a set of routines to help with formatting and displaying the glyphs from the OTF file.

VexFlow 5
* All SIL Open SMuFL fonts are welcome
* Use FontFace instead of redistributing the paths extracted from the OTFs

What do you think?

tests/offscreencanvas_tests.ts
```diff
@@ -29,9 +29,7 @@ function simpleTest(): void {
   // eslint-disable-next-line
   // @ts-ignore
   const offscreenCanvas = new OffscreenCanvas(550, 200);
-  // eslint-disable-next-line
-  // @ts-ignore
-  const offscreenCtx: OffscreenCanvasRenderingContext2D = offscreenCanvas.getContext('2d');
```
ronyeh, line 34: Why were these removed? 

I think we had some build errors earlier. That's why I added a @ts-ignore here.

Removing the type name results in a build error in some instances.

tests/vexflow_test_helpers.ts
```diff
@@ -198,6 +198,7 @@ export class VexFlowTests {
   static FONT_STACKS: Record<string, string[]> = {
     Bravura: ['Bravura', 'Gonville', 'Custom'],
     Gonville: ['Gonville', 'Bravura', 'Custom'],
+    Gootville: ['Gootville', 'Bravura', 'Custom'],
```
ronyeh, line 201: How does adding Gootville impact our test page's loading times? 

What about grunt test and `grunt generate:current`? Do they take much more time to run when we include another font?

I'm wondering if someday, we have 10 or 20 SMuFL fonts.... will our testing process become too cumbersome?

ronyeh (changes requested):
The PR looks good to me.

However, we should decide and agree on VexFlow's overall strategy here. Is the long term plan to include all SIL Open SMuFL fonts? Are we bumping the major revision to 5.0 to support this shift? 

We already added Leland.... I don't see why one SMuFL font should be preferred over another. We might as well include them all in VexFlow and then let the developer decide which one(s) they want to include in their final vexflow distribution!

tests/vexflow_test_helpers.ts
rvilarl: `time npm run test:reference` 
**without Gootville**
```
real    1m7,523s
user    2m27,025s
sys     0m20,205s
```
**with Gootville**
```
real    1m16,106s
user    2m33,854s
sys     0m24,427s
```

tests/offscreencanvas_tests.ts
rvilarl: I do not remember how I got the complain. I have changed it now to cast on its usage to avoid any problem
```
const ctx = new CanvasContext(offscreenCtx as OffscreenCanvasRenderingContext2D);
```

rvilarl:
@ronyeh can you have a look please? the files required to generate the gootville only files were missing. It does generate now the files in esm but not the ones in cjs.

rvilarl:
> VexFlow 5
> 
> * All SIL Open SMuFL fonts are welcome
> * Use FontFace instead of redistributing the paths extracted from the OTFs
> 
> What do you think?

I agree, Ibelieve that this is the approach that we should follow. @AaronDavidNewman @mscuthbert @sschmidTU  what do you think?

sschmidTU:
@rvilarl answering the ping:
I'm not sure what exactly `Use FontFace instead of redistributing the paths extracted from the OTFs` means, but if it means that vexflow can load the font from an OTF reference/file, that sounds fine for me, if it doesn't complicate the build/usage for the developer using OSMD too much. (will a developer using OSMD have to provide the OTF file themselves? Or can we 'bake it in' for OSMD?)

In general, I would ideally imagine having a vexflow release (or if not possible, then a way to build) that has only one selected font, or possibly a set of fonts (though for OSMD we'd probably use a gonville-only build, or possibly offer some other fonts in a separate build, if that increases build size).

rvilarl:
> @rvilarl answering the ping: I'm not sure what exactly `Use FontFace instead of redistributing the paths extracted from the OTFs` means, but if it means that vexflow can load the font from an OTF reference/file, that sounds fine for me, if it doesn't complicate the build/usage for the developer using OSMD too much. (will a developer using OSMD have to provide the OTF file themselves? Or can we 'bake it in' for OSMD?)

yes, it means reading the OTF directly rather than having the path embeded in VexFlow, this reduces a lot the size of VexFlow and also the size of the rendered HTMLs, see first comment in #1479
The aim is to bake it in.

ronyeh:
> the files required to generate the gootville only files were missing. It does generate now the files in esm but not the ones in cjs.

You need to also update the Gruntfile.js.

Search for VEX_BRAVURA for vexflow-bravura and follow the pattern.

ronyeh:
@rvilarl let me know when it's ready for testing. I'll be mostly away from GitHub activities until Jan 5th.

I am okay with adding as many SMuFL fonts as we can find, as long as it's easy for developers to pick and choose which font(s) they want in their final distribution.

Since this is a slight shift in our philosophy, let's call this next release VexFlow 5.

A major revision bump will also give us a chance to clean up some APIs, if desired....

rvilarl:
@ronyeh ready for review.

mscuthbert:
My feeling is really that out of box Vexflow should load in the JS one font( or two--i.e., a handwritten alternative--at max). The loadFont can be set up to access several more in the default installation (npm or `<script>`) -- Leland is already similar enough to Bravura/Gonville, that I don't know what is gained by having it packaged by default.

When we look at https://elbsound.studio/music-font-comparison.php -- the number of SMuFL-compliant music fonts is really becoming staggering.  Easy loading of another favorite (probably by reading in the associated json files as well) seems to me to be the only long-term sustainable way of keeping the compiled code small and the test suite managable.

PR #1495, opened 2022-12-17
GlyphArticulation support

### LINKED ISSUE #1494: Other articulations
The MusicXML format defines other articulations 
https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/other-articulation/ 
where any Glyph can be used.

I will submit a PR defining a new class `GlyphArticulation` to provide this functionality.

rvilarl:
fixes #1494 

ready to review


rvilarl:
I have added the glyph fermataVeryShort to better show the feature.
Bravura
GlyphArticulation GlyphArticulation___Vertical_Placement Gonville
Gonville
GlyphArticulation GlyphArticulation___Vertical_Placement Bravura

This reveals a bug in Gonville, the inverted double angled fermata must be rotated.

rvilarl:
@mscuthbert @ronyeh @rvilarl @sschmidTU any feedback? do you find it useful? I do.

ronyeh:
Questions:

1) should this just be an overload of the Articulation constructor? What if
you add optional params to the Articulation constructor?

2) Or, can we add a static method to the Articulation class? Or modify the
Factory.Articulation(...) builder method to support the glyph code param?


The feature seems useful. I'm just wondering whether we need the subclass.


3) If you really think we need a separate subclass for this, should we call
it OtherArticulation to match the musicxml spec?

AaronDavidNewman:
lol that is a pretty obscure articulation.  Is it idiomatic to a particular instrument?  I've never seen that thing.

ditto the comment about articulations, can't we just have it in articulations.ts?  You can still have your own test file for it, if you don't want to mess with existing visual diffs.

ronyeh:
It won't mess with the visual diffs if we just add a new test case in the
articulations test file. It should create separate images.

rvilarl:
I was thinking about the two options. I decided to do a new class so as to leave Articulation.ts untouched and to follow the approach of Note with GlyphNote.

AaronDavidNewman:
Can't you just add a new code to `Tables.articulations`?  I don't think you have to change the articulation class at all.

`GlyphArticulations` isn't a good name for a class because 1) all articulations are glyphs, and 2) GlyphArticulation doesn't do anything special with 'glyph', the way GlyphNote does.

It also seems wrong to add a new class that inherits, but doesn't add any logic to the original class.

mscuthbert:
Are the labels reversed?  It looks like the bug is in Bravura.  https://w3c.github.io/smufl/latest/tables/holds-and-pauses.html -- but I've only ever seen short -- I've never seen very short (if I wanted that, I'd use _ten._ )

rvilarl:
> Can't you just add a new code to `Tables.articulations`? I don't think you have to change the articulation class at all.

For the very short fermata this is possible, but I wanted to provide a way to comply with the MusicXML spec. In Vexflow there are also today elements that are considered articulations in MusicXML but are implemented as TextNote. i.e.: breath-mark and caesura.
 
> `GlyphArticulations` isn't a good name for a class because 1) all articulations are glyphs, and 2) GlyphArticulation doesn't do anything special with 'glyph', the way GlyphNote does.

That is fine. I can also add the minor extra logic to Articulations.

> It also seems wrong to add a new class that inherits, but doesn't add any logic to the original class.

I would say it adds minor extra logic. But as said I can place the logic in Articulations.

rvilarl:
> Are the labels reversed? It looks like the bug is in Bravura. https://w3c.github.io/smufl/latest/tables/holds-and-pauses.html -- but I've only ever seen short -- I've never seen very short (if I wanted that, I'd use _ten._ )

**Gonville**
GlyphArticulation GlyphArticulation___Vertical_Placement Gonville
 
This is the one that is wrong both very short fermatas are oriented in the same direction.

rvilarl:
@ronyeh could you merge this one if ok and create a new beta?

ronyeh (approved):

PR #1501, opened 2022-12-24
fix scale calculation

### LINKED ISSUE #1126: SVGContext's scale(x, y) works differently from CanvasContext's scale(x,y)
While migrating tests, I encountered a weird bit of code in NoteHeadTests:

https://github.com/0xfe/vexflow/blob/b37ddd7a228c69b165c96f4431598b703585fc37/tests/notehead_tests.ts#L34-L62

The setupContext() scales the context to 0.9, 0.9. The basic() test case applies another scale of 2.0, 2.0. (It also ignores the stave created by setupContext and creates its own stave, but that's another story.)

The scaling transformation should be multiplied against the existing scale:
https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/scale

So 0.9 * 2.0 => 1.8 final scale.

`flow.html` shows that the Canvas renderer works as expected, but the SVG renderer sets the final scale to 2.0, 2.0.

The fix might be straightforward. In SVGContext's scale method, we have:
`this.state.scale = { x, y };`

We probably just need to multiply it into the existing scale.

This bug has existed for a while, as it repros on vexflow.com: https://www.vexflow.com/tests/?module=NoteHead

You can verify the bug by doing something like this:
```
  ctx.scale(2.0, 2.0);
  ctx.scale(0.5, 0.5);
  ctx.scale(2.0, 2.0);
  ctx.scale(0.5, 0.5);
```

The final output will look like:

S 2021-09-01 at 10 25 22

rvilarl:
fix #1126 
@ronyeh do these changes address the issue? There are no visual differences.

### DIFF
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -13,6 +13,8 @@ export type Attributes = {
   'font-size'?: string | number;
   'font-style'?: string;
   'font-weight'?: string | number;
+  scaleX?: number;
+  scaleY?: number;
 };
 
 /** For a particular element type (e.g., rect), we will not apply certain presentation attributes. */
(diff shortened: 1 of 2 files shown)


mscuthbert:
It looks good to me -- if there are no visual tests that change, can we add the test from #1126 to the suite so it doesn't break in the future?

rvilarl:
Test modified as prepared by @ronyeh, no visual differences.

Merging.

PR #1511, opened 2022-12-31
Fix/1504 consider actual clef width

### LINKED ISSUE #1504: Too much space after percussion clef
In the default font set on Vexflow 4.1, there is much too much space after the percussion clef and before the key/time signature compared to 1.2.9, looking a bit like something is meant to go there and is missing:

Screenshot 2022-12-27 at 14 17 09

Just doing a lot of visual regression testing -- 95% are improvements or at least a matter of taste.  Will be flagging the things that I think are a step backwards.

rvilarl:
fix #1504

Vissual differences on al tests including clefs

### DIFF
--- a/src/clef.ts
+++ b/src/clef.ts
@@ -143,7 +143,7 @@ export class Clef extends StaveModifier {
 
     this.setPosition(StaveModifierPosition.BEGIN);
     this.setType(type, size, annotation);
-    this.setWidth(Tables.currentMusicFont().lookupMetric(`clef_${this.size}.width`));
+    this.setWidth(Glyph.getWidth(this.clef.code, this.clef.point, `clef_${this.size}`));
     L('Creating clef:', type);
   }
 
--- a/src/stave.ts
+++ b/src/stave.ts
@@ -854,6 +854,45 @@ export class Stave extends Element {
   }
 
   static formatBegModifiers(staves: Stave[]): void {
+    const adjustCategoryStartX = (category: Category) => {
+      let minStartX = 0;
+      // Calculate min start X for the category
+      staves.forEach((stave) => {
+        const modifiers = stave.getModifiers(StaveModifierPosition.BEGIN, category);
+        // Consider only the first instance
+        if (modifiers.length > 0 && modifiers[0].getX() > minStartX) minStartX = modifiers[0].getX();
+      });
+      let adjustX = 0;
+      staves.forEach((stave) => {
+        adjustX = 0;
+        const modifiers = stave.getModifiers(StaveModifierPosition.BEGIN, category);
+        // Calculate adjustement required for the stave
+        modifiers.forEach((modifier) => {
+          if (minStartX - modifier.getX() > adjustX) adjustX = minStartX - modifier.getX();
+        });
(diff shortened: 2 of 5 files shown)


rvilarl:
All notes moves a bit to the left gClef and cClef

Annotation Bottom_Annotation Bravura_current
Annotation Bottom_Annotation Bravura_reference

More with TAB
Annotation Harmonics Bravura_current
Annotation Harmonics Bravura_reference


With Multiple Staves they are aligned to the widest one
Clef_Keys Minor_Key_Clef_Test Bravura_current

mscuthbert:
This looks PERFECT to me.  Thank you!  And i really like the new helper tool.

I wonder if we can change the multistaves key signature tests to add braces or something for staves that are formatted together.  It's a little hard to be sure which go together.  

This is SWEET!

mscuthbert (approved):
Wonderful!  Approved

rvilarl:
> Wonderful! Approved

@mscuthbert did you see https://github.com/0xfe/vexflow/issues/1504#issuecomment-1368254784 ?

mscuthbert:
> > Wonderful! Approved
> 
> @mscuthbert did you see [#1504 (comment)](https://github.com/0xfe/vexflow/issues/1504#issuecomment-1368254784) ?

Answered in #1504.  It's fine.  We just might need more tests as well.

mscuthbert:
More significantly (and much harder) is to eventually get some notion of perceptual distance between elements.  For instance, the distance between the percussion clef and the flat keys looks too little, because the flat is in the same perceptual space as the clef itself.  Whereas the sharp looks great.  (In practice, no one is going to use key signatures w/ percussion clef, so it's a moot point).  Similarly, I think that there's too much space between a Time Signature and the next note when the note is say F# in line 1, but it's just right if it's F-natural. The space between time signature and stem is the most perceptually salient point of distance, not from last pixel of one element to first pixel of the next.

ronyeh:
> And i really like the new helper tool.

@mscuthbert which helper tool?

ronyeh:
I'm reviewing the output now and can merge if no issues.

ronyeh:
Looks OK to me, but I noticed that now Gonville and Petaluma clefs push the notes a bit further to the right, while Bravura and Leland clefs are narrower than before, so they pull the first measure's notes to the left a bit.

I mostly checked the Bach Demo and the percussion tests, and also noticed that some percussion tests now include a 4/4 time signature.

ronyeh:
A random aside:

I notice that `grunt generate:current` sometimes doesn't generate all the required images. It'll stop at say, 2506 images instead of the full 3857 images.

Will probably file a bug if I can get any sort of repro or guess at what's happening. Maybe the main script finishes before all the processes are done generating images?

Also, if we are defaulting to puppeteer, maybe we should remove the pptr- prefix from the image file names.

We can do a single PR to fix these two issues.

sschmidTU:
> Maybe the main script finishes before all the processes are done generating images?

That may very well be, I think I remember an issue like that from using puppeteer image generation with OSMD (we now use a node-only script which is much faster, though the rendering can be slightly different).
It might also be an async method finishing before or after another and so on.

Issue #1498, opened 2022-12-23
VexFlow 5


ronyeh:
List your suggestions / requests in this thread!

Here are some of mine:

* The master branch is renamed to `main`.
* camelCase and UpperCamelCase / PascalCase throughout the code base.
* Support all open SMuFL fonts (that are free for us to redistribute).
* Transition to a VexFlow organization on GitHub: https://github.com/vexflow/vexflow
* Host a font comparison tool (either on vexflow.com or github pages) so developers can quickly compare VexFlow-rendered scores between SMuFL fonts. (Maybe base it on our current [image comparison tool](https://github.com/0xfe/vexflow/tree/master/tools/compare)).

## main branch
The git command line now defaults to `main`, and many top projects have switched over to using `main` (e.g., [TypeScript](https://github.com/microsoft/TypeScript/tree/main), [React](https://github.com/facebook/react/tree/main), [QUnit](https://github.com/qunitjs/qunit/tree/main), [jQuery](https://github.com/jquery/jquery/tree/main), [Stable Diffusion](https://github.com/Stability-AI/stablediffusion/tree/main))

See: https://github.com/github/renaming

For now, we will keep VexFlow version 4.x in the `master` branch  for any important bug fixes. Eventually, we can tag the final long-term support version of 4.x and then retire / delete the `master` branch.

## camelCase

For camelCase conventions, we follow the Google Typescript style guide:

https://google.github.io/styleguide/tsguide.html#identifiers

If anything is still unclear, we fall back to:

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions

As an additional fallback, we can use this guide for contributors to the TypeScript compiler (even though Microsoft says to not use it):

https://github.com/microsoft/TypeScript/wiki/Coding-guidelines

## SMuFL fonts
We will not redistribute commercial SMuFL fonts, but we can try our best to make VexFlow 5 compatible with them.

We will make it easy for a developer to choose which font(s) to include in their final vexflow build.

## VexFlow organization
As the community grows, we can indicate our long term support for this library by housing the official repository under a GitHub "organization". We would want to get our founder Mohit's blessing for this 🤞😁. Maybe we can designate Rodrigo as the lead developer, since he has been doing much of the work for the past almost-two years (see: PR https://github.com/0xfe/vexflow/pull/851).

mscuthbert:
[...]

PR #1513, opened 2022-12-31
Time Signature simplifications

mscuthbert:
To the extent possible without breaking backwards compatibility of public interfaces, this PR implements a little bit of my (second) comment on #1498 which @sschmidTU gave a thumbs up to: deemphasizing using structs in favor of objects themselves on TimeSignature and TimeSigNote.

Places line, glyph, num (now "is_numeric") attributes directly on the TimeSignature object (where the crucial .point) already is, rather than on a .info interface.   (Note that the name of this struct differs among modifiers.  On the Clef object it is ".clef", so changing one type of Modifier without changing all should not break code expecting a standard .info struct.)

TimeSigNote now stores the reference to the TimeSignature object rather than the .info struct, so that it can be modified up until the point it is formatted/drawn.

Feel free to comment if this is the right direction or not.  I won't be offended (too much.  :-D ) if it's not where the project should go.  But class-objects are quite lightweight and as long as we're creating the TimeSignature object anyhow, might as well hang onto it until we're ready.

All visual tests pass.

### DIFF
--- a/src/timesignature.ts
+++ b/src/timesignature.ts
@@ -62,14 +62,13 @@ export class TimeSignature extends StaveModifier {
 
   protected timeSpec: string = '4/4';
   protected line: number = 0;
-  protected glyph: Glyph;
+  protected glyph!: Glyph;
   protected is_numeric: boolean = true;
   protected validate_args: boolean;
 
   constructor(timeSpec: string = '4/4', customPadding = 15, validate_args = true) {
     super();
     this.validate_args = validate_args;
-    this.timeSpec = timeSpec;
 
     const padding = customPadding;
 
@@ -81,13 +80,7 @@ export class TimeSignature extends StaveModifier {
     this.topLine = 2 + fontLineShift;
     this.bottomLine = 4 + fontLineShift;
     this.setPosition(StaveModifierPosition.BEGIN);
-
-    // violates DRY w/ setTimeSig(timeSpec) but needed to convince TypeScript that all is well.
-    const info = this.parseTimeSpec(timeSpec);
-    this.glyph = info.glyph;
-    this.is_numeric = info.num;
-    this.line = info.line;
-
+    this.setTimeSig(timeSpec);
     this.setWidth(defined(this.glyph.getMetrics().width));
     this.setPadding(padding);
   }
--- a/src/timesignature.ts
+++ b/src/timesignature.ts
@@ -36,6 +36,11 @@ const assertIsValidTimeSig = (timeSpec: string) => {
   });
 };
 
+/**
+ * A TimeSignature is a StaveModifier that can make its appropriate Glyphs directly from
+ * a provided "timeSpec" such as "4/4", "C|" (cut time), or even something more advanced
+ * such as "3/4(6/8)" or "2/4+5/8".
+ */
 export class TimeSignature extends StaveModifier {
   static get CATEGORY(): string {
     return Category.TimeSignature;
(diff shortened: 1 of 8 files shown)


mscuthbert:
(if this were a 5.0 PR, I'd get rid of `getInfo()`, `TimeSignatureInfo` and unify `setTimeSig` to `setTimeSpec`, but those aren't backwards compatible).

src/timesignature.ts
```diff
@@ -69,14 +72,16 @@ export class TimeSignature extends StaveModifier {
...
     this.point = musicFont.lookupMetric('digits.point');
+
     const fontLineShift = musicFont.lookupMetric('digits.shiftLine', 0);
     this.topLine = 2 + fontLineShift;
     this.bottomLine = 4 + fontLineShift;
     this.setPosition(StaveModifierPosition.BEGIN);
-    this.info = this.parseTimeSpec(timeSpec);
-    this.setWidth(defined(this.info.glyph.getMetrics().width));
+    this.setTimeSig(timeSpec);
+    this.setWidth(defined(this.glyph.getMetrics().width));
```
rvilarl, line 84: Where is `this.glyph` assigned a value?

src/timesignature.ts
```diff
@@ -106,26 +112,67 @@ export class TimeSignature extends StaveModifier {
...
-    this.info = this.parseTimeSpec(timeSpec);
+    this.timeSpec = timeSpec;
+    const info = this.parseTimeSpec(timeSpec);
+    this.glyph = info.glyph;
+    this.is_numeric = info.num;
+    this.line = info.line;
     return this;
   }
 
+  getTimeSpec(): string {
```
rvilarl, line 137: Comments for the new functions are missing.

src/timesignature.ts
```diff
@@ -106,26 +112,67 @@ export class TimeSignature extends StaveModifier {
...
+  setLine(line: number) {
+    this.line = line;
+  }
+
+  getGlyph(): Glyph {
+    return this.glyph;
+  }
+
+  setGlyph(glyph: Glyph) {
+    this.glyph = glyph;
```
rvilarl, line 154: After setting a new glyph the width should be updated, should not it?

rvilarl (changes requested):

src/timesignature.ts
mscuthbert: in setTimeSig.

src/timesignature.ts
mscuthbert: will add -- I wasn't sure if we were doing them for getter/setter type methods, but will get.

src/timesignature.ts
mscuthbert: ah, yes.  I'll put that in and in fact use the setGlyph() routine within setTimeSig.

mscuthbert:
Done -- and in fact in fixing setting width on change of glyph, uncovered another bug in the current system:

The two-system StaveModifier tests work by first adding all the modifiers, including addTimeSignature() and then work by calling setTimeSignature() which reuses the current TimeSignature object but changes the timeSpec.  Since the width was (previously) set only in the constructor, changing the timeSpec did not change the width.  This can most easily been seen when changing the final time signature from 9/8 (a thin TS) to C (a wider TS).

Current master:
StaveModifier Begin___End_StaveModifier_Test Bravura_reference

PR:
StaveModifier Begin___End_StaveModifier_Test Bravura_current


In Petaluma it's even more obvious (again master followed by PR):

StaveModifier Begin___End_StaveModifier_Test Petaluma_reference

StaveModifier Begin___End_StaveModifier_Test Petaluma_current

That there should be a little bit of empty staff between the end TS and the end of the line is confirmed by Gould Behind Bars:
Screenshot 2023-01-01 at 11 28 07

mscuthbert:
Notice that the last mentioned bug still exists on TimeSigNote since changing the TimeSigNote.timeSig.setGlyph() will only change the width of TimeSigNote.timeSig not of TimeSigNote itself.  Will discuss more in the context of mixins.

rvilarl (approved):

Issue #441, opened 2016-08-29
Style Audit - Bach Minuet in G


Silverwolf90:
@0xfe has been [working on a simple full-score example in VexFlow](https://github.com/0xfe/vexflow/pull/437). Even though it's still in progress I thought this was a great  opportunity to assess VexFlow's default style.  Almost everything in the example consists of defaults except for the  slurs' (ie `Curve`) control points, which are manually set.

![Bach Minuet in G](https://camo.githubusercontent.com/8b7356acc32db4764828e5c34012ae701f15805a/687474703a2f2f696d6775722e636f6d2f596646766a44442e706e67)

Stylistic Issues:

* Key signature too far away from the clef
* Time signature too far away from key signature
* Time signature numbers are too small
* Fingering 2 not centered over noteheads (Bar 6)
* Gracenote is a bit too far away from the note (Bar 8)
* Left modifiers increasing initial stave padding by too much (Bar 9, Bar 15)
* Beams not connected to staff lines
* Beam angles are too aggressive
* Beams feel a bit thin
* Metronome mark too close to tempo text
* Too much spacing between beamed inner notes. (Bar 14)
* Too much spacing between last two notes (Bar 15)
* Slurs are too thick
* Slurs endpoints should be centered on the notehead
* Slur endpoints shouldn't start on staff lines
* Slur endpoints should go above staccato dots
* Augmentation dots are a bit close to notehead
* The thin/thick repeat barlines are too close to each other
* Repeat barline dots too close to barline
* Notes before/after repeat are too close to the barline

*Disclaimer: This is naturally a bit opinionated, but I don't think I've said anything too controversial.*

mscuthbert:
Fantastic.

M. 8 -- the "grace" note should not have a slash through it -- it's an appogiatura.  (B-quarter note + A half note).

0xfe:
Thanks for the audit, Cyril. This is exactly what I was hoping to get from the demo. Do you mind numbering them, so it's easier to refer to them in comments?

Silverwolf90:
Done!

Silverwolf90:
I'm changing:

> Left modifiers increasing initial stave padding by too much (Bar 9, Bar 15)

to 

> Left modifiers increase spacing between objects when not necessary or by too much

rvilarl:
[...]

PR #1521, opened 2023-01-11
Addresses a grace note issue mentioned in the #441 thread.

ronyeh:
A

The only difference is in measure 8.


mscuthbert:
Thanks!

PR #1536, opened 2023-03-18
point fixed on rests

rvilarl:
Rest use now default point size

### DIFF
--- a/src/fonts/common_metrics.ts
+++ b/src/fonts/common_metrics.ts
@@ -427,14 +427,7 @@ export const CommonMetrics = {
         shiftX: -8,
       },
     },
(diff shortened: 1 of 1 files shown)


rvilarl:
Rests are now slightly bigger
pptr-canvas_CrossBeam Single_clef_mixed_1 Bravura_current
pptr-canvas_CrossBeam Single_clef_mixed_1 Bravura_reference

ronyeh:
Thanks.

I think we should engrave a simple score (like 4 measures) with VexFlow and then compare the visual output with LilyPond, MuseScore, Dorico, etc. It would be cool to know what our scores look like next to the other top engraving software.

PR #1540, opened 2023-03-19
point fixed on tremolo and parenthesis

rvilarl:
No visual differences on Paranthesis. Tremolos are now bigger.

### DIFF
--- a/src/fonts/common_metrics.ts
+++ b/src/fonts/common_metrics.ts
@@ -217,11 +217,9 @@ export const CommonMetrics = {
 
   parenthesis: {
     default: {
-      point: 39,
       width: 7,
     },
     gracenote: {
-      point: (39 * 3) / 5,
       width: 3,
     },
   },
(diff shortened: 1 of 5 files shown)


rvilarl:
pptr-Percussion Percussion_Snare2 Bravura svg_current
pptr-Percussion Percussion_Snare2 Bravura svg_reference

rvilarl:
Captured from Musescore, I would say that size is right but position is and was too close to notehead. Musescore extends the stem if required. We do not.

image

rvilarl:
@AaronDavidNewman @mscuthbert @ronyeh @sschmidTU I would like to get your oppinion on this one. I go next week on holidays and it would be great to have the new beta ready. I think that the new size is better and closer to the one present on other rendering tools. I also believe that we introduced the big variant (feature request) because the originals wer too small.

ronyeh:
Seeing the MuseScore output, it is clear that our solution should adjust the length of the stem to improve the spacing. Otherwise the bigger tremolo makes it look too crowded, since it is right next to the notehead.

ronyeh:
There is a 
```
const start = note.getModifierStartXY(this.position, this.index);
let x = start.x;
```

Is there something we can do if we use the start.y? Maybe we should adjust it to be farther away from the note head, so we can accommodate the heavier tremolo strokes.

rvilarl:
I have uncreased the distance to the notehead, @ronyeh thaks for the feedback

pptr-Percussion Percussion_Snare2 Bravura svg_current

Increasing the stem lenght should be another PR, merging

PR #1541, opened 2023-03-20
gonville with common metrics

### LINKED ISSUE #1523: Proposal: Unifying smufl metrics
Today we have a metrics file per font to adjust margins, scales,...

I want to propose to:
- use the same metrics for all the fonts (this is an important step to allow the easy addition of new fonts once we move to FontFace).
- resolve the scaling issues modifying the OTF files following the approach I used for Gonville.

rvilarl:
fixes #1523


rvilarl:
gonville now ready for the new beta release. Lots of minor changes, I think that it looks good.

ronyeh:
So how did you accomplish this?

Is this a custom "Rodrigo's" version of Gonville where you have tweaked the size and origin of every glyph?

How would we replicate the work that you did? Like, if we downloaded a fresh version of Gonville from the source website, how do we end up with the glyph files you have here? 

We need to make sure it's reproducible.

ronyeh:
I see you're overwriting GonvilleSmufl.otf. I know it's in the git history, but do we need to keep different "versioned" GonvilleSmufl.otf files? Like Maybe GonvilleSmufl_1.0.otf and GonvilleSmufl_1.0.1.otf, or a dated version like GonvilleSmufl_2023_03_31.otf

How did you produce this GonvilleSmufl.otf file? 

I know in a previous PR, you used https://github.com/rvilarl/gonville/blob/remappingOnly/gonville2smufl.py

Is there a python script that you wrote that converts the Gonville from this webpage: https://www.chiark.greenend.org.uk/~sgtatham/gonville/

...into the gonville_glyphs.ts file in this PR? Can you include that tool in this PR, so we can reproduce the conversion if needed?

rvilarl:
yes I can add the tool to /tools. This rodrigo's version is much closer to the original. Probably you remember that I had to scale all the glyphs to keep the original metrics. This rodrigo's version is not scaled.

rvilarl:
@ronyeh script is now in

PR #1544, opened 2023-03-23
fix notehead bounding box

### LINKED ISSUE #556: Can't obtain Notehead Bounding Box
After everything is formatted, my_notehead.getBoundingBox() gives the error "Can't call
getBoundingBox on an unformatted note.", because my_notehead.preFormatted is set to false.
However, there are four variables needed for the calculation, all of which are returning good values -- therefore the bounding box should be obtainable!

The variables are:
my_notehead.stave.getSpacingBetweenLines() 
my_notehead.getAbsoluteX()
my_notehead.width
my_notehead.y

oriented
(The attached picture shows why this is needed -- this illustration has to do with melodic pitch-shape: lines between notes should be oriented down or up, according to melodic direction, but this would only work with the bounding box on the notehead, not on the stavenote!)

rvilarl:
fixes #556

- noteheads of stavenotes are not preformatted.
- stavenot reset should use setstave to recalculate properly.

### DIFF
--- a/src/notehead.ts
+++ b/src/notehead.ts
@@ -10,7 +10,7 @@ import { Stave } from './stave';
 import { Stem } from './stem';
 import { Tables } from './tables';
 import { Category } from './typeguard';
-import { defined, log, RuntimeError } from './util';
+import { defined, log } from './util';
 
 // eslint-disable-next-line
 function L(...args: any[]) {
@@ -232,9 +232,6 @@ export class NoteHead extends Note {
 
   /** Get the `BoundingBox` for the `NoteHead`. */
   getBoundingBox(): BoundingBox {
-    if (!this.preFormatted) {
-      throw new RuntimeError('UnformattedNote', "Can't call getBoundingBox on an unformatted note.");
-    }
     const spacing = this.checkStave().getSpacingBetweenLines();
     const half_spacing = spacing / 2;
     const min_y = this.y - half_spacing;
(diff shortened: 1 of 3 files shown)


rvilarl:
Test changed to use also StaveNotes, this one wouldfail without this PR (current first)
pptr-NoteHead Bounding_Boxes Bravura svg_current
pptr-NoteHead Bounding_Boxes Bravura svg_reference

Differences due to StaveNote change (current first)
pptr-Auto_Beaming Break_Beams_on_Middle_Rests_Only Bravura svg_current
pptr-Auto_Beaming Break_Beams_on_Middle_Rests_Only Bravura svg_reference
pptr-Auto_Beaming Beam_Across_All_Rests_with_Stemlets Bravura svg_current
pptr-Auto_Beaming Beam_Across_All_Rests_with_Stemlets Bravura svg_reference
pptr-Auto_Beaming Beam_Across_All_Rests Bravura svg_current
pptr-Auto_Beaming Beam_Across_All_Rests Bravura svg_reference
pptr-Tuplet Complex_Tuplet Bravura svg_current
pptr-Tuplet Complex_Tuplet Bravura svg_reference

rvilarl:
@sschmidTU The first test, which was modified to use StaveNotes rather than NoteHeads crashes without the modification. 
The last beam with rests in the middle is lower.
The 3 on the last image is very slightly lower.

PR #1545, opened 2023-03-23
unison configurable

### LINKED ISSUE #1507: Unison noteheads too aggressive
(Vexflow 4 regression)

VF4 (3?) allows for StaveNotes in different voices to be formatted so that they can share noteheads in certain cases in the great #1263 PR.  I've noticed however that the algorithm is somewhat too aggressive.  I've just been looking at the two-voice code so far.

It checks for stemmed vs unstemmed, notes vs rests, white noteheads vs black, number of dots, but there are a number of things that are not yet working that I think needs a partial rewrite (started in "fix_unison_noteheads" branch).

(1) start with a conservative approach: assume notes with any overlap cannot be at the same xShift unless proven innocent.  :-)  That's the biggest change that seems to be the major change here.

(2) the method of counting number of Dot modifiers on each note and seeing if they are equal is not robust enough.  Here a two-note chord (F-C) with one dot on each note is merged with a single double-dotted C.

Screenshot 2022-12-28 at 16 29 50

(3) different notehead glyphs except for white vs. black noteheads are not taken into account.  Harmonics must be shifted from regular notes, etc.  (whole notes and breves seem to work great except that the shorter note should be the one shifted right):

Screenshot 2022-12-28 at 16 47 27

(4) noteheads with different styles (colors, etc.) should not share the same notehead.

(5) (very important): notes with different accidentals should never share a notehead unless either both accidentals are displayed and the notes are the exact same duration/lyrics/instrument/etc. (so it doesn't matter which accidental goes with which note).

Gould 55-60 shows how many cases are ambiguous.  Thus one of the most important things we can do is have a system where these formatter rules can be turned on and off (on a per note and system-wide setting).  Probably my first PR will just be to add a way of disabling the unison notehead setting.

When it's done right, sharing a notehead is very cool, so super thankful for OSMD and @rvilarl for implementing; but it's never absolutely necessary, so I wonder if we can walk back a little bit on this until bugs are worked out a bit more.

(note that rules for notehead sharing are very different for keyboard/harp music; there noteheads are shared more aggressively.  See Gould 307-8; but this is given as an exception not the general rule)

Thanks!

rvilarl:
fixes #1507 

No visual differences with `Tables.UNISON=true` which is the default value.

### DIFF
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -279,8 +279,8 @@ export class StaveNote extends StemmableNote {
             } //Very close whole notes
           } else if (lineDiff < 1) {
             xShift = voiceXShift + 2;
-            if (noteU.stemDirection === noteL.stemDirection) {
-              // upper voice is middle voice, so shift it right
+            if (noteU.note.duration < noteL.note.duration) {
+              // upper voice is shorter, so shift it right
               noteU.note.setXShift(xShift);
             } else {
               // shift lower voice right
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -231,35 +231,29 @@ export class StaveNote extends StemmableNote {
           //If we are sharing a line and in the same voice, only then offset one note
           const lineDiff = Math.abs(noteU.line - noteL.line);
           if (noteU.note.hasStem() && noteL.note.hasStem()) {
-            //If we have different dot values, must offset
-            //Or If we have a white mixed with a black notehead, must offset
-            let whiteNoteHeadCount = 0;
-            let blackNoteHeadCount = 0;
-            if (Tables.durationToNumber(noteU.note.duration) === 2) {
-              whiteNoteHeadCount++;
-            } else if (Tables.durationToNumber(noteU.note.duration) > 2) {
-              blackNoteHeadCount++;
-            }
-            if (Tables.durationToNumber(noteL.note.duration) === 2) {
-              whiteNoteHeadCount++;
-            } else if (Tables.durationToNumber(noteL.note.duration) > 2) {
-              blackNoteHeadCount++;
-            }
+            const noteUHead = Tables.codeNoteHead(
+              noteU.note.sortedKeyProps[0].keyProps.code ?? 'N',
+              noteU.note.duration
+            );
+            const noteLHead = Tables.codeNoteHead(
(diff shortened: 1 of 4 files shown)


rvilarl:
with `Tables.UNISON=false` (reference first then current)
pptr-StaveNote No_Padding Bravura svg_reference
pptr-StaveNote No_Padding Bravura svg_current
pptr-StaveNote Note_Heads_Placement___Simple Bravura svg_reference
pptr-StaveNote Note_Heads_Placement___Simple Bravura svg_current
pptr-Three_Voice_Rests Three_Voices____2_Complex Bravura svg_reference
pptr-Three_Voice_Rests Three_Voices____2_Complex Bravura svg_current

rvilarl:
I have added the cases mentioned by @mscuthbert in #1507 these will be fixed on another PR 

pptr-Unison Accidentals_true_ Bravura svg
pptr-Unison Breve_true_ Bravura svg
pptr-Unison Dots_true_ Bravura svg
pptr-Unison Style_true_ Bravura svg

rvilarl:
Unison tests fixed

pptr-Unison Accidentals_true_ Bravura svg
pptr-Unison Dots_true_ Bravura svg
pptr-Unison Breve_true_ Bravura svg
pptr-Unison Style_true_ Bravura svg

rvilarl:
@mscuthbert for completion the results with unison disabled
pptr-Unison Accidentals_false_ Bravura svg
pptr-Unison Dots_false_ Bravura svg
pptr-Unison Breve_false_ Bravura svg
pptr-Unison Style_false_ Bravura svg

rvilarl:
@mscuthbert ask discussed in #1507 I would like to get your feedback about the accidentals. I will look into the breve topic

rvilarl:
@mscuthbert the rule "the shorter note should be the one shifted right" is now implemented
 
pptr-Unison Breve_true_ Bravura svg

mscuthbert:
> @mscuthbert for completion the results with unison disabled
> pptr-Unison Accidentals_false_ Bravura svg

acceptable and probably best for stems in same direction.  Stems in opposite direction, the second accidental should separate the notes, but no one implements that algorithmically...

> pptr-Unison Dots_false_ Bravura svg

Looking good, except that the algorithm should take into account dots on the first note in figuring out how far to shift the second note right (but that's a different error altogether)

> pptr-Unison Style_false_ Bravura svg

Great!  This is really the one I was hoping for!  :-)

PR #1552, opened 2023-04-06
clef annotations size refactor

rvilarl:
Four more points removed from the metrics. I have also used the same scale related to the size of the clef.

### DIFF
--- a/src/clef.ts
+++ b/src/clef.ts
@@ -23,7 +23,6 @@ export interface ClefMetrics {
   width: number;
   annotations: {
     [key: string]: {
-      point: number;
       [type: string]: { line?: number; shiftX?: number } | number;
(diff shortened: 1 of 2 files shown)


rvilarl:
current first

pptr-canvas_Clef Clef_Change_Test Bravura_current
pptr-canvas_Clef Clef_Change_Test Bravura_reference

pptr-canvas_Clef Clef_Test Bravura_current
pptr-canvas_Clef Clef_Test Bravura_reference

rvilarl:
This makes the default clef annotation a little bigger but making the constant smaller makes the annotation, in my oppinion too small for change clefs. Merging, the code chage is minor and the constant can be adjusted on the visual review.

PR #1554, opened 2023-04-07
coda point standard

rvilarl:
the point used with coda was 20, now making it standard to 39 they look much bigger but they follow now the creators' design.

### DIFF
--- a/src/staverepetition.ts
+++ b/src/staverepetition.ts
@@ -15,7 +15,7 @@ export class Repetition extends StaveModifier {
 
   static TEXT_FONT: Required<FontInfo> = {
     family: Font.SERIF,
-    size: 12,
+    size: Tables.NOTATION_FONT_SCALE / 3,
     weight: FontWeight.BOLD,
     style: FontStyle.NORMAL,
   };
@@ -192,7 +192,9 @@ export class Repetition extends StaveModifier {
       this.y_shift +
       Tables.currentMusicFont().lookupMetric('staveRepetition.symbolText.offsetY');
     if (draw_coda) {
-      Glyph.renderGlyph(ctx, symbol_x, y, 40, 'coda', { category: 'coda' });
+      Glyph.renderGlyph(ctx, symbol_x, y, Font.convertSizeToPointValue(this.textFont?.size) * 2, 'coda', {
+        category: 'coda',
+      });
     }
 
     ctx.fillText(text, text_x, y + 5);
--- a/src/fonts/common_metrics.ts
+++ b/src/fonts/common_metrics.ts
@@ -344,7 +344,6 @@ export const CommonMetrics = {
   // Details in Glyph.lookupFontMetrics.
   glyphs: {
     coda: {
-      point: 20,
       shiftX: -7,
       shiftY: 8,
     },


rvilarl:
Bravura (current first)
pptr-Stave Multiple_Staves_Volta_Test Bravura svg_current
pptr-Stave Multiple_Staves_Volta_Test Bravura svg_reference
Gonville (current first)
pptr-Stave Multiple_Staves_Volta_Test Gonville svg_current
pptr-Stave Multiple_Staves_Volta_Test Gonville svg_reference
Leland (current first)
pptr-Stave Multiple_Staves_Volta_Test Leland svg_current
pptr-Stave Multiple_Staves_Volta_Test Leland svg_reference
Petaluma (current first)
pptr-Stave Multiple_Staves_Volta_Test Petaluma svg_current
pptr-Stave Multiple_Staves_Volta_Test Petaluma svg_reference

sschmidTU:
image

I find it unlikely this is the "creators' design", it's ridiculously large, I've never seen a coda symbol remotely as big in any score.
It's also overlapping with the volta.

Here's one example of coda signs in a physical edition I found:
image
Even the second coda symbol near the end looks like it could be smaller.

rvilarl:
@sschmidTU I agree with you in Bravura but what about Gonville? I think that the current is better.

In the physical edition that you posted there are two interesting things:
1) There are two sizes of coda, what is the reason behind that
2) The big coda and the segno are quite similar in size. I would say same font size.

The Petaluma key numbers are also very big.

AaronDavidNewman:
>> There are two sizes of coda, what is the reason behind that

The coda in the 'to coda' symbol is often a little smaller than the symbol marking the start of the 'coda'.  There is only one of each in any piece of music, so you don't gain much by making it hard to see. 

We need better test cases.  In most of the music I've seen, the start of the coda is left-flush with the page, and vertically spaced from the rest of the tune.  Not sure why you would have a coda under a volta...

rvilarl:
@AaronDavidNewman thanks for the info. Here some examples from Musescore

Bravura
image
Gonville
image
Emmentaler
image

rvilarl:
@sschmidTU it seems that it is indeed the intention see

https://github.com/0xfe/vexflow/pull/1554

ronyeh:
Regardless, we should allow the vexflow developer to customize the default size of each glyph easily for each font. 

Right now we have a single common metrics file, but it should also load a Bravura, Petaluma, Gonville, Leland, etc file that overrides the common metrics defaults. 

These files can be empty by default, with comments to indicate how to override individual glyphs. 

In the past, each file was different, but in the current master we should start each file as an empty json object.

rvilarl:
> Regardless, we should allow the vexflow developer to customize the default size of each glyph easily for each font.
> 
> Right now we have a single common metrics file, but it should also load a Bravura, Petaluma, Gonville, Leland, etc file that overrides the common metrics defaults.
> 
> These files can be empty by default, with comments to indicate how to override individual glyphs.
> 
> In the past, each file was different, but in the current master we should start each file as an empty json object.

I am aiming to get rid of the Glyphs files in Vexflow5 in order to be able to load any Font. I am also aiming to use music and text fonts in the same way, it is difficult to support to have each letter with a differen size on a text in htmtl.

We will also have to move more in the direction of run-time calculations. I did already quite a few to get common metrics, but if we want to allow any glyph to have any size, it will be a step further.

rvilarl:
The size of the text in StaveRepetition was hardcoded. I have changed it to a run time calculation and have set the coda to double the calculated size.

pptr-Stave Multiple_Staves_Volta_Test Bravura svg_current
pptr-Stave Multiple_Staves_Volta_Test Gonville svg_current
pptr-Stave Multiple_Staves_Volta_Test Leland svg_current
pptr-Stave Multiple_Staves_Volta_Test Petaluma svg_current

sschmidTU:
@rvilarl Yes, Gonville looks better in current.
I'm not interested in Petaluma though. For Bravura, it's a different question.

ronyeh:
It's good that we are slowly moving toward the default sizes in the SMuFL fonts, but we need to make sure we aren't breaking the layouts (e.g., coda overlapping with volta brackets). 

If we continue accepting all these visual changes we will need to be clear that the 4.2 release is actually a big / breaking release. I hesitate to call it 5.0 even though it should be if we were following semver. 

In 4.2 we should allow people to revert to the older layout if possible (via metrics files that are empty by default).

ronyeh:
> Not sure why you would have a coda under a volta...

Right. :-) I agree with needing better test cases. Many of our test cases are contrived and shove as many symbols as possible into the same staff. So it's possible that we'd never run into the coda x volta layout issue in real life settings, haha.

mscuthbert:
> > Not sure why you would have a coda under a volta...
> 
> Right. :-) I agree with needing better test cases. Many of our test cases are contrived and shove as many symbols as possible into the same staff. So it's possible that we'd never run into the coda x volta layout issue in real life settings, haha.

(Aside): This is a real problem in lots of music tests.  For instance, Byrd's proposed OMRTestBed where some of the examples are things that I would hope any musically intelligent OMR system pauses before trying to parse, like line one where the same whole rest is represented 4 different ways without reason (A smart system: are we sure one of those rests isn't a note?) or the last line (What instrument uses alto clef, one staff, and pedal marks?  One of those must be wrong).  And I don't even know if the same-pitch notes on line 2 are meant to be slurred or tied?  (If slurred, why doesn't the slur fully encompass the notes?  If tied, why not just write them as a half note?)

TP3_QT3_600

ronyeh:
If one of the goals of these PRs is to follow the font designer's original design, we should find engraved examples from the font creator to compare our VexFlow output with.

For example, Bravura was designed for Dorico.

https://www.smufl.org/fonts/

Can we get an example engraved with Dorico that includes the coda symbols. Are they also huge in Dorico?

rvilarl:
@ronyeh see https://github.com/steinbergmedia/petaluma/issues/17

ronyeh:
> @ronyeh see [steinbergmedia/petaluma#17](https://github.com/steinbergmedia/petaluma/issues/17)

But comment was for Petaluma. I am using Dorico (which I believe defaults to Bravura) and the default Coda symbol is approximately the size of the capital letter C. 

I'll try to generate an example to share.

Regardless, it seems useful to allow the developer to customize the size if they desire (although we should pick a good default).

ronyeh:
Here are the two Dorico projects:
[Coda.zip](https://github.com/0xfe/vexflow/files/11424105/Coda.zip)

The way I compare the size of coda glyphs is to compare it with the vertical distance between any two staff lines. So how many line spaces tall should a coda glyph be?

I think the examples in this PR are too big (at least the first coda sign). The SMuFL fonts only provide a single coda glyph. How come we make one coda glyph bigger than the other?

rvilarl:
> The way I compare the size of coda glyphs is to compare it with the vertical distance between any two staff lines. So how many line spaces tall should a coda glyph be?

I would say 2 in dorico.

> I think the examples in this PR are too big (at least the first coda sign). The SMuFL fonts only provide a single round coda glyph (and one rectangular coda glyph). How come we make one coda glyph bigger than the other in our tests?

Simon (sschmidTU) provide an example above with 2 sizes. Codas are used in two clases `TextNote` and `StaveRepetition`. `TextNote` is using the default and `StaveNote` the size of the text * 2.

src/staverepetition.ts
```diff
@@ -192,7 +192,9 @@ export class Repetition extends StaveModifier {
       this.y_shift +
       Tables.currentMusicFont().lookupMetric('staveRepetition.symbolText.offsetY');
     if (draw_coda) {
-      Glyph.renderGlyph(ctx, symbol_x, y, 40, 'coda', { category: 'coda' });
+      Glyph.renderGlyph(ctx, symbol_x, y, Font.convertSizeToPointValue(this.textFont?.size) * 2, 'coda', {
```
ronyeh, line 195: Maybe the hardcoded 2 should be customizable in the future? Perhaps I don't want my coda to be that huge? I could set it to 1 instead....

ronyeh (approved):

PR #1564, opened 2023-05-01
use assert in qunit calls

rvilarl:
1st step to update QUnit:

- use assert parameter.

No visual differences.


rvilarl:
This change is required to generate the refence images once QUnit is updated.

PR #1572, opened 2023-05-30
feat(CanvasContext): Added optional devicePixelRatio argument to resize

saitonakamura:
It's mainly useful in context of rendering in web worker, because web worker doesn't have access to devicePixelRatio, so resize code will be split between
1. main thread asking for a render, providing devicePixelRatio
2. Worker rendering and resizing CanvasContext w/ provided ratio
3. Worker giving back main thread the resize target
4. Main thread applying css width/height provided by the worker

### DIFF
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -169,19 +169,19 @@ export class CanvasContext extends RenderContext {
     return this;
   }
 
-  resize(width: number, height: number): this {
+  resize(width: number, height: number, devicePixelRatio?: number): this {
     const canvas = this.context2D.canvas;
-    const devicePixelRatio = globalObject().devicePixelRatio || 1;
+    const dpr: number = devicePixelRatio ?? globalObject().devicePixelRatio ?? 1;
 
     // Scale the canvas size by the device pixel ratio clamping to the maximum supported size.
-    [width, height] = CanvasContext.sanitizeCanvasDims(width * devicePixelRatio, height * devicePixelRatio);
+    [width, height] = CanvasContext.sanitizeCanvasDims(width * dpr, height * dpr);
 
     // Divide back down by the pixel ratio and convert to integers.
-    width = (width / devicePixelRatio) | 0;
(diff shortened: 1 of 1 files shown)


rvilarl:
Do you use the SVGContext? Would something similar be required there?

ronyeh:
Thanks!

I think we have been slowly going away from checks against `null` and using the `??` operator where possible. 

Does this code work for you?

```typescript

  resize(width: number, height: number, devicePixelRatio?: number): this {
    const canvas = this.context2D.canvas;
    const dpr: number = devicePixelRatio ?? globalObject().devicePixelRatio ?? 1;

    // Scale the canvas size by the device pixel ratio clamping to the maximum supported size.
    [width, height] = CanvasContext.sanitizeCanvasDims(width * dpr, height * dpr);

    // Divide back down by the pixel ratio and convert to integers.
    width = (width / dpr) | 0;
    height = (height / dpr) | 0;

    canvas.width = width * dpr;
    canvas.height = height * dpr;

    // The canvas could be an instance of either HTMLCanvasElement or an OffscreenCanvas.
    // Only HTMLCanvasElement has a style attribute.
    if (isHTMLCanvas(canvas)) {
      canvas.style.width = width + 'px';
      canvas.style.height = height + 'px';
    }

    return this.scale(dpr, dpr);
  }

```

If it works for you, can you resubmit the PR with this approach? I'll be happy to merge it.

Thanks for your contribution!

saitonakamura:
> Do you use the SVGContext? Would something similar be required there?

Thanks for the feedback! I personally don't, but I don't think SVGContext takes any advantage of the devicePixelRatio, it's usually only the canvas thing in my experience. Also don't see it in the svgcontext code

> I think we have been slowly going away from checks against null. Instead we try to use the ?? operator where possible.

Thanks for the feedback too! Works for me, updated

rvilarl (approved):
LGTM thanks.

ronyeh:
Thanks!

Issue #1571, opened 2023-05-25
Bundle qunit when we publish to NPM


ronyeh:
"tests/support" no longer exists. flow.html currently accesses the qunit files in node_modules/.

When shipping to NPM, we need to somehow bundle those qunit files alongside flow.html, so that it can continue to be accessed on unpkg.com. For example:

https://unpkg.com/vexflow@4.1.0/tests/flow.html

Maybe this option will work:
https://docs.npmjs.com/cli/v9/configuring-npm/package-json#bundledependencies

ronyeh:
@rvilarl Does flow.html need to work while your computer is offline?

If not.... we should consider following the approach from the QUnit getting started pages:
https://qunitjs.com/intro/

They recommend just including the QUnit JS and CSS from the jQuery CDN. I think that'll make it simpler.

```
<!DOCTYPE html>
<meta charset="utf-8">
<title>Test Suite</title>
<link rel="stylesheet" href="https://code.jquery.com/qunit/qunit-2.19.4.css">
<body>
  <div id="qunit"></div>
  <div id="qunit-fixture"></div>
  <script src="https://code.jquery.com/qunit/qunit-2.19.4.js"></script>
</body>
```

Our local grunt test can still use the local node_modules/qunit/, but I think flow.html should load from CDN. That way, when we deploy on unpkg.com, the test page will magically work.

ronyeh:
OK, I'm actually reintroducing a local copy of (the most recent version of) qunit.css and qunit.js in the `vexflow/tests/qunit/`. That's how we did it before, and I think that's the best option.

I think I found a "bug" in the qunit.js that limits the dropdown to 20 modules.

ronyeh:
Here's the fix. QUnit was chopping off the dropdown to the first 20 items.

```js

diff --git a/tests/qunit/qunit.css b/tests/qunit/qunit.css
index d445a567..5d8b3964 100644
--- a/tests/qunit/qunit.css
+++ b/tests/qunit/qunit.css
@@ -236,6 +236,9 @@
 	color: #0D3349;
 	background-color: #F5F5F5;
 	z-index: 99;
+
+    max-height: 80vh;
+    overflow: auto;
 }
 
 #qunit-modulefilter-actions {
diff --git a/tests/qunit/qunit.js b/tests/qunit/qunit.js
index 5e48b793..a01121e1 100644
--- a/tests/qunit/qunit.js
+++ b/tests/qunit/qunit.js
[...]

PR #1573, opened 2023-06-02
Qunit Fixes/Improvements

### LINKED ISSUE #1570: Flow.html dropdown does not show all unit test modules.
There is some nice new styling for qunit, but the modules dropdown now I can't access many test cases from the 'module' dropdown. I tried it in both Chrome and Firefox on windows 11.

This is changeset from master https://github.com/0xfe/vexflow/commit/fff37a0e55cd980c41c0d4c5d82a64c747bf0599 dated 5/21.

image

ronyeh:
This fixes #1570 and #1571 

The newest version of QUnit was limiting the dropdown to the first 20 modules. We have way more modules....

```js
diff --git a/tests/qunit/qunit.css b/tests/qunit/qunit.css
index d445a567..5d8b3964 100644
--- a/tests/qunit/qunit.css
+++ b/tests/qunit/qunit.css
@@ -236,6 +236,9 @@
 	color: #0D3349;
 	background-color: #F5F5F5;
 	z-index: 99;
+
+    max-height: 80vh;
+    overflow: auto;
 }
 
 #qunit-modulefilter-actions {
diff --git a/tests/qunit/qunit.js b/tests/qunit/qunit.js
index 5e48b793..a01121e1 100644
--- a/tests/qunit/qunit.js
+++ b/tests/qunit/qunit.js
@@ -5509,7 +5509,7 @@
           // module names, indicating how the interface works. This also makes
           // for a quicker interaction in the common case of small projects.
           // Don't mandate typing just to get the menu.
-          results = dropdownData.options.slice(0, 20).map(function (obj) {
+          results = dropdownData.options.map(function (obj) {
             // Fake empty results. https://github.com/farzher/fuzzysort/issues/41
             return {
               obj: obj
@@ -5517,7 +5517,7 @@
           });
         } else {
           results = fuzzysort.go(searchText, dropdownData.options, {
-            limit: 20,
+            limit: 1000,
             key: 'name',
             allowTypo: true
           });
```

### DIFF
--- a/tests/flow.html
+++ b/tests/flow.html
@@ -14,19 +14,20 @@
     <meta charset="UTF-8" />
   </head>
   <body>
+    <div style="text-align: center; margin-top: 180px">
+      <h2>[ <a href="https://vexflow.com">Home</a> ] [ <a href="https://github.com/0xfe/vexflow">GitHub</a> ]</h2>
+      <h3>
+        Please wait a moment while the tests run....<br /><br />
+        See the: <a id="vex-src" target="_blank"></a>. Don't forget to run the
+        <a href="https://github.com/0xfe/vexflow/wiki/Visual-Regression-Tests">Visual Regression Tests</a>!
+      </h3>
+    </div>
     <div style="text-align: center">
       <div id="qunit"></div>
       <div id="qunit-fixture"></div>
-      <div>
-        <h2>[ <a href="https://vexflow.com">Home</a> ] [ <a href="https://github.com/0xfe/vexflow">GitHub</a> ]</h2>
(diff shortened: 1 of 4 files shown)


tests/flow.html
```diff
@@ -164,7 +162,6 @@ <h3>
       // Show only failed tests.
       QUnit.config.hidepassed = true;
       QUnit.config.noglobals = true;
-      QUnit.start();
```
ronyeh, line 167: QUnit autostarts....

rvilarl:
Thanks @ronyeh !

Issue #1595, opened 2023-09-14
Flow.keyProperties hangs in certain contexts


jaredjj3:
**TL;DR**: Doing arithmetic using `-0` seems to vary by platform. I recommend that `Table.keyProperties` to be updated to remove one of its [instances](https://github.com/0xfe/vexflow/blob/207dc59e3a6659854898c488a0a186dbcc189acd/src/tables.ts#L618).

[vexflow-key-properties-issue](https://github.com/jaredjj3/vexflow-key-properties-issue) is the reproduction repo, which also has a README that describes the symptoms and how to run the examples.

@rvilarl and I have been working on a MusicXML -> VexFlow library called `vexml`. I'm having trouble with https://github.com/stringsync/vexml/pull/96. It's a really esoteric issue that seems orthogonal to vexflow, but I figured I'd reach out to see if you all could find something I'm doing wrong.

I observed that `Flow.keyProperties` cannot be called >O(1000) times in a test run ([example](https://github.com/jaredjj3/vexflow-key-properties-issue/blob/23786c3fdddd6a9c969101bbe6109a6bf41a718b/src/flow.test.js)), _depending on the platform_. I narrowed down the issue to [this line in Table.keyProperties](https://github.com/0xfe/vexflow/blob/207dc59e3a6659854898c488a0a186dbcc189acd/src/tables.ts#L618).

If I change the implementation from

```ts
octave += -1 * options.octave_shift;
```

to

```ts
octave -= options.octave_shift;
``` 

the tests work. I suspect that it's some issue propagating `-0` on some platforms. However, when I explode out the operations explicitly, I'm unable to reproduce it.

Would you accept a PR to make the change I mentioned?

rvilarl:
Of course we would. Thanks Jared!

ronyeh:
🤯

Wow. Thanks for the catch and fix. If you ever figure out why this occurs, I'd love to learn about it.

PR #1596, opened 2023-09-15
Prevent octave from being -0 when calculating key properties

jaredjj3:
Fixes https://github.com/0xfe/vexflow/issues/1595. -0 seems to cause issues on certain platforms/environments. There are other instances where -0 is possible throughout the codebase, but I wanted this PR to be as minimal as possible.


rvilarl (approved):

mscuthbert:
Did this also get into the v5 main branch?

rvilarl:
There are two things that I have in my todo list, this one and a return value that was changed on reques of another user.

PR #1598, opened 2023-10-03
Fix ornament delayed draw, such as on a grand staff.

RyoSusami:
When building a TickContext with multiple voices, such as a grand staff, the order of ticks was not continuous, so detection of the end of voice element failed and the Ornament display position was incorrect.

**Before**
image

**After**
image

### DIFF
--- a/src/ornament.ts
+++ b/src/ornament.ts
@@ -303,7 +303,7 @@ export class Ornament extends Modifier {
     // Ajdust x position if ornament is delayed
     if (this.delayed) {
       let delayXShift = 0;
-      const startX = glyphX - (stave.getX() - 10);
+      const startX = glyphX - stave.getNoteStartX();
       if (this.delayXShift !== undefined) {
         delayXShift = this.delayXShift;
       } else {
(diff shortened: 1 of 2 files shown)


rvilarl:
Hi @RyoSusami, thanks for that! Would it be possible to add the testcase?

RyoSusami:
Additionally, fix considering the modifier width at the beginning of stave.

**Before**
image

**After**
image

RyoSusami:
@rvilarl
I have never created a VexFlow test case, but I will check it and try.
Is there anything I should be careful about?

rvilarl:
just start duplicating one within ornament_test.ts

RyoSusami:
@rvilarl
Added test case.
This is what the changes with this fix look like.

**Before**
image

**After**
image

RyoSusami:
I noticed that in the case above, the TickID order of TickContexts is 0,8192,4096, and the next context at the beginning of the lower voice is 8192, so the turn position is closer to the right than expected.
Need different approach than TickContext.getNextContext(context)?

RyoSusami:
Improved selection of nextContext.

image

rvilarl:
Thanks a lot @RyoSusami !

PR #1615, opened 2024-04-02
Factory initialization directly with Renderer

### LINKED ISSUE #580: Factory, DOM element and React
I am wondering about the EasyScore example, and how you configure the renderer by means of the Factory.

Because I am using React I don't want to use `getElementById`, but get the DOM element with React Component's `ref` attribute

Although the Factory explicitly asks for a `elementId`, I see in Renderer that it is still ok to pass the element itself.

```
    this.element = document.getElementById(elementId);
    if (!this.element) this.element = elementId;
```

Can I safely assume this option will remain available in the future?

MGreek:
I made some changes to `Factory` such that `FactoryOptions` can now accept an instance of `Renderer`.

This fixes #580.

### DIFF
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -40,6 +40,14 @@ interface OffscreenCanvasRenderingContext2D
  * underlying CanvasRenderingContext2D object, part of the browser's API.
  */
 export class CanvasContext extends RenderContext {
+  getHeight(): number {
+    return this.context2D.canvas.height;
+  }
+
+  getWidth(): number {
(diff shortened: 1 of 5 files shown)


ronyeh:
Minor comment: Can you fix the typo in the second commit message?

```
allow initilalization => allow initialization
```

Do you have a small-ish example of using this new API? Either a JSFiddle or a snippet of code from a website you're building would help.

Thanks!

MGreek:
Fixed the typo. I'll also try to do an example next.

PR #1618, opened 2024-04-19
fix: BarNote getSVGElement always returning undefined

judocode:
BarNote is not opening its own context group which means it fails to perform operations requiring the id of the Element such as getSVGElement, addClass, removeClass etc.

I also added barline as a property of barnote to memoize it and have direct access to the barline if one needs it.

It looks like there are not any tests for the barnote directly but I added a test for the error scenario I was running into where  getSVGElement was always returning undefined for rendered BarNotes

### DIFF
--- a/src/barnote.ts
+++ b/src/barnote.ts
@@ -30,6 +30,7 @@ export class BarNote extends Note {
   protected metrics: { widths: Record<string, number> };
   // Initialized by the constructor via this.setType(type)
   protected type!: BarlineType;
+  public barline!: Barline;
 
   constructor(type: string | BarlineType = BarlineType.SINGLE) {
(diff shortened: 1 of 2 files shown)


AaronDavidNewman:
Hi @judocode I just pulled the latest version of your branch, did npm install and grunt webpack:debug.  I see the following error in Barline tests.

I'm thinking your check should skip the undefined check if the renderer is canvas.

image

judocode:
Ahh thanks @AaronDavidNewman - my tests were timing out for an unrelated issue it seems, but I added that check and was able to run all tests 👍

AaronDavidNewman:
This looks good, thanks for your contribution.

PR #1648, opened 2025-02-07
Always truthy error

mscuthbert:
Remove part of a line in tables.ts that will never be executed -- raises a compiler error in Typescript 5.6 and above.  Probably a result of a refactor.  Will commit also to VF5 repo.


mscuthbert:
(Error does not exist in VF5)

PR #1653, opened 2025-03-05
Added 256th note required glyphs and fixed 1/2 rest dot rendering.

TuxIceland:
Added required 256th glyphs for Bravura, Petaluma, Leland fonts. With these changes, when a 256th note (or rest) is rendered, the corresponding glyph is displayed instead of the quarter glyph.
I also fixed a small issue with the alignment of double whole rest dots.

### Rests Preview
Screenshot 2025-03-05 alle 22 50 48


rvilarl (approved):
Thanks!

